PR #13, opened 2015-10-30
add "baselineUpdateConfig" task to download latest config files

bdworth:
to .baseline/ , where plugin looks for its config

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -37,3 +37,14 @@ subprojects {
     version = flexVersion()
     description = 'A Gradle plugin for applying Baseline-recommended build and IDE settings'
 }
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -37,3 +37,14 @@ subprojects {
     version = flexVersion()
     description = 'A Gradle plugin for applying Baseline-recommended build and IDE settings'
 }
+
+def baselineDependency = dependencies.create("com.palantir:gradle-baseline-java-config:+@zip")
```
uschi2000: this doesn't work: it's not visible to users of the plugin.

PR #17, opened 2015-11-07
Baseline-eclipse merges jdt.core config rather than replacing it

uschi2000:
@bdworth @cjp39 take a look please.

### DIFF
--- a/gradle-baseline-java-config/resources/eclipse/org.eclipse.jdt.core.prefs
+++ b/gradle-baseline-java-config/resources/eclipse/org.eclipse.jdt.core.prefs
@@ -1,4 +1,3 @@
-eclipse.preferences.version=1
 org.eclipse.jdt.core.codeComplete.argumentPrefixes=
 org.eclipse.jdt.core.codeComplete.argumentSuffixes=
 org.eclipse.jdt.core.codeComplete.fieldPrefixes=
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineEclipse.groovy
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineEclipse.groovy
@@ -16,6 +16,7 @@
 
 package com.palantir.baseline.plugins
 
+import groovy.text.SimpleTemplateEngine
 import org.gradle.api.Project
 import org.gradle.api.plugins.JavaPlugin
 import org.gradle.plugins.ide.eclipse.EclipsePlugin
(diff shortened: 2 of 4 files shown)


cjp39:
Looks better, thanks! Can we remove the overwrite logic entirely?

uschi2000:
I don't think we can remove it since the files that get written (or overwritten if they exist) don't have corresponding configuration in Gradle's eclipse plugin, AFAIK. This seems to be doing the same thing that gradle-processors does with org.eclipse.jdt.apt.core.prefs, if I understand the code right at first glance?

cjp39:
Then thumbs up.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineEclipse.groovy
```diff
@@ -25,11 +26,61 @@ import org.gradle.plugins.ide.eclipse.EclipsePlugin
...
+    /**
+     * Copies all name/value pairs from {@code from} to {@code into}, replacing variable names (e.g., $var or ${var})
+     * occurring in values via a {@link SimpleTemplateEngine} and the given binding.
+     */
+    static def mergeProperties(Properties from, Properties into, Map binding) {
+        from.stringPropertyNames().each { property ->
+            def propertyValue = from.getProperty(property)
+            try {
+                // Not all properties are well-formed; attempt to parse and fall-back to original value.
+                // Sadly, this is outrageously slow.
```
bdworth: How slow is "outrageously slow"?

gradle-baseline-java-config/resources/eclipse/org.eclipse.jdt.core.prefs
```diff
@@ -93,8 +92,8 @@ org.eclipse.jdt.core.formatter.comment.preserve_white_space_between_code_and_lin
 org.eclipse.jdt.core.formatter.compact_else_if=true
 org.eclipse.jdt.core.formatter.continuation_indentation=2
 org.eclipse.jdt.core.formatter.continuation_indentation_for_array_initializer=2
-org.eclipse.jdt.core.formatter.disabling_tag=@formatter\\:off
-org.eclipse.jdt.core.formatter.enabling_tag=@formatter\\:on
+# org.eclipse.jdt.core.formatter.disabling_tag=@formatter\\:off
+# org.eclipse.jdt.core.formatter.enabling_tag=@formatter\\:on
```
bdworth: Do these changes have a different effect than deleting the lines? If not, let's delete them.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineEclipse.groovy
uschi2000: Didn't measure, but on the order of 2-3 seconds for the 300 lines, vs "immediate" without the template scanning.

gradle-baseline-java-config/resources/eclipse/org.eclipse.jdt.core.prefs
uschi2000: Ah, my bad. Fixed.

PR #20, opened 2015-11-20
Eclipse formatter prefers wrapping entire assignment rather than subs…

uschi2000:
…et of parameters

Example
Before:
  int foo = bar(a
      b, c);

After:
  int foo =
      bar(a, b, c);


uschi2000:
@bdworth

PR #35, opened 2016-01-20
Remove VariableDeclarationUsageDistance check

uschi2000:
… section

### DIFF
--- a/gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
+++ b/gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
@@ -295,9 +295,6 @@
             <property name="processJavadoc" value="true"/>
         </module>
         <module name="UpperEll"/> <!-- Java Style Guide: Numeric Literals -->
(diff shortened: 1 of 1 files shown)


uschi2000:
@derekcicerone

derekcicerone-zz:
Does anyone actually modify the defaults in practice? Like how useful is it to have the 2 different sections?

jmcampanini:
question: does this mean that it'll still be applied by default, but you can easily delete/restore it?

**re: 2 different states:** the original pt-gradle/pt-checkstyle plugins had strict/non-strict, and it just turned into people not using strict. it was not well received. it also feels really weird saying `baseline-strict` and `baseline-

uschi2000:
non-strict mode is currently opt-in: you comment out the respective section from .baseline/checkstyle/checkstyle.xml .

uschi2000:
After email discussion, removed the check entirely.

jmcampanini:
:+1:

jnew528:
+1

PR #36, opened 2016-01-22
Upgrade Gradle to 2.10

markelliot:


uschi2000:
Groovy conflict

markelliot:
Yeah....nebula.test only goes up to 2.8-rc1 (which I think is when testing made it into core Gradle) and Gradle 2.9 bumped to Groovy 2.4, so the nebula.test version of spock just isn't recent enough. Thinking about just punting on the upgrade.

Started a thread internally re: the changes I actually wanted to make (and for which these were pre-commits).

uschi2000:
Straight to 2.12: https://github.com/palantir/gradle-baseline/pull/60

PR #39, opened 2016-01-29
Automatically extract configuration, and allow for overrides

markelliot:
Throwing this out there mostly as an intermediate state so we can discuss some implementation details.


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineCheckstyle.groovy
```diff
@@ -31,6 +31,7 @@ class BaselineCheckstyle extends AbstractBaselinePlugin {
     void apply(Project project) {
         this.project = project
 
+		project.rootProject.apply BaselineConfig
```
markelliot, line 34: I'm not sure this is kosher, but can't think of another way. In addition to adding this plugin to the root project, I think I need to require each project's checkstyle to depend on `:baselineUpdateConfig` (which should skip once it runs once, per changes)

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/AbstractBaselinePlugin.groovy
```diff
@@ -35,4 +36,21 @@ abstract class AbstractBaselinePlugin implements Plugin<Project> {
...
+    /**
+     * Returns the absolute path of the Baseline configuration extract, 
+     * i.e., the directory '.bline' in the root directory of this project.
+     */
+    protected final String getExtractDir() {
+        return Paths.get(project.rootDir.toString(), BaselineParameters.DEFAULT_EXTRACT_DIR).toString()
+    }
+
+	protected final File resolveConfigPath(String path) {
+		File file = project.file(getConfigDir() + '/' + path)
```
uschi2000, line 49: should use NIO path things instead of "/"

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineCheckstyle.groovy
uschi2000, line 34: I believe it's OK to apply plugins to other projects; you may need/want to check for duplicates though.

uschi2000:
closing for now, can re-open when we consider this again. @markelliot

PR #42, opened 2016-02-20
Checkstyle rule updates

bdworth:
- sort checks
- add Float to IllegalInstantiation
- add most Guava Multimap classes to IllegalTypes


uschi2000:
Thanks, Brian!

PR #44, opened 2016-03-01
exclude certain findbugs patterns from Immutables-generated classes

gracew:
Before change:
screen shot 2016-02-29 at 9 12 20 pm

After change:
screen shot 2016-02-29 at 9 11 13 pm

### DIFF
--- a/gradle-baseline-java-config/resources/findbugs/excludeFilter.xml
+++ b/gradle-baseline-java-config/resources/findbugs/excludeFilter.xml
@@ -18,4 +18,16 @@
             <Bug pattern="URF_UNREAD_PUBLIC_OR_PROTECTED_FIELD"/> <!-- @Rule ivars can be public and unused -->
         </Or>
     </Match>
+
+    <!-- Ignore the following bug patterns in Immutables generated code -->
+    <!-- (i.e., classes beginning with 'Immutable' and inner classes of same) -->
+    <Match>
+        <Class name="~.*Immutable.*(\$.*)?" />
+        <Or>
+            <Bug pattern="DLS_DEAD_LOCAL_STORE"/>
+            <Bug pattern="ES_COMPARING_PARAMETER_STRING_WITH_EQ"/>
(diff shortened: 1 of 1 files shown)


gradle-baseline-java-config/resources/findbugs/excludeFilter.xml
```diff
@@ -18,4 +18,16 @@
             <Bug pattern="URF_UNREAD_PUBLIC_OR_PROTECTED_FIELD"/> <!-- @Rule ivars can be public and unused -->
         </Or>
     </Match>
+
+    <!-- Ignore the following bug patterns in Immutables generated code -->
+    <!-- (i.e., classes beginning with 'Immutable' and inner classes of same) -->
+    <Match>
+        <Class name="~.*Immutable.*(\$.*)?" />
```
uschi2000, line 25: is there a chance that people will use ImmutableFoo when it's not generated? I wonder if we should move to a prefix convention like "Immutables_" instead. Or maybe this block doesn't actually belong in baseline because not everyone uses Immutables.

gracew, line 25: yes, I was also worried about that (non-generated classes starting with Immutable).

actually it looks like Immutables has added FB suppressions in 2.1.12 (released 2 days ago!) so I'll try upgrading to that instead.

See: https://github.com/immutables/immutables/issues/263

uschi2000, line 25: ah, cool!

gracew, line 25: unfortunately upgrading didn't do what I wanted since it only adds @SuppressFBWarnings when the findbugs annotations jar is on the classpath. Since not everyone uses Immutables, I'll just add this block to the excludeFilter.xml in my project.

uschi2000:
ok, thanks Grace!

PR #45, opened 2016-03-21
Set default JUnit directory for IDEA to be module directory

nmiyake:


nmiyake:
@uschi2000 

Flying a bit blind here since I don't know how to build and test the Gradle plugin locally... However, did confirm that manually adding the specified config to to a local `.ipr` file has the desired behavior.

Most unit tests assume that the module is the current working directory -- this is the standard behavior for "gradle test" and Eclipse JUnit tests. However, without this modification, IntelliJ uses the workspace directory as the working directory, which makes many test path resolutions fail.

This configuration setting is outlined in http://stackoverflow.com/a/5638830.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
```diff
@@ -132,6 +133,19 @@ class BaselineIdea extends AbstractBaselinePlugin {
     }
 
     /**
+     * Configure the default working directory of JUnit tests to be the module directory.
+     */
+    private void addJUnitWorkingDirectory(node) {
+        node.append(new XmlParser().parseText('''
+            <component name="RunManager">
+                <configuration default="true" type="JUnit" factoryName="JUnit">
```
uschi2000, line 144: beautiful. should we add this for the Application configuration as well?

nmiyake, line 144: Good call -- if we do it for debug, would make sense for the Application configuration as well. Will add that.

uschi2000, line 144: cheers

PR #46, opened 2016-03-21
Set IDEA settings on iws instead of ipr

nmiyake:


nmiyake:
@uschi2000 

Still flying blind, but based on Gradle docs seems like this should work... Can you verify?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
```diff
@@ -41,6 +41,10 @@ class BaselineIdea extends AbstractBaselinePlugin {
                     addCopyright(node)
                     addCheckstyle(node)
                     addGit(node)
+                }
+
+                ideaRootModel.workspace.iws.withXml { provider ->
+                    def node = provider.asNode()
                     addJUnitWorkingDirectory(node)
```
uschi2000: method name is stale

nmiyake:
OK, actually ran and tested this one, and it should be good (and more correct than the previous approach :)

uschi2000:
OK, let's try this one. Thanks Nick!

uschi2000:
For me, the ipr now contains the right configuration, but the iws still does not. (Gradle 2.11). Any idea, @nmiyake ?

nmiyake:
Hmm, what project are you trying it on?

uschi2000:
foundry

nmiyake:
OK cool, let me try it over there. I just verified locally again and it works (I'm using `http-remoting`, which is on Gradle 2.7).

This is how I'm testing this:
- Check out `gradle-baseline` locally
- Run `./gradlew publishToMavenLocal`
- Check out develop of another project (I used `http-remoting`)
- Open `build.gradle`, add `mavenLocal()` to `repositories()` block (in `buildscript` and at top level), and set version of Baseline to be `0.3.0-rc1` (since that's what's published locally): `"com.palantir:gradle-baseline-java:0.3.0-rc1"`
- Run `./gradlew idea`
- Open `http-remoting.iws` to verify output

When I do so, I see that the `.iws` file has the expected values set.

nmiyake:
OK, checked out Foundry and confirmed that it's not working as I would expect there -- will investigate.

nmiyake:
@uschi2000 , did a bit of investigating and I think it's an issue with external dependency resolution -- it's the difference between `gradle.plugin.com.palantir` and `com.palantir` -- the former does not show the new behavior, while the latter does.

This dependency does not exhibit the new behavior:

```
classpath "gradle.plugin.com.palantir:gradle-baseline-java:0.3.0-rc1"
```

But this one does:

```
classpath "com.palantir:gradle-baseline-java:0.3.0-rc1"
```

Does this have something to do with external publishing?

uschi2000:
Ah, sheit. bintray artifacts are mutable and gradle.plugin.com artifacts
are not ... I republished the plugin under the same 0.3.0-rc1 name and this
screwed me :(
Sorry for the confusion. I'll release 0.3.0-rc2 and we should be all good.
  R

PR #48, opened 2016-04-05
Add Assume and Truth to valid static imports

NeilRickards:


uschi2000:
Very nice, thanks Neil!

PR #54, opened 2016-04-18
Remove requirement to use ImmutableList.<Foo, Bar>builder() instead of new ImmutableList.Builder<Foo, Bar>()

j-baker:
Doesn't feel like this fits the criteria of being a 'baseline requirement' - in Java 6 there's an advantage to the former, but in Java 7+ there's no practical advantage to using this method, it's just a matter of taste.

### DIFF
--- a/gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
+++ b/gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
@@ -238,18 +238,6 @@
             <property name="format" value="CoreMatchers\.notNull"/>
             <property name="message" value="Use better assertion method(s): Assert.assertEquals(), assertNull(), assertSame(), etc."/>
         </module>
-        <module name="RegexpSinglelineJava"> <!-- Java Coding Guidelines: Avoid Generics clutter where possible -->
-            <property name="format" value="ImmutableList\.Builder.*new ImmutableList\.Builder"/>
-            <property name="message" value="Use ImmutableList.builder() for variable assignment."/>
-        </module>
-        <module name="RegexpSinglelineJava"> <!-- Java Coding Guidelines: Avoid Generics clutter where possible -->
-            <property name="format" value="ImmutableMap\.Builder.*new ImmutableMap\.Builder"/>
-            <property name="message" value="Use ImmutableMap.builder() for variable assignment."/>
-        </module>
-        <module name="RegexpSinglelineJava"> <!-- Java Coding Guidelines: Avoid Generics clutter where possible -->
-            <property name="format" value="ImmutableSet\.Builder.*new ImmutableSet\.Builder"/>
-            <property name="message" value="Use ImmutableSet.builder() for variable assignment."/>
-        </module>
         <module name="RegexpSinglelineJava"> <!-- Java Coding Guidelines: Check parameters for validity -->
             <property name="format" value="Preconditions\.checkNotNull\((?!.*,)([^()]*(\(([^()]*|\(([^()]*|\([^()]*\))*\))*\)[^()]*)*)\)"/>
             <property name="message" value="Use Preconditions.checkNotNull(Object, String)."/>


j-baker:
To be clear, this doesn't remove any actual generics clutter, as:
- If you are assigning the result to a list, you must still provide generic parameters.
- If you are assigning the result to a builder, you just use diamond notation.

uschi2000:
re clutter: agreed. re taste, is there a reason we should not "prefer" (in terms of taste) `ImmutableList.builder()`? to be clear, a substantial part of baseline is exactly about taste, it's just a matter of finding an agreeable taste.

j-baker:
I guess my thoughts are that baseline
1. Provides rules and tools that attempt to stop people from writing crappy code or accidentally introducing bugs. Stuff like hiddenfield, fallthrough, emptycatchblock fall under this, and perhaps designforextension for all its sins (and obviously findbugs + jacoco etc).
2. Provides consistency for consistency's sake where it helps comprehension, in order to make it easier for people to swap teams and contribute across the company, etc. For me, this refers to code shape more than anything else, but also to large pieces of tech that you have to spend a lot of time understanding. So, for example, use Jackson not GSON, use JUnit and not TestNG. Prefer Java to Scala.

I think you're coming from the perspective of 'baseline should probably have an opinion here' - I don't think that having an opinion adds any readability either way (because the code shape is identical), and so enforcing consistency isn't a win for Baseline.

I think that this sort of rule can be very powerful, but I think that it could be better used to indicate strictly better alternatives. For example, I'd love to see a rule that enforces the usage of LinkedHashMap instead of HashMap, for example (thoughts?), or maybe even one that warns when you define an array - but in that case you're doing it for entirely different reasons.

uschi2000:
Trying to understand... We agree that if one alternative is strictly better than another, then Baseline should advocate for the better one. Conversely, if two alternatives are functionality equivalent, isn't it still preferable to just agree on using one? What do you win by this ambiguity, what do you loose by asking people to use one of them (mostly arbitrarily)?

A slightly more dramatic example: (2-space vs 4-space) and (tabs vs spaces) are each functionally equivalent and a matter of "code shape". Would you advocate for removing those recommendations from Baseline?

j-baker:
I guess the primary difficulty is knowing what's been agreed upon. For example, here, I discovered that it had been agreed on 8 minutes into my PR build, because I simply didn't know this rule existed, which wasted my time. I use these builders infrequently enough that I've not been dinged for this at least the last 6 months, and it felt weird that Baseline had an opinion on these three classes specifically.

I suspect that the 'right' check in this case would be one like 'use static builder() methods instead of new Foo.Builder()' - which could easily be enforced with FindBugs.

On Mon, 18 Apr 2016 at 21:47 Robert Fink <notifications@github.com<mailto:notifications@github.com>> wrote:

Trying to understand... We agree that if one alternative is strictly better than another, then Baseline should advocate for the better one. Conversely, if two alternatives are functionality equivalent, isn't it still preferable to just agree on using one? What do you win by this ambiguity, what do you loose by asking people to use one of them (mostly arbitrarily)?

A slightly more dramatic example: (2-space vs 4-space) and (tabs vs spaces) are each functionally equivalent and a matter of "code shape". Would you advocate for removing those recommendations from Baseline?

uschi2000:
OK, suggestion noted. I will be in London next week and would love to collect Baseline feedback and suggestions, maybe we can chat in person and hash out what you'd to see (and not see)?

NeilRickards:
Spaces and tabs is a different case, as they don't interact well.
Where there are two alternatives that are functionally equivalent and interoperate fine, my preference would be to be less restrictive.

uschi2000:
Let's get two more opinions: @derekcicerone @joelea

joelea:
I entirely agree with @NeilRickards comment.

derekcicerone-zz:
Rob, I'll defer to you but personally I'm in sync with removing this restriction if there is no functionality advantage in Java 8+.

uschi2000:
deal.

uschi2000:
Thanks all!

PR #58, opened 2016-05-03
Suppress checkstyle for */generated/** files

uschi2000:


bavardage:
👍

uschi2000:
added path-based exclusions for findbugs.
@jmcampanini @chrisalice could you take a look please?

gradle-baseline-java-config/resources/checkstyle/checkstyle-suppressions.xml
```diff
@@ -19,6 +19,7 @@
     <suppress files="\.(bdr|eot|gif|gzip|jar|json|otf|png|svg|ttf|woff|zip)$" checks="NewlineAtEndOfFile" />
 
     <!-- Generated code should not be subjected to checkstyle. -->
+    <suppress files="[/\\].*[/\\]generated[/\\]" checks="." />
     <suppress files="[/\\].*[/\\]generated_src[/\\]" checks="." />
```
jmcampanini, line 23: do we still need these? can we remove `generated_[src|testSrc]`?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/BaselineFindBugsExtension.groovy
```diff
@@ -0,0 +1,11 @@
+package com.palantir.baseline
+
+import java.util.regex.Pattern
+
+class BaselineFindBugsExtension {
+    def List<Pattern> exclusions = [Pattern.compile("/generated/")]
```
jmcampanini, line 6: do we want to be more specific to match JUST `/generated/` ?

uschi2000, line 6: don't understand, what do you mean?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/BaselineFindBugsExtension.groovy
```diff
@@ -0,0 +1,11 @@
+package com.palantir.baseline
+
+import java.util.regex.Pattern
+
+class BaselineFindBugsExtension {
```
jmcampanini, line 5: can we add directly to the findbugs extension? not saying we should, just curious if that's even an option...

jmcampanini:
in general 👍 

but i'd prefer if we picked one way of representing generated sources and i'd vote for `/src/generated/*`

gradle-baseline-java-config/resources/checkstyle/checkstyle-suppressions.xml
uschi2000, line 23: that's what gradle-processors uses

gradle-baseline-java/src/main/groovy/com/palantir/baseline/BaselineFindBugsExtension.groovy
uschi2000, line 5: hmm, I don't think so

gradle-baseline-java-config/resources/checkstyle/checkstyle-suppressions.xml
jmcampanini, line 23: right. i'm hoping to consolidate to a single location. seems simpler/cleaner

gradle-baseline-java/src/main/groovy/com/palantir/baseline/BaselineFindBugsExtension.groovy
jmcampanini, line 5: i think you could with `MetaClass` in groovy? but i doubt it's the right thing to do.. was just curious

gradle-baseline-java/src/main/groovy/com/palantir/baseline/BaselineFindBugsExtension.groovy
jmcampanini, line 6: apologies. comment doesn't make sense. please ignore :)

gradle-baseline-java/src/main/groovy/com/palantir/baseline/BaselineFindBugsExtension.groovy
uschi2000, line 5: ah, yes. I'd prefer the "standard" gradle way: extension classes.

gradle-baseline-java-config/resources/checkstyle/checkstyle-suppressions.xml
uschi2000, line 23: ok. I will leave these here for now and see if we can get gradle-processors to use generated instead of generated_src. @chrisalice ?

alicederyn, line 23: What's the motivation behind putting generated files under the src directory? It seems to violate convention. Generated code should live in separate build directories.

alicederyn, line 23: Oh, I think I may have misunderstood this a bit. If gradle-processors starts writing to `build/generated/main/java` and `build/generated/test/java`, for instance, would that conform?

jmcampanini, line 23: `build/generated/*` seems better than `{projectDir}/generated_src/`. my only concern is how that feels when working on things from an IDE (since `build/` is not an intellij/eclipse concept). i assume it should be fine, but curious to how it affects the developer experience - since functionally all 3 options are equivalent.

my vote: having generated code go in `src/generated/` gives enough signal that the code is generated and not to be modified. other pros: it lives next to my other source code and it's easy to find in my IDE. also flagging the directory as "Generated Source Code" in intellij makes it even clearer.

alicederyn, line 23: We should stick to maven convention unless there's a really good reason, and maven dictates that src is the stuff that lives in source control.

jmcampanini, line 23: +1

Do you have a link to the conventions? Only one I could find doesn't reference generated source directly.

alicederyn, line 23: https://maven.apache.org/guides/introduction/introduction-to-the-standard-directory-layout.html

> The target directory is used to house all output of the build.

uschi2000, line 23: how do you then deal with generated stuff that lives in source control? e.g., protobuf-generated code.

jmcampanini, line 23: if i understand the maven conventions, if you check it in, then `src/*` (and i would vote for `src/generated/*`), and if you don't it should go in `build/*` (the gradle equivalent of `target/`)

uschi2000:
closing since no movement

PR #65, opened 2016-05-11
Upgrade to Gradle 2.13

punya:
@uschi2000


jmcampanini:
missing the `gradlew` and `gradlew.bat` files :-\

punya:
When I ran `./gradlew wrapper --gradle-version=2.13`,
- `gradlew` was unchanged
- `gradlew.bat` had whitespace changes only (windows newlines instead of unix)

Happy to include that if you prefer.

jmcampanini:
hrmm. i had these changes when i ran 2.13 wrapper: https://github.com/palantir/dropwizard-web-security/pull/59/files

i had to delete my existing gradlew + gradlew.bat files before i saw any change.

punya:
Maybe because you were upgrading from 2.10 and this is upgrading from 2.12?

jmcampanini:
def could be it. if it's unchanged, 👍 from me :)

punya:
Nope, I was just plain wrong :) After upgrading the wrapper to 2.13 I ran the thing again and there were changes. Sounds like it was a bootstrapping problem.

uschi2000:
will this break pre-2.13 users?

punya:
Nope, this only changes how the plugin is built, not how it can be consumed.

uschi2000:
That's not true, Punya. build.gradle pulls in the Gradle API version that the build is running with: https://github.com/palantir/gradle-baseline/blob/develop/gradle-baseline-java/build.gradle#L12 . So in theory you could start relying on new behaviour/API when upgrading the gradle version. I guess there are no code changes though, so this should be safe. thanks!

PR #66, opened 2016-05-13
Gradle 2.13 correctly handles compileOnly

robert3005:
This causes massive confusion for gradle and it assigns all compile dependencies as PROVIDED which breaks any launch configuration in intellij.

After looking at https://github.com/gradle/gradle/blob/master/subprojects/ide/src/integTest/groovy/org/gradle/plugins/ide/idea/IdeaIntegrationTest.groovy it turns out this should already be handled. I guess this workaround is only necessary for 2.12. Confirmed working on contour repo.

If someone has an idea how to make it work for 2.12 and 2.13 at the same time please contribute. Added a test case to ensure we don't have any non solutions.
cc @ash211  @uschi2000


jmcampanini:
running into this too - almost shipped it with highline :(. downgrading to 0.4.0 was our immediate fix.

jmcampanini:
ping?

uschi2000:
sorry, missed this one.

uschi2000:
just to clarify: in 2.13, compileOnly works out of the box with intellij?
then I think it's the right decision to merge this PR.

PR #68, opened 2016-06-11
Exclude 'com.sun' and 'java.awt' packages from auto-complete in IDEA

nmiyake:


nmiyake:
@markelliot @uschi2000 

Fixes annoying issue where when you try to import things like `List` or `Map` it shows multiple matches in irrelevant packages like `com.sun` and `java.awt`. I think excluding auto-import for these is reasonable since we already have a rule against `com.sun` imports and we generally don't write awt code.

Although Eclipse has the equivalent setting, it's a workspace-level setting rather than a project one, so can't make an equivalent Baseline-generated rule :(

uschi2000:
hmm, i'm a little torn -- what's the chance of hitting a false positive?
i think it may be too restrictive, given that some people may actually want to write awt apps.

nmiyake:
For clarification, this just hides from auto-complete, so doesn't completely stop ability to write code.

Also, I know it's hard to speak in generalizations, but would be really skeptical that there are many people writing AWT apps -- even for native stuff, people are almost always on Swing.

If we don't want to exclude all of AWT from auto-complete, can we narrow down and just exclude `java.awt.List`? Super annoying that it overloads that, and I would venture that 99% of time it's not what people want.

uschi2000:
so alt-enter wouldn't show those any longer, correct?

nmiyake:
Correct

uschi2000:
Ok, then I think we should not do this as part of the baseline-idea plugin.
I know those cases will be rare, but if someone actually wants to import
those packages it will be a very frustrating experience. We could make a
separate plugin or explicit config option, but it should be opt-in.

nmiyake:
I think the trade-off comes down to whether to optimize for an exceedingly rare large frustration where the work-around is not immediately obvious (because it's an omission) versus optimizing for a minor frustration that happens pretty much daily for all projects but where the immediate work-around is obvious (select the import directly). The long-term work-around (set ignores as part of IDE) can manually be applied as well, but it requires users to think about it and it's easy to forget to do when moving to a new computer/workspace etc.

I do agree that customizing whether or not it's applied can make sense (much as we have ways to work around it if users really want to use an illegal import).

My preferences in order:
(1) Include configuration that controls whether or not these classes are excluded, but have them _excluded by default_. If a project maintainer has to remember to flip this on, it's way too easy to forget. Note that this will happen in the docs.
(2) Similar to (1), but narrow the target more aggressively to the individual class levels (instead of blanket ban on all `sun` and `awt` classes:
- To disambiguate `Map`: `com.sun.javafx.collections.MappingChange.Map`, `com.sun.xml.internal.xsom.impl.scd.Iterators.Map`
- To disambiguate `List`: `java.awt.List`, `com.sun.tools.javac.util.List`, `com.sun.xml.internal.bind.v2.schemagen.xmlschema.List`
- To disambituate `Set`: `com.sun.tools.classfile.Opcode.Set`

Other places in Baseline make principled stands about making it an error to use `com.sun` classes and concrete collection classes, so seems like it would be consistent/okay to be opinionated about this (although I do understand the argument of discoverable versus non-discoverable rules/failure).

(3) Write another plugin that does just this, and call it something like "recommended-idea-defaults" or "palantir-idea-defaults". Should be fine, although it does have the overhead of writing/maintaining another plugin.
(4) Fork Baseline to include this change and I'll just use the forked version in all the projects I control.

Thoughts?

Also looping in @jmcampanini since I'm interested in his take on the trade-offs in situations like these as someone who's been involved with DevX for a while.

uschi2000:
My preference order: 2, 1, 3. (4 is obviously not an option ;) )
I could be convinced that (1) makes sense, let's hear what Javi thinks.

nmiyake:
Yup, understood :)

Will wait to see what Javi thinks. If we go with (2), will echo back to try to get consensus on what the format/specifics should look like

jmcampanini:
doesn't the checkstyle rule block all `sun.*` packages and not `com.sun.*`? maybe i'm missing something there?

---

it feels like this is trying to take the illegal package rule we have in checkstyle and leverage it to convenience users in their IDE. if that's the case, why not power the IDE settings from the checkstyle rule? that way there's a single source of truth. without that, i feel that the plugin and the XML will skew, causing pain.

nmiyake:
@jmcampanini great catch, you're totally right -- `com.sun` is actually distinct from `sun.`, and the former is legal. Agree that this fact makes this PR much more tenuous -- "exclude the classes that are disallowed in checkstyle" seems conceptually consistent/defensible from an API and design standpoint, but since this doesn't square with that, it becomes much more tenuous.

I do personally want this functionality because it trips me up, but agree that Baseline isn't the place for it as this is currently constituted -- maybe I'll just write a separate plugin. Will close the PR for now -- thanks!

uschi2000:
Thanks Nick.

PR #70, opened 2016-06-20
Fix GStringImpl cast to String issue in Eclipse plugin

### LINKED ISSUE #69: gradle eclipse issue with 2.14
getting the following error when trying to run gradle eclipse using 2.14. Any idea @uschi2000 ?

```

* What went wrong:
Execution failed for task ':dummyproject-api:eclipseClasspath'.
> org.codehaus.groovy.runtime.GStringImpl cannot be cast to java.lang.String
```

qinfchen:
fixes https://github.com/palantir/gradle-baseline/issues/69

### DIFF
--- a/gradle-baseline-java/build.gradle
+++ b/gradle-baseline-java/build.gradle
@@ -14,6 +14,7 @@ dependencies {
     compile 'com.google.guava:guava:18.0'
 
     testCompile 'com.netflix.nebula:nebula-test:4.0.0'
+    testCompile 'org.apache.commons:commons-io:1.3.2'
     testCompile 'junit:junit:4.11'
     testCompile 'net.lingala.zip4j:zip4j:1.3.2'
     testCompile 'org.hamcrest:hamcrest-all:1.3'
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineEclipse.groovy
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineEclipse.groovy
@@ -120,9 +120,10 @@ class BaselineEclipse extends AbstractBaselinePlugin {
             // Override default Eclipse JRE.
             if (project.plugins.hasPlugin(JavaPlugin)) {
                 project.tasks.eclipseClasspath.doFirst {
+                    String eclipseClassPath = "org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-" + project.sourceCompatibility;
                     project.eclipse.classpath {
                         containers.clear()
-                        containers.add("org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-${project.sourceCompatibility}")
+                        containers.add(eclipseClassPath)
                     }
                 }
             }


qinfchen:
@uschi2000

gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineEclipseIntegrationTest.groovy
```diff
@@ -20,7 +20,7 @@ import com.google.common.base.Charsets
 import com.google.common.io.Files
 import nebula.test.IntegrationSpec
 import nebula.test.functional.ExecutionResult
-import org.apache.commons.io.FileUtils
+import org.gradle.internal.impldep.org.apache.commons.io.FileUtils
```
uschi2000: is this a 2.14-only class?

qinfchen: They removed the dependency on `org.apache.commons.io.FileUtils` and added `org.gradle.internal.impldep.org.apache.commons.io.FileUtils` instead.

qinfchen: should we pull in org.apache.commons.io.FileUtils? since it is for tests, i'd prefer pulling it directly

uschi2000: yes, I was just going to suggest that. can we stay on gradle 2.13 for a bit too, please? (don't want people to upgrade unnecessarily often.)

qinfchen:
ok, remove the gradle version in baseline. I think we should fix this regardless, so if people do decide to use new version of gradle, this plugin would still work

uschi2000:
sorry, what exactly is broken against 2.14?

qinfchen:
From the release note below, which broke the GStringImpl that we use in Eclipse plugin

```
A number of plugins were converted from Groovy to Java

The following plugins were fully converted to Java: jacoco, scala, osgi, javascript, distribution and announce.

Some other plugins were partially converted to Java, keeping tasks types as Groovy classes: init, checkstyle, codenarc, findbugs, pmd, jdepend, java, war, ear, application, signing, comparison, idea and eclipse. For the latter two, plugin types have also been kept in Groovy.

Existing builds and plugins should continue to work.

API method changes

org.gradle.api.artifacts.DependencyArtifact has setters for all properties to make it clearer that this type is mutable.
org.gradle.api.java.archives.ManifestMergeSpec now has setContentCharset() and getContentCharset() methods.
org.gradle.plugins.ide.eclipse.model.Facet.FacetType is no longer decorated with Groovy enum extensions.
org.gradle.testing.jacoco.plugins.JacocoTaskExtension.Output is no longer decorated with Groovy enum extensions.
org.gradle.nativeplatform.toolchain.plugins.MicrosoftVisualCppPlugin has been renamed to org.gradle.nativeplatform.toolchain.plugins.MicrosoftVisualCppCompilerPlugin.
org.gradle.testing.jacoco.plugins.JacocoPluginExtension.applyTo(JavaForkOptions) signature changed to make it clear this method accepts types that extend Task and JavaForkOptions.
```

uschi2000:
ok, so with this PR merged the eclipse error should be fixed when using 2.14, correct?

qinfchen:
yes, this plugin should work with 2.14 now.

uschi2000:
cool, thanks

qinfchen:
can we get a patch release out :)? Would love to use 2.14 and this hehe

uschi2000:
done

qinfchen:
👍 thanks!

PR #84, opened 2016-08-05
Make the maximum file size length 200 lines

j-baker:
As discussed internally, a reasonable default for file length is
200 lines.

### DIFF
--- a/gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
+++ b/gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
@@ -16,6 +16,7 @@
 
     <module name="FileLength"> <!-- Java Style Guide: File length: 2000 lines -->
         <property name="fileExtensions" value="java"/>
+        <property name="max" value="200"/>
     </module>
     <module name="FileTabCharacter"/> <!-- Java Style Guide: Whitespace characters -->
     <module name="NewlineAtEndOfFile"> <!-- Java Style Guide: Line ending: LF -->


gracew:
200 lines seems really aggressive. the apache license and java imports can easily take up 50 lines. maybe 500 is a more reasonable default?

gracew:
I'm also fine with deleting this check altogether (#85)

iamdanfox:
The project I work on has ended up with pretty nice readable classes and I think this check would actually prompt us to keep this trend.  The histogram of class length is:

screen shot 2016-08-05 at 16 10 21

(This excludes Immutables generated classes obviously)

j-baker:
License is a valid concern, although I think we have few enough OSS projects that those projects could easily just add 40 lines or whatever to their limit.

Wanted to share figures for my internal project (@hpryce created these fancy graphs, btw) - 

screen shot 2016-08-05 at 16 22 46

screen shot 2016-08-05 at 16 24 00

and simultaneously wanted to flag that i categorise our outliers as specifically the parts of the project most in need of a refactor.

Additionally, four of the larger classes are upgrade tasks.

j-baker:
screen shot 2016-08-05 at 16 33 57

this by the way is our 'volume of code bucketed by class size' graph.

hpryce:
The volume of code for the project Dan posted about looks like this, which I thought was interesting as it shows the bulk of the work done in the project is in classes below 100 lines:

screen shot 2016-08-05 at 16 37 36

jared2501:
Running a few quick bash commands on stemma: ~400 java classes, 60 of which are >= 200 lines. I think 200 is far too aggressive, and we shouldn't be enforcing this type of rule with checkstyle.

eshlin:
I usually just lurk and agree with intent of rule, but also worried about 200 lines being too aggressive, so wanted to chime in. Running across Foundry Core with bash as a rough check, I'm seeing about 300 classes (excluding generated ones, etc.) with about 45 classes >= 200 lines. More broadly, awesome we're using data, but we should probably sample across a broader range of projects, including ones that have been around longer as well.

jared2501:
Also not really in sync with this check existing at all.. no matter what number it is. Mainly since I don't think big classes are bad, and discouraging people from creating them means that people prematurely break up code into many classes that are actually less understandable that a larger one.

j-baker:
Closed as the alternative merged.

uschi2000:
cheers guys. I think 200 is a little harsh for most projects, so let's march down the yolo route instead.

PR #86, opened 2016-08-15
Replacing SimpleTemplateEngine with basic string replace to improve performance in mergeProperties

pjvelez:
Tested against our Emcee project (which has 97 projects). `./gradlew eclipse` runtime drops from 4 minutes to 40 seconds.

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineEclipse.groovy
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineEclipse.groovy
@@ -31,36 +31,37 @@ class BaselineEclipse extends AbstractBaselinePlugin {
      */
     static def mergeProperties(Properties from, Properties into, Map binding) {
         from.stringPropertyNames().each { property ->
-            def propertyValue = replace(from.getProperty(property), binding)
-            into.setProperty(property, propertyValue)
+            try {
+                def propertyValue = replace(from.getProperty(property), binding)
+                into.setProperty(property, propertyValue)
+            }
+            catch(Exception e) {
+                e.printStackTrace()
+            }
         }
     }
 
     static String replace(String str, Map binding) {
         StringBuilder sb = new StringBuilder()
         int start = 0
-        while (start < str.length()) {
-            int begin = str.indexOf("\${", start)
-            if (begin < 0) {
-                sb.append(str.substring(start))
-                break
-            }
 
-            int end = str.indexOf('}', begin)
-            if (end < 0) {
-                sb.append(str.substring(start))
-                break
-            }
+        int begin = str.indexOf("\${", start)
+        if (begin < 0) {
+            sb.append(str.substring(start))
+            return sb.toString()
+        }
 
-            String replacement = binding.get(str.substring(begin + 2, end))
-            if (replacement != null) {
-                sb.append(str.substring(start, begin))
-                sb.append(replacement)
-                start = end + 1
-            } else {
-                sb.append(str.substring(start, begin + 2))
-                start += 2
-            }
+        int end = str.indexOf('}', begin)
+        if (end < 0) {
+            throw new Exception("Missing '}' for property: " + str)
+        }
+
+        String replacement = binding.get(str.substring(begin + 2, end))
+        if (replacement != null) {
+            sb.append(str.substring(start, begin))
+            sb.append(replacement)
+        } else {
+            throw new Exception("Replacement missing from binding for property: " + str)
         }
         return sb.toString()
     }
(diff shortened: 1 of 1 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineEclipse.groovy
```diff
@@ -27,23 +27,42 @@ import org.gradle.plugins.ide.eclipse.EclipsePlugin
...
+            int begin = str.indexOf("\${", start)
+            if (begin < 0) {
+                sb.append(str.substring(start))
+                break
             }
 
-            into.setProperty(property, propertyValue)
+            int end = str.indexOf('}', begin)
+            if (end < 0) {
+                sb.append(str.substring(start))
```
uschi2000: should this be an error case?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineEclipse.groovy
```diff
@@ -27,23 +27,42 @@ import org.gradle.plugins.ide.eclipse.EclipsePlugin
...
+                break
+            }
+
+            String replacement = binding.get(str.substring(begin + 2, end))
+            if (replacement != null) {
+                sb.append(str.substring(start, begin))
+                sb.append(replacement)
+                start = end + 1
+            } else {
+                sb.append(str.substring(start, begin + 2))
```
uschi2000: should this be an error case?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineEclipse.groovy
```diff
@@ -27,23 +27,42 @@ import org.gradle.plugins.ide.eclipse.EclipsePlugin
...
-                def template = new SimpleTemplateEngine().createTemplate(from.getProperty(property))
-                propertyValue = template.make(binding).toString()
-            } catch (Exception e) {
-                propertyValue = from.getProperty(property)
+            def propertyValue = replace(from.getProperty(property), binding)
+            into.setProperty(property, propertyValue)
+        }
+    }
+
+    static String replace(String str, Map binding) {
```
uschi2000, line 44: this method supports multiple replacements, correct? do we really need that complication? if yes, then there should be a test for the behaviour.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineEclipse.groovy
pjvelez: Do you mean should this throw an error? Or account for this situation in a try/catch block? `indexOf` will return -1 if the character isn't present and not an error, so it'd only hit a catch block if we manually throw an error.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineEclipse.groovy
pjvelez: Similar to above, we can surround this with a try/catch. But then we'd have to manually throw the error. Is this preferred?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineEclipse.groovy
pjvelez, line 44: This version does support multiple replacements. `Template.make` was already doing this in the previous iteration however (it replaces all the bindings it finds in the input). So I don't believe this changes existing behavior.

uschi2000, line 44: Yep, but we replaced a documented library with custom code.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineEclipse.groovy
uschi2000: No, I'm wondering if this case should generate an exception. In other words, when we encounter a string `huihi ${foobar lalala`, should we silently leave it as is, or throw an error?

pjvelez: Ah I understand. Exception has been added in most recent commit.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineEclipse.groovy
pjvelez: Exception added in most recent commit.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineEclipse.groovy
pjvelez, line 44: Multiple replacements removed in most recent commit.

uschi2000:
Thanks!

uschi2000:
will be available as 0.7.2

pjvelez:
Sweet, thanks @uschi2000 !

PR #87, opened 2016-08-19
Throwables.propagate check

a10y:
Prefer using `Throwables.propagate()` over wrapping with `RuntimeException`. Can still do things like `throw new RuntimeException("custom error message")`.

Before:

```
catch (SomeException e) {
    throw new RuntimeException(e);
}
```

After:

```
catch (SomeException e) {
    throw Throwables.propagate(e);
}
```

### DIFF
--- a/gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
+++ b/gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
@@ -27,6 +27,11 @@
         <property name="format" value="@VisibleForTesting\s+(protected|public)"/>
         <property name="message" value="@VisibleForTesting members should be package-private."/>
     </module>
+    <module name="RegexpMultiline"> <!-- Using Guava's Throwables.propagate() is preferred over wrapping with RuntimeException -->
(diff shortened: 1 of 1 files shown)


a10y:
@uschi2000 when you get the chance

uschi2000:
sorry, this fell off my radar, looking now.

uschi2000:
This smells like a controversial change; can you please first hit the internal Baseline mailing list with this proposal, @andreweduffy ?

uschi2000:
(in particular, there are projects with no or limited access to Guava)

mateuszlitwin:
http://google.github.io/guava/releases/snapshot/api/docs/com/google/common/base/Throwables.html#propagate(java.lang.Throwable)

Throwables.propagate will be removed from Guava, it is recommended to use `new RuntimeException()` or `throwIfUnchecked()`; we should actually do opposite check if any.

a10y:
@uschi2000 I'm back in school now but someone should tell SMS people about this since Jared originally wanted to use Throwables.propagate everywhere.

uschi2000:
@jared2501 see above

PR #94, opened 2016-08-31
Disable -Xdoclint for Java 8+ projects, fix typo

a10y:
@uschi2000 

We check the javadoc with checkstyle, so we can disable -Xdoclint as it's purely console spam.


uschi2000:
lgtm, but can you send an email to pd-baseline-users, proposing this change?

a10y:
Sure

a10y:
No one had anything to say about this so I take that as an all clear

Issue #96, opened 2016-09-12
Block import of Clearspring stub versions of Guava classes


ash211:
E.g. http://grepcode.com/file/repo1.maven.org/maven2/com.clearspring.analytics/stream/2.9.0/com/clearspring/analytics/util/Lists.java and adjacent files

Add `com.clearspring.analytics.util` to `IllegalImport` / `illegalPkgs` block list

PR #97, opened 2016-09-12
Block clearspring stub packages of Guava

ash211:
Fixes https://github.com/palantir/gradle-baseline/issues/96


ash211:
My main concern here is whether the other parts of this package (that aren't stubs of well-known Guava classes) are used in other places.  Can we use this checkstyle rule on specific classes, and not just packages?

uschi2000:
there is this: http://checkstyle.sourceforge.net/config_imports.html#ImportControl

uschi2000:
looking through https://github.com/addthis/stream-lib/tree/master/src/main/java/com/clearspring/analytics/util , it seems that none of that stuff should get imported by anyone.

darora:
🎉

PR #107, opened 2016-11-14
Work around gradle/gradle#855

punya:
This showed up in an internal project when upgrading to Baseline 0.9.

(Also, to appease Github's linkification, gradle/gradle#855)


uschi2000:
Thanks, Punya!

PR #112, opened 2016-12-07
spark guava repackaged

bavardage:
cc @robert3005


bavardage:
lol @robert3005 beat me with #111

PR #130, opened 2017-02-24
Add checkstyle rule for test setup/teardown method names

uschi2000:


gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
```diff
@@ -307,6 +307,10 @@
             <property name="format" value="(\/\/TODO)|(\/\/ TODO:)|(\/\/ TODO\(\))|(\/\/ TODO\(.*\):[^ ])|(\/\/ TODO\(.*\)[^:])"/>
             <property name="message" value="TODO format: // TODO(flastname): explanation"/>
         </module>
+        <module name="RegexpSinglelineJava">
+            <property name="format" value="(void setUp\(\))|(void setup\(\))|(void setupStatic\(\))|(void setUpStatic\(\))|(void beforeTest\(\))|(void teardown\(\))|(void tearDown\(\))"/>
```
mnazbro: Why not just enforce one of `setUp` and `setup`?

uschi2000:
It's inverse, this rule throws up when it sees one of those patterns.

gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
```diff
@@ -307,6 +307,10 @@
...
+            <property name="message" value="Test setup/teardown methods are called before(), beforeStatic(), after(), afterStatic(), but not setUp, teardown, etc."/>
```
ellisjoe: `beforeClass()` and `afterClass()`?

uschi2000:
Ah,yes. Thanks

schlosna:
Honestly curious what value this adds? Specifically I'm thinking about case of having multiple independent @After that release resources and allow JUnit to handle ensuring shutdown appropriately. I guess you could have each one be a @Rule, but seems heavyweight and friction for little gain.

uschi2000:
I haven't seen a class with multiple `@After`, do you have an example?
The benefits of this change are: it doesn't matter what you call this method, so let's agree on one canonical name for the 99% cases (`before()`) so that we don't need to argue on PRs and can find the setup method quickly.

jmcampanini:
+1. instead of multiple `@After` methods, could you put them in a single one? it would also disambiguate what order those methods are ran in, since order isn't guaranteed by junit

PR #135, opened 2017-03-07
Remove implicit dependency on gradle-baseline-java-config

uschi2000:


markelliot (approved):

uschi2000:
Adding a few tests in second commit.

PR #145, opened 2017-03-11
Add baseline-errorprone project with slf4j check

uschi2000:
FYI @markelliot

### DIFF
--- a/baseline-errorprone/src/main/java/com/palantir/errorprone/slf4j/Slf4jConstantLogMessage.java
+++ b/baseline-errorprone/src/main/java/com/palantir/errorprone/slf4j/Slf4jConstantLogMessage.java
@@ -1,0 +1,65 @@
+/*
+ * Copyright 2017 Palantir Technologies, Inc. All rights reserved.
+ */
+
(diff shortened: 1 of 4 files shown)


baseline-errorprone/src/main/java/com/palantir/errorprone/slf4j/Slf4jConstantLogMessage.java
```diff
@@ -0,0 +1,65 @@
+/*
+ * Copyright 2017 Palantir Technologies, Inc. All rights reserved.
```
markelliot, line 2: Need to adjust for open source

uschi2000:
ACK

baseline-errorprone/src/main/java/com/palantir/errorprone/slf4j/Slf4jConstantLogMessage.java
uschi2000: fixed in https://github.com/palantir/gradle-baseline/pull/146

PR #146, opened 2017-03-12
Add baseline-error-prone

uschi2000:


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.groovy
```diff
@@ -0,0 +1,43 @@
...
+        this.project = project
+
+        if (!project.configurations.findByName("processor")) {
+            throw new GradleException("Expected to find 'processor' configuration for dependencies of the annotation" +
+                    "processor classpath. Consider adding the 'org.inferred.processors' Gradle plugin.")
+        }
+
+        project.plugins.apply(ErrorPronePlugin)
+        project.dependencies {
+            // TODO(rfink): This is somewhat ugly. Is there a better to add the processor dependency on the library?
```
uschi2000, line 35: any idea for making this better, @markelliot ?

markelliot: Could do it directly as a JavaCompile argument tweak?

uschi2000: I don't mind the `processor` thing. The requirement to know the plugin version at runtime is ugly.

markelliot: Can do latest.release and then assume a dep manager is going to override it?

Other way is via jar introspection?

uschi2000: Tried the jar introspection thing, but it's terrible in testing (because you're operating on classes, not jars ...). I'll leave it as it is.

PR #151, opened 2017-03-20
Reorder gradle in README so that baseline method is available

TheBeruriahIncident:
As is, the `baseline` method is not available. It comes from the baseline-config, so putting that above the use of `baseline` fixes things.

I'm very much not a gradle-wiz, but @uschi2000 seemed to agree with this issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,9 @@ repositories {
     jcenter()
 }
 
+// Apply for baselineUpdateConfig task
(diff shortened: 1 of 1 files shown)


uschi2000 (approved):
Thanks!

uschi2000:
Sorry, this fell off my radar.

PR #153, opened 2017-03-30
Use errorprone plugin's built-in configuration

markelliot:
Apply baseline-error-prone plugin using the errorprone configuration.

Resolve compilation problems when applying this plugin and working in Eclipse due to Eclipse classpath mangling with the processors plugin, which correctly applies the configuration's classpath to Eclipse's single classpath and where this plugin should not appear on the actual compilation classpath.

In particular, when using other processor plugins, this change will correct errors such as:

    Errors running builder 'Java Builder' on project 'glyph'.
        java.lang.ClassCastException: org.immutables.value.internal.$processor$.$Processor cannot be cast to javax.annotation.processing.Processor


markelliot:
see also https://github.com/tbroyer/gradle-errorprone-plugin#usage

uschi2000:
Oh, yes this was definitely a bug. Thanks.

markelliot:
cool, can I get a release with this included?

markelliot:
nvm, turns out i can cut releases

PR #156, opened 2017-04-08
Apply a couple of fixes we've found useful.

j-baker:
Fix 1 ensures that resources directories are marked as such.
Fix 2 ensures that the classpath as understood by IntelliJ is unified
with the classpath as understood by Gradle.

These are a couple of changes we've benefited from internally.


uschi2000 (approved):

uschi2000:
Cheers, James.

PR #158, opened 2017-04-12
bump gradle-errorprone-plugin to 0.0.10

bavardage:


bavardage:
this should add support for gradle 3.5

PR #161, opened 2017-04-23
Add error-prone check for unsafe usage of JsonTypeInfo

uschi2000:


markelliot (approved):

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DangerousJsonTypeInfoUsage.java
```diff
@@ -0,0 +1,69 @@
...
+import com.sun.source.tree.ExpressionTree;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "DangerousJsonTypeInfoUsage",
+        category = Category.ONE_OFF,
+        severity = SeverityLevel.ERROR,
+        summary = "Disallow usage of Jackson's JsonTypeInfo.Id.CLASS annotation for security reasons, "
+                + "cf. https://github.com/FasterXML/jackson-databind/issues/1599")
+public final class DangerousJsonTypeInfoUsage extends BugChecker implements BugChecker.AnnotationTreeMatcher {
```
j-baker, line 40: hey, I think this check is overly broad which is bad in the long run because it leads to poor understanding of why things are problems.

Specifically, my understanding is that the deserialization vulnerability is bad when you're using `@JsonTypeInfo` on a _field_ where the type is something overly broad, e.g. Object - because you can instruct Jackson to deserialize into an unsafe class.

I think we should be prohibiting this only on fields. `@JsonTypeInfo` with `Id.CLASS` is fine for security in a bunch of other places - possibly bad for maintainability, but not something that should be checked for by static analysis.

For an example, I worked on an internal product with `Id.CLASS` that remains unaffected by the Jackson security stuff.

> As to vulnerability, this only applies to polymorphic type handling via default typing (or especially annotated @JsonTypeInfo on java.lang.Object type property) -- and obviously JSON crafted by untrusted third party.

j-baker: Like, the pattern we should be trying to instil into people here is:

If you have a metaprogramming tool which will use reflection to call random constructors matching patterns, you should make sure that it can only call constructors which you have verified are safe to call.

Whereas I worry that checks like this instil the behaviour of 'Id.CLASS is dangerous'.

markelliot: I think this is appropriately broad. There is risk in saying that the JSON can self-declare which Java class to load, even if no known exploit is available today, there's real potential for one in the future. So, to future-proof a bit, it'd be much safer to use `JsonTypeInfo` with `Id.NAME` and simply using `JsonTypeName(<Type>.class)` on valid deserialization targets as this restricts the potential deserialization targets to a whitelist.

gsheasby: For AtlasDB, we're running into this too, with [RowColumnRangeIterator](https://github.com/palantir/atlasdb/blob/develop/atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/RowColumnRangeIterator.java)

markelliot: Why doesn't the recommendation above work for you?

gsheasby: We used the workaround (as you will see if you click the above link again). I was simply flagging an example of where Id.CLASS was not dangerous (at least, not today...).

Issue #162, opened 2017-04-25
Don't clobber existing RunManager configurations' WORKING_DIRECTORY values


henryptung:
`BaselineIdea` currently iterates all `RunManager` configurations and sets `WORKING_DIRECTORY` to `file://$MODULE_DIR$`. While this is appropriate for most projects, some need working directories other than the module directory (e.g. multiple runtime environments modeled by different working directories).

In particular, I'm not sure the plugin should act to modify any run configurations that may already exist (populating defaults is fine); otherwise, they might be corrupted by reexecution of the plugin.

uschi2000:
Can you point me to an example for: `e.g. multiple runtime environments modeled by different working directories`? (by email if it's internal)

iamdanfox:
Closing out stale issue (no activity in >1 year)

PR #163, opened 2017-04-26
Only modify default Application and JUnit run configs in BaselineIdea

henryptung:
Don't mutate existing non-default configs.

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
@@ -164,13 +164,14 @@ class BaselineIdea extends AbstractBaselinePlugin {
         def runManager = node.component.find { it.'@name' == 'RunManager' }
 
         if (runManager) {
-            // if RunManager configuration exists, set all WORKING_DIRECTORY options to module
-            def items = runManager.configuration.option.findAll { it.'@name' == 'WORKING_DIRECTORY' }
-            items.each { item ->
-                item.'@value' = 'file://$MODULE_DIR$'
-            }
+            // if RunManager configuration exists, set Application, JUnit WORKING_DIRECTORY default to $MODULE_DIR$
+            def appJunitDefaults = new NodeList(runManager.configuration
+                    .findAll { it.'@default' == 'true' && it.'@type' in ['Application', 'JUnit'] })
+            def workingDirectories = new NodeList(appJunitDefaults.option
+                    .findAll { it.'@name' == 'WORKING_DIRECTORY' })
+            workingDirectories.each { it.'@value' = 'file://$MODULE_DIR$' }
         } else {
-            // if RunManager configuration does not exist, add it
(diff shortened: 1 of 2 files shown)


henryptung:
Targets #162

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
```diff
@@ -164,13 +164,14 @@ class BaselineIdea extends AbstractBaselinePlugin {
...
 
         if (runManager) {
-            // if RunManager configuration exists, set all WORKING_DIRECTORY options to module
-            def items = runManager.configuration.option.findAll { it.'@name' == 'WORKING_DIRECTORY' }
-            items.each { item ->
-                item.'@value' = 'file://$MODULE_DIR$'
-            }
+            // if RunManager configuration exists, set Application, JUnit WORKING_DIRECTORY default to $MODULE_DIR$
+            def appJunitDefaults = new NodeList(runManager.configuration
+                    .findAll { it.'@default' == 'true' && it.'@type' in ['Application', 'JUnit'] })
```
uschi2000: can you pull out a variable `def configurations = ['Application', 'JUnit']` and use it here and in the else block?

uschi2000 (commented):
one nitpick, otherwise +1. thanks, Henry!

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
henryptung: Hmm, if we're talking about code duplication, how do you feel about this?

```
    /**
     * Configure the default working directory of RunManager configurations to be the module directory.
     */
    private static void setRunManagerWorkingDirectory(Node node) {
        def runTypes = ['Application', 'JUnit'] as Set

        def runManager = matchOrCreateChild(node, 'component', [name: 'RunManager'])
        runTypes.each { runType ->
            def configuration = matchOrCreateChild(runManager, 'configuration',
                    [default: 'true', type: runType],
                    [factoryName: runType])
            def workingDirectory = matchOrCreateChild(configuration, 'option', [name: 'WORKING_DIRECTORY'])
            workingDirectory.'@value' = 'file://$MODULE_DIR$'
        }
    }

    private static Node matchOrCreateChild(Node base, String name, Map attributes = [:], Map defaults = [:]) {
        def child = base[name].find { it.attributes().entrySet().containsAll(attributes.entrySet()) }
        if (child) {
            return child
        }

        return base.appendNode(name, attributes + defaults)
    }
```

uschi2000: yep, that's cool

henryptung: pushed fix, thanks!

uschi2000 (approved):
cheers, Henry.

PR #170, opened 2017-05-10
Improve generated code compiler arg

mnazbro:
The default README only described how to disable errorprone warnings in non-test code.


mnazbro:
Thanks @gracew for pointing this out

gracew (approved):

uschi2000 (approved):

PR #172, opened 2017-05-15
allow setup() and tearDown() for JMH benchmarks

ellisjoe:


ellisjoe:
eg: http://grepcode.com/file/repo1.maven.org/maven2/org.openjdk.jmh/jmh-core/1.10.3/org/openjdk/jmh/annotations/Setup.java?av=f

markelliot:
It'd be better if you could scope this; have a strong desire to take the debate out of test method names.

uschi2000:
+1

ellisjoe:
think this needs to be an error-prone check then

uschi2000:
can you not gate on the java file name? FooTest/FooTests vs FooBenchmarks?

iamdanfox:
@ellisjoe looks like this PR hasn't moved in a while - do you want to pick it up again or are you OK with just using the weird annotation/naming for now?

```
@Setup
public void before() {

}
```

iamdanfox:
Closing stale PR

PR #175, opened 2017-05-26
fix assumption in ValidateConstantMessage that message is latest string arg

gracew:
will try to submit a fix -- hopefully easy, as the PreconditionsConstantMessage check doesn't seem to exhibit this behavior.


baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/ValidateConstantMessageTests.java
```diff
@@ -118,6 +119,8 @@ public void negative() throws Exception {
                         "    Validate.notEmpty(mapArg, \"message %s %s\", \"msg\", \"msg\");",
                         "    Validate.notEmpty(stringArg, \"message %s %s\", \"msg\", \"msg\");",
                         "    Validate.notBlank(stringArg, \"message %s %s\", \"msg\", \"msg\");",
+                        "    Validate.notBlank(stringArg, \"message %s %s\", \"msg\", compileTimeConstant);",
+                        "    Validate.notBlank(stringArg, \"message %s %s\", \"msg\", localConstant);",
```
gracew: this test fails. usage of `compileTimeConstant` passes.

gracew:
fixes #176

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/ValidateConstantMessageTests.java
```diff
@@ -109,8 +109,11 @@ public void negative() throws Exception {
                         "  void f(boolean bArg, int iArg, Object oArg, Integer[] arrayArg, "
                             + "Collection<String> collectionArg, Map<String, String> mapArg, String stringArg, "
                             + "Iterable<String> iterableArg, double dArg) {",
+                        "    String localConstant = \"constant\";",
                         "    Validate.isTrue(bArg, \"message %d\", 123L);",
                         "    Validate.isTrue(bArg, \"message %f\", 0.0);",
+                        "    Validate.notBlank(stringArg, \"message %s\", compileTimeConstant);",
+                        "    Validate.notBlank(stringArg, \"message %s\", localConstant);",
```
uschi2000: these are not really constants, right?

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ValidateConstantMessage.java
```diff
@@ -45,6 +46,19 @@
     private static final Matcher<ExpressionTree> VALIDATE_METHODS =
             MethodMatchers.staticMethod()
                     .onClassAny("org.apache.commons.lang3.Validate", "org.apache.commons.lang.Validate");
+    private static final Matcher<ExpressionTree> VALIDATE_METHODS_MESSAGE_THIRD_PARAM =
+            Matchers.anyOf(
+                    MethodMatchers.staticMethod()
+                            // all org.apache.commons.lang.Validate methods have the message as the second arg
+                            .onClassAny("org.apache.commons.lang3.Validate")
+                            .withNameMatching(
+                                    Pattern.compile("validIndex|matchesPattern|isInstanceOf|isAssignableFrom")));
```
uschi2000: seems like this is missing a lot of test coverage.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ValidateConstantMessage.java
```diff
@@ -45,6 +46,19 @@
     private static final Matcher<ExpressionTree> VALIDATE_METHODS =
             MethodMatchers.staticMethod()
                     .onClassAny("org.apache.commons.lang3.Validate", "org.apache.commons.lang.Validate");
+    private static final Matcher<ExpressionTree> VALIDATE_METHODS_MESSAGE_THIRD_PARAM =
+            Matchers.anyOf(
+                    MethodMatchers.staticMethod()
+                            // all org.apache.commons.lang.Validate methods have the message as the second arg
```
uschi2000: i don't quite understand how this comment relevant here

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ValidateConstantMessage.java
gracew: just had that thought as well, will add some.

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/ValidateConstantMessageTests.java
gracew: what do you mean? 

the arguments to the message template have to be non-compile time constants in order to demonstrate the bug in the previous code.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ValidateConstantMessage.java
gracew: this check runs on methods from both `org.apache.commons.lang.Validate` and`org.apache.commons.lang3.Validate`, but I'm only including certain `lang3` methods here, because all the `lang` methods fall under the "normal" case where the message string is the second arg

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/ValidateConstantMessageTests.java
uschi2000: ok, so they're just variables then? :)

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ValidateConstantMessage.java
uschi2000: oh, i see

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/ValidateConstantMessageTests.java
gracew: haha yes. I switched to just using `stringArg` rather than defining a new local constant

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ValidateConstantMessage.java
gracew: actually there's one, `allElementsOfType`, where the message is the third arg... updated matcher

iamdanfox:
Closing stale PR

PR #178, opened 2017-06-07
apply latest baseline (0.14.0)

gracew:


.baseline/copyright/999_palantir.txt
```diff
@@ -0,0 +1 @@
+Copyright ${today.year} Palantir Technologies, Inc. All rights reserved.
```
gracew, line 1: this file shouldn't have been added =/ 

Excavator should send a PR soon removing it

gracew: removed in #231

PR #186, opened 2017-06-16
Allow static imports of Collections and Collectors

j-baker:
Collectors and Collections are classes that are pleasant to use as static imports. We can avoid boilerplate by allowing their use, and disallowing them is not helpful because the method names aren't likely to be confused with other things.

Example usage

```java
return foos.stream().flatMap(Collection::stream).collect(toList());
```

```java
when(foo.doSomething()).thenReturn(emptyMap());
```


uschi2000:
i'm fine with this. @gracew ?

gracew:
sgtm

PR #192, opened 2017-06-27
don't wrap fields with single annotation

ellisjoe:
Changes style from:
```
@Mock
private Service service;
```
to:
```
@Mock private Service service
```

which we've found to be much nicer in tests.


gradle-baseline-java-config/resources/idea/intellij-java-palantir-style.xml
```diff
@@ -44,6 +44,9 @@
             </value>
           </option>
         </GroovyCodeStyleSettings>
+        <JavaCodeStyleSettings>
```
j-baker: will this affect what IntelliJ and Eclipse do? I don't think it's cool for autoformat to cause failures

j-baker: ah sorry, thought htis wa sa checkstyle rule.

ellisjoe: "Changes style" above should really be "Changes intellij autoformatter". Will make the corresponding change to eclipse

uschi2000:
i'm not a huge fan of this.
- fair amount of code churn (that users will have to reconcile manually)
- behaves different than https://github.com/google/google-java-format (I suspect)
- does the same rule hold for class-level annotations?

would propose to run an internal thread to see if we really do want this.

ellisjoe:
Re google-java-format: looks like they have a test to support this case:
input: https://github.com/google/google-java-format/blob/d360f2bf500198282c1aeb436e85394b7f33541e/core/src/test/resources/com/google/googlejavaformat/java/testdata/B20577626.input
output: https://github.com/google/google-java-format/blob/d360f2bf500198282c1aeb436e85394b7f33541e/core/src/test/resources/com/google/googlejavaformat/java/testdata/B20577626.output

Re code churn: do you have examples of other common field annotations that you expect this change to affect?

Re class-level: I'm happy with class-level annotations on a new line. In particular class-level annotations don't suffer from the same repetitive nature as field-level annotations.

ellisjoe:
Looks like their guide also says field params may go on the same line: https://google.github.io/styleguide/javaguide.html#s4.8.5-annotations

uschi2000:
ah, i got confused between method and field annotations, sorry. i'm fine with this for fields. there isn't any automation around this any longer, unfortunately, but we'll need to verify that eclipse/intellij/checkstyle behave consistently.

ellisjoe:
I've manually verified both intellij and eclipse. I'm fine leaving checkstyle as is, which currently allows annotations on the same line or on their own.

uschi2000 (approved):

PR #195, opened 2017-07-05
Use Java's built-in Objects.requireNonNull instead of Guava's checkNotNull

iamdanfox:
I stumbled across this method earlier and figured it's always nice to reduce dependence on Guava where you can!

### DIFF
--- a/gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
+++ b/gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
@@ -39,6 +39,10 @@
         <property name="format" value="\s+$"/>
         <property name="message" value="Whitespace at end-of-line"/>
     </module>
+    <module name="RegexpSingleline">
+        <property name="format" value="Preconditions\.checkNotNull"/>
+        <property name="message" value="Use Java's built-in Objects.requireNonNull instead of Guava's checkNotNull"/>
+    </module>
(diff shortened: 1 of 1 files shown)


j-baker (changes requested):
I think this is too much effort to fix for too little benefit. If we had a bot that could open all the PRs then fine.

j-baker:
(there are 11k references to this internally)

iamdanfox:
Fair enough - to be honest, I was just planning on doing a massive `sed -i` on our repo.

j-baker:
Yeah, but then you'll get the imports wrong. Like, if there's an upgrade task we can run then that's fine for me - but like, the internal project I used to be on has 400 cases of this, and it's a pain to fix this kind of papercut.

gracew:
re: imports, you can select the entire project in eclipse/intellij and organize imports.

I think that soon (on the order of weeks) we will be able to automate these types of simple refactors with the internal excavator project.

j-baker:
Yep - once we can automate these refactors we can and should do these kinds of changes. But before then, the cost/benefit seems skewed.

j-baker:
(saying this as someone who once submitted a 10k line 'get back up to date with baseline' diff, which although almost totally formed of automated changes did still take a whole weekend).

j-baker:
It's like, fixing this on most repos if your muscle to do so is trained is possibly 10 minutes of work. But once you add 20 more similar changes, it becomes soulcrushing.

PR #196, opened 2017-07-10
allow IO abbreviation

ellisjoe:


j-baker (approved):

uschi2000 (commented):
Don't think we should take this. The current definition of camel case is clear and simple.

j-baker:
Clear and simple, but painful when you have an `IoCacheKey` or a `testHandlesIoExceptionProperly`

uschi2000:
`IoCacheKey` is exactly what I would write intuitively. Really don't set the a precedent for customizations on this one, sorry.

ellisjoe:
I'd argue the JDK set the precedent here. From the docs: http://www.oracle.com/technetwork/java/codeconventions-135099.html

> Class names should be nouns, in mixed case with the first letter of each internal word capitalized. Try to keep your class names simple and descriptive. Use whole words-avoid acronyms and abbreviations (unless the abbreviation is much more widely used than the long form, such as URL or HTML).

uschi2000:
cool, so you should call it `InputOutputCacheKey` then?

uschi2000:
if you feel strongly about this, please put it up for a vote in the internal list. my vote is going to be "no".

j-baker:
And a UniformResourceLocatorCacheKey?

uschi2000:
well, I just followed joe's JDK precedent quote, i'm not saying i'd want that :) i am personally very happy with UrlFoo and IoFoo. If we support IOFoo, then we'll soon also have to support URLFoo etc. the current convention is the verbatim google one: https://google.github.io/styleguide/javaguide.html#s5.3-camel-case .

ellisjoe:
meh `IO` is the only abbreviation that actually bothers me in cases like `testBlah_handlesIOException` or `interface IOFunction<T, R> { R apply(T in) throws IOException }` where I'd like to match `IOException`'s casing. Don't feel like fighting the oracle vs. google naming battle though.

iamdanfox:
Closing out as the current definition is concise and special-casing "IO" opens the door for lots more PRs to add special-casing for other abbreviations that are "more widely used than the long form" such as "HTML", "URL", "IP", "XML" or "JSON".

PR #205, opened 2017-08-07
Update checkstyle 7.5.1 -> 8.1

ashrayjain:
This was changed in Checkstyle 8.1

http://checkstyle.sourceforge.net/releasenotes.html#Release_8.1#Release_8.1
https://github.com/checkstyle/checkstyle/issues/4714

Causes errors like the following in IntelliJ

screen shot 2017-08-07 at 3 47 18 pm

### DIFF
--- a/gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
+++ b/gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
@@ -43,12 +43,6 @@
         <property name="file" value="${config_loc}/checkstyle-suppressions.xml"/>
     </module>
     <module name="SuppressWarningsFilter"/> <!-- baseline-gradle: README.md -->
-    <module name="SuppressionCommentFilter"/> <!-- baseline-gradle: README.md -->
-    <module name="SuppressionCommentFilter">
-        <property name="offCommentFormat" value="CHECKSTYLE.OFF\: ([\w\|]+)"/>
-        <property name="onCommentFormat" value="CHECKSTYLE.ON\: ([\w\|]+)"/>
-        <property name="checkFormat" value="$1"/>
-    </module>
     <module name="TreeWalker">
         <module name="AbbreviationAsWordInName"> <!-- Java Style Guide: Camel case : defined -->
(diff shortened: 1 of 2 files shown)


uschi2000:
Will have to update the checkstyle Gradle tooling too, right?

ashrayjain:
@uschi2000 not sure I follow? I thought updating this file will push this change to people when they upgrade baseline and run `./gradlew baselineUpdateConfig`?

uschi2000:
but don't you need checkstyle 8.1 for this syntax to make sense? the checkstyle version is encoded in the baseline-checkstyle plugin

ashrayjain:
Oh I see. So currently it's set to 7.5.1 here https://github.com/palantir/gradle-baseline/blob/dfc293e7a56d06ed3ad58c280b70367b33f9c29a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineCheckstyle.groovy#L30

Would bumping that to 8.1 be enough?

uschi2000:
There's a pretty big change list between 7.5.1 and 8.1. Did you go through them and verify that (1) we're comfortable with the changes, (2) they align with our Baseline docs, (3) they are compatible with the IntelliJ and Eclipse formatting, (4) they require no other config change?

yiweig:
@ashrayjain fwiw, a quick temporary workaround to suppress the Checkstyle notifications being spammed every 60 seconds (without turning off Checkstyle notifications completely) is to explicitly bump down IntelliJ's Checkstyle version to `8.0`

screen shot 2017-08-23 at 10 14 37 am

ashrayjain:
Yea that is what i did. Sorry I didn't have the bandwidth atm to go through the changes and actually bump baseline to 8.1 for realz. :(

CRogers:
See #212 where I fix the root of the problem.

uschi2000:
0.16.1 will fix this.

uschi2000:
(should still take the 8.x upgrade eventually, so leaving this open for now)

iamdanfox:
Closing out in favour of https://github.com/palantir/gradle-baseline/pull/313

PR #211, opened 2017-09-20
Logsafe errorprone check

scottsj:
Requires `com.palantir.logsafe.arg` for slf4j log arguments:

Failing example:
```
log.warn("Could not find job with ID {}", jobId);
```
Passing example:
```
log.warn("Could not find job with ID {}", Safeargs.of(jobId));
```

I'm curious if anyone knows a way to prevent this check if safe-logging is not on the classpath?

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/Slf4jLogsafeArgs.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/Slf4jLogsafeArgs.java
@@ -40,7 +40,7 @@ import java.util.regex.Pattern;
         category = Category.ONE_OFF,
         link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
         linkType = BugPattern.LinkType.CUSTOM,
-        severity = SeverityLevel.ERROR,
+        severity = SeverityLevel.WARNING,
         summary = "Allow only com.palantir.logsafe.Arg types as parameter inputs to slf4j log messages.")
 public final class Slf4jLogsafeArgs extends BugChecker implements MethodInvocationTreeMatcher {
 
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/Slf4jLogsafeArgs.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/Slf4jLogsafeArgs.java
@@ -78,10 +78,10 @@ public final class Slf4jLogsafeArgs extends BugChecker implements MethodInvocati
 
         if (badArgs.isEmpty()) {
             return Description.NO_MATCH;
+        } else {
+            return buildDescription(tree)
+                    .setMessage("slf4j log statement does not use logsafe parameters for arguments " + badArgs)
+                    .build();
         }
-
-        return buildDescription(tree)
-                .setMessage("slf4j log statement does not use logsafe parameters for arguments " + badArgs)
-                .build();
     }
 }
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/Slf4jLogsafeArgs.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/Slf4jLogsafeArgs.java
@@ -38,6 +38,8 @@ import java.util.regex.Pattern;
 @BugPattern(
         name = "Slf4jLogsafeArgs",
         category = Category.ONE_OFF,
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
         severity = SeverityLevel.ERROR,
         summary = "Allow only com.palantir.logsafe.Arg types as parameter inputs to slf4j log messages.")
 public final class Slf4jLogsafeArgs extends BugChecker implements MethodInvocationTreeMatcher {
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/Slf4jLogsafeArgs.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/Slf4jLogsafeArgs.java
@@ -1,0 +1,87 @@
+/*
+ * Copyright 2017 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.baseline.errorprone;
+
+import com.google.auto.service.AutoService;
+import com.google.common.collect.ImmutableList.Builder;
+import com.google.errorprone.BugPattern;
+import com.google.errorprone.BugPattern.Category;
+import com.google.errorprone.BugPattern.SeverityLevel;
+import com.google.errorprone.VisitorState;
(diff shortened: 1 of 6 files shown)


palantirtech:
Thanks for your interest in [palantir/gradle-baseline](https://github.com/palantir/gradle-baseline), @scottsj! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

scottsj:
Could someone please take a look @gracew, @jmcampanini, @uschi2000?

uschi2000:
I won't have time today, sorry.

Sent from Email+ secured by MobileIron

-------- Original Message --------

scottsj:
TODO: allow compile-time constants

scottsj:
Hi @uschi2000, what about this week?

uschi2000:
This week! Thanks for reminding me.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/Slf4jLogsafeArgs.java
```diff
@@ -0,0 +1,87 @@
...
+        Builder<Integer> badArgsBuilder = new Builder<>();
+        for (int i = startArg; i <= endArg; i++) {
+            if (!ASTHelpers.isCastable(ASTHelpers.getType(allArgs.get(i)),
+                    state.getTypeFromString("com.palantir.logsafe.Arg"), state)) {
+                badArgsBuilder.add(i);
+            }
+        }
+        List<Integer> badArgs = badArgsBuilder.build();
+
+        if (badArgs.isEmpty()) {
```
uschi2000, line 81: prefer symmetry for return cases:
```
Foo foo() {
  if (!precondition) {
    return ...; // early return is OK
  }

  // lots of logic

  // return something depending on logic
  if (something) {
    return ...;
  } else {
    return ...;
  }
}
```

uschi2000:
The code looks OK, but as it stands, I think this is too restrictive to be enforced for all Baseline users. Can discuss three alternatives:
(1) Explicit opt-in to this check through Baseline config
(2) Auto-opt-in when SafeArg is on the classpath
(3) Use WARNING instead of ERROR

Option (2) looks reasonable to me. @gracew ? @jared2501 thoughts?
@scottsj , to implement (2) you can `try` to create an instance of `SafeArg` in order to determine whether it's on the classpath.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/Slf4jLogsafeArgs.java
scottsj: Changed to...
```
        if (badArgs.isEmpty()) {
            return Description.NO_MATCH;
        } else {
            return buildDescription(tree)...
        }
```

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/util/ClasspathUtils.java
```diff
@@ -0,0 +1,42 @@
...
+
+public final class ClasspathUtils {
+
+    private ClasspathUtils() {
+        // no instantiation
+    }
+
+    public static boolean existsInClasspath(String jarNamePrefix) {
+
+        ClassLoader classLoader = ClassLoader.getSystemClassLoader();
```
uschi2000: would prefer to not go down to the classloader level. would the following work?
```
boolean exists(String className) {
  try {
    Class.forName(className);
    return true;
    return false;
  }
}
```

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/Slf4jLogsafeArgs.java
```diff
@@ -0,0 +1,91 @@
...
+import java.util.List;
+import java.util.regex.Pattern;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "Slf4jLogsafeArgs",
+        category = Category.ONE_OFF,
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
+        severity = SeverityLevel.WARNING,
```
uschi2000, line 43: @gracew thoughts on making this WARNING vs ERROR?

scottsj:
(1) I couldn't determine how to do this. BugPattern does not support `SeverityLevel.OFF`, and disabling occurs in `gradle.build` so I didn't know how to ship with this...
```
    tasks.withType(JavaCompile) {
            options.compilerArgs += ['-Xep:Slf4jLogsafeArgs:OFF']
```

(2) I tried to test for Classpath existence (commit: [e2dd4ac](https://github.com/palantir/gradle-baseline/pull/211/commits/e2dd4acffb0c59e53b8e78a1c1e311a79450bb30) but this isn't working on my external project - a match is never returned, I believe because I'm checking the classpath of gradle-baseline, not the project being tested. It's unclear to me how to invoke an instance of `Arg<T>` from the expression tree.

(3) Used WARNING instead of ERROR

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/util/ClasspathUtils.java
scottsj: // sorry I missed this in my recent comment

Neither approach will work as it will call the classpath of baseline.

uschi2000:
I'm fine with this. @gracew ?

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/Slf4jLogsafeArgsTest.java
```diff
@@ -0,0 +1,171 @@
...
+        test("\"constant {} {}\", \"string\", SafeArg.of(\"name\", \"string\")", "[1]");
+
+        // log.<>(Marker, String, Object, Arg<T>)"
+        test("new DummyMarker(), \"constant {} {}\", \"string\", SafeArg.of(\"name\", \"string\")", "[2]");
+
+        // log.<>(Marker, String, Object, Arg<T>, Exception)"
+        test("new DummyMarker(), \"constant {} {}\", \"string\", UnsafeArg.of(\"name\", \"string\"), new Exception()",
+                "[2]");
+
+        // log.<>(String, Object, Arg<T>, Exception)"
```
gracew, line 85: think you meant to include an exception below

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/Slf4jLogsafeArgsTest.java
```diff
@@ -0,0 +1,171 @@
...
+                        "",
+                        "    // log.<>(Marker, String)",
+                        "    log." + logLevel + "(new DummyMarker(), \"constant\");",
+                        "",
+                        "    // log.<>(Marker, String, Arg<T>)",
+                        "    log." + logLevel + "(new DummyMarker(), \"constant {}\",",
+                        "        SafeArg.of(\"name\", \"string\"));",
+                        "",
+                        "    // log.<>(Marker, String, Arg<T>, Arg<T>)",
+                        "    log." + logLevel + "(\"constant {} {}\",",
```
gracew: marker is not used here

gracew (approved):
sweet! lgtm other than some test comments.

jmcampanini:
you can test that the project's classpath by writing a [gradle integration test][1] (different jvm or at least a different classloader) for a project both with and without it on the classpath.

[1]: https://github.com/palantir/gradle-baseline/blob/develop/gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineConfigIntegrationTest.groovy

scottsj:
@uschi2000 seeking your help to rebuild -> approve -> merge -> release.

uschi2000:
@scottsj I don't have permissions to rebuild either :/
You can either push an empty commit, or I can take over this branch into the repository and build.

scottsj:
Rebuild succeeded 👍 

@uschi2000 do you have write access permission in order to Approve?

uschi2000 (approved):

PR #212, opened 2017-09-29
Configure the intellij checkstyle plugin to use the correct checkstyle version

CRogers:
Currently, we do not set the CheckStyle-IDEA version, so it defaults to the latest, 8.1. There was a [breaking change introduced in 8.1](https://github.com/checkstyle/checkstyle/issues/4714) that makes the baseline checkstyle files invalid. This means you get an error message and the checkstyle plugin no longer works in IDEA. It's also sadness to have the IDE and gradle running different checkstyle versions.

screen shot 2017-09-29 at 2 50 12 pm

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
@@ -108,11 +108,13 @@ class BaselineIdea extends AbstractBaselinePlugin {
 
         project.logger.debug "Baseline: Configuring Checkstyle for Idea"
         def checkstyleFile = "LOCAL_FILE:\$PRJ_DIR\$.baseline/checkstyle/checkstyle.xml"
+
         node.append(new XmlParser().parseText("""
             <component name="CheckStyle-IDEA">
               <option name="configuration">
                 <map>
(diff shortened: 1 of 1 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
```diff
@@ -108,11 +108,13 @@ class BaselineIdea extends AbstractBaselinePlugin {
 
         project.logger.debug "Baseline: Configuring Checkstyle for Idea"
         def checkstyleFile = "LOCAL_FILE:\$PRJ_DIR\$.baseline/checkstyle/checkstyle.xml"
+
```
ashrayjain: nit: extra newline?

ashrayjain (approved):
Nice!!

CRogers:
@ashrayjain Any chance of a merge + release?

ashrayjain:
I don't have write access to this repo. That would be @uschi2000 I believe.

uschi2000:
Can take a a look in a little bit. Just getting of a plane

Sent from Email+ secured by MobileIron

-------- Original Message --------

uschi2000 (approved):

uschi2000:
0.16.1 building

PR #220, opened 2017-10-18
allow both old and new copyright styles

gracew:


gracew:
@jmcampanini @uschi2000

jmcampanini:
is the plan to keep both forever? or to use have both temporarily to allow for easier upgrades of our  codebase?

jmcampanini (approved):

uschi2000:
hmm. I think I'd use the Baseline upgrade as a forcing function to adopt the correct/new format?

gracew:
our legal team said it was ok to keep the old format for existing files. I'm fine with removing support for the old format once we have automatically converted everyone's copyrights to the new format, but I feel strongly that we should merge this change for the interim because currently people cannot upgrade to baseline 0.17.0 without converting all their existing files. doing this on a per-repo basis is a waste of time.

j-baker:
Yeah maybe. Like, it's really really really easy to replace all the copyrights.

gatesn:
Someone want to craft some `sed` and paste it here?

gracew:
ok. I think we can automatically convert everyone's copyrights while also upgrading their baselines.

I'll close this and I'm going to release 0.17.1 off of develop so at least there is a release with consistent checkstyle + new file copyrights.

PR #223, opened 2017-10-26
remove comma from copyright

ellisjoe:


uschi2000:
try gradle-baseline-java-config/resources

PR #224, opened 2017-11-01
Todo bracket capture is no longer greedy

felixdesouza:
Failing case:

```
// TODO(felixdesouza): This is a comment (with some more information).
```

By default `.*` is greedy, it's going to match the last parentheses, despite us complying with the rule.

What we want is the `Possessive` quantifier i.e. `.*+`, which will be lazy, **but not** backtrack. 

This matches the semantics we want: TODO followed by a left bracket, then a name, then a right bracket. Only failure mode would be brackets inside the name, but that's nonsensical so I think this change is okay.

### DIFF
--- a/gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
+++ b/gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
@@ -328,7 +328,7 @@
             <property name="message" value="Redundant ''static'' modifier."/>
         </module>
         <module name="RegexpSinglelineJava">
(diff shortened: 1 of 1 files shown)


uschi2000:
I don't think we need to allow `TODO(my full name with spaces): foo`, so maybe we can simplify this?

felixdesouza:
sure, makes sense, I can change

felixdesouza:
I'm not really sure it simplifies tbh, trying to match on having spaces (to trigger) seems orthogonal to the failure mode I'm seeing with the current matchers and if anything it's an *extra* case. I've added it nonetheless.

uschi2000 (approved):

PR #232, opened 2017-11-17
Allow static import of assertj Guava extensions

uschi2000:


uschi2000:
See https://github.com/joel-costigliola/assertj-guava

gracew (approved):

PR #237, opened 2018-01-16
More careful checkstyle regex for TODOs

kyegupov:
The current regex (because of greedy matching) erroneously catches:

`// TODO(dtrump): Make america() return GREAT again`


kyegupov:
How do I request a review? @uschi2000 ?

iamdanfox (approved):

PR #239, opened 2018-02-05
Allow remoting assertions

j-baker:
For same reason as other assertions libraries.


iamdanfox:
This rule has so many exceptions I don't think it's a valid heuristic. @uschi2000 what do you think about relaxing it so that static imports are allowed for tests?

uschi2000:
Allowing for tests sounds good to me

iamdanfox:
Closing in favour of https://github.com/palantir/gradle-baseline/pull/240

PR #242, opened 2018-02-17
Remove TreeMultimap from disallowed types

jared2501:
Since `com.google.common.collect.TreeMultimap` has `NavigableSet` as the `#keySet` return type.

@uschi2000

### DIFF
--- a/.baseline/checkstyle/checkstyle.xml
+++ b/.baseline/checkstyle/checkstyle.xml
@@ -154,7 +154,7 @@
             <property name="message" value="Avoid using corresponding octal or Unicode escape."/>
         </module>
         <module name="IllegalType"> <!-- Java Coding Guide: Limit coupling on concrete classes -->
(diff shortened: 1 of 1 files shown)


palantirtech:
Thanks for your interest in [palantir/gradle-baseline](https://github.com/palantir/gradle-baseline), @jared2501! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

markelliot:
@jared2501 I think you want https://google.github.io/guava/releases/23.0/api/docs/com/google/common/collect/Maps.html#newTreeMap--

jared2501:
Hmm I have:

```
TreeMultimap<Long, Long> foo = TreeMultimap.create();

....

foo.keySet().headSet(val);
```

And that `headSet` method is only available if your Multimap is a TreeMultimap (via a NavigableSet). I.e. here's the code that does the casting:

image

jared2501:
Or are you saying the pattern of requesting a headSet on a TreeMultimap is not encouraged?

markelliot:
ah my bad, yeah; we should drop anything that's actually an interface from this list

markelliot (approved):

PR #240, opened 2018-02-06
Allow static imports in tests

iamdanfox:
Follow-up to https://github.com/palantir/gradle-baseline/pull/239

Rationale is that many testing libraries rely on static imports - maintaining an increasingly large list of exceptions just suggests that banning static imports in test code doesn't make sense in the first place.

PR #243, opened 2018-02-21
suppress test classes that are in a "tests" folder

tusharnarayan:
for example, integration tests that live in `src/integrationTests/java`

noticed because after https://github.com/palantir/gradle-baseline/pull/240, static imports in classes within my `integrationTests` folder were causing checkstyle violations - figured we want to make the `s` optional for test all suppression rules

### DIFF
--- a/gradle-baseline-java-config/resources/checkstyle/checkstyle-suppressions.xml
+++ b/gradle-baseline-java-config/resources/checkstyle/checkstyle-suppressions.xml
@@ -8,10 +8,10 @@
  for your changes to take effect in its Checkstyle integration. -->
 <suppressions>
     <!-- Suppress test classes -->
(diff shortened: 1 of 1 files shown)


tusharnarayan:
@uschi2000 and @iamdanfox for review

iamdanfox:
@tusharnarayan would you consider renaming your source set to `integrationTest` for consistency with the existing singular `test` one?  It seems kinda preferrable to unify on singular names rather than supporting both singular and plural?

tusharnarayan:
yeah i poked around a couple projects first, seems like a number of folks are using `integrationTests` - i'm down to rename if that's something we want to standardize on

iamdanfox:
Cool, I think it makes sense to just match the existing singular noun approach and use `src`, `test`, `integrationTest`/`integTest` and `ete` rather than plurals

PR #249, opened 2018-02-28
Try casting the last logged argument to Throwable instead of Exception

ashrayjain:
This matches the slf4j api which allows one to log a throwable, not just an exception.

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/Slf4jLogsafeArgs.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/Slf4jLogsafeArgs.java
@@ -66,7 +66,7 @@ public final class Slf4jLogsafeArgs extends BugChecker implements MethodInvocati
                 state) ? 2 : 1;
         int endArg = ASTHelpers.isCastable(
                 ASTHelpers.getType(allArgs.get(lastIndex)),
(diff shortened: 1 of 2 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/Slf4jLogsafeArgs.java
```diff
@@ -66,7 +66,7 @@ public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState
                 state) ? 2 : 1;
         int endArg = ASTHelpers.isCastable(
                 ASTHelpers.getType(allArgs.get(lastIndex)),
-                state.getTypeFromString("java.lang.Exception"),
+                state.getTypeFromString("java.lang.Throwable"),
```
uschi2000, line 69: can you add a test (or change an existing one) to cover this?

ashrayjain: Sure

ashrayjain: Done

uschi2000 (approved):

PR #250, opened 2018-02-28
Clarify that PreconditionsConstantMessage allows for a template string

TheBeruriahIncident:
From the current error message, I thought it truly wanted a constant, since I didn't realize Preconditions supported templates

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreconditionsConstantMessage.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreconditionsConstantMessage.java
@@ -69,6 +69,6 @@ public final class PreconditionsConstantMessage extends BugChecker implements Bu
         }
 
         return buildDescription(tree).setMessage(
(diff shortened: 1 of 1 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreconditionsConstantMessage.java
```diff
@@ -69,6 +69,6 @@ public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState
         }
 
         return buildDescription(tree).setMessage(
-                "Preconditions.checkX() statement uses a non-constant message").build();
+                "Preconditions.checkX() statement uses a non-constant message. Consider using a template string").build();
```
uschi2000: `Consider using a template string with '%s'.`

TheBeruriahIncident: Done, thanks!

uschi2000 (approved):

PR #254, opened 2018-03-15
Require that Files#list appear within a try-with-resources block

gatesn:
It's a common mistake to use `Files#list` without closing the returned Stream. This check ensures that this method is only ever used inside a try-with-resources block and therefore the stream is auto-closed.

We could one day be more lenient here and allow it to be outside iff the `close` method is called on the returned value. But that's harder to implement and perhaps not necessary initially.

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/UnclosedFilesStreamUsage.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/UnclosedFilesStreamUsage.java
@@ -38,7 +38,7 @@ import java.util.regex.Pattern;
         name = "UnclosedFilesStreamUsage",
         category = Category.ONE_OFF,
         severity = SeverityLevel.ERROR,
-        summary = "Ensure a stream returned by java.nio.file.Files#{list,walk} "
+        summary = "Ensure a stream returned by java.nio.file.Files#{find,lines,list,walk} "
                 + "is closed to prevent leaking file descriptors.")
 public final class UnclosedFilesStreamUsage extends BugChecker implements BugChecker.MethodInvocationTreeMatcher {
 
@@ -46,7 +46,7 @@ public final class UnclosedFilesStreamUsage extends BugChecker implements BugChe
 
     private static final MethodMatchers.MethodNameMatcher filesStreamMatcher = MethodMatchers.staticMethod()
             .onClass("java.nio.file.Files")
-            .withNameMatching(Pattern.compile("list|walk"));
+            .withNameMatching(Pattern.compile("find|lines|list|walk"));
 
     private static final Matcher<Tree> tryResourcesMatcher = (Matcher<Tree>)
             (methodInvocationTree, state) -> Optional.ofNullable(state.findEnclosing(TryTree.class))
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/UnclosedFilesStreamUsage.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/UnclosedFilesStreamUsage.java
@@ -73,7 +73,8 @@ public final class UnclosedFilesStreamUsage extends BugChecker implements BugChe
             return Description.NO_MATCH;
         }
         return buildDescription(tree)
-                .setMessage("java.nio.file.Files must be called within a try-with-resources block")
+                .setMessage("Methods returning a Stream<> in java.nio.file.Files must be called "
+                        + "within a try-with-resources block")
                 .build();
     }
 
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/UnclosedFilesStreamUsage.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/UnclosedFilesStreamUsage.java
@@ -35,17 +35,17 @@ import java.util.regex.Pattern;
 
 @AutoService(BugChecker.class)
 @BugPattern(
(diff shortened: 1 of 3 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/UnclosedFilesStreamUsage.java
```diff
@@ -0,0 +1,79 @@
...
+import com.sun.source.util.TreePath;
+import java.util.Optional;
+import java.util.regex.Pattern;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "UnclosedFilesStreamUsage",
+        category = Category.ONE_OFF,
+        severity = SeverityLevel.ERROR,
+        summary = "Ensure a stream returned by java.nio.file.Files#{list,walk} is closed.")
```
iamdanfox: What about "Ensure a stream returned by java.nio.file.Files#{list,walk} is closed to prevent leaking file descriptors"?

justinuang (approved):

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/UnclosedFilesStreamUsage.java
```diff
@@ -0,0 +1,79 @@
...
+public final class UnclosedFilesStreamUsage extends BugChecker implements BugChecker.MethodInvocationTreeMatcher {
+
+    private static final long serialVersionUID = 1L;
+
+    private static final MethodMatchers.MethodNameMatcher filesStreamMatcher = MethodMatchers.staticMethod()
+            .onClass("java.nio.file.Files")
+            .withNameMatching(Pattern.compile("list|walk"));
```
schlosna: Probably also want to check for Files.find and Files.lines

https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#find-java.nio.file.Path-int-java.util.function.BiPredicate-java.nio.file.FileVisitOption...-

https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#lines-java.nio.file.Path-

ashrayjain: Also Files.walk

gatesn: I couldn't find a way to match on static methods with return type "Stream" - do you know of one?

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/UnclosedFilesStreamUsage.java
```diff
@@ -0,0 +1,79 @@
...
+        return Optional.empty();
+    }
+
+    @Override
+    public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
+        if (!Matchers.allOf(filesStreamMatcher, Matchers.not(tryResourcesMatcher)).matches(tree, state)) {
+            return Description.NO_MATCH;
+        }
+        return buildDescription(tree)
+                .setMessage("java.nio.file.Files must be called within a try-with-resources block")
```
uschi2000: Methods returning a Stream<> in java.nio.file.Files must ...

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/UnclosedFilesListUsageTests.java
```diff
@@ -0,0 +1,89 @@
...
+        compilationHelper
+                .addSourceLines(
+                        "Test.java",
+                        "import java.nio.file.Files;",
+                        "import java.nio.file.Path;",
+                        "import java.nio.file.Paths;",
+                        "import java.util.stream.Stream;",
+                        "class Test {",
+                        "  void f(String param) {",
+                        "    try (Stream<Path> files = Files.list(Paths.get(\"/tmp\"))) {",
```
uschi2000, line 108: how about this case:
```
try (Writer writer = new MyWriter()) {
  Stream<Path> files = Files.list(...);
  files.forEach(writer.write(f));
} catch (IOException e) {}
```

gatesn: The Tree in try's resources cannot be reached from the `Files.list` method invocation, so it still correctly warns.  Added a test case for it too

uschi2000 (approved):

uschi2000:
Thanks!

ashrayjain:
Does this handle using Files.{list,walk,etc} inside streams using `flatMap` correctly? ie, does it alert in that case, because it shouldn't since `flatMap` actually does close the underlying stream.

From https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html#flatMap-java.util.function.Function-

```Each mapped stream is closed after its contents have been placed into this stream.```

**Comment by gatesn** [comment]
@ashrayjain no it doesn't. I do acknowledge it's not an ideal check and is overly strict so open to improving it over time. But how do you deal with things like:
```
Stream<Path> files = Files.list(...);
return otherStream.flatMap(files);
```

uschi2000:
I'm happy to outlaw this. Leaking resources through API boundaries isn't really something we need to endorse.

gatesn:
Sooooo turns out this is built-in! http://errorprone.info/bugpattern/StreamResourceLeak

Filed here: https://github.com/palantir/gradle-baseline/issues/271

PR #255, opened 2018-03-16
Added BaselineClasspathConflict plugin

carterkozak:


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineClasspathConflict.groovy
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * Adds a simple test for shadowed classes.
+ */
+class BaselineClasspathConflict extends AbstractBaselinePlugin  {
+
+    @Override
+    void apply(Project project) {
+        if (project.plugins.hasPlugin(JavaPlugin)) {
+            def silentConflict = project.tasks.create("silentConflict") {
+                dependsOn: project.configurations.testRuntime
```
uschi2000, line 35: testRuntime?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineClasspathConflict.groovy
```diff
@@ -0,0 +1,62 @@
...
+                doLast {
+                    SetMultimap<String, File> classToJarMap = HashMultimap.create()
+                    project.configurations.testRuntime.getResolvedConfiguration().getFiles().each { File jarFile ->
+                        new JarFile(jarFile).entries()
+                                // Only check class uniquness for now
+                                .find({JarEntry entry -> entry.getName().endsWith(".class")})
+                                .each {JarEntry entry ->
+                            classToJarMap.put(entry.getName(), jarFile)
```
uschi2000, line 43: nit: indentation is a little off

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineClasspathConflict.groovy
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * Adds a simple test for shadowed classes.
+ */
+class BaselineClasspathConflict extends AbstractBaselinePlugin  {
+
+    @Override
+    void apply(Project project) {
+        if (project.plugins.hasPlugin(JavaPlugin)) {
+            def silentConflict = project.tasks.create("silentConflict") {
```
uschi2000, line 34: maybe call this something like `checkClasspathIsDuplicateFree`?

gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineClasspathConflictTest.groovy
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+package com.palantir.baseline
+
+import com.palantir.baseline.plugins.BaselineClasspathConflict
+import org.gradle.api.Project
+import org.gradle.testfixtures.ProjectBuilder
+import spock.lang.Specification
+
+class BaselineClasspathConflictTest extends Specification {
```
markelliot, line 24: ideally the tests would show that this plugin finds a conflict when there is one and does not when there is not one

uschi2000 (commented):
Code looks simple enough, but I think we'll need some tests.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineClasspathConflict.groovy
uschi2000: think we should set the default, but let plugin users specify the configuration (possibly multiple)

gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineClasspathConflictTest.groovy
iamdanfox: I believe there are identically named classes in the following jars:

- https://mvnrepository.com/artifact/javax.el/javax.el-api/3.0.0
- https://mvnrepository.com/artifact/javax.servlet.jsp/jsp-api/2.1

e.g. `javax/el/ArrayELResolver.class`

berler:
Would be really nice to have this. For example I just spent most of a day debugging an issue that would be caught by this. There are several cases where open source libraries change ownership and end up being published with multiple versions under different groups. Gradle version pinning doesn't prevent you from ending up with both libs (since they have different groups), and at runtime you can error when it decides to load the "wrong" version of the class. As a concrete example this happens with the feign library which is published under com.netflix.feign and the newer io.github.openfeign. If you ever end up with both of this on your classpath, there is a good chance you will have problems (and might not realize it until you get unlucky with class loading).

iamdanfox:
Closing in favour of https://github.com/palantir/gradle-baseline/pull/267 - gonna address comments and add tests there.

PR #256, opened 2018-03-26
Include check for logsafe Preconditions

cbrockington:
Safe loggable Preconditions should also (if not more so) have a constant message since the message is intended to be safely loggable.

https://github.com/palantir/safe-logging/pull/28

### DIFF
--- a/.baseline/checkstyle/checkstyle.xml
+++ b/.baseline/checkstyle/checkstyle.xml
@@ -297,11 +297,6 @@
             <property name="message" value="Use ? rather than ? extends Object."/>
         </module>
         <module name="RegexpSinglelineJava">
-            <property name="format" value="\bPreconditions\.check.*\{\}"/>
-            <property name="message" value="Use printf-style formatting in Preconditions, not '{}' style formatting."/>
-            <property name="ignoreComments" value="true"/>
-        </module>
-        <module name="RegexpSinglelineJava">
             <property name="format" value="(?i)log(ger)?\.(debug|info|warn|error)\(.*%[sd]"/>
             <property name="message" value="SLF4J loggers support '{}' style formatting."/>
             <property name="ignoreComments" value="true"/>
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/GuavaPreconditionsMessageFormat.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/GuavaPreconditionsMessageFormat.java
@@ -1,5 +1,5 @@
 /*
- * (c) Copyright 2017 Palantir Technologies Inc. All rights reserved.
+ * (c) Copyright 2018 Palantir Technologies Inc. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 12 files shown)


gracew:
can you add a test case for this? let's also add `com.palantir.logsafe.Preconditions` to the checkstyle exclusion for AvoidStaticImport

cbrockington:
Just updating on this, added tests and realised there are some more changes needed as the RegexpSingleline check for "do you have {} on the same line as the word Preconditions" is no longer satisfactory, since Guava Preconditions and LogSafe Preconditions use different formatting.

Fixing that up, will let you know when it's done.

cbrockington:
OK, this should now work properly.  Most of the changes are splitting out the RegexpSinglelineJava Preconditions formatting check into two checks: `GuavaPreconditionsMessageFormat` and `LogSafePreconditionsMessageFormat`, since a simple regex can not determine the difference between an imported guava Preconditions vs logsafe Preconditions (and they have opposite formatting expectations.

.baseline/checkstyle/checkstyle.xml
```diff
@@ -296,11 +296,6 @@
             <property name="format" value="\? extends Object\W"/>
             <property name="message" value="Use ? rather than ? extends Object."/>
         </module>
-        <module name="RegexpSinglelineJava">
```
uschi2000, line 299: I don't think we should remove this. It catches many bugs that vanilla Guava users produce.

uschi2000: oh, you replaced it with error-prone. got it.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/GuavaPreconditionsMessageFormat.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    @Override
+    protected Description matchMessageFormat(MethodInvocationTree tree, String message, VisitorState state) {
+        if (!message.contains("{}")) {
+            return Description.NO_MATCH;
+        }
+
+        return buildDescription(tree).setMessage(
+                "Use printf-style formatting in Preconditions, not '{}' style formatting.").build();
```
uschi2000: "in Guava Preconditions"

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/LogSafePreconditionsMessageFormat.java
```diff
@@ -0,0 +1,65 @@
...
+import java.util.regex.Pattern;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "LogSafePreconditionsMessageFormat",
+        category = Category.ONE_OFF,
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = LinkType.CUSTOM,
+        severity = SeverityLevel.ERROR,
+        summary = "logsafe Preconditions.checkX() methods must have '{}' style formatting.")
```
uschi2000: @gracew why do we accept parameters at all?

iamdanfox: Seems like params are just more helpful information that devs can use to try to figure out what went wrong when a stack trace appears in their low-side logs...

why would we not want them?

uschi2000: we don't inline parameters for slf4j log calls, so we would we support them here?

iamdanfox: Oh right, you mean just getting rid of the `{}`... yeah that seems fine to me!  As long as people can keep tacking on values to the varargs call and they end up on disk somehow :)

cbrockington: The logsafe preconditions doesn't inline them either, but I do think that even if they aren't actually substituted they can be useful (particularly when named) as they are a visual indicator "you should look in the params block if this is important to you", without them you don't necessarily know if a noun is just a noun, or a placeholder for a param.

cbrockington: OK, I changed the check to merely say what it was actually doing anyway, which is flagging you should not be using print-f style formatting, regardless of using any braces.

Is that fair?

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/LogSafePreconditionsMessageFormat.java
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    @Override
+    protected Description matchMessageFormat(MethodInvocationTree tree, String message, VisitorState state) {
+        if (!message.contains("%s")) {
+            return Description.NO_MATCH;
+        }
+
+        return buildDescription(tree).setMessage(
+                "Do not use printf-style formatting in logsafe Preconditions.").build();
```
iamdanfox, line 63: ...use further `SafeArg` or `UnsafeArg` arguments instead?

uschi2000 (approved):

uschi2000:
cheers, @cbrockington

PR #260, opened 2018-04-18
Reference issues, not authors, in TODO blocks

alicederyn:
Authors tend to change projects frequently at Palantir, leaving a bunch of TODOs behind with their name. We've found it useful to get TODOs linked to issues, so they can be more easily tracked, explained and discussed.

I _think_ the existing regex allows this format, so it's just a messaging change.


iamdanfox:
In the spirit of not keeping PRs floating forever, I'm of the disposition to merge this since it doesn't actually change any regexes, just recommends a new approach going forward (and people can always ignore this messaging and do the old thing).

@uschi2000 let me know if you've got serious reservations!

iamdanfox (approved):

PR #261, opened 2018-04-23
Allow empty lambdas

gregakinman:
Only thing here is that it requires checkstyle 6.14 since that's when this property was introduced.


gregakinman:
@uschi2000 for thoughts

uschi2000 (approved):

PR #266, opened 2018-04-26
Simplify format of TODO regex

cbrockington:
* Rather than try to invididually describe every incorrect case (which actually
   misses cases), use a negative look-ahead to match against anything which isn't
   correct.  A single OR is still needed as anything without a space between the
   comment markers and the TODO is incorrect regardless of what proceeds it.


iamdanfox:
I could't get this to work on https://regex101.com/... could you screenshot some valid cases?

cbrockington:
Did you copy-paste it right?

Every case apart from the correct one matches for me
image

iamdanfox (approved):

PR #267, opened 2018-04-30
Plugin to validate all classes on the classpath are uniquely named

iamdanfox:
## Usage

For now, I'm suggesting people explicitly opt-in to this plugin.  It adds a new task called `checkClassUniqueness` which runs as a dependency of `check`.  This task should fail if any jars in the specified classpath happen to contain identically named classes (ie same package and class name).

```gradle
apply plugin: 'com.palantir.baseline-class-uniqueness'
```

## Example output

```
> Task :checkClassUniqueness FAILED
26 Identically named classes with differing impls found in [javax.servlet.jsp:jsp-api:2.1, javax.el:javax.el-api:3.0.0]: [javax.el.ExpressionFactory, javax.el.BeanELResolver, javax.el.CompositeELResolver$CompositeIterator, javax.el.ELContextListener, javax.el.MethodExpression, javax.el.PropertyNotWritableException, javax.el.FunctionMapper, javax.el.MethodInfo, javax.el.BeanELResolver$BeanProperty, javax.el.ELException, javax.el.MapELResolver, javax.el.ArrayELResolver, javax.el.CompositeELResolver, javax.el.ListELResolver, javax.el.Expression, javax.el.ELUtil, javax.el.BeanELResolver$BeanProperties, javax.el.ELResolver, javax.el.MethodNotFoundException, javax.el.ELUtil$1, javax.el.ValueExpression, javax.el.ELContextEvent, javax.el.ResourceBundleELResolver, javax.el.ELContext, javax.el.PropertyNotFoundException, javax.el.VariableMapper]

* What went wrong:
Execution failed for task ':checkClassUniqueness'.
> 'runtime' contains multiple copies of identically named classes - this may cause different runtime behaviour depending on classpath ordering.
  To resolve this, try excluding one of the following jars:
  
  	(26 classes)  javax.servlet.jsp:jsp-api:2.1 javax.el:javax.el-api:3.0.0  
```

Followup to https://github.com/palantir/gradle-baseline/pull/255.

## Future work

- More speed
- handle `project(':foo')` dependencies

_Note: I intentionally didn't include 'classpath' in the name because java10 replaces the concept of a 'classpath' with a 'modulepath' and I think this plugin should be able to cover both._


carterkozak (approved):
lgtm.
May need to aded a configuration block for allowed duplicates, there are some instances that can be difficult to deduplicate. If we can get away with not providing that functionality, that's likely much better.

iamdanfox:
I think this features actually needs more work - I just ran this on a bigger project and got a pretty intimidating failure. I think the summary needs more information. Also unclear how the singleton set `[org.apache.xmlbeans:xmlbeans:2.6.0]` appeared in this list:

EDIT, updated sorted table:

```
'testRuntime' contains multiple copies of identically named classes - this may cause different runtime behaviour depending on classpath ordering.
  To resolve this, try excluding one of the following jars, changing a version or shadowing:
  
        (4 classes)   com.google.code.findbugs:annotations:3.0.0        com.github.stephenc.jcip:jcip-annotations:1.0-1   
        (35 classes)  com.google.code.findbugs:annotations:3.0.0        com.google.code.findbugs:jsr305:3.0.0             
        (1 classes)   com.palantir.tritium:tritium-api:0.9.0            com.palantir.tritium:tritium-core:0.9.0           
        (18 classes)  com.sun.jersey:jersey-server:1.9                  org.glassfish.jersey.core:jersey-server:2.25.1    
        (6 classes)   commons-logging:commons-logging:1.2               org.slf4j:jcl-over-slf4j:1.7.25                   
        (26 classes)  javax.servlet.jsp:jsp-api:2.1                     org.glassfish:javax.el:3.0.0                      
        (29 classes)  log4j:log4j:1.2.17                                org.slf4j:log4j-over-slf4j:1.7.25                 
        (8 classes)   org.apache.xmlbeans:xmlbeans:2.6.0                
        (6 classes)   org.glassfish.hk2.external:javax.inject:2.5.0-b32 javax.inject:javax.inject:1                       
        (35 classes)  org.hamcrest:hamcrest-all:1.3                     org.hamcrest:hamcrest-library:1.3                 
        (45 classes)  org.hamcrest:hamcrest-core:1.3                    org.hamcrest:hamcrest-all:1.3     
```

UPDATE 2: I think some of these are actually false positives because the duplicated classes might be identical. For example, the following two jars both contain this guy:

- com.google.code.findbugs:annotations:3.0.0
- com.google.code.findbugs:jsr305:3.0.0

```java
//
// Source code recreated from a .class file by IntelliJ IDEA
// (powered by Fernflower decompiler)
//

package javax.annotation;

import java.lang.annotation.Documented;
import java.lang.annotation.Retention;
import java.lang.annotation.RetentionPolicy;
import javax.annotation.meta.TypeQualifier;
import javax.annotation.meta.When;

@Documented
@TypeQualifier(
    applicableTo = CharSequence.class
)
@Retention(RetentionPolicy.RUNTIME)
public @interface Syntax {
    String value();

    When when() default When.ALWAYS;
}
```

gracew:
is there a reason we can't use https://github.com/nebula-plugins/gradle-lint-plugin/wiki/Duplicate-Classes-Rule instead of implementing this ourselves?

README.md
```diff
@@ -255,6 +256,13 @@ checks](https://errorprone.info):
 - Slf4jLogsafeArgs: Allow only com.palantir.logsafe.Arg types as parameter inputs to slf4j log messages. More information on
 Safe Logging can be found at [github.com/palantir/safe-logging](https://github.com/palantir/safe-logging).
 
+
+### Class Uniqueness Plugin (com.palantir.baseline-class-uniqueness)
+
+Run `./gradlew checkClassUniqueness` to scan all jars on the `testRuntime` classpath for identically named classes.
```
gracew: the plugin configures the task with the `runtime` configuration

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineClassUniquenessPlugin.java
```diff
@@ -0,0 +1,34 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.baseline.plugins;
+
+import com.palantir.baseline.tasks.CheckClassUniquenessTask;
+import org.gradle.api.Project;
+
+@SuppressWarnings("checkstyle:designforextension") // making this 'final' breaks gradle
```
gracew: you can make the `apply` method final and get rid of this suppression

gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckClassUniquenessTask.java
```diff
@@ -0,0 +1,120 @@
...
+import java.util.TreeMap;
+import java.util.stream.Collectors;
+import org.gradle.api.DefaultTask;
+import org.gradle.api.artifacts.Configuration;
+import org.gradle.api.artifacts.ModuleVersionIdentifier;
+import org.gradle.api.tasks.Input;
+import org.gradle.api.tasks.OutputFile;
+import org.gradle.api.tasks.TaskAction;
+
+@SuppressWarnings("checkstyle:designforextension") // making this 'final' breaks gradle
```
gracew: again, should be able to make the individual methods final. also, maybe include some javadoc mentioning the gradle lint rule and differences in our implementation (e.g. hashing of class files)?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckClassUniquenessTask.java
```diff
@@ -0,0 +1,120 @@
...
+                Set<String> classes = analyzer.getDuplicateClassesInProblemJars(problemJars);
+                getLogger().error("Identically named classes found in {} jars ({}): {}",
+                        problemJars.size(), problemJars, classes);
+            });
+
+            throw new IllegalStateException(String.format(
+                    "'%s' contains multiple copies of identically named classes - "
+                            + "this may cause different runtime behaviour depending on classpath ordering.\n"
+                            + "To resolve this, try excluding one of the following jars, "
+                            + "changing a version or shadowing:\n\n%s",
```
gracew: I don't really want to encourage changing a version or shadowing =/

gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckClassUniquenessTask.java
```diff
@@ -0,0 +1,120 @@
...
+                TreeMap::new));
+
+        StringBuilder builder = new StringBuilder();
+        sortedTable.forEach((jars, classes) ->
+                builder.append(String.format("\t%-14s", "(" + classes + " classes) ") + jars + "\n"));
+        return builder.toString();
+    }
+
+    /**
+     * This only exists to convince gradle this task is incremental.
```
gracew, line 103: don't think this comment is strictly necessary -- I feel this is pretty standard/well understood for people who write gradle plugins

gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckClassUniquenessTask.java
```diff
@@ -0,0 +1,120 @@
...
+                entry -> entry.getKey().stream().map(jar -> String.format(format, jar)).collect(Collectors.joining()),
+                entry -> entry.getValue().size(),
+                (first, second) -> {
+                    throw new RuntimeException("Unexpected collision: " + first + ", " + second);
+                },
+                TreeMap::new));
+
+        StringBuilder builder = new StringBuilder();
+        sortedTable.forEach((jars, classes) ->
+                builder.append(String.format("\t%-14s", "(" + classes + " classes) ") + jars + "\n"));
```
gracew: why not include "(" and ")" within the string.format template?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/ClassUniquenessAnalyzer.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    public void analyzeConfiguration(Configuration configuration) {
+        Instant before = Instant.now();
+
+        Set<ResolvedArtifact> dependencies = configuration
+                .getResolvedConfiguration()
+                .getResolvedArtifacts();
+
+        Map<String, Set<ModuleVersionIdentifier>> tempClassToJarMap = new HashMap<>();
```
gracew: nit, the other one is called classToJarsMap (plural jars)

gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/ClassUniquenessAnalyzer.java
```diff
@@ -0,0 +1,113 @@
...
+                    if (!jarEntry.getName().endsWith(".class")) {
+                        continue;
+                    }
+
+                    multiMapPut(tempClassToJarMap,
+                            jarEntry.getName().replaceAll("/", ".").replaceAll(".class", ""),
+                            resolvedArtifact.getModuleVersion().getId());
+                }
+            } catch (Exception e) {
+                log.error("Failed to read JarFile {}", resolvedArtifact, e);
```
gracew: the resolvedArtifact might not be a jar -- we should probs filter those out first, a la https://github.com/nebula-plugins/gradle-lint-plugin/blob/master/src/main/groovy/com/netflix/nebula/lint/rule/dependency/DependencyService.groovy#L92-L94

gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/ClassUniquenessAnalyzer.java
```diff
@@ -0,0 +1,113 @@
...
+                throw new RuntimeException(e);
+            }
+        });
+
+        tempClassToJarMap.entrySet().stream()
+                .filter(entry -> entry.getValue().size() > 1)
+                .forEach(entry -> {
+                    // add to the top level map
+                    entry.getValue().forEach(value -> multiMapPut(classToJarsMap, entry.getKey(), value));
```
gracew: not sure why we need to save off this collection if we're not returning it anywhere -- getProblemJars could be implemented with jarsToClasses.keySet(), i believe

gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineClassUniquenessPluginIntegrationTest.groovy
```diff
@@ -0,0 +1,154 @@
...
+        then:
+        def result = with('check', '--stacktrace').build()
+        result.task(':checkClassUniqueness').outcome == TaskOutcome.SUCCESS
+    }
+
+    def 'detect duplicates in two external jars'() {
+        when:
+        buildFile << standardBuildFile
+        buildFile << """
+        dependencies {
```
gracew, line 53: i'd prefer to trim this test case down to a more minimal one

gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineClassUniquenessPluginIntegrationTest.groovy
```diff
@@ -0,0 +1,154 @@
...
+            compile 'com.palantir.tritium:tritium-api:0.9.0'
+            compile 'com.palantir.tritium:tritium-core:0.9.0'
+        }   
+        """.stripIndent()
+        BuildResult result = with('checkClassUniqueness').buildAndFail()
+
+        then:
+        result.getOutput().contains("Identically named classes found in 2 jars ([javax.servlet.jsp:jsp-api:2.1, javax.el:javax.el-api:3.0.0]): [javax.")
+        result.getOutput().contains("'runtime' contains multiple copies of identically named classes")
+        result.getOutput().contains("(4 classes)   com.google.code.findbugs:annotations:3.0.1      net.jcip:jcip-annotations:1.0                   com.github.stephenc.jcip:jcip-annotations:1.0-1");
```
gracew: I feel like people would want to know the class names that are duplicated -- thoughts on listing that out? (it would be more verbose =/)

gracew:
talked offline, one benefit cited for this approach over the nebula lint rule is that this PR hashes class files, not erroring in cases where identical classes are found on the classpath (this is considered unhygienic but isn't actively harmful). 

however, after reading through the PR I'm not sure where that hashing of class file content is happening -- ClassUniquenessAnalyzer only seems to handle jar/class names?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckClassUniquenessTask.java
iamdanfox: The `%-14s` handles right-padding with spaces, so I want to supply the fully parenthesised numbers as input and have the padding appended after the last paren.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckClassUniquenessTask.java
iamdanfox: I think it's helpful to reassure readers of this code that the file we write really is inconsequential and not magically wired to something elsewhere in the project.

gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineClassUniquenessPluginIntegrationTest.groovy
iamdanfox: Lots of entries here allows me to verify the sorted table works nicely.

EDIT have thrown away the complicated sorting and split these unit tests.

gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineClassUniquenessPluginIntegrationTest.groovy
iamdanfox: These appear in log lines (at ERROR) level, so they'll appear in the console

gradle-baseline-java/build.gradle
```diff
@@ -10,6 +10,7 @@ repositories {
 dependencies {
     compile gradleApi()
     compile 'net.ltgt.gradle:gradle-errorprone-plugin'
+    compile 'com.google.guava:guava'
```
gracew, line 13: let's remove the testCompile dep in line 17

gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineClassUniquenessPluginIntegrationTest.groovy
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+
+    def 'ignores duplicates when the implementations are identical'() {
+        when:
+        buildFile << standardBuildFile
+        buildFile << """
+        dependencies {
+            compile 'com.palantir.tritium:tritium-api:0.9.0'
+            compile 'com.palantir.tritium:tritium-core:0.9.0'
```
gracew, line 74: what are the duplicate classes here? i'd expect tritium-core to pull in tritium-api and therefore not have any duplicate classes

gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/ClassUniquenessAnalyzer.java
```diff
@@ -0,0 +1,155 @@
...
+                .getResolvedArtifacts();
+
+        // we use these temporary maps to accumulate information as we process each jar,
+        // so they may include singletons which we filter out later
+        Map<String, Set<ModuleVersionIdentifier>> classToJars = new HashMap<>();
+        Map<String, Set<HashCode>> tempClassToHashCodes = new HashMap<>();
+
+        dependencies.stream().forEach(resolvedArtifact -> {
+            File file = resolvedArtifact.getFile();
+            if (!file.exists()) {
```
gracew, line 66: probably wanna filter out non-jars too to avoid spurious errors in line 92

gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/ClassUniquenessAnalyzer.java
```diff
@@ -0,0 +1,155 @@
...
+                            inputStream.hash());
+                }
+            } catch (IOException e) {
+                log.error("Failed to read JarFile {}", resolvedArtifact, e);
+                throw new RuntimeException(e);
+            }
+        });
+
+        // discard all the classes that only come from one jar - these are completely safe!
+        classToJars.entrySet().stream()
```
gracew, line 98: don't think we need this collection -- finding which entries of `tempClassToHashCodes` have value of size > 1 (which we're doing a few lines down) should be sufficient

gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineClassUniquenessPluginIntegrationTest.groovy
iamdanfox: Just a package-info.java

gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/ClassUniquenessAnalyzer.java
iamdanfox: If we delete `classToJars` then I don't really see how we can accumulate information about where a class came from and then derive the `Map<Set<ModuleVersionIdentifier>, Set<String>> jarsToClasses` we need to report problems?

People need to know which are the offending jars!

gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/ClassUniquenessAnalyzer.java
iamdanfox: TBH I'd kinda like to just get an MVP released and get a few repos to pick it up and see what we encounter!  The `-sources.jar` and `-javadoc.jar` thing doesn't seem to cause a problem because we only consider files ending in `.class`.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/ClassUniquenessAnalyzer.java
gracew: oh whoops I didn't read this closely enough

gracew (approved):

PR #268, opened 2018-04-30
Upgrade tooling

iamdanfox:
- switch to circle2
- actually collect test reports
- parallelize docs vs gradle stuff
- cache dependencies
- upgrade nebula-test

No changes to prod code, just trying to make it easier to contribute to this repo!


.circleci/config.yml
```diff
@@ -0,0 +1,54 @@
+version: 2
+jobs:
+
+  build:
+    docker:
+      - image: circleci/openjdk:8-node
```
gracew, line 6: can we use a plain ol' java image for this job? (i.e. no node)

.circleci/config.yml
```diff
@@ -0,0 +1,54 @@
...
+          key: gradle-wrapper-{{ checksum "gradle/wrapper/gradle-wrapper.properties" }}
+          paths:
+            - ~/.gradle/wrapper
+      - save_cache:
+          key: gradle-cache-{{ checksum "versions.props" }}-{{ checksum "build.gradle" }}
+          paths:
+            - ~/.gradle/caches
+
+      - run: ./gradlew publishToMavenLocal --parallel  # ensure artifacts are available to tests
+      - run: ./gradlew clean build
```
gracew, line 27: `--parallel`? and why do we need the clean

gradle/java.gradle
```diff
@@ -6,10 +6,6 @@ apply plugin: 'com.palantir.baseline-idea'
 
 sourceCompatibility = 1.8
 
-checkstyle {
-    configDir = rootProject.file(".baseline")
-}
-
-if (System.env.CIRCLE_TEST_REPORTS) {
-    test.reports.junitXml.destination = new File(System.env.CIRCLE_TEST_REPORTS, project.getName())
+if (System.env.CI) {
```
gracew, line 9: I prefer to have something like the following in the circle config.yml so the magic path only shows up in one place 
```
       - run:
          command: |
            mkdir -p ~/junit
            find . -type f -regex ".*/build/.*TEST.*xml" -exec cp {} ~/junit/ \;
          when: always
      - store_test_results:
          path: ~/junit 
```

gracew (approved):
looks good, comments are non-blocking

.circleci/config.yml
iamdanfox: So weirdly circleci don't offer a pre-built image with java but no node: https://circleci.com/docs/2.0/circleci-images/

I figured by using the official circleci ones we're more likely to be cached :)

.circleci/config.yml
iamdanfox: --parallel resulted in builds hanging at >8 minutes.  No idea why the clean was there.

PR #274, opened 2018-05-07
Ignore generated_schema folder in checkstyle

tdeitch:
We use generated-schema folders to store generated AtlasDB schema. Use
of AtlasDB is common enough that I propose we include this rule by
default.

### DIFF
--- a/gradle-baseline-java-config/resources/checkstyle/checkstyle-suppressions.xml
+++ b/gradle-baseline-java-config/resources/checkstyle/checkstyle-suppressions.xml
@@ -21,7 +21,8 @@
     <suppress files="\.(bdr|eot|gif|gzip|jar|json|otf|png|svg|ttf|woff|zip)$" checks="NewlineAtEndOfFile" />
 
     <!-- Generated code should not be subjected to checkstyle. -->
+    <suppress files="[/\\].*[/\\]generated[/\\]" checks="." />
+    <suppress files="[/\\].*[/\\]generated_schema[/\\]" checks="." />
     <suppress files="[/\\].*[/\\]generated_src[/\\]" checks="." />
     <suppress files="[/\\].*[/\\]generated_testSrc[/\\]" checks="." />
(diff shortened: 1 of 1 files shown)


gradle-baseline-java-config/resources/checkstyle/checkstyle-suppressions.xml
```diff
@@ -24,4 +24,5 @@
     <suppress files="[/\\].*[/\\]generated_src[/\\]" checks="." />
     <suppress files="[/\\].*[/\\]generated_testSrc[/\\]" checks="." />
     <suppress files="[/\\].*[/\\]generated[/\\]" checks="." />
+    <suppress files="[/\\].*[/\\]generated-schema[/\\]" checks="." />
```
gracew: I've seen many projects generate their atlas schema code into the `generated` source set -- is there a reason for preferring `generated-schema`?

diogoholanda: It does not work if the schema itself depends on some generated classes, like some immutables objects.

diogoholanda: I mean, I think you can make it "work", but in a manual cumbersome way. When generating the schema files, you already need the other depended generated classes to be present. The schema files are generated by a gradle task, not by annotation processing. I've tried setting some proper gradle dependencies over a year ago, but couldn't make it work in those situations (especially when trying to clean up the schema files before re-generating them) and gave up because the setup with an extra folder was just so much simpler.

iamdanfox:
@tdeitch I've found six code results for "generated-schema" in gradle files internally... if we're going to add this to baseline, I'd prefer to go with `generated_schema` for consistency (and make PRs to fix up the internal usages)!

tdeitch:
OK, I switched to `generated_schema` and sorted things in that block.

iamdanfox (approved):

PR #282, opened 2018-06-04
Suppress Checkstyle on generated code in src/gen/

TheBeruriahIncident:
This is where bird based internal product generates code. CC @ossher


iamdanfox:
Baseline is really intended to be an anti-bikeshedding device, so when all the options are reasonably similar, we generally prefer to just pick one and run with it.  Could we change that internal product to write code to `src/generated` instead?

TheBeruriahIncident:
Looking into that

PR #283, opened 2018-06-06
Checkstyle ignores all directories beginning with generated

ellisjoe:


ellisjoe:
@uschi2000 @iamdanfox We have several projects the generate code and some gradle projects need to use multiple of these projects. Having them all generate into `src/generated/java` makes cleanup impossible, so I'd instead like to have them generate into separate `src/generated_*/java` folders which should get ignored by checkstyle.

gradle-baseline-java-config/resources/checkstyle/checkstyle-suppressions.xml
```diff
@@ -21,8 +21,5 @@
     <suppress files="\.(bdr|eot|gif|gzip|jar|json|otf|png|svg|ttf|woff|zip)$" checks="NewlineAtEndOfFile" />
 
     <!-- Generated code should not be subjected to checkstyle. -->
-    <suppress files="[/\\].*[/\\]generated[/\\]" checks="." />
-    <suppress files="[/\\].*[/\\]generated_schema[/\\]" checks="." />
-    <suppress files="[/\\].*[/\\]generated_src[/\\]" checks="." />
-    <suppress files="[/\\].*[/\\]generated_testSrc[/\\]" checks="." />
+    <suppress files="[/\\].*[/\\]generated.*[/\\]" checks="." />
```
iamdanfox, line 24: this would allow people to generate code to `generated_src`, `generated-src` and `generated_src`...

Thoughts on forcing people to pick one and run with it?

ellisjoe: I've seen `generated_*` and been generally happy with that

iamdanfox: :+1: as long as we also tolerate `src/generated/java` which I think immutables does :)

ellisjoe: Do you want stricter enforcement here? Or are we ok saying "any directory that begins with 'generated' will be excluded from checkstyle"?

iamdanfox (approved):
Let's ship this, we can always tighten it up later!

PR #284, opened 2018-06-06
Clarify why the sun imports are banned

TheBeruriahIncident:
It's not because they're repackaged, it's because they're implementation details.


uschi2000 (approved):

uschi2000:
Thanks!

PR #287, opened 2018-06-11
Bintray plugin 1.6 -> 1.8.1

iamdanfox:
This is necessary to make publishing work again with gradle 4.8.

Matches https://github.com/palantir/http-remoting/pull/728


qinfchen:
another issue with the compatibility of bintray and gradle 4.8 https://github.com/bintray/gradle-bintray-plugin/issues/239

PR #298, opened 2018-06-28
Upgrade error-prone's EqualsHashCode and EqualsIncompatibleType from Warn -> Error

### LINKED ISSUE #297: Make Java compiler warnings for equality checks ERROR by default
Error prone provides a bunch of checks via warnings and errors. Given that most warnings are ignored, by making these errors, users would be able to improve their code quality.

mnazbro:
Fixes #297

Enables [`EqualsHashCode`](http://errorprone.info/bugpattern/EqualsHashCode) and [`EqualsIncompatibleType`](http://errorprone.info/bugpattern/EqualsIncompatibleType) as `ERROR` for Errorprone.

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.groovy
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.groovy
@@ -19,6 +19,7 @@ package com.palantir.baseline.plugins
 import net.ltgt.gradle.errorprone.ErrorPronePlugin
 import org.gradle.api.GradleException
 import org.gradle.api.Project
+import org.gradle.api.tasks.compile.JavaCompile
 
 class BaselineErrorProne extends AbstractBaselinePlugin {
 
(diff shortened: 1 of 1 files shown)


ferozco:
can you update the readme to include steps to disable?

mnazbro:
There is current discussion as to whether or not this is the right approach. This may be changed to a list of warnings to check instead.

iamdanfox:
@mnazbro it seems like the prevailing feeling is that this will have too much of an adverse impact on developer productivity as not all warnings are easily actionable and may contain false positives.  Do you want to include just the specific ones that triggered this whole thing?

j-baker:
fwiw we have it enabled internally in most of our repositories (as in, on teams I'm on). Baseline disables a bunch of the more onerous checks so far as I can see.

iamdanfox:
FutureReturnValueIgnored is still a funny one - seems like there are a lot of false positives...

mnazbro:
I've changed this to be focused around equality checks which caused the original movement behind this.

iamdanfox (approved):

PR #301, opened 2018-07-12
Catch block contains log line that doesn't log exception

gatesn:
If you log something from within your catch block, it ought to log the exception.

This fails:
```java
try {
    // blah
    log.info("Something went wrong");
}
```

This is ok:
```java
try {
    // blah
    log.info("Something went wrong", e);
}
```

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/CatchBlockLogException.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/CatchBlockLogException.java
@@ -1,0 +1,71 @@
+/*
+ * (c) Copyright 2018 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.baseline.errorprone;
+
+import com.google.auto.service.AutoService;
+import com.google.errorprone.BugPattern;
+import com.google.errorprone.BugPattern.Category;
+import com.google.errorprone.BugPattern.LinkType;
+import com.google.errorprone.BugPattern.SeverityLevel;
+import com.google.errorprone.VisitorState;
+import com.google.errorprone.bugpatterns.BugChecker;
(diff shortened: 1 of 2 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/CatchBlockLogException.java
```diff
@@ -0,0 +1,71 @@
...
+import com.sun.source.tree.Tree;
+import java.util.regex.Pattern;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "CatchBlockLogException",
+        category = Category.ONE_OFF,
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = LinkType.CUSTOM,
+        severity = SeverityLevel.WARNING,
```
robert3005: IMHO if there's no false positives it should be an ERROR since you always should be logging the exception?

ferozco (approved):
LGTM

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/CatchBlockLogExceptionTest.java
```diff
@@ -0,0 +1,78 @@
...
+    @Test
+    public void testNoLogException() {
+        test(RuntimeException.class, "log.info(\"hello\");", Optional.of(errorMsg));
+    }
+
+    @Test
+    public void testNoLogStatement() {
+        test(RuntimeException.class, "// Do nothing", Optional.empty());
+    }
+
```
uschi2000, line 58: I think it'd be good to add a test case like this, mostly for documentation purposes:
```
try {
  foo();
} catch (IOException e) {
  throw new RuntimeException("Failed to foo()", e);
}
```

uschi2000 (approved):

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/CatchBlockLogExceptionTest.java
gatesn: I wonder if we should have a similar check to ensure you're passing the exception as a cause? Slightly more complicated since not all exception constructors take a cause

dansanduleac: We could at least have it for the builtin exception types that people might be tempted to reach for in ways like `new RuntimeException()`

samrogerson:
This currently rejects `log.error(msg, UnsafeArg.of(e))`, where you may prefer this over `log.error(msg, e)` because the `UnsafeArg` wrapped logs a structured stacktrace instead.

helenyuyu:
This currently rejects 
```
try {
    // blah
    log.info("Something went wrong", e, SafeArg.of("someOtherArg", otherArg));
}
```

markelliot:
It's way more efficient for the exception to be the last argument.

PR #306, opened 2018-07-25
Upgrade error prone to 2.3.1

robert3005:


robert3005:
fixes outstanding issues from #286. @iamdanfox @dansanduleac

baseline-error-prone/build.gradle
```diff
@@ -13,3 +13,11 @@ dependencies {
 
     processor 'com.google.auto.service:auto-service'
 }
+
```
robert3005, line 16: this can be dropped once we release this change and pickup new baseline in baseline

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.groovy
```diff
@@ -32,12 +33,19 @@ class BaselineErrorProne extends AbstractBaselinePlugin {
             errorprone "com.palantir.baseline:baseline-error-prone:${extractVersionString()}"
         }
 
-        project.tasks.withType(JavaCompile) {
-            options.compilerArgs += [
-                    "-XepDisableWarningsInGeneratedCode",
-                    "-Xep:EqualsHashCode:ERROR",
-                    "-Xep:EqualsIncompatibleType:ERROR",
-            ]
+        project.afterEvaluate { p ->
```
robert3005, line 35: according to error-prone docs you need to configure bootclasspath with java8 since error-prone shades some of the jdk (sun.misc) dependencies with java9 versions

robert3005: Unfortunately calling `getAsPath` on a configuration will force evaluation and we need to delay it. There's no api in JavaCompile nor Test task to set it lazily

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.groovy
```diff
@@ -32,12 +33,19 @@ class BaselineErrorProne extends AbstractBaselinePlugin {
...
+            p.tasks.withType(JavaCompile) {
+                options.compilerArgs += [
+                        "-Xbootclasspath/p:${project.getConfigurations().getByName("errorprone").getAsPath()}",
```
iamdanfox, line 38: Does the -> arrow syntax make this any more lazy? We've used this in [conjure](https://github.com/palantir/conjure-java/blob/develop/conjure-java-verifier/build.gradle#L90) occasionally:

```groovy
 "-Xbootclasspath/p:${-> project.getConfigurations().getByName("errorprone").getAsPath()}",
```

iamdanfox (approved):
I'd love to get rid of the `afterevaluate`, but if it's really not possible then let's just :ship: :it:

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.groovy
robert3005: didn't make it work - I think the assignment forces evaluation

iamdanfox (approved):

PR #307, opened 2018-07-27
Add circleci plugin

robert3005:
Applies CircleStylePlugin by default, persists junit html reports to CIRCLE_ARTIFACTS/junit and build profiles to CIRCLE_ARTIFACTS/profile


versions.props
```diff
@@ -6,6 +6,7 @@ com.google.guava:guava = 21.0
 com.palantir.baseline:* = 0.24.0
 com.palantir.safe-logging:* = 1.4.0
 org.slf4j:slf4j-api = 1.7.24
+gradle.plugin.com.palantir:gradle-circle-style = 1.2.0-rc6
```
robert3005: we should release 1.2.0 once we have confidence it works. We shouldn't let all the features there just rot

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineCircleCi.groovy
```diff
@@ -0,0 +1,61 @@
...
+import org.gradle.profile.ProfileListener
+import org.gradle.profile.ProfileReportRenderer
+
+class BaselineCircleCi extends AbstractBaselinePlugin {
+    private static final SimpleDateFormat FILE_DATE_FORMAT = new SimpleDateFormat("yyyy-MM-dd-HH-mm-ss")
+
+    @Override
+    void apply(Project project) {
+        project.pluginManager.apply CircleStylePlugin
+        final String circleArtifactsDir = System.getenv("CIRCLE_ARTIFACTS")
```
qinfchen: Note that `CIRLCE_ARTIFACTS` and `CIRCLE_TEST_REPORTS` environment variables are no longer available in circle2

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineCircleCi.groovy
```diff
@@ -0,0 +1,61 @@
...
+        if (circleArtifactsDir == null) {
+            return
+        }
+
+        project.rootProject.allprojects({Project p ->
+            p.getTasks().withType(Test.class, {Test test ->
+                File junitReportsDir = new File(circleArtifactsDir, "junit")
+                for (String component : test.getPath().substring(1).split(":")) {
```
qinfchen: nit: use `Paths.get(...).toFile()`?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineCircleCi.groovy
```diff
@@ -0,0 +1,61 @@
...
+        })
+
+        if (project.getGradle().startParameter.profile) {
+            project.getGradle().addListener(new ProfileListener() {
+                @Override
+                void buildFinished(BuildProfile buildProfile) {
+                    ProfileReportRenderer renderer = new ProfileReportRenderer()
+                    File file = new File(circleArtifactsDir, "profile/profile-" +
+                            FILE_DATE_FORMAT.format(new Date(buildProfile.getBuildStarted())) + ".html")
+                    renderer.writeTo(buildProfile, file)
```
qinfchen: should we add a report url for people to access it?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineCircleCi.groovy
robert3005: Sure, we need a convention though and @ferozco @iamdanfox and me agree on this convention.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineCircleCi.groovy
robert3005: Not sure how `Paths.get` helps here. I have to keep on converting between path -> file -> string

robert3005:
```
File junitReportsDir = Paths.get(circleArtifactsDir, "junit").toFile()
for (String component : test.getPath().substring(1).split(":")) {
    junitReportsDir = Paths.get(junitReportsDir.toString(), component).toFile()
}
```

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineCircleCi.groovy
robert3005: I always find the output directories confusing since they're not what's accessible to you. We can print some comment about where to find the profiles but we might as well include it in the readme of this repo

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineCircleCi.groovy
iamdanfox: In our internal circle2 build image, we actually set these environment variables now, so they're available internally!

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineCircleCi.groovy
```diff
@@ -0,0 +1,63 @@
...
+                for (String component : test.getPath().substring(1).split(":")) {
+                    junitReportsDir = junitReportsDir.resolve(component)
+                }
+                test.getReports().getHtml().setEnabled(true)
+                test.getReports().getHtml().setDestination(junitReportsDir.toFile())
+            })
+        })
+
+        if (project.getGradle().startParameter.profile) {
+            project.getGradle().addListener(new ProfileListener() {
```
iamdanfox: <3 I am so happy about this - solves https://github.com/palantir/gradle-circle-style/issues/28

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineCircleCi.groovy
```diff
@@ -0,0 +1,63 @@
...
+    private static final SimpleDateFormat FILE_DATE_FORMAT = new SimpleDateFormat("yyyy-MM-dd-HH-mm-ss")
+
+    @Override
+    void apply(Project project) {
+        project.pluginManager.apply CircleStylePlugin
+        final String circleArtifactsDir = System.getenv("CIRCLE_ARTIFACTS")
+        if (circleArtifactsDir == null) {
+            return
+        }
+
```
iamdanfox: Could we also do `Files.createDirectories` on the circleArtifactsDir, so that we don't fail if users forgot to do a `mkdir -p`??

iamdanfox (approved):
Thanks for putting this together! Can we have a little heading in the README then let's merge?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineCircleCi.groovy
qinfchen: i think  you can do something like this without the for loop. 
```
Paths.get(junitReportsDir.toString(), test.getPath().substring(1).split(":")).toFile()
```

PR #311, opened 2018-08-01
Change plugin application order in baseline-circleci

robert3005:


qinfchen (approved):

iamdanfox (commented):
Let’s also add an info log line before that early return, to avoid the same confusion again!

robert3005:
We should also add a test

PR #312, opened 2018-08-02
New `com.palantir.baseline-versions` plugin

iamdanfox:
This plugin means people no longer need to directly depend on nebula dependency recommender.  I think it's desirable to present our own abstraction because:

1. nebula dependency recommender is technically [deprecated](https://github.com/nebula-plugins/nebula-dependency-recommender-plugin), so we can swap out our implementation without breaking people
1. the setup gradle is copied and pasted everywhere

I've published this locally and it tried it out and it makes a top-level build.gradle look nicer :)

Questions:

- [x] decide a on a good name, can we do better than `com.palantir.baseline-versions-props`? Possibly just `com.palantir.baseline-versions`?
- integration tests??
- Should we take this opportunity to switch from versions.props -> versions.properties?? (This would get us syntax highlighting in IDEs)


j-baker:
How should I specify that I want to apply some BOMs in this world?

j-baker:
image

robert3005:
as @j-baker suggests - you need to expose an extension to let us configure bom(b)s

iamdanfox:
To maintain our existing BOM workflow, I think we could just re-use nebula's extension for now? Users would have to add the following snippet. (The `addFirst` method here ensures that versions.props file can override things from the bom)

```gradle
dependencyRecommendations {
    addFirst mavenBom(module: 'com.palantir.wc:wc-core-bom')
}
```

The obvious downside here is that re-exposing a bit of nebula makes it harder to completely change the implementation in the future, but perhaps this is not such a big deal?

dansanduleac (approved):

dansanduleac:
Noting that `addFirst` is not necessary since the _MavenBomRecommendationProvider_ is already always [added as the first provider](https://github.com/nebula-plugins/nebula-dependency-recommender-plugin/blob/1c52e5c27f1014f3183f1c4d9d4af909122d4c17/src/main/groovy/netflix/nebula/dependency/recommender/provider/RecommendationProviderContainer.java#L60), and all `mavenBom` does is it adds dependencies to a configuration read by that provider.

PR #313, opened 2018-08-02
Improve baseline-checkstyle implementation and upgrade to 8.12

### LINKED ISSUE #198: Design for extension is absurd for test methods
> Method 'smoke_test' is not designed for extension - needs to be abstract, final or empty.

screen shot 2017-07-14 at 14 58 36

I appreciate this is very minor, but it is a paper-cut that will encourage people to either: not write tests or disable checkstyle.  I can't see any upside.

### LINKED ISSUE #207: checkstyle.xml rules don't work with checkstyle 8.1
I get this with the latest version of intellij 2017.2

```
2:51 PM	The Checkstyle rules file could not be parsed.
				SuppressionCommentFilter is not allowed as a child in Checker
				The file has been blacklisted for 60s.
```

As per http://checkstyle.sourceforge.net/releasenotes.html, we need to move the filter under TreeWalker, but that breaks compatibility with checkstyle 7.5.1, which is the default now.

robert3005:
@iamdanfox 

fixes https://github.com/palantir/gradle-baseline/issues/207
fixes https://github.com/palantir/gradle-baseline/issues/198


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineCheckstyle.java
```diff
@@ -0,0 +1,89 @@
...
+        // Set default version (outside afterEvaluate so it can be overridden).
+        project.getExtensions().configure(CheckstyleExtension.class,
+                ext -> ext.setToolVersion(DEFAULT_CHECKSTYLE_VERSION));
+
+        // We use the "JavadocMethod" module in our Checkstyle configuration, making
+        // Java 8+ new doclint compiler feature redundant.
+        project.getTasks().withType(Javadoc.class, javadoc -> {
+            if (project.getConvention().getPlugin(JavaPluginConvention.class)
+                    .getSourceCompatibility().isJava8Compatible()) {
+                javadoc.options(minimalJavadocOptions -> ((StandardJavadocDocletOptions) minimalJavadocOptions)
```
robert3005: this is really annoying but looks like the api is a little bs here

iamdanfox: If the `if` statement is always going to return the same value, could we move it outside the `withType(Javadoc.class, ->` bit?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineCheckstyle.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    private void configureCheckstyle() {
+        // Configure checkstyle
+        project.getExtensions().getByType(CheckstyleExtension.class)
+                .setConfigDir(project.file(Paths.get(getConfigDir(), "checkstyle").toString()));
+        project.getTasks().withType(Checkstyle.class, checkstyle -> {
+            // Make checkstyle include files in src/main/resources and src/test/resources, e.g., for whitespace checks.
+            checkstyle.source("src/main/resources");
+            checkstyle.source("src/test/resources");
```
iamdanfox: Is it worth iterating through sourcesets here?  Repos commonly add `integTest` too

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineCheckstyle.java
```diff
@@ -0,0 +1,85 @@
...
+import org.gradle.external.javadoc.StandardJavadocDocletOptions;
+import org.gradle.plugins.ide.eclipse.model.EclipseModel;
+import org.gradle.plugins.ide.eclipse.model.EclipseProject;
+
+/**
+ * Configures the Gradle "checkstyle" task with Baseline settings.
+ */
+public final class BaselineCheckstyle extends AbstractBaselinePlugin {
+
+    private static final String DEFAULT_CHECKSTYLE_VERSION = "8.11";
```
iamdanfox: I guess we should probably cut and RC so that we can try out the 7.5.1 -> 8.11 upgrade on a few internal projects?

iamdanfox:
So just to double check, this PR is essentially:

1. convert groovy -> java 
1. upgrade checkstyle 7.5.1 -> 8.11
1. remove some afterEvaluate
1. remove workaround for now fixed https://github.com/gradle/gradle/issues/855

robert3005:
Yes, I am trying to go through some of the plugins here and make them more modern/conforming to current practices.

The checkstyle bump is here since in #205 we expressed desire to do so and we never did overall and current version is 1.5 year old now

uschi2000:
Repeating my question from #205 : 

There's a pretty big change list between 7.5.1 and 8.1. Did you go through them and verify that (1) we're comfortable with the changes, (2) they align with our Baseline docs, (3) they are compatible with the IntelliJ and Eclipse formatting, (4) they require no other config change?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineCheckstyle.java
uschi2000: there's like an ordering problem here ...

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineCheckstyle.java
```diff
@@ -0,0 +1,86 @@
...
+        configureCheckstyleForEclipse();
+    }
+
+    private void configureCheckstyle() {
+        // Configure checkstyle
+        project.getExtensions().getByType(CheckstyleExtension.class)
+                .setConfigDir(project.file(Paths.get(getConfigDir(), "checkstyle").toString()));
+        project.getTasks().withType(Checkstyle.class, checkstyle -> {
+            // Make checkstyle include files in src/main/resources and src/test/resources, e.g., for whitespace checks.
+            project.getConvention().getPlugin(JavaPluginConvention.class).getSourceSets()
```
robert3005: need to verify this is the correct path

robert3005:
Run this locally - need to update some checkstyle config to match new values.

Things I found that are different:
`SummaryJavaDoc` - your javadoc needs to have a summary sentence - sounds like a good change
`UnnecessaryParentheses` - now catches unnecessary parens for lambda arguments - good as well
`ImportOrder` - Doesn't allow empty line between import static and import - should we change it or not?
`DesignForExtension` - catches violations in tests that weren't previously being caught - might want to suppress for tests

robert3005:
The DFE thing is only because the repo I looked at is using junit5 and defaults don't account for it

robert3005:
After adjusting the checkstyle config to match what we expected before I get only `SummaryJavadoc` and `UnnecessaryParentheses` violations which look like good changes and could be considered bugs in checkstyle overall

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineCheckstyle.java
robert3005: It is the correct path though it's impossible to test in unit tests. Will add an integration test

gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
```diff
@@ -433,6 +433,10 @@
             <property name="format" value="^[a-z][a-zA-Z0-9]+$"/>
             <message key="name.invalidPattern" value="Parameter name ''{0}'' must match pattern ''{1}''."/>
         </module>
+        <module name="LambdaParameterName"> <!-- Java Style Guide: Parameter names -->
```
robert3005: Adds parameter name checkstyle for lambda parameters. There's issues with ParameterName right now for lambdas that it doesn't correctly detect lambdas (and after applying this I can confirm that I have few violations of lambda parameter names that weren't caught before). From 8.12 checkstyle ParameterName will completely ignore lambda parameters and this check is preferred.

ferozco (approved):
LGTM, lets cut an RC and test this out in a few places before rolling out more widely

robert3005:
@uschi2000 how do you feel about the current state?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineCheckstyle.java
uschi2000: want to go straight to 8.12?

uschi2000:
Looks good, thanks @robert3005 . Can you include the 8.11/8.12 bump in the commit message, please?

robert3005:
We're going to remove LambdaParameterName from the rules since in practice we found way too many violations for this rule to be meaningful

PR #316, opened 2018-08-02
Fix javadoc task after error-prone upgrade

robert3005:
@iamdanfox found this after upgrading internal projects

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.groovy
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.groovy
@@ -49,6 +49,8 @@ class BaselineErrorProne extends AbstractBaselinePlugin {
             }
 
             p.tasks.withType(Javadoc) {
+                // Add error-prone to bootstrap classpath of javadoc task.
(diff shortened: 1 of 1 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.groovy
```diff
@@ -32,18 +33,24 @@ class BaselineErrorProne extends AbstractBaselinePlugin {
...
             }
 
             p.tasks.withType(Test) {
-                jvmArgs "-Xbootclasspath/p:${project.getConfigurations().getByName("errorprone").getAsPath()}"
+                jvmArgs "-Xbootclasspath/p:${errorProneConf.getAsPath()}"
+            }
+
+            p.tasks.withType(Javadoc) {
+                options.bootClasspath.addAll(errorProneConf.resolve())
+                options.bootClasspath.addAll(System.properties["sun.boot.class.path"].split(File.pathSeparator).collect{new File(it)})
```
iamdanfox: can we put some explanatory comments around this stuff? Perhaps just a note to say why it's necessary here or where it came from? I wouldn't really feel confident to touch any of this stuff myself for fear of breaking people

iamdanfox (approved):

PR #319, opened 2018-08-03
Merge gradle-circle-style into this repo

robert3005:
Takes unreleased 1.2.0 branch from gradle-circle-style and merges it into gradle-baselineci plugin. The enhancements include javac errors as circle reports as well as setting junit xml report destination setting.

This pr should not be squashed to preserve original commit history. Didn't want to make deeper changes to minimize churn.


robert3005:
Squashed all the existing commits from gradle-circle-style into one and rebased (to avoid separate root). Please do not squash when merging

ferozco (approved):

PR #322, opened 2018-08-06
Fix application of Baseline plugin

### LINKED ISSUE #181: Applying all baseline tasks to subprojects
The example in the link below results in an error because you can't `apply plugin: 'com.palantir.baseline'` to subprojects:
https://github.com/palantir/gradle-baseline#applying-baseline-plugins-selectively-or-all-at-once

```
* What went wrong:
A problem occurred evaluating root project 'monocle'.
> Failed to apply plugin [class 'com.palantir.baseline.plugins.BaselineConfig']
   > com.palantir.baseline.plugins.BaselineConfig plugin can only be applied to the root project
```

I think the functionality of being able to apply all checks is nice as you don't have to keep up-to-date with new checks being added, so it would be nice if there were a way to separate the root plugins from subproject plugins.

robert3005:
it's enough to just apply at top level now
```
plugins {
  id 'com.palantir.baseline'
}
```

fixes https://github.com/palantir/gradle-baseline/issues/181


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/Baseline.java
```diff
@@ -0,0 +1,43 @@
...
+import org.gradle.api.Plugin;
+import org.gradle.api.Project;
+
+/**
+ * A Plugin that configures a project with all Baseline settings.
+ */
+public final class Baseline implements Plugin<Project> {
+
+    @Override
+    public void apply(Project project) {
```
iamdanfox, line 28: Could we include a warning line so that people don't think they have to apply this to subprojects?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/Baseline.java
```diff
@@ -0,0 +1,43 @@
...
+        Project rootProject = project.getRootProject();
+        rootProject.getPluginManager().apply(BaselineConfig.class);
+        rootProject.getPluginManager().apply(BaselineCircleCi.class);
+        rootProject.allprojects(proj -> {
+            proj.getPluginManager().apply(BaselineCheckstyle.class);
+            proj.getPluginManager().apply(BaselineEclipse.class);
+            proj.getPluginManager().apply(BaselineIdea.class);
+            proj.getPluginManager().apply(BaselineErrorProne.class);
+            proj.getPluginManager().apply(BaselineVersions.class);
```
qinfchen, line 43: should this be applied at the root?

iamdanfox: No, it's fine to apply `com.palantir.baseline-versions` to allprojects:

https://github.com/palantir/gradle-baseline/blob/develop/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineVersions.java

robert3005: it should be applied to every project I think because of how dependency recommender works?

iamdanfox (approved):

uschi2000:
Please update the readme, it's out of date now: https://github.com/palantir/gradle-baseline/#applying-baseline-plugins-selectively-or-all-at-once

PR #324, opened 2018-08-07
Fix readme about baseline plugin application

robert3005:
@uschi2000


robert3005:
This is partially broken because of checkstyle plugin so until #313 we can't do this

robert3005:
```
plugins {
  id 'com.palantir.baseline'
}
```
doesn't work right now since this project isn't published correctly for pluginmanagement to resolve it

ferozco (approved):

PR #326, opened 2018-08-09
Upgrade baseline and publish to plugins.gradle.org

robert3005:
Not sure why we weren't publishing to plugins.gradle.org before. @iamdanfox @dansanduleac not sure I am doing publishing right but it should be enough to just do full publish on top of bintrayUpload?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,10 +30,7 @@ jobs:
           path: /tmp/store_test_results
 
       - deploy:
-          command: |
-            if [[ "${CIRCLE_TAG}" =~ [0-9]+(\.[0-9]+)+(-[a-zA-Z]+[0-9]*)* ]]; then
-              ./gradlew --info bintrayUpload
-            fi
+          command: ./gradlew --info --continue publish
 
   docs:
     docker:
--- a/gradle-baseline-java-config/build.gradle
+++ b/gradle-baseline-java-config/build.gradle
@@ -11,20 +11,22 @@ task zipConfig(type: Zip) {
     }
 }
 
-bintrayUpload.dependsOn { generatePomFileForConfigPublicationPublication }
-bintrayUpload.dependsOn { zipConfig }
-
 bintray {
-    user = System.getenv('BINTRAY_USERNAME')
-    key = System.getenv('BINTRAY_PASSWORD')
+    user = System.env.BINTRAY_USERNAME
+    key = System.env.BINTRAY_PASSWORD
     publish = true
     pkg {
         repo = 'releases'
-        name = 'gradle-baseline'
+        name = rootProject.name
         userOrg = 'palantir'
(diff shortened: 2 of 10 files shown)


settings.gradle
```diff
@@ -1,3 +1,5 @@
+enableFeaturePreview("STABLE_PUBLISHING")
```
ferozco, line 2: can we specify the root project name?

gradle-baseline-java-config/build.gradle
```diff
@@ -11,22 +11,24 @@ task zipConfig(type: Zip) {
     }
 }
 
-bintrayUpload.dependsOn { generatePomFileForConfigPublicationPublication }
-bintrayUpload.dependsOn { zipConfig }
-
 bintray {
-    user = System.getenv('BINTRAY_USERNAME')
-    key = System.getenv('BINTRAY_PASSWORD')
+    user = System.env.BINTRAY_USERNAME
```
ferozco, line 15: it might be worthwhile to factor out the bintray config into a shared script

.circleci/config.yml
```diff
@@ -30,10 +30,7 @@ jobs:
           path: /tmp/store_test_results
 
       - deploy:
-          command: |
-            if [[ "${CIRCLE_TAG}" =~ [0-9]+(\.[0-9]+)+(-[a-zA-Z]+[0-9]*)* ]]; then
-              ./gradlew --info bintrayUpload
-            fi
+          command: ./gradlew --info --continue publish
```
ferozco, line 33: seems like the only difference is 
```
> :baseline-error-prone:publish SKIPPED
> :gradle-baseline-java:publish SKIPPED
> :gradle-baseline-java-config:publish SKIPPED
```

which i don't think is a real problem

robert3005: looks like there's more options to configure to get it to publish

robert3005:
We do not have GRADLE credentials though for the moment so can't merge it yet.

@uschi2000 why did we stop publishing to plugins.gradle.org?

uschi2000:
@robert3005 because the `plugins {}` mechanism wasn't really flexible enough to apply plugins in subprojects selectively, so we never used it.

robert3005:
That makes sense - I don't think there's that much of optionality anyway and this pr still preserves old publishing to bintray (essentially double publish). I think only eclipse/idea and errorprone do anything on their own other plugins just configure existing ones with sane defaults. I think we should try encourage usage of `plugins {}` and there's still an escape hatch for those who want current behaviour

robert3005:
also we can just publish to gradle.org with `mavenCoordinate` override but then we'll have to wait for approval from someone from gradle

robert3005:
I will split this pr in two for easier readability and we can decouple baseline upgrade from publishing since I understand the main issue with publishing now

PR #333, opened 2018-08-16
Don't apply baseline-error-prone checks to baseline error prone project

robert3005:
This avoids circular dependency due to bootclasspath setting

@iamdanfox @ferozco


iamdanfox (approved):
This should unblock publishing tags of this project again, so no more red builds like https://circleci.com/gh/palantir/gradle-baseline/915

PR #339, opened 2018-08-20
New com.palantir.baseline-format plugin allows `./gradlew format`

iamdanfox:
Currently, people laboriously edit up their code to comply with checkstyle.  It would be much nicer if we could provide an automatic way to fix up most of the trivial problems (e.g. whitespace).

This new `./gradlew format` task only affects your imports, it does not attempt to format anything else (we reverted the google-java-format/eclipse formatter version in the spirit of making small incremental progress).


uschi2000:
Directionally, I'd prefer moving towards `google-java-format`.

iamdanfox:
@uschi2000 the big difference I can see is that the eclipse one allows 120 character lines whereas the google-java-format forces you to use 100 character lines.  You can try this out pretty easily locally, by editing a build.gradle and then running `./gradlew spotlessApply`:

```gradle
buildscript { 
  dependencies { 
    classpath 'com.diffplug.spotless:spotless-plugin-gradle:3.14.0'
  }
}

apply plugin: 'com.diffplug.gradle.spotless'

spotless {
    java {
        target 'src/main/java/**/*.java', 'src/main/test/**/*.java'
        googleJavaFormat().aosp()
        removeUnusedImports()
        trimTrailingWhitespace()
        indentWithSpaces(4)
        endWithNewline()
    }
}
```

uschi2000:
OK, let me rephrase my previous comment: I'm in favour of using an auto-formatter and don't care too much whether this is the Eclipse formatter (with some config file) or google-java-format (with zero-config). If we do this, though, then we should simultaneously delete all formatting-related config and code from the `idea`, `eclipse`, and `checkstyle` Baseline plugins.

uschi2000:
The 100 line thing is a bit annoying, but not a deal-breaker for me personally. Others might have other opinions.

robert3005:
Can you show examples on some internal repos of what would change with this? Afaik google java format struggles with line breaks on lambdas

iamdanfox:
I've just run `./gradlew format` on a large internal repo: f/f/pull/3870, which interestingly resulted in +2,744 −3,369 lines.... I'm actually pleasantly surprised, it seems to have made mostly sensible decisions.

mnazbro:
How often do people actually debate over formatting in PRs? I've never really had too much of an issue here, so this feels to be solving a non-issue.

Also, are we at all concerned that there can be unnecessary changes to code because the formatter decides it needs to be formatted differently? I've noticed that prettier on the Typescript side often causes changes based on lambdas, lengths of jsx elements, numbers of imports, number of arguments to functions, etc.

a10y:
Personally, I'm actually not sure if I've worked on many projects internally where style isn't enforced as part of the code review process. I've both been on the receiving end and dished out a decent amount of `nit: ____` comments, whether its whitespace, parameter grouping, etc. I think there's value in centralizing/automatting these style decisions for the same reasons why we have baseline to begin with.

To your point about prettier, from running it on an internal project we haven't bumped into spurious code changes and I don't foresee any. There is an escape hatch via magic comments same as checkstyle in case you strongly disagree with the formatter's choice, but I've only had to use that in one place where there were deeply nested parameterized types (and that code got removed recently so we no longer use the escape hatch comments).

uschi2000:
The goal of formatting rules in described in the [Baseline docs](https://github.com/palantir/gradle-baseline/tree/develop/docs/java-style-guide):
```
Code is more often read than written.
Style guidelines exist to make reading code more pleasant and 
efficient by standardizing formatting across development teams.
```
A second goal to avoid syntax debates in PRs and instead focus on the subject matter.

Most new/modern languages (e.g., Go, Rust) have a canonical format. Java, unfortunately does not. As described in Baseline docs, it's more important to agree on *a* format than to agree its particular shape. The simplest thing is let the computer do the work and auto-format everything; the <1% edge cases of suboptimal formatting (in the eye of some) are an unfortunate yet acceptable consequence.

qinfchen:
I've removed the eclipse.xml for now and only kept the less contentious style changes. Can we merge this and have a separate pr to pick up either google or eclipse formatter along with the removal of formatter relatedd baseline configs?

qinfchen (approved):

iamdanfox:
This looks great to me.  It allows us to more easily roll out automation that changes people's imports, without imposing opinionated styles on people yet!

Issue #216, opened 2017-10-09
Do not require explicitly setting the baseline version in two places


leeavital:
Users need to specify the baseline version in two places:

```
buildscript {

   classpatch "baseline:<version>
}

// and then again
dependencies {
  baseline "baseline:<version>@zip"
}
```

users are choosing to either factor out a variable into gradle.properties or leave a comment reminding maintainers to keep the values in sync. Both of these don't gel with our gradle linter which expects version numbers to live in a version.props file. If would be great if baseline could automatically pull down the right version of the baseline configuration if the explicit setting in the dependencies block was ommited.

markelliot:
There’s not really a great way to version the buildscript dependencies since the recommender is itself a buildscript dependency.

leeavital:
I think it's fine to version the buildscript dependency explicitly.  But I think it should be possible to automatically configure the `baseline` dependency that accompanies it. Something like:

```
afterEvaluate {
    if (!baselineConfigured()) {
       project.configurations.add("baseline").add("baseline:" + getBaselineVersion() + "@zip");
   }
}
```

in the plugin (I may be mixed up about gradle lifecycle things)

jmcampanini:
configurations have a concept of `defaultDependencies` which lets plugin authors add specific dependencies if none are provided: https://docs.gradle.org/current/javadoc/org/gradle/api/artifacts/Configuration.html#defaultDependencies(org.gradle.api.Action)

qinfchen:
A related ticket, https://github.com/palantir/gradle-baseline/issues/193.  

> The default was removed in order to make the configuration space more
democratic and not favour a single configuration.

iamdanfox:
Fixed in [0.30.0](https://github.com/palantir/gradle-baseline/releases/tag/0.30.0)

PR #340, opened 2018-08-20
baseline "com.palantir.baseline:gradle-baseline-java-config:x.y.z@zip" no longer required

### LINKED ISSUE #193: make dependencies.baseline default to same version of configuration files
Currently, it is required to specify the baseline configuration files in dependencies block. Can we make it default to the same version of configuration files if it is not being set? The benefit is simplifying the config block since it is fairly common that we use the same version for the plugin and configuration file. 

```
dependencies {
    // Adds a dependency on the Baseline configuration files. Typically use 
    // the same version as the plugin itself.
    baseline "com.palantir.baseline:gradle-baseline-java-config:<version>@zip"
}
```

iamdanfox:
## Before this PR

Every user of baseline has the same version number in _two places_ in their build.gradle.

1. once in the buildscript dependency
2. again in the `baseline` configuration

(This functionality was originally implemented so that users could substitute their own config but this has not been heavily used, and is really just boilerplate at this point.)

## After this PR

The gradle-baseline-java-config dependency is assumed by default, so users no longer need to specify it explicitly! Note: users can still override this if they choose.

```diff
-dependencies {
-    baseline "com.palantir.baseline:gradle-baseline-java-config:x.y.z@zip"
-}
```

fixes #193, #216


ferozco:
I thought we had explicitly wanted this to decouple the configuration from the plugin. What prompted the change in posture?

iamdanfox:
Posture has not changed - the configuration remains decoupled from the plugin.  The only thing changed in this PR is that every repo that uses the default baseline dep (ie same version as the plugin) can delete three lines from their build.gradle.

gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineConfigIntegrationTest.groovy
```diff
@@ -54,7 +54,7 @@ class BaselineConfigIntegrationTest extends AbstractPluginTest {
 
         then:
         with('baselineUpdateConfig').buildAndFail().output.contains(
-                "Expected to find exactly one config dependency in the 'baseline' configuration, found: []")
+                "Could not find com.palantir.baseline:gradle-baseline-java-config:.")
```
ferozco: could we add a passing test where we don't add the dependency explicitly?

ferozco (approved):
looks good! can we excavate to remove the cruft from users repos as a follow up?

iamdanfox:
Yeah definitely... I think I might add one log.warn so that users realise they can remove it themselves too

robert3005:
@dfox - I was about to make a pr for this. Can we instead use default dependencies? Or at least unify behaviour with baseline error prone? Happy to discuss and probably the ideal solution involves teaching buildscript dependencies about nebula recommendations

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineConfig.java
```diff
@@ -0,0 +1,61 @@
...
+        if (!rootProject.equals(rootProject.getRootProject())) {
+            throw new IllegalArgumentException(
+                    BaselineConfig.class.getCanonicalName() + " plugin can only be applied to the root project.");
+        }
+
+        Configuration configuration = rootProject.getConfigurations().create("baseline");
+
+        // users can still override this default dependency, it just reduces boilerplate
+        Optional<String> version = Optional.ofNullable(getClass().getPackage().getImplementationVersion());
+        configuration.defaultDependencies(d -> d.add(project.getDependencies().create(String.format(
```
iamdanfox, line 40: @robert3005 this is the `defaultDependencies` approach that I'm currently using... did you have something else in mind?

PR #342, opened 2018-08-21
[feature] Add linting tasks for versions.props

rubenfiszel:
A PR that adds the tasks `checkBomConflict`, `checkNoUnusedPin`, and `checkVersionsProps`

`checkBomConflict`: That task checks and lists all pins that override recommendations.
It fails the build if there are any "critical" conflicts. A critical conflict is 
an override that only override a recommendation with the same version. If it's a wildcard 
that also override a non-recommended artifact, it is not considered critical. 
It will help you clean and constrain your versions.props to overrides that
are actually necessary.

`checkNoUnusedPin`: That task list all pinning that are never used and 
that you can remove safely. If there are any, the build fails.

The `checkVersionsProps` depend on those two tasks. 

The `check` task depends on `checkVersionsProps` so if check is already part of your build, the check will happen as part of your build without further modifications

It handles the wildcard `*` from versions.props.

You can ignore properties using: 
`# linter:OFF` and reactivating `# linter:ON`.

`checkNoUnusedPin` example:

```
$ ./gradlew checkNoUnusedPin

* What went wrong:
Execution failed for task ':checkNoUnusedPin'.
> There are unused pins in your versions.props: 
  com.palantir.foundry:foundry-testing-lib
  com.palantir.ls:*
  com.palantir.sql.translate:sql-translate
  org.assertj:assertj-guava
```

`checkBomConflict` example:

```
$ ./gradlew checkBomConflict

There are conflicts between versions.props and the bom:
bom:            com.palantir.transforms.worker:transforms-spark-module-lib 1.73.0
versions.props: com.palantir.transforms.worker:transforms-spark-module-lib 1.73.0

...

* What went wrong:
> Critical conflicts between versions.props and the bom (overriding with same version)
  bom:            com.palantir.transforms.worker:transforms-spark-module-lib 1.73.0
  versions.props: com.palantir.transforms.worker:transforms-spark-module-lib 1.73.0
  
  bom:            com.palantir.foundry.build2:build-foundry-core-api-objects 0.66.0
  versions.props: com.palantir.foundry.build2:* 0.66.0
  
  bom:            com.palantir.foundry.build2:workers 0.66.0
  versions.props: com.palantir.foundry.build2:* 0.66.0
  
  bom:            com.palantir.foundry.build2:foundry-core-adapter-workers 0.66.0
  versions.props: com.palantir.foundry.build2:* 0.66.0
  
  bom:            com.palantir.foundry.build2:foundry-core-workers 0.66.0
  versions.props: com.palantir.foundry.build2:* 0.66.0
  
  bom:            com.palantir.foundry.build2:build2-api-objects 0.66.0
  versions.props: com.palantir.foundry.build2:* 0.66.0
  
  bom:            com.palantir.foundry.build2:build2-api-jersey 0.66.0
  versions.props: com.palantir.foundry.build2:* 0.66.0
  
  bom:            com.palantir.conjure.java:conjure-lib 1.0.0
  versions.props: com.palantir.conjure.java:conjure-lib 1.0.0
  
  bom:            com.google.code.gson:gson 2.2.4
  versions.props: com.google.code.gson:gson 2.2.4
  
  ...
```


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineVersions.java
```diff
@@ -53,19 +60,33 @@
...
 
         // get dependencyRecommendations extension
-        RecommendationProviderContainer extension = project.getExtensions()
+        RecommendationProviderContainer extension = project
+                .getExtensions()
                 .getByType(RecommendationProviderContainer.class);
 
-        extension.setStrategy(RecommendationStrategies.OverrideTransitives); // default is 'ConflictResolved'
+        extension.setStrategy(RecommendationStrategies.OverrideTransitives); // default is 'ConflictResolved';
+        extension.whenObjectAdded(recommendationProvider -> applyPropsFile(project));
```
robert3005: this is actually a bit more annoying since if you have multiple boms it's still not what you expect - happy to outlaw multiple boms anyway

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineVersions.java
```diff
@@ -53,19 +60,33 @@
...
 
-        extension.propertiesFile(ImmutableMap.of("file", rootVersionsPropsFile));
+        project.getTasks().create("checkBomConflict", BomConflictCheckTask.class);
```
robert3005: this task needs to declare inputs since otherwise it's never up to date

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineVersions.java
```diff
@@ -53,19 +60,33 @@
...
+        project.getTasks().create("checkNoUnusedPin", NoUnusedPinCheckTask.class);
+        project.getPlugins().apply(BasePlugin.class);
```
robert3005: `getPluginManager` instead

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineVersions.java
```diff
@@ -53,19 +60,33 @@
...
+        project.getTasks().create("checkVersionsProps");
+        project.getTasks().getByName("checkVersionsProps").dependsOn("checkBomConflict", "checkNoUnusedPin");
```
robert3005: these two should happen as a configure action on `checkVersionProps`. You can do `project.getTasks().register(...)`

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineVersions.java
```diff
@@ -84,4 +105,60 @@ private static File rootVersionsPropsFile(Project project) {
...
+                } catch (Exception e) {
+                    throw new RuntimeException("Error during resolution of the artifacts of all"
+                            + "configuration from all subprojcts", e);
+                }
+            });
+        });
+        return artifacts;
+    }
+
+    public static void checkVersionsProp(Project project, Function<Pair<String, String>, Void> function) {
```
robert3005: There's a BiFunction you can use here

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BomConflictCheckTask.java
```diff
@@ -0,0 +1,136 @@
...
+
+    }
+
+    private String conflictsToString(List<Conflict> conflicts, Map<String, String> resolvedConflicts) {
+        return conflicts.stream()
+                .map(conflict -> conflict.details(resolvedConflicts))
+                .collect(Collectors.joining("\n"));
+    }
+
+    private static class Conflict {
```
robert3005, line 113: use immutables for this?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineVersions.java
robert3005: Actually - bom recommendations are always first so no need to do this actually

rubenfiszel: Ok. I still need to make sure the mavenBom configuration have been applied so I probably want to make my task check task depend on "resolveConfiguration". 
Could you point me to the logic where you found that the mavenBom recommendations are always applied first ? I remember looking at it and it was a bit unclear if it was or not.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BomConflictCheckTask.java
rubenfiszel: It's not already a dependency, not sure I want to add it as a dependency for removing a few lines of boilerplate

robert3005: Fair

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BomConflictCheckTask.java
```diff
@@ -0,0 +1,149 @@
...
+import java.util.Set;
+import java.util.stream.Collectors;
+import javax.inject.Inject;
+import netflix.nebula.dependency.recommender.provider.RecommendationProviderContainer;
+import org.gradle.api.DefaultTask;
+import org.gradle.api.tasks.InputFile;
+import org.gradle.api.tasks.TaskAction;
+
+
+public final class BomConflictCheckTask extends DefaultTask {
```
robert3005: This also needs to depend on the maven bom itself - since you can change the one you depend on and it will consider things up to date

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineVersions.java
```diff
@@ -84,4 +98,55 @@ private static File rootVersionsPropsFile(Project project) {
...
+            });
+        });
+        return artifacts;
+    }
+
+    public static void checkVersionsProp(File propsFile, BiFunction<String, String, Void> function) {
+        boolean active = true;
+        if (propsFile.exists()) {
+            try {
+                List<String> lines = Files.newBufferedReader(propsFile.toPath()).lines().collect(
```
robert3005: you're leaking the handle here. You can just do `Files#readAllLines`

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BomConflictCheckTask.java
```diff
@@ -0,0 +1,149 @@
...
+                .getMavenBomProvider()
+                .getRecommendations();
+        List<Conflict> conflicts = new LinkedList<>();
+        Set<String> artifacts = BaselineVersions.getResolvedArtifacts(getProject());
+        Map<String, String> resolvedConflicts = new HashMap<>();
+        BaselineVersions.checkVersionsProp(getPropsFile(),
+                (propName, propVersion) -> {
+                    String regex = propName.replaceAll("\\*", ".*");
+                    artifacts.forEach(artifactName -> {
+                        if (artifactName.matches(regex)) {
```
robert3005: `stream().filter(...).forEach(...)`?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineVersions.java
j-baker: Don't outlaw multiple boms! It's been useful a number of times for me

robert3005: this logic https://github.com/nebula-plugins/nebula-dependency-recommender-plugin/blob/master/src/main/groovy/netflix/nebula/dependency/recommender/provider/RecommendationProviderContainer.java#L59 - the provider is always added upon creation of extension and subsequent invocations are just adding to the NEBULA_RECOMMENDER_BOM configuration

rubenfiszel: I see. I remember not understanding at first read why they treated the mavenbomrecommendation provider differently than the others (as in, not just adding it as yet another recommendation provider) but that must be it. Thanks!

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineVersions.java
```diff
@@ -53,23 +59,31 @@
...
         project.getPluginManager().apply(DependencyRecommendationsPlugin.class);
 
         // get dependencyRecommendations extension
-        RecommendationProviderContainer extension = project.getExtensions()
+        RecommendationProviderContainer extension = project
+                .getExtensions()
                 .getByType(RecommendationProviderContainer.class);
 
-        extension.setStrategy(RecommendationStrategies.OverrideTransitives); // default is 'ConflictResolved'
+        extension.setStrategy(RecommendationStrategies.OverrideTransitives); // default is 'ConflictResolved';
```
robert3005: revert this semicolon change

robert3005:
One last thing - tests please

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineVersions.java
```diff
@@ -53,23 +58,30 @@
...
 
+        extension.propertiesFile(ImmutableMap.of("file", rootVersionsPropsFile));
         // allow nested projects to specify their own nested versions.props file
         if (project != project.getRootProject() && project.file("versions.props").exists()) {
             extension.propertiesFile(ImmutableMap.of("file", project.file("versions.props")));
         }
+        project.getTasks().create("checkBomConflict", BomConflictCheckTask.class, rootVersionsPropsFile);
+        project.getTasks().create("checkNoUnusedPin", NoUnusedPinCheckTask.class, rootVersionsPropsFile);
+        project.getPluginManager().apply(BasePlugin.class);
+        project.getTasks().register("checkVersionsProps").configure(task ->
```
robert3005: break on `.configure`

robert3005: Actually, just use second parameter to register as shorthand

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineVersions.java
```diff
@@ -53,23 +58,30 @@
...
 
-        extension.propertiesFile(ImmutableMap.of("file", rootVersionsPropsFile));
+        File rootVersionsPropsFile = rootVersionsPropsFile(project);
 
+        extension.propertiesFile(ImmutableMap.of("file", rootVersionsPropsFile));
         // allow nested projects to specify their own nested versions.props file
         if (project != project.getRootProject() && project.file("versions.props").exists()) {
             extension.propertiesFile(ImmutableMap.of("file", project.file("versions.props")));
         }
+        project.getTasks().create("checkBomConflict", BomConflictCheckTask.class, rootVersionsPropsFile);
```
robert3005: should also use register instead of create

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineVersions.java
```diff
@@ -53,23 +58,30 @@
...
+        project.getTasks().create("checkNoUnusedPin", NoUnusedPinCheckTask.class, rootVersionsPropsFile);
+        project.getPluginManager().apply(BasePlugin.class);
+        project.getTasks().register("checkVersionsProps").configure(task ->
+                task.dependsOn("checkBomConflict", "checkNoUnusedPin"));
+        project.getTasks().getByName("check").dependsOn("checkVersionsProps");
```
robert3005: `named` instead of `getByName`

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineVersions.java
```diff
@@ -84,4 +96,54 @@ private static File rootVersionsPropsFile(Project project) {
         }
         return file;
     }
+
+    public static Set<String> getResolvedArtifacts(Project rootProject) {
+        Set<String> artifacts = new HashSet<>();
+        rootProject.getAllprojects().forEach(project -> {
```
robert3005: `allprojects(Action)` instead

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineVersions.java
```diff
@@ -84,4 +96,54 @@ private static File rootVersionsPropsFile(Project project) {
         }
         return file;
     }
+
+    public static Set<String> getResolvedArtifacts(Project rootProject) {
```
robert3005: this is not a root project so you need to make sure you're using a root project here first

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineVersions.java
```diff
@@ -84,4 +96,54 @@ private static File rootVersionsPropsFile(Project project) {
...
+        Set<String> artifacts = new HashSet<>();
```
robert3005: Use guava for this, Sets.newHashSet()

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineVersions.java
```diff
@@ -84,4 +96,54 @@ private static File rootVersionsPropsFile(Project project) {
...
+        rootProject.getAllprojects().forEach(project -> {
+            project.getConfigurations().stream().forEach(configuration -> {
```
robert3005: this should totally return the artifact set and not do a foreach adding to hashset

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineVersions.java
robert3005: And if you follow comment below than you would replace this with `rootProject.getAllProjects().stream().flatMap(...).collect(Collectors.toSet())`

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/NoUnusedPinCheckTask.java
```diff
@@ -0,0 +1,68 @@
...
+                (propName, propVersion) -> {
+                    String regex = propName.replaceAll("\\*", ".*");
+                    if (!artifacts.stream().anyMatch(artifact -> artifact.matches(regex))) {
+                        unusedProps.add(propName);
+                    }
+                    return null;
+                });
+
+        if (!unusedProps.isEmpty()) {
+            String unusedPropsString = unusedProps.stream().collect(Collectors.joining("\n"));
```
robert3005: Just use `String#join`

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BomConflictCheckTask.java
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    @InputFile
+    public final File getPropsFile() {
+        return propsFile;
+    }
+
+    @TaskAction
+    public final void checkBomConflict() {
+        final Map<String, String> recommendations = getMavenBomRecommendations();
```
robert3005: final shouldn't be necessary here - java should figure out that the value is effectively final

gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineVersionsIntegrationTest.groovy
```diff
@@ -0,0 +1,149 @@
...
+                        <artifactId>scala-library</artifactId>
+                        <version>2.12.5</version>
+                      </dependency>
+                    </dependencies>
+                  </dependencyManagement>
+                </project>
+            """.stripIndent()
+        Path bomPath = Paths.get(projectDir.toString(), "maven", "com", "palantir", "product", "your-bom", bomVersion)
+        java.nio.file.Files.createDirectories(bomPath)
+        Files.write(bomContent.getBytes(StandardCharsets.UTF_8), bomPath.resolve("your-bom-${bomVersion}.pom").toFile())
```
robert3005: actually - java.nio.file.Files is better here. No reason to use the guava equivalent

gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineVersionsIntegrationTest.groovy
```diff
@@ -0,0 +1,149 @@
...
+            compile 'org.scala-lang:scala-library'
+        }
+        
+        dependencyRecommendations {
+            mavenBom module: 'com.palantir.product:your-bom:${bomVersion}'
+        }
+        """.stripIndent()
+    }
+
+    def setupBom() {
```
robert3005: No need to split this out

robert3005 (commented):
A few style changes

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineVersions.java
dansanduleac: Why? Isn't that deprecated against `new HashSet<>()`?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineVersions.java
```diff
@@ -53,23 +59,30 @@
...
         // allow nested projects to specify their own nested versions.props file
         if (project != project.getRootProject() && project.file("versions.props").exists()) {
             extension.propertiesFile(ImmutableMap.of("file", project.file("versions.props")));
         }
+        project.getTasks().register("checkBomConflict", BomConflictCheckTask.class, rootVersionsPropsFile);
+        project.getTasks().register("checkNoUnusedPin", NoUnusedPinCheckTask.class, rootVersionsPropsFile);
+        project.getPluginManager().apply(BasePlugin.class);
+        project.getTasks().register("checkVersionsProps",
+                task -> task.dependsOn("checkBomConflict", "checkNoUnusedPin"));
+        project.getTasks().named("check").get().dependsOn("checkVersionsProps");
```
robert3005: don't do a get - configure instead

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BomConflictCheckTask.java
```diff
@@ -0,0 +1,155 @@
...
+                                resolvedConflicts.remove(entry.getKey());
+                            });
+                    return null;
+                });
+
+        //Critical conflicts are versions.props line that only override bom recommendations with same version
+        //so it should avoid considering the case where a wildcard also pin an artifact not present in the bom
+        List<Conflict> critical = conflicts.stream()
+                .filter(c -> c.getBomVersion().equals(c.getPropVersion()))
+                .filter(c -> !resolvedConflicts.containsValue(c.getPropName()))
```
robert3005: would be best to avoid `containsValue` since you already have a hashmap

robert3005: Can you make conflicts indexed over the artifact as well?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BomConflictCheckTask.java
```diff
@@ -0,0 +1,155 @@
...
+        }
+
+        public String getBomVersion() {
+            return bomVersion;
+        }
+
+        public String criticalString(Map<String, String> resolvedConflicts) {
+            if (!getBomVersion().equals(getPropVersion())) {
+                return "non critical: prop version not equals to bom version. (remove if unnecessary override)";
+            } else if (resolvedConflicts.containsValue(getPropName())) {
```
robert3005: using contains value indicates something is broken

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BomConflictCheckTask.java
```diff
@@ -0,0 +1,155 @@
...
+                    artifacts.stream()
+                            .filter(artifactName -> artifactName.matches(regex))
+                            .forEach(artifactName -> resolvedConflicts.put(artifactName, propName));
+
+                    recommendations.entrySet()
+                            .stream()
+                            .filter(entry -> entry.getKey().matches(regex))
+                            .forEach(entry -> {
+                                conflicts.add(new Conflict(propName, propVersion, entry.getKey(), entry.getValue()));
+                                resolvedConflicts.remove(entry.getKey());
```
robert3005: you already have an artifact here

gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineVersionsIntegrationTest.groovy
```diff
@@ -0,0 +1,144 @@
...
+
+    def setupVersionsProps(String propsContent) {
+        Files.write(projectDir.toPath().resolve("versions.props"), propsContent.getBytes(StandardCharsets.UTF_8))
+    }
+
+    def checkVersionsPropsSucceed() {
+        try {
+            with('checkVersionsProps', '--stacktrace').build()
+            ""
+        } catch (Exception e) {
```
robert3005: Can you split build and buildAndFail?

gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineVersionsIntegrationTest.groovy
```diff
@@ -0,0 +1,144 @@
...
+        }
+    }
+
+    def 'Override version conflict should succeed'() {
+        when:
+        setupVersionsProps("org.scala-lang:scala-library = 2.12.6")
+        buildFile << standardBuildFile(projectDir)
+
+        then:
+        checkVersionsPropsSucceed() == ""
```
robert3005: you should check that the task outcome is actually success

gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineVersionsIntegrationTest.groovy
robert3005: so don't do a catch and return empty strings

gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineVersionsIntegrationTest.groovy
```diff
@@ -0,0 +1,144 @@
...
+
+    }
+
+    def 'Unused version should fail'() {
+        when:
+        setupVersionsProps("notused:atall = 42.42")
+        buildFile << standardBuildFile(projectDir)
+
+        then:
+        checkVersionsPropsSucceed()
```
robert3005: check that the task you expect to fail actually failed

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BomConflictCheckTask.java
rubenfiszel: But why ? containsValue is somewhat inefficient but we have few values anyway.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BomConflictCheckTask.java
rubenfiszel: why ?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BomConflictCheckTask.java
rubenfiszel: ?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BomConflictCheckTask.java
robert3005: why is it a map then? You're just building a set

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BomConflictCheckTask.java
robert3005: you're using a hashmap as a set - let's either use a hashmap and not lookup over values or use a set

rubenfiszel: I need to be able to remove values based on the keys

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BomConflictCheckTask.java
rubenfiszel: I am not, I need to be able to remove values based on the key

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineVersions.java
```diff
@@ -84,4 +97,52 @@ private static File rootVersionsPropsFile(Project project) {
...
+                    if (line.equals("# linter:ON")) {
+                        active = true;
+                    } else if (line.equals("# linter:OFF")) {
+                        active = false;
+                    }
+                    if (active && line.length() > 0 && line.charAt(0) != '#' && line.matches(".*:.*\\s*=\\s*.*")) {
+                        String[] split = line.split("\\s*=\\s*");
+                        String propName = split[0];
+                        String propVersion = split[1];
+                        function.apply(propName, propVersion);
```
dansanduleac: This function should return a list rather than passing values to a function, and use a Stream for the maps and filters before converting to a list: 
```java
try (Stream<String> lines = Files.lines(...))
```

And rename the function to `readVersionsProps`

dansanduleac: Prefer to not have this statefulness about linter:ON / OFF, rather create an extension containing a set `whitelistedVersionPins` and ask people to add their 'unlintable' forces there.
e.g.
```
baselineVersions {
  whitelistedVersionPins 'foo:*' //adds the 'foo:*' pin to the list of version pins that won't be linted
}
```

dansanduleac (changes requested):

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineVersions.java
```diff
@@ -84,4 +97,52 @@ private static File rootVersionsPropsFile(Project project) {
...
+            project1.getConfigurations().stream().flatMap(configuration -> {
+                try {
+                    return configuration
+                            .getResolvedConfiguration()
+                            .getResolvedArtifacts().stream()
+                            .map(resolvedArtifact ->
+                                    resolvedArtifact.getModuleVersion().getId().getGroup() + ":"
+                                            + resolvedArtifact.getName());
+                } catch (IllegalStateException e) {
+                    //in every case so far, tnose IleegalStateException are ignorable. It's just the specific
```
dansanduleac: tnose -> those

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineVersions.java
```diff
@@ -84,4 +97,52 @@ private static File rootVersionsPropsFile(Project project) {
         }
         return file;
     }
+
+    public static Set<String> getResolvedArtifacts(Project project) {
+        return project.getRootProject().getAllprojects().stream().flatMap(project1 ->
+            project1.getConfigurations().stream().flatMap(configuration -> {
+                try {
```
dansanduleac: can we check Configuration.isCanBeResolved here?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineVersions.java
dansanduleac: I'm a bit wary of swallowing these exceptions, I think instead we should check for isCanBeResolved  and not catch this at all

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BomConflictCheckTask.java
dansanduleac: I see what you're trying to do here, but I find the add, then remove logic confusing.

The yes/no conditions could be collapsed into multiple .filter() invocations and I think that reads better:

```java
Map<String, String> resolvedConflicts = BaselineVersions.readVersionsProps(getPropsFile())
                .stream()
                .flatMap(pair -> {
                    String propName = pair.getLeft();
                    String propVersion = pair.getRight();
                    String regex = propName.replaceAll("\\*", ".*");

                    Set<String> recommendationConflicts = recommendations
                            .entrySet()
                            .stream()
                            .filter(entry -> entry.getKey().matches(regex))
                            .map(entry -> {
                                conflicts.add(new Conflict(propName, propVersion, entry.getKey(), entry.getValue()));
                                return entry.getKey();
                            })
                            .collect(Collectors.toSet());

                    return artifacts.stream()
                            .filter(artifactName -> artifactName.matches(regex))
                            .filter(artifactName -> !recommendationConflicts.contains(artifactName))
                            .map(artifactName -> Pair.of(artifactName, propName));
                })
                .collect(Collectors.toMap(Pair::getLeft, Pair::getRight));
```

robert3005: Please don't use apache commons pair - it's not a very java way.

rubenfiszel: Adressed the review. Kept the pair. What is the data structure that you suggest instead @robert3005 ?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BomConflictCheckTask.java
dansanduleac: luckily no need to remove values anymore!

dansanduleac (approved):

PR #346, opened 2018-08-27
Adds baseline-scalastyle plugin that configures sane scala defaults

robert3005:
Not attached to the scalastyle config here - this is slightly modified version of default one that's friendly for java interop (i.e. NullChecker is off)

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineConfig.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineConfig.java
@@ -59,12 +59,14 @@ class BaselineConfig extends AbstractBaselinePlugin {
                     copySpec.exclude("**/scalastyle_config.xml");
                     copySpec.setIncludeEmptyDirs(false);
                 });
-                rootProject.copy(copySpec -> {
-                    copySpec.from(rootProject.zipTree(configuration.getSingleFile())
-                            .filter(file -> file.getName().equals("scalastyle_config.xml")));
-                    copySpec.into(rootProject.getRootDir().toPath().resolve("project"));
-                    copySpec.setIncludeEmptyDirs(false);
-                });
+                if (rootProject.getAllprojects().stream().anyMatch(p -> p.getPluginManager().hasPlugin("scala"))) {
+                    rootProject.copy(copySpec -> {
+                        copySpec.from(rootProject.zipTree(configuration.getSingleFile())
+                                .filter(file -> file.getName().equals("scalastyle_config.xml")));
+                        copySpec.into(rootProject.getRootDir().toPath().resolve("project"));
+                        copySpec.setIncludeEmptyDirs(false);
+                    });
+                }
             });
         });
     }
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineConfig.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineConfig.java
@@ -37,23 +37,33 @@ class BaselineConfig extends AbstractBaselinePlugin {
 
         // users can still override this default dependency, it just reduces boilerplate
         Optional<String> version = Optional.ofNullable(getClass().getPackage().getImplementationVersion());
-        configuration.defaultDependencies(d -> d.add(project.getDependencies().create(String.format(
+        configuration.defaultDependencies(d -> d.add(rootProject.getDependencies().create(String.format(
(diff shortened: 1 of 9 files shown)


iamdanfox (commented):
My main concern with this is that adding support for another language to baseline might be interpreted as a sufficient 'blessing' that people start using it more widely.  (I really don't want to fragment the nice java homogeneity that we have internally)

In this specific case though, I think spark's commitment to scala forces our hand. Consolidating this config _somewhere_ is definitely preferrable to having it duplicated across repos.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineScalastyle.java
```diff
@@ -0,0 +1,98 @@
+/*
```
iamdanfox, line 1: General concern - does this addition make the `gradle-baseline-java` and `gradle-baseline-java-config` jar names semantically wrong now? Should we add `gradle-baseline-scala` / `gradle-baseline-scala-config`?

robert3005: uh, yes? Usually you have both java and scala in one project anyway. Should we move BaselineIdea/BaselineEclipse out to it's own subproject as well?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineScalastyle.java
```diff
@@ -0,0 +1,98 @@
...
+                            .findFirst()
+                            .orElseGet(() -> rootNode.appendNode(
+                                    "component", ImmutableMap.of("name", "ScalaCompilerConfiguration")));
+                    Node compilerOrder = (Node) scalaCompilerConf.getAt(new QName("option"))
+                            .stream()
+                            .filter(o -> ((Node) o).attributes().get("name").equals("compileOrder"))
+                            .findFirst()
+                            .orElseGet(() -> scalaCompilerConf.appendNode("option"));
+                    compilerOrder.attributes().put("name", "compileOrder");
+                    compilerOrder.attributes().put("value", "JavaThenScala");
```
robert3005: this has to be configurable in case you want mixed compilation mode

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineConfig.java
```diff
@@ -37,23 +37,33 @@ public void apply(Project rootProject) {
...
-                    copySpec.from(project.zipTree(configuration.getSingleFile()));
+                    copySpec.from(rootProject.zipTree(configuration.getSingleFile()));
                     copySpec.into(getConfigDir());
+                    copySpec.exclude("**/scalastyle_config.xml");
+                    copySpec.setIncludeEmptyDirs(false);
+                });
+                rootProject.copy(copySpec -> {
+                    copySpec.from(rootProject.zipTree(configuration.getSingleFile())
+                            .filter(file -> file.getName().equals("scalastyle_config.xml")));
+                    copySpec.into(rootProject.getRootDir().toPath().resolve("project"));
```
iamdanfox: Let's put a tiny comment to explain why we're putting stuff into "$rootProject/project/scalastyle_config.xml"

gradle-baseline-java/src/main/resources/META-INF/gradle-plugins/com.palantir.baseline-scalastyle.properties
```diff
@@ -0,0 +1 @@
+implemenetation-class=com.palantir.baseline.plugins.BaselineScalastyle
```
iamdanfox: s/implemenetation/implementation/

iamdanfox (changes requested):
Can we also make sure that pure java repos never notice the existence of a scalastyle file?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineConfig.java
uschi2000: does this mean that I'll have a new directory, `project` in every repo from now on? not a huge fan of that churn, given we have only a very small amount of scala repos.

robert3005: I pushed a fix for this - the file is only created if there exists a project that has scala plugin applied to it. Should also make sure that scala style is applied as well

iamdanfox (approved):

PR #348, opened 2018-08-27
Configurations are filecollections

robert3005:
as such they should be either inputfiles or classpath inputs


robert3005:
You can also see this being flagged by ValidateTaskProperties from gradle

iamdanfox (approved):

PR #351, opened 2018-08-29
[improvement] Avoid shaded guava from avro

a-k-g:
## Before this PR
Users could accidentally import repackaged guava classes such as `avro.shaded.com.google.common.collect.ImmutableMap`

## After this PR
The Checkstyle rule wouldn't allow them to do this.

### DIFF
--- a/gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
+++ b/gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
@@ -123,11 +123,17 @@
             <message key="import.illegal" value="Use JUnit 4-style (org.junit.*) test classes and assertions instead of JUnit 3 (junit.framework.*)."/>
         </module>
         <module name="IllegalImport"> <!-- Java Coding Guidelines: Import the canonical package -->
-            <property name="illegalPkgs" value="org.elasticsearch.common.base, jersey.repackaged.com.google.common, com.google.api.client.repackaged, org.apache.hadoop.thirdparty.guava, com.clearspring.analytics.util, org.spark_project.guava, avro.shaded.com.google.common"/>
+            <property name="illegalPkgs" value="org.elasticsearch.common.base, com.clearspring.analytics.util, org.spark_project.guava"/>
+            <message key="import.illegal" value="Must not import repackaged classes."/>
+        </module>
+        <module name="IllegalImport"> <!-- Java Coding Guidelines: Import the canonical package -->
+            <property name="illegalPkgs" value=".*.(repackaged|shaded|thirdparty)"/>
+            <property name="regexp" value="true" />
             <message key="import.illegal" value="Must not import repackaged classes."/>
         </module>
         <module name="IllegalImport">
-            <property name="illegalPkgs" value="^org\.gradle\..*internal"/>
+            <property name="illegalPkgs" value="^org.gradle..*internal"/>
+            <property name="regexp" value="true" />
             <message key="import.illegal" value="Do not rely on gradle internal classes as these may change in minor releases - use org.gradle.api versions instead."/>
         </module>
         <module name="IllegalImport">
--- a/gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
+++ b/gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
@@ -123,7 +123,7 @@
             <message key="import.illegal" value="Use JUnit 4-style (org.junit.*) test classes and assertions instead of JUnit 3 (junit.framework.*)."/>
         </module>
         <module name="IllegalImport"> <!-- Java Coding Guidelines: Import the canonical package -->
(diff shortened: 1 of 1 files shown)


gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
```diff
@@ -123,7 +123,7 @@
             <message key="import.illegal" value="Use JUnit 4-style (org.junit.*) test classes and assertions instead of JUnit 3 (junit.framework.*)."/>
         </module>
         <module name="IllegalImport"> <!-- Java Coding Guidelines: Import the canonical package -->
-            <property name="illegalPkgs" value="org.elasticsearch.common.base, jersey.repackaged.com.google.common, com.google.api.client.repackaged, org.apache.hadoop.thirdparty.guava, com.clearspring.analytics.util, org.spark_project.guava"/>
+            <property name="illegalPkgs" value="org.elasticsearch.common.base, jersey.repackaged.com.google.common, com.google.api.client.repackaged, org.apache.hadoop.thirdparty.guava, com.clearspring.analytics.util, org.spark_project.guava, avro.shaded.com.google.common"/>
```
j-baker: we should probably add `.*shaded.*, .*thirdparty.*` as illegal imports!

markelliot: `.*repackaged.*`, then, too

a-k-g: Do we want that to be done in this PR then?

j-baker: It would simplify the current rules considerably!

iamdanfox: I actually tried to use regexes to ban stuff for gradle internal, but it didn't work with out old version of checkstyle.  This might have changed now that we've upgraded checkstyle... can you screenshot this rule working?

a-k-g: Yes, I had to bump the checkstyle version. I used 7.8.2. I didn't realize that the repo had upgraded to 8.12.

gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
```diff
@@ -123,11 +123,17 @@
             <message key="import.illegal" value="Use JUnit 4-style (org.junit.*) test classes and assertions instead of JUnit 3 (junit.framework.*)."/>
         </module>
         <module name="IllegalImport"> <!-- Java Coding Guidelines: Import the canonical package -->
-            <property name="illegalPkgs" value="org.elasticsearch.common.base, jersey.repackaged.com.google.common, com.google.api.client.repackaged, org.apache.hadoop.thirdparty.guava, com.clearspring.analytics.util, org.spark_project.guava"/>
+            <property name="illegalPkgs" value="org.elasticsearch.common.base, com.clearspring.analytics.util, org.spark_project.guava"/>
+            <message key="import.illegal" value="Must not import repackaged classes."/>
+        </module>
+        <module name="IllegalImport"> <!-- Java Coding Guidelines: Import the canonical package -->
+            <property name="illegalPkgs" value=".*.(repackaged|shaded|thirdparty)"/>
```
j-baker: is this a prefix, or do we need a suffix too?

a-k-g: We don't need a suffix

markelliot: Should the second dot be escaped?

markelliot: Would be ideal if we had tests for this.

a-k-g: Escaping it actually breaks the rule.

a-k-g:
image
Screenshot for the gradle internal ^

gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
markelliot: Is the “.*.” component even necessary then? (ie is this actually a regex?)

a-k-g: As per checkstyle docs, it is treated as a regex when we set regexp to true. From testing it out, the `.*.` component does seem necessary.

a-k-g: Seems like I was mistaken about the dot escaping. It works without escaping the dot, but escaping it doesn't break the rule.

markelliot: I think it needs to be escaped or these will also match “xrepackaged”

iamdanfox (approved):

iamdanfox:
Thanks for pushing this through @a-k-g!

JacekLach:
```

import com.palantir.codex.io.arrow20.ArrowDataframeTimeseriesIoFactory;
import com.palantir.shaded.arrow.v020.org.apache.arrow.memory.RootAllocator;
```


:( Can we make this exclude `com.palantir` shaded things?

PR #354, opened 2018-09-03
[break] Errorprone gradle integration using Xplugin option

### LINKED ISSUE #350: Switch to https://github.com/tbroyer/gradle-errorprone-javacplugin-plugin
Uses errorprone as a java plugin which should improve performance compared to the current one

robert3005:
Replaces `net.ltgt.errorprone` with the newer `net.ltgt.errorprone-javacplugin`.  This should drastically improve compilation speed (original issue: https://github.com/google/error-prone/issues/994).

fixes #350 

## Manual-tasks

- users must upgrade to `gradle.plugin.org.inferred:gradle-processors:1.2.18` to avoid seeing `error: plug-in not found: ErrorProne`
- users must change their suppressions:

```diff
 tasks.withType(JavaCompile).configureEach {
-    options.compilerArgs += ['-Xep:Slf4jLogsafeArgs:OFF']
+    options.errorprone.errorproneArgs += ['-Xep:Slf4jLogsafeArgs:OFF']
 }
```


robert3005:
This is potentially a break since people will have to configure errorprone options differently using the configuration provided by this plugin

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
```diff
@@ -21,47 +21,73 @@
...
+            project.getPluginManager().apply(ErrorProneJavacPluginPlugin.class);
+            JavaPluginConvention javaConvention = project.getConvention().getPlugin(JavaPluginConvention.class);
+            String version = Optional.ofNullable(getClass().getPackage().getImplementationVersion())
+                    .orElse("latest.release");
+            project.getDependencies().add(
+                    ErrorProneJavacPluginPlugin.CONFIGURATION_NAME,
+                    "com.palantir.baseline:baseline-error-prone:" + version);
+
+
+            // This plugin interferes with gradles native annotation processor path configuration so need to configure
```
j-baker: any chance you could just fix the processors plugin?

robert3005: this is to avoid lockstep upgrading if you use that plugin and this. The real fix is to switch to gradles native annotationprocessor handling

ferozco (approved):

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
robert3005: In particular from gradle 4.6 you can do https://docs.gradle.org/current/userguide/java_plugin.html#example_declaring_annotation_processors which should do all the things the plugin does

robert3005: Actually - will push a fix there since this is a gradle feature in general

robert3005: Created https://github.com/palantir/gradle-processors/pull/81 which should allow us to remove this code

robert3005:
Fixes issues reported in https://github.com/google/error-prone/issues/994

PR #357, opened 2018-09-04
[improvement] Checkstyle bans ObjectMapper#findAndRegisterModules()

iamdanfox:
## Before this PR

A few products internally use ObjectMapper's `findAndRegisterModules` method.  This resulted in changed behaviour when I upgraded jackson version and different modules appeared on the classpath.

https://github.com/FasterXML/jackson-databind/blob/jackson-databind-2.9.6/src/main/java/com/fasterxml/jackson/databind/ObjectMapper.java#L1027

## After this PR

This method is banned because I think it is a foot-gun.


ferozco (approved):
Can we add an issue tracking that this should be removed in favour of a more robust solution

iamdanfox:
Will open the issue if it turns out we get a false positive on this, otherwise I think this approach is actually kinda OK.

PR #358, opened 2018-09-05
Chevrons are illegal characters as values of xml properties

robert3005:
@iamdanfox


ferozco (approved):

ferozco:
we can follow up with tests later

PR #359, opened 2018-09-05
Only register checkNoUnusedPin task if current project is root + task fix

rubenfiszel:
This task only makes sense as a rootProject's task

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineVersions.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineVersions.java
@@ -72,19 +72,21 @@ public final class BaselineVersions implements Plugin<Project> {
         File rootVersionsPropsFile = rootVersionsPropsFile(project);
         extension.propertiesFile(ImmutableMap.of("file", rootVersionsPropsFile));
 
+        project.getTasks().register("checkBomConflict", BomConflictCheckTask.class, rootVersionsPropsFile);
         if (project != project.getRootProject()) {
             // allow nested projects to specify their own nested versions.props file
             if (project.file("versions.props").exists()) {
                 extension.propertiesFile(ImmutableMap.of("file", project.file("versions.props")));
             }
+            project.getTasks().register("checkVersionsProps",
+                task -> task.dependsOn("checkBomConflict"));
         } else {
             project.getTasks().register("checkNoUnusedPin", NoUnusedPinCheckTask.class, rootVersionsPropsFile);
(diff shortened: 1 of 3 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineVersions.java
```diff
@@ -70,14 +70,18 @@ public void apply(Project project) {
...
+            }
+        } else {
+            project.getTasks().register("checkNoUnusedPin", NoUnusedPinCheckTask.class, rootVersionsPropsFile);
         }
+
         project.getTasks().register("checkBomConflict", BomConflictCheckTask.class, rootVersionsPropsFile);
-        project.getTasks().register("checkNoUnusedPin", NoUnusedPinCheckTask.class, rootVersionsPropsFile);
         project.getPluginManager().apply(BasePlugin.class);
         project.getTasks().register("checkVersionsProps",
                 task -> task.dependsOn("checkBomConflict", "checkNoUnusedPin"));
```
robert3005: isn't this wrong now?

rubenfiszel: ?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/NoUnusedPinCheckTask.java
```diff
@@ -38,7 +38,9 @@ public NoUnusedPinCheckTask(File propsFile) {
 
     @Input
     public final Set<String> getResolvedArtifacts() {
-        return BaselineVersions.getResolvedArtifacts(getProject());
+        return getProject().getAllprojects().stream()
```
robert3005, line 41: I am pretty sure that this is infinite recursion as getAllProjects returns the project itself.

rubenfiszel:
That's an interesting remark. I didn't think of it. Fortunately, I tested
it and it works. I also think that would be true only if a project
registered as part of getAllProject after resolving every Gradle fields but
that is probably not true as there is no real reason of requiring that from
the Gradle standpoint.

robert3005:
Nah, I confused the flow - I thought we're returning to the same function but we're not. Sorry.

robert3005 (approved):

iamdanfox (approved):
LGTM.  I think now that we've got some non-trivial functionality here, I'd like to get into the pattern of always adding some tests that demonstrate the tasks passing and failing, to make sure we don't unintentionally change stuff!  I think we can add a nebula.test rig in a separate PR.

PR #365, opened 2018-09-10
[pure-refactor] Use String#format instead of string concatenation

robert3005:


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BomConflictCheckTask.java
```diff
@@ -93,8 +93,8 @@ public final void checkBomConflict() {
                 .collect(Collectors.toList());
 
         if (!conflicts.isEmpty()) {
-            System.out.println("There are conflicts between versions.props and the bom:");
-            System.out.println(conflictsToString(conflicts, resolvedConflicts));
```
iamdanfox, line 97: Filed https://github.com/palantir/gradle-baseline/issues/366 so we catch this automatically next time

iamdanfox (approved):

PR #368, opened 2018-09-10
Readme

iamdanfox:
**[Rendered version](https://github.com/palantir/gradle-baseline/blob/readme/README.md)**

## Before this PR

- Our README didn't have a clear structure, 
- inconsistent titles
- overview didn't actually mention newer functionality
- claimed compatibility with gradle 4.0 when it actually fails below 4.9

## After this PR

- README now has a quick overview table and a section per plugin


README.md
```diff
@@ -230,8 +131,30 @@ checks](https://errorprone.info):
...
+```Java
+@SuppressWarnings("checkstyle:magicnumber")
+```
+
+Checkstyle rules can also be suppressed using comments, which is useful for checks such as `IllegalImport` where
+annotations cannot be used to suppress the violation. To suppress checks for particular lines, add the comment
+`// CHECKSTYLE:OFF` before the first line to suppress and add the comment `// CHECKSTYLE:ON` after the last line.
 
+To disable certain checks for an entire file, apply [custom suppressions](http://checkstyle.sourceforge.net/config.html)
+in `.baseline/checkstyle/checkstyle-suppressions`.
```
dansanduleac: checkstyle-suppressions.xml

README.md
```diff
@@ -294,7 +216,8 @@ dependencyRecommendations {
 }
 ```
 
-### Copyright Checks
+Adds the following tasks:
 
-By default Baseline enforces Palantir copyright at the beginning of files. To change this, edit the template copyright
-in `.baseline/copyright/*.txt` and the RegexpHeader checkstyle configuration in `.baseline/checkstyle/checkstyle.xml`
+- `checkVersionsProps` - A catch-all task to lint your versions.props file.
+- `checkBomConflict` - Ensures your versions.props don't conflict with versions sourced from a BOM.
```
dansanduleac: can we say, ... "your versions.props pins don't force the same version that is already recommended by a BOM"?

dansanduleac (approved):

PR #370, opened 2018-09-11
[improvement] Ignore palantir-prefixed shaded jars

JacekLach:
If a project purposefully uses a shaded dependency, slapping `//CHECKSTYLE:OFF IllegalImport` everywhere is tedious. It seems unlikely that there would be a `import com.palantir.shaded.` thing that is likely used by mistake.

## Before this PR

Any imports of internally shaded libs, like `com.palantir.shaded.foo`, would trigger the checkstyle failure.

## After this PR

Palantir-specific shaded imports are no longer flagged.


j-baker:
Internally, any chance you could send me cases where this is a problem?

j-baker:
Probably the only thing worse than shading is relying on someone else's shaded deps...

JacekLach:
On recommendation resolved this via a suppression

PR #371, opened 2018-09-11
[fix] don't fail critically for wildcard forcing versions outside of bom

rubenfiszel:


dansanduleac (approved):

dansanduleac:
I'd like to make this also lint `versions.props` inside subprojects, rather than only the root level one, but for now this is fine.

PR #376, opened 2018-09-19
[fix] checkstyle path for Checkstyle-IDEA 5.21.1+

dtobin:
## Before this PR

As of Checkstyle-IDEA 5.21.1, $PRJ_DIR$ is replaced with the IDEA
variable $PROJECT_DIR$, which does not include a trailing slash.
Thus a trailing slash is required when substituting $PRJ_DIR$ into a
path. Baseline currently generates an invalid checkstyle.xml path
for Checkstyle-IDEA 5.21.1+.

## After this PR

Add the missing trailing slash when producing the checkstyle.xml path.

Because the use of $PRJ_DIR$ in Checkstyle-IDEA is deprecated, also
switch to using $PROJECT_DIR$.

See https://github.com/jshiell/checkstyle-idea/issues/404


dtobin:
Workaround for other people running into this issue:
```
ln -s {project_dir}/.baseline {project_dir}.baseline
```
for all of your projects :)

robert3005 (approved):

PR #380, opened 2018-09-21
Handle versions.props conflicts

### LINKED ISSUE #379: Duplicate key error when checking artifact versions
Given the following `versions.props`:
```
com.fasterxml.jackson.*:* = 2.9.6
com.fasterxml.jackson.datatype:jackson-datatype-jdk7 = 2.6.7
```

Running `checkBomConflict` will result in the following error:
```
> Duplicate key com.fasterxml.jackson.*:*
```

This is because we do not supply a merging function to the map collector here:
https://github.com/palantir/gradle-baseline/blob/0.32.3/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BomConflictCheckTask.java#L99

Many times these conflicts are indicative of a non-ideal version props, but in certain cases overriding specific subpackages is desired. Given the example above, the only alternative is to individually specify each of `com.fasterxml.jackson.core:*`, `com.fasterxml.jackson.databind:*`, etc.

Ideally we would have a way to identify the "more specific" version of a particular artifact. However it's probably reasonable to assume that a single artifact will never map to more than 2 entries. So it would be sufficient to resolve conflicts only if one of the entries exactly matches the artifact.

pkoenig10:
Fixes #379

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BomConflictCheckTask.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BomConflictCheckTask.java
@@ -19,12 +19,10 @@ package com.palantir.baseline.plugins;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Lists;
 import java.io.File;
-import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
-import java.util.stream.Stream;
 import javax.inject.Inject;
 import netflix.nebula.dependency.recommender.DependencyRecommendationsPlugin;
 import netflix.nebula.dependency.recommender.provider.RecommendationProviderContainer;
@@ -102,14 +100,7 @@ public class BomConflictCheckTask extends DefaultTask {
                         Pair::getLeft,
                         Pair::getRight,
                         // Resolve conflicts by choosing the longer entry because it is more specific
-                        (propName1, propName2) -> {
-                            if (propName1.equals(propName2)) {
-                                throw new RuntimeException("Duplicate versions.props entry: " + propName1);
-                            }
-                            return Stream.of(propName1, propName2)
-                                    .max(Comparator.comparingLong(String::length))
-                                    .get();
-                        }));
+                        (propName1, propName2) -> propName1.length() > propName2.length() ? propName1 : propName2));
 
         Set<String> versionPropConflictingLines = ImmutableSet.copyOf(resolvedConflicts.values());
 
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BomConflictCheckTask.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BomConflictCheckTask.java
@@ -23,6 +23,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import javax.inject.Inject;
 import netflix.nebula.dependency.recommender.DependencyRecommendationsPlugin;
 import netflix.nebula.dependency.recommender.provider.RecommendationProviderContainer;
@@ -96,7 +97,15 @@ public class BomConflictCheckTask extends DefaultTask {
                             .filter(artifactName -> !recommendationConflicts.contains(artifactName))
                             .map(artifactName -> Pair.of(artifactName, propName));
                 })
-                .collect(Collectors.toMap(Pair::getLeft, Pair::getRight));
+                .collect(Collectors.toMap(
+                        Pair::getLeft,
+                        Pair::getRight,
+                        // Resolve conflicts by choosing the entry that exactly matches the artifact
+                        (propName1, propName2) -> Stream.of(propName1, propName2)
+                                .filter(propName -> !propName.contains("\\*"))
+                                .findFirst()
+                                .orElseThrow(() -> new IllegalStateException(
+                                        String.format("Conflicts in versions.props:\n%s\n%s", propName1, propName2)))));
 
         Set<String> versionPropConflictingLines = ImmutableSet.copyOf(resolvedConflicts.values());
 
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BomConflictCheckTask.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BomConflictCheckTask.java
@@ -19,6 +19,7 @@ package com.palantir.baseline.plugins;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Lists;
 import java.io.File;
+import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineVersionsIntegrationTest.groovy
```diff
@@ -151,6 +151,30 @@ class BaselineVersionsIntegrationTest  extends AbstractPluginTest {
...
+            compile 'com.fasterxml.jackson.core:jackson-core'
+        }"""
+
+        then:
+        buildAndFailWith("")
+    }
+
+    def 'Version props conflict with exact artifact should succeed'() {
+        when:
+        setupVersionsProps("com.fasterxml.jackson.*:* = 2.9.6\ncom.fasterxml.jackson.core:jackson-core = 2.9.5")
```
qinfchen: what happens if you have `com.fasterxml.jackson.core:jackson-core = 2.9.6\ncom.fasterxml.jackson.core:jackson-core = 2.9.5"`

rubenfiszel:
@pkoenig10 That's a cool idea, and I have a suggestion. Pick the most specialized pattern by counting the number of artifacts it corresponds to instead of `findFist`.

pkoenig10:
@rubenfiszel that's a bit difficult to do here, it would require creating an intermediate map that mapped prop -> artifact.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BomConflictCheckTask.java
```diff
@@ -96,7 +97,15 @@ public final void checkBomConflict() {
                             .filter(artifactName -> !recommendationConflicts.contains(artifactName))
                             .map(artifactName -> Pair.of(artifactName, propName));
                 })
-                .collect(Collectors.toMap(Pair::getLeft, Pair::getRight));
+                .collect(Collectors.toMap(
+                        Pair::getLeft,
+                        Pair::getRight,
+                        // Resolve conflicts by choosing the entry that exactly matches the artifact
+                        (propName1, propName2) -> Stream.of(propName1, propName2)
+                                .filter(propName -> !propName.contains("\\*"))
```
rubenfiszel: @JacekLach proposed an elegant solution to just pick the longest one since it is the most specialized:
We could just replace this by 
```
                .map(propName -> Pair.of(propName, propName.length)))
                .max(Comparator.comparing(Pair::getRight))
                .findFirst().get().getLeft()
```

gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineVersionsIntegrationTest.groovy
```diff
@@ -151,6 +151,30 @@ class BaselineVersionsIntegrationTest  extends AbstractPluginTest {
...
+    def 'Version props conflict without exact artifact should fail'() {
+        when:
+        setupVersionsProps("com.fasterxml.jackson.*:* = 2.9.6\ncom.fasterxml.jackson.core:* = 2.9.5")
+        buildFile << standardBuildFile(projectDir) + """
+        dependencies {
+            compile 'com.fasterxml.jackson.core:jackson-core'
+        }"""
+
+        then:
+        buildAndFailWith("")
```
rubenfiszel: With the proposed solution that wouldn't fail anymore :)

gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineVersionsIntegrationTest.groovy
rubenfiszel: With the solution proposed above, that wouldn't fail but it probably should. We should likely fail on conflicts that are as specialized.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BomConflictCheckTask.java
rubenfiszel: Actually, regarding @qinfchen comments below. We probably should sort every conflict for the artifact and pick the max if it's a proper max (no other prop of the same length/as specialized), fail otherwise.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BomConflictCheckTask.java
```diff
@@ -100,12 +101,15 @@ public final void checkBomConflict() {
...
-                                .filter(propName -> !propName.contains("\\*"))
-                                .findFirst()
-                                .orElseThrow(() -> new IllegalStateException(
-                                        String.format("Conflicts in versions.props:\n%s\n%s", propName1, propName2)))));
+                        // Resolve conflicts by choosing the longer entry because it is more specific
+                        (propName1, propName2) -> {
+                            if (propName1.equals(propName2)) {
+                                throw new RuntimeException("Duplicate versions.props entry: " + propName1);
+                            }
+                            return Stream.of(propName1, propName2)
```
rubenfiszel: `propName1.length > propName2.length ? propName1 : propName2`
:D

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BomConflictCheckTask.java
```diff
@@ -100,12 +101,15 @@ public final void checkBomConflict() {
...
                         Pair::getRight,
-                        // Resolve conflicts by choosing the entry that exactly matches the artifact
-                        (propName1, propName2) -> Stream.of(propName1, propName2)
-                                .filter(propName -> !propName.contains("\\*"))
-                                .findFirst()
-                                .orElseThrow(() -> new IllegalStateException(
-                                        String.format("Conflicts in versions.props:\n%s\n%s", propName1, propName2)))));
+                        // Resolve conflicts by choosing the longer entry because it is more specific
+                        (propName1, propName2) -> {
+                            if (propName1.equals(propName2)) {
```
rubenfiszel: The idea behind the check is good but since that function is applied 2-by-2 but not exhaustively and the keys are not sorted  it's possible that some duplicates slip through. We can always save the previous stream as a list and apply separately a check for that and convert to a map.

pkoenig10: I'd rather just remove this check. The dependency recommender plugin throws the versions into a `Properties` so it will just take the last one.

pkoenig10:
It seems to me like we are duplicating a lot of work of the dependency recommender here. Why don't we just get the recommended dependencies using only the bom and using only the properties file and then check for conflicts?

rubenfiszel:
I believe the reason I don't do that is that it's not easy to get
recommendations from the bom without the props, and then apply later the
props. Configurations have a lifecycle for their dependency resolution.

iamdanfox (approved):

dansanduleac:
Sorry I didn't look at this earlier, but this is actually incorrect, because that's not how nebula picks the override. It just picks the latest override that matches the artifact.

pkoenig10:
@dansanduleac yes that's right. Avoiding these inconsistencies is why I suggested we just use the recommended dependencies here instead of trying to parse the props file ourselves.

rubenfiszel:
@dsanduleac Not sure but if that's the case, assuming the collection is
traversed in order, it should be enough to always pick the second element?

PR #383, opened 2018-09-24
[fix] applies the baseline-format plugin only when java plugin is applied

qinfchen:
## Before this PR
Always applies the spotless plugin even if the java plugin is not applied

## After this PR
Only applies the spotless plugin when the java plugin is applied.

### DIFF
--- a/gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineFormatIntegrationTest.groovy
+++ b/gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineFormatIntegrationTest.groovy
@@ -48,7 +48,7 @@ class BaselineFormatIntegrationTest extends AbstractPluginTest {
         with('format', '--stacktrace').build()
     }
 
-    def 'cannot apply plugin when java plugin is missing'() {
+    def 'cannot run format task when java plugin is missing'() {
         when:
         buildFile << noJavaBuildFile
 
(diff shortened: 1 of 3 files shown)


gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineFormatIntegrationTest.groovy
```diff
@@ -0,0 +1,73 @@
...
+
+    def 'can apply plugin'() {
+        when:
+        buildFile << standardBuildFile
+
+        then:
+        with('format', '--stacktrace').build()
+    }
+
+    def 'cannot apply plugin when java plugin is missing'() {
```
iamdanfox: tweak

gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineFormatTest.groovy
```diff
@@ -0,0 +1,45 @@
...
+    def setup() {
+        project = ProjectBuilder.builder().build()
+        project.plugins.apply 'java'
+        project.plugins.apply BaselineFormat
+        project.evaluate()
+    }
+
+    def baselineFormatPluginApplied() {
+        expect:
+        assertTrue project.plugins.hasPlugin(BaselineFormat.class)
```
iamdanfox: Not sure these tests are really worth it - you apply it above and now just check it's applied?? I don't anticipate this catching bugs.

qinfchen: oops, updated

iamdanfox (approved):

PR #388, opened 2018-09-28
[improvement] Disallow System.out and System.err.

jkozlowski:
## Before this PR
It is possible to ship code that has System.out and System.err usage.

## After this PR
It is not possible to ship code that has System.out and System.err usage.

## Worries
When there are some CLIs that need it, the will need to disable this rule, which is probably the right tradeoff.

### DIFF
--- a/gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
+++ b/gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
@@ -335,8 +335,15 @@
             <property name="ignoreComments" value="true"/>
         </module>
         <module name="RegexpSinglelineJava">
+            <property name="id" value="BanSystemOut"/>
             <property name="format" value="System\.out\."/>
-            <property name="message" value="System.out usage is not generally allowed"/>
+            <property name="message" value="System.out usage is not generally allowed. Please use appropriate logging methods"/>
+            <property name="ignoreComments" value="true"/>
+        </module>
+        <module name="RegexpSinglelineJava">
+            <property name="id" value="BanSystemErr"/>
+            <property name="format" value="System\.err\."/>
+            <property name="message" value="System.err usage is not generally allowed. Please use appropriate logging methods"/>
             <property name="ignoreComments" value="true"/>
         </module>
         <module name="RegexpSinglelineJava">
--- a/gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
+++ b/gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
@@ -335,6 +335,11 @@
             <property name="ignoreComments" value="true"/>
         </module>
         <module name="RegexpSinglelineJava">
+            <property name="format" value="System\.out\."/>
+            <property name="message" value="System.out usage is not generally allowed"/>
+            <property name="ignoreComments" value="true"/>
+        </module>
+        <module name="RegexpSinglelineJava">
             <property name="format" value="\bCharsets\."/>
             <property name="message" value="Use JDK StandardCharsets instead of alternatives."/>
         </module>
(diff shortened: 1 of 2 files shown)


gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
```diff
@@ -334,6 +334,11 @@
             <property name="message" value="printStackTrace is not generally allowed."/>
             <property name="ignoreComments" value="true"/>
         </module>
+        <module name="RegexpSinglelineJava">
```
iamdanfox, line 337: Please give this an `id` so that it can be suppressed using an `@SuppressWarnings` annotation, e.g.

```
<property name="id" value="BanSystemOut"/>
```

gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
```diff
@@ -334,6 +334,11 @@
...
+            <property name="format" value="System\.out\."/>
+            <property name="message" value="System.out usage is not generally allowed"/>
```
iamdanfox: The error message should explain why this is bad and what people should do instead.

iamdanfox:
I'd kinda prefer to make this a main-only checkstyle rule, as I think System.out has legitimate reasons to exist in occasional test code.

Also what about banning `System.err`?

gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
jkozlowski: Nice, I didn't know how to do that.

jkozlowski: Ok done, have a look

gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
jkozlowski: I added a reference to logging, have a look.

jkozlowski:
@iamdanfox can you give some examples of when you would use System.out in test code? You mean like unit tests?

gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
```diff
@@ -334,6 +334,18 @@
             <property name="message" value="printStackTrace is not generally allowed."/>
             <property name="ignoreComments" value="true"/>
         </module>
+        <module name="RegexpSinglelineJava">
+            <property name="id" value="BanSystemOut"/>
+            <property name="format" value="System\.out\."/>
+            <property name="message" value="System.out usage is not generally allowed. Please use appropriate logging methods"/>
```
iamdanfox: I still don't think this error message gives an uninformed user enough information about what they should do (or why it's important). What about something like:

```
Logging with System.out is not allowed because it has no metadata and can't be configured at runtime. Please use an SLF4J logger instead, e.g. log.info("Message").
```

jkozlowski:
Cool, will do that on Monday

jkozlowski:
Although tbh, not sure why we’re aiming for an uninformed user. I used the same message as most of the other checks

jkozlowski:
@iamdanfox ping here

jkozlowski:
@iamdanfox ping here :)

iamdanfox:
Hey @jkozlowski I've seen System.out.println lines mostly used as quick debugging aids to sanity check things in unit tests (e.g. "started", "running", "stopped"), especially when there isn't a proper SLF4J implementation in tests.

Empirically, searching all the repos I have checked out I've found very few main usages (except in CLIs) and quite a lot of test usages.  If you want to enforce the status quo, then I think banning System.out.println in main usage would be sufficient.  I'm mainly concerned that banning System.out.println in test code will annoy people without much upside.

jkozlowski:
righty-ho @iamdanfox, done! Can we get a release?

iamdanfox (approved):

iamdanfox:
0.35.0 building now: https://circleci.com/gh/palantir/gradle-baseline/1386, excavator should roll it out automatically for you :)

PR #391, opened 2018-10-05
[improvement] Update errorprone plugin to reduce false positive logging on missing javac configuration

robert3005:


iamdanfox (approved):
Sounds good, just looks like some classes also got moved

PR #397, opened 2018-10-08
[improvement] Detect possible NPEs using Uber's NullAway error-prone plugin

### LINKED ISSUE #273: Consider adding the uber/NullAway error-prone plugin
After removing Findbugs from the tritium repo in https://github.com/palantir/tritium/pull/76, it seems like we might actually now be exposed to a few more bugs.

Apparently https://github.com/uber/NullAway catches silly mistakes like:

```java
static void log(Object x) {
    System.out.println(x.toString());
}
static void foo() {
    log(null);
}
```

cc @schlosna 

Other things worth investigating:

- google/error-prone#44
- google/error-prone#280
- https://checkerframework.org/

iamdanfox:
https://github.com/uber/NullAway

## Before this PR

People could write silly code containing obvious null pointer bugs which would compile and pass all our current static analysis.

## After this PR

Anyone who applies `com.palantir.baseline-error-prone` would get Uber's NullAway checks for free. Initially these would just be flagged as warnings, when we've got a bit of confidence and fixed [silly things](https://github.com/palantir/safe-logging/pull/70) then I'd like to upgrade them to errors.

fixes #273


dansanduleac:
// excavate me pls

PR #413, opened 2018-10-10
checkVersionsProps can --fix problematic lines

dansanduleac:
## Before this PR

* `checkBomConflict` task was created on all subprojects
* `checkBomConflict`, `checkNoUnusedPin` would report errors but can't fix them automatically

## After this PR

* `checkBomConflict` (and `checkVersionsProps` which depends on the other two checks) only created in root project. This means that only BOM recommendations that are applied to root project will be checked. This is not a problem if configuring `dependencyRecommendations` in `allprojects {}`.
* `checkBomConflict`, `checkNoUnusePin`, `checkVersionsProps` all get a `--fix` option. Running like this will rewrite your `versions.props` file to get rid of the critical BOM conflicts and/or unused pins.
* add troubleshooting block to `com.palantir.gradle-version` description in the readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,6 +223,8 @@ Adds the following tasks:
 - `checkBomConflict` - Ensures your versions.props pins don't force the same version that is already recommended by a BOM.
 - `checkNoUnusedPin` - Ensures all versions in your versions.props correspond to an actual gradle dependency.
 
+Run `./gradlew checkVersionsProps --fix` to solve the problems flagged by the above tasks.
+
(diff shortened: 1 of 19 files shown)


iamdanfox (approved):
Looks solid. Let's do a final publishToMavenLocal and make sure this behaves on some real-world projects (and also add it to excavator!)

gregakinman:
Just want to say `--fix` is super slick - thanks guys 😄.

PR #418, opened 2018-10-11
Fix BaselineVersions regression

dansanduleac:
## Before this PR

#413 introduced a regression where we made `check` depend on a non-existent task in subprojects.

## After this PR

Fixed the regression.


robert3005 (approved):

dansanduleac:
plugins.gradle.org keeps flaking with 502s :(

PR #419, opened 2018-10-11
[improvement] baseline-format collects java files from all source sets

dansanduleac:
## Before this PR

baseline-format only targets java files in the main and test source dirs

## After this PR
 
it targets `*.java` files in all source dirs

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineFormat.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineFormat.java
@@ -36,7 +36,7 @@ class BaselineFormat extends AbstractBaselinePlugin {
                         .getPlugin(JavaPluginConvention.class)
                         .getSourceSets()
                         .stream()
-                        .map(SourceSet::getJava)
+                        .map(SourceSet::getAllJava)
                         .toArray();
 
                 java.target(allJavaFiles);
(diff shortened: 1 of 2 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineFormat.java
```diff
@@ -29,8 +31,15 @@ public void apply(Project project) {
             project.getPluginManager().apply("com.diffplug.gradle.spotless");
 
             project.getExtensions().getByType(SpotlessExtension.class).java(java -> {
-                // TODO(dfox): apply this to all source sets, not just 'main' and 'test'
-                java.target("src/main/java/**/*.java", "src/main/test/**/*.java");
+                Object[] allJavaFiles = project
```
robert3005: is this the right way to wire it up?

dansanduleac: it takes vararg objects, and accepts FileCollection as one of the possible types. I'm passing in SourceDirectorySet objects which <: FileTree <: FileCollection

iamdanfox (approved):

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineFormat.java
```diff
@@ -29,8 +31,15 @@ public void apply(Project project) {
...
             project.getExtensions().getByType(SpotlessExtension.class).java(java -> {
-                // TODO(dfox): apply this to all source sets, not just 'main' and 'test'
-                java.target("src/main/java/**/*.java", "src/main/test/**/*.java");
+                // Configure a lazy FileCollection then pass it as the target
+                ConfigurableFileCollection allJavaFiles = project.files();
+                project
+                        .getConvention()
+                        .getPlugin(JavaPluginConvention.class)
+                        .getSourceSets()
+                        .all(sourceSet -> allJavaFiles.from(sourceSet.getAllJava()));
```
qinfchen, line 40: @dansanduleac, this breaks all internal projects with generated objects.
```
sourceSets {
    generatedObjects
    generatedJersey
}
```

PR #423, opened 2018-10-16
Fix versions.props handling of partially overlapping forces

### LINKED ISSUE #379: Duplicate key error when checking artifact versions

dansanduleac:
Fixes #379 properly

## Before this PR

In `checkBomConflict`, multiple forces from `versions.props` that match the same artifact would pick the longest force as the one that was "used".
This is not always the the one that nebula picks, because nebula just picks the last line that matches the artifact.

A similar mistake occurs in `checkNoUnusedPin` where _every_ force is individually treated as used if it matches even a single artifact, disregarding other forces in the file that might have overridden it.

## After this PR

* `checkBomConflict`: Resolve conflicts by associating an artifact with the last `versions.props` line that matches it.
* `checkNoUnusedPin`: when considering a resolved artifact, only treat the *last* matching force as used.

This now matches nebula's behaviour.

### DIFF
--- a/gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineVersionsIntegrationTest.groovy
+++ b/gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineVersionsIntegrationTest.groovy
@@ -97,9 +97,9 @@ class BaselineVersionsIntegrationTest  extends AbstractPluginTest {
         result
     }
 
-    def buildAndFailWith(String error) {
+    void buildAndFailWith(String error) {
         BuildResult result = with('checkVersionsProps', '--stacktrace').buildAndFail()
-        result.output.contains(error)
+        assert result.output.contains(error)
     }
 
     def buildWithFixWorks() {
@@ -162,7 +162,8 @@ class BaselineVersionsIntegrationTest  extends AbstractPluginTest {
     def 'Same version conflict but wildcard override at least one should succeed'() {
         when:
         setupVersionsProps("org.scala-lang:scala-* = 2.12.5")
-        buildFile << standardBuildFile(projectDir) + """
+        buildFile << standardBuildFile(projectDir)
+        buildFile << """
         dependencies {
             compile 'org.scala-lang:scala-reflect'
             compile 'org.scala-lang:scala-compiler'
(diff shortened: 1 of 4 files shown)


ferozco (approved):

ferozco:
discussed offline, would like to see an additional test showing the previously broken behaviour

dansanduleac:
@ferozco actually quite hard to make up a test for `checkBomConflict` that doesn't also end up with unused forces that get picked up by the other task.
In the meanwhile though, I noticed `checkNoUnusedPins` was also broken in a similar way.
Updated the description, fixed it and added tests for that ([here](https://github.com/palantir/gradle-baseline/pull/423/files#diff-1b3ac1518abcb74fddedc83185b367e8R196)).

As part of this, noticed the original test 'Version props conflict should succeed' was wrong! The first star force wasn't actually being used anywhere, so the new logic correctly marked it as unused. I therefore changed the test to use jackson and make sure the star force is also used.

gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineVersionsIntegrationTest.groovy
```diff
@@ -97,9 +97,9 @@ class BaselineVersionsIntegrationTest  extends AbstractPluginTest {
         result
     }
 
-    def buildAndFailWith(String error) {
+    void buildAndFailWith(String error) {
         BuildResult result = with('checkVersionsProps', '--stacktrace').buildAndFail()
-        result.output.contains(error)
+        assert result.output.contains(error)
```
dansanduleac, line 102: better error messages! groovy will actually destructure your command when you use assert, just like it does when putting boolean expressions in a `then:` / `expect:` block.

gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineVersionsIntegrationTest.groovy
```diff
@@ -174,21 +175,42 @@ class BaselineVersionsIntegrationTest  extends AbstractPluginTest {
...
-        setupVersionsProps("com.google.guava:* = 22.0\ncom.google.guava:guava = 23.0")
-        buildFile << standardBuildFile(projectDir) + """
+        setupVersionsProps("""
+            com.fasterxml.jackson.*:* = 2.9.3
+            com.fasterxml.jackson.core:jackson-annotations = 2.9.5
+        """.stripIndent())
+        buildFile << standardBuildFile(projectDir)
+        buildFile << """
         dependencies {
-            compile 'com.google.guava:guava'
```
dansanduleac, line 185: original test was failing with the newly corrected logic, because `com.google.guava:*` is not being used by any artifact.

dansanduleac:
@ferozco had an even better idea of what we can do here.
To avoid PR inflation, I'll file an issue with the follow-up.

PR #428, opened 2018-10-16
 [fix] Ignore errors in generated code

ellisjoe:
## Before this PR
A recent changed caused the spotless plugin to be applied to all java source files which would fail on generated code which we don't want to validate or format.

## After this PR
Java source sets that begin with `generated` will be ignored by the spotless plugin.


ellisjoe:
Two options here are to ignore any java file that has `generated` in it's path relative to the project. The alternative, which I think is a bit nicer, is that we agree on the convention that generated code should end up in a `generated*` source set (potentially `generated_*` to force the use of `_`). This is essentially what our checkstyle ignore does at the moment, though it will ignore any path with `generated` vs just source sets.

robert3005:
I think the convention is right though our checkstyle just does `generated.*` (without underscore). Might be easiest to go with that since otherwise there will be a lot of code churn to move things around

qinfchen (approved):

PR #429, opened 2018-10-16
Revert #428 and #419 

qinfchen:
## Before this PR

https://github.com/palantir/gradle-baseline/pull/428 didn't account for generated src in main.  As a result, dev workflow is still broken
## After this PR

Revert to the behavior before https://github.com/palantir/gradle-baseline/pull/419 until we have a better long term solution


dansanduleac:
Found a better fix :)
https://github.com/palantir/gradle-baseline/pull/431

PR #431, opened 2018-10-17
Don't format java files that contain '/generated' in their path

dansanduleac:


iamdanfox (approved):

iamdanfox:
I've triggered a high priority run for the main foundry org, so it's churning through 320 repos now.

PR #434, opened 2018-10-17
[fix] BaselineVersions: resolve configurations without downloading artifacts

dansanduleac:
## Before this PR

`checkVersionsProps` task would resolve and download the artifacts of all configurations in all projects. This can be very slow and it is unnecessary since all we need is the module identifiers (`group:name`) that were resolved.

## After this PR

We only resolve the dependency graph (without downloading artifacts) and look at the resolved [components](https://github.com/gradle/gradle/blob/a64dee21188920c6725db433ef61a7aecb532457/subprojects/core-api/src/main/java/org/gradle/api/artifacts/result/ResolutionResult.java#L74) that are external dependencies to derive the information we need.


iamdanfox (approved):
Could we have a test with a jar that can't actually be downloaded?

PR #435, opened 2018-10-17
[fix] Use exact matching for unused pin check

pkoenig10:
#423 modified this check to use `Pattern#asPredicate` which uses `Matcher#find` which checks if there is a subsequence that matches, not if the entire pattern matches.

This causes issues when one artifact is a substring of another.


dansanduleac:
Yep, didn't realize this :(
However, can we keep the original `Predicate<String>` and just construct the matcher ourselves rather than calling `.asPredicate()`

dansanduleac (approved):

PR #440, opened 2018-10-23
[feature] errorprone check for SafeLoggable exceptions

### LINKED ISSUE #344: error-prone FR: prefer SafeRuntimeException over RuntimeException
We redact all exception messages by default, but the safe-logging project has the blessed SafeRuntimeException, SafeIllegalArgumentException, SafeIOException, SafeIllegalStateException.  We should stop people writing the plain variants which are guaranteed to be redacted!

```diff
+throw new SafeRuntimeException("something went wrong", e)
-throw new RuntimeException("something went wrong", e)
```

iamdanfox:
## Before this PR

People might write code like this that uses regular RuntimeExceptions etc:
```
throw new RuntimeException("Something went wrong", cause);
```

Exceptions like these will be redacted because they are not SafeLoggable.

## After this PR

People will get a 'suggestion' to use safe loggable exceptions. I've also implemented the auto-fix thing, so if we ever wire that up properly it should give people a one-click fix.

fixes https://github.com/palantir/gradle-baseline/issues/344


robert3005 (approved):

iamdanfox:
I'm gonna do a publishToMavenLocal and run this locally somewhere before actually publishing...

PR #441, opened 2018-10-24
Polish PreferSafeLoggableExceptions

iamdanfox:
Follow-up to https://github.com/palantir/gradle-baseline/pull/440 after I tried it locally:

- adds a nice link
- turns it off in test code to minimize noise

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferSafeLoggableExceptions.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferSafeLoggableExceptions.java
@@ -37,6 +37,7 @@ import java.util.Optional;
 @AutoService(BugChecker.class)
 @BugPattern(
         name = "PreferSafeLoggableExceptions",
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
         category = BugPattern.Category.ONE_OFF,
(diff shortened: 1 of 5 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferSafeLoggableExceptions.java
```diff
@@ -85,4 +95,10 @@ public Description matchNewClass(NewClassTree tree, VisitorState state) {
                 .findAny()
                 .orElse(Description.NO_MATCH);
     }
+
+    private static boolean isTestCode(VisitorState state) {
+        return Streams.stream(state.getPath().iterator())
+                .filter(ancestor -> ancestor instanceof ClassTree)
+                .anyMatch(ancestor -> JUnitMatchers.hasJUnit4TestCases.matches((ClassTree) ancestor, state));
```
robert3005: please make this handle junit5

robert3005 (changes requested):

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferSafeLoggableExceptions.java
iamdanfox: done

robert3005 (approved):

PR #444, opened 2018-11-01
[improvement] Implement DangerousThrowableMessageSafeArg error-prone check

carterkozak:
SafeArg.of(any, Throwable.getMessage) is not allowed,
SafeLoggable.getLogMessage is preferrable when safe data is
needed.


iamdanfox (approved):
Thanks @cakofony!

PR #445, opened 2018-11-01
[improvement] Ban internal palantir packages

dansanduleac:
## Before this PR

You could use `com.palantir.<product>.internal` classes.

## After this PR

Checkstyle will prevent you from importing classes from any `com.palantir.<product>.internal` package.


iamdanfox:
Let's make sure this doesn't give us a false postive on `com.palantir.internalpublish.Foo`

iamdanfox:
I tried this locally and found that `com.palantir.product.internal.Foo` was flagged but `com.palantir.product.internalpublish.Bar` was not flagged!

iamdanfox (approved):

PR #454, opened 2018-11-13
Improve intellij-java-palantir-style.xml

diogoholanda:


uschi2000:
Can you provide some examples of the before/after here? Do we also need to adapt checkstyle or the Eclipse formatting rules?

diogoholanda:
There is no need to adapt checkstyle rules. This is already running with the default checkstyle rules in a couple of internal repos.
I have not looked into equivalent eclipse formatting rules.

Main differences here are related to default indentation of params when they can't be all in the same line, preferring to split one param per line. Similarly for ternary operator. It also minimizes the amount of unnecessary empty lines

uschi2000:
The original Baseline Eclipse&IntelliJ formatter settings were chosen such that there as minimal changes between IntelliJ and Eclipse formatting, so that an "autoformat" step before a commit doesn't yield a large diff across IDEs. This meant that some formatting choices couldn't be done automatically, namely in cases where the are no compatible Eclipse&IntelliJ settings.

I would prefer we validate that this property still holds after this change.

uschi2000:
(Another path forward to kill baseline-eclipse)

diogoholanda:
Some example changes:

```
    // before
    private void doSomething(String longStringParamName, Long longLongParamName, Double foo, Object myRandomObject,
            Object anotherRandomObject) {
        return;
    }

    // after
    private void doSomething(
            String longStringParamName,
            Long longLongParamName,
            Double foo,
            Object myRandomObject,
            Object anotherRandomObject) {
        return;
    }
```

```
    // before
    private MyObject getMyObject() {
        return MyObject.builder().id("MyObjectId").shape(Shape.Round).colour(Colour.RED).displayName(
                "My Object Display Name").build();
    }

    // after
    private MyObject getMyObject() {
        return MyObject.builder()
                .id("MyObjectId")
                .shape(Shape.Round)
                .colour(Colour.RED)
                .displayName("My Object Display Name")
                .build();
    }
```

```
    // before
    private void doSomething() {
        Object myObject = returnNewObject ? veryVeryVeryLongMethodNameToGetNewObject()
                : veryVeryVeryLongMethodNameToGetOldObject();
    }

    // after
    private void doSomething() {
        Object myObject = returnNewObject
                ? veryVeryVeryLongMethodNameToGetNewObject()
                : veryVeryVeryLongMethodNameToGetOldObject();
    }
```

diogoholanda:
Looking at parity with eclipse is probably more than what I am willing to do at this moment. I can close this PR and re-open it at a later time when I can look into it if this is a requirement.

diogoholanda:
Notice that even if you reformat your whole code, you won't get to the new formatting. Intellij only reformats (for the changes above) if the original code is exceding the 120 character line limit.

uschi2000:
`Intellij only reformats (for the changes above) if the original code is exceding the 120 character line limit.`
OK, perfect.

uschi2000 (approved):

PR #455, opened 2018-11-13
[fix] Add errorprone javac only if compiler is Java 8

robert3005:
## Before this PR

We were deciding whether or not to add the errorprone javac implementation based on the configured `sourceCompatibility` of the project.
However, if the compiler is Java 9+, the source compatibility doesn't matter - you'd never need to add this compiler. Additionally, the `sun.boot.class.path` won't be set, so our code that configures the test/javadoc tasks would break on parsing the bootstrapClasspath with a NullPointerException.
See the [check](https://github.com/tbroyer/gradle-errorprone-plugin/blob/d36cf009776592cd6bed22d49a3265583a3d8d2f/src/main/kotlin/net/ltgt/gradle/errorprone/ErrorPronePlugin.kt#L79) that the ErrorPronePlugin itself performs.

## After this PR

Only add the errorprone javac compiler to the bootstrap classpaths of Test/Javadoc tasks if the compiler itself is older than Java 9.

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
@@ -26,6 +26,7 @@ import java.util.stream.Collectors;
 import net.ltgt.gradle.errorprone.CheckSeverity;
 import net.ltgt.gradle.errorprone.ErrorProneOptions;
 import net.ltgt.gradle.errorprone.ErrorPronePlugin;
+import org.gradle.api.JavaVersion;
 import org.gradle.api.Plugin;
 import org.gradle.api.Project;
 import org.gradle.api.file.FileCollection;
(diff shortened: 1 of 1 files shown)


iamdanfox:
Can you flesh out what the broken behaviour you were seeing?  I think I've been using latest baseline with Java10 on my machine locally and it seems to have worked fine?

robert3005:
@dfox there are 3 cases to consider and I will contest that 1 is broken (the one I hit) and another one works by accident while other one is fine

| source compatibility version\javac version  |  1.8 |  11 |
|---|---|---|
|  1.8  | ✓  | ✖  |
|  11  | not possible  | ✓ |

What's happening from my understanding of the code is that we're most likely not configuring 1.8 source compatibility in time so you never hit the case and just get java 11 binaries. In rare cases where it is configured before baseline-errorprone is applied you see issue with reading bootstrap classpath

dansanduleac:
👍

PR #465, opened 2018-11-20
Don't configure nebula propertiesFile if core bom support is on

dansanduleac:
Using `propertiesFile` or any other recommendation source other than `mavenBom` will [always throw if core BOM support is on](https://github.com/nebula-plugins/nebula-dependency-recommender-plugin/blob/v7.1.2/src/main/groovy/netflix/nebula/dependency/recommender/provider/RecommendationProviderContainer.java#L96).

cc @robert3005


ferozco (approved):

robert3005 (approved):

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/versions/BaselineVersions.java
```diff
@@ -67,6 +70,11 @@ public void apply(Project project) {
         // apply plugin: "nebula.dependency-recommender"
         project.getPluginManager().apply(DependencyRecommendationsPlugin.class);
 
+        if (Boolean.getBoolean("nebula.features.coreBomSupport")) {
```
uschi2000, line 73: no test?

PR #473, opened 2018-11-27
Ban Runtime#addShutdownHook.

jkozlowski:
## Before this PR
Use of Runtime#addShutdownHook is allowed, which in libraries can cause memory leaks if the library does not deregister the hooks promptly.

## After this PR
Devs are prompted to not use the shutdown hook functionality directly, but instead rely on their webserver's managed lifecycle functionality.

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ShutdownHook.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ShutdownHook.java
@@ -22,12 +22,10 @@ import com.google.errorprone.VisitorState;
 import com.google.errorprone.bugpatterns.BugChecker;
 import com.google.errorprone.matchers.Description;
 import com.google.errorprone.matchers.Matcher;
-import com.google.errorprone.matchers.Matchers;
 import com.google.errorprone.matchers.method.MethodMatchers;
 import com.google.errorprone.predicates.TypePredicates;
 import com.sun.source.tree.ExpressionTree;
 import com.sun.source.tree.MethodInvocationTree;
-import com.sun.source.tree.Tree;
 import java.util.regex.Pattern;
 
 @AutoService(BugChecker.class)
@@ -48,9 +46,6 @@ public final class ShutdownHook extends BugChecker implements BugChecker.MethodI
     private static final Matcher<ExpressionTree> shutdownHookMethods = MethodMatchers.instanceMethod()
             .onClass(TypePredicates.isExactType("java.lang.Runtime"))
             .withNameMatching(Pattern.compile("addShutdownHook|removeShutdownHook"));
-
-    private static final Matcher<Tree> containsShutdownHookMethod = Matchers.contains(
-            Matchers.toType(ExpressionTree.class, shutdownHookMethods));
 
     @Override
     public Description matchMethodInvocation(
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ShutdownHook.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ShutdownHook.java
@@ -1,0 +1,66 @@
+/*
+ * (c) Copyright 2018 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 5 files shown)


jkozlowski:
cc @j-baker

gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
```diff
@@ -370,6 +370,11 @@
             <property name="format" value="(void setUp\(\))|(void setup\(\))|(void setupStatic\(\))|(void setUpStatic\(\))|(void beforeTest\(\))|(void teardown\(\))|(void tearDown\(\))|(void beforeStatic\(\))|(void afterStatic\(\))"/>
             <property name="message" value="Test setup/teardown methods are called before(), beforeClass(), after(), afterClass(), but not setUp, teardown, etc."/>
         </module>
+        <module name="RegexpSinglelineJava">
```
j-baker: we should make sure it's still possible to call this in test code (e.g. DFE is not enforced on test code), since JUnit provides no easy way to run something after all tests

jkozlowski: Is @SuppressWarnings("BanShutdownHooks") not enough? I think I can also add it to the list of suppressions.

jkozlowski: Done

iamdanfox (dismissed):

iamdanfox (commented):
This seems like a good idea any chance we could turn it into an errorprone check instead?  I'd like to try to push checkstyle towards code formatting/style!

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ShutdownHook.java
```diff
@@ -0,0 +1,66 @@
...
+
+    private static final long serialVersionUID = 1L;
+    private static final String errorMsg = "Use your webserver's managed resource functionality "
+            + "instead of using java.lang.Runtime#addShutdownHook directly.";
+
+    private static final Matcher<ExpressionTree> shutdownHookMethods = MethodMatchers.instanceMethod()
+            .onClass(TypePredicates.isExactType("java.lang.Runtime"))
+            .withNameMatching(Pattern.compile("addShutdownHook|removeShutdownHook"));
+
+    private static final Matcher<Tree> containsShutdownHookMethod = Matchers.contains(
```
iamdanfox: unused?

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/ShutdownHookTests.java
```diff
@@ -0,0 +1,59 @@
...
+        test("addShutdownHook(new Thread())", Optional.of(errorMsg));
+        test("removeShutdownHook(new Thread())", Optional.of(errorMsg));
+    }
+
+    @Test
+    public void testDoesNotUseShutdownHooks() {
+        test("availableProcessors()", Optional.empty());
+    }
+
+    private void test(String method, Optional<String> error) {
```
iamdanfox: I'd prefer not to introduce new indirection in the unit-tests, I think the errorprone library is already super readable (even if there is a little bit of duplication).

jkozlowski: I copied this setup from the other tests. Happy to just copy the method.

jkozlowski: Donezo

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ShutdownHook.java
jkozlowski: Donezo

jkozlowski:
@iamdanfox done, wanna have another look?

iamdanfox (approved):
Looks great, thanks @jkozlowski!

jkozlowski:
@iamdanfox can you do a release?

iamdanfox:
Tagged: https://github.com/palantir/gradle-baseline/releases/tag/0.42.0

PR #474, opened 2018-11-27
Patch 1

a10y:


iamdanfox (changes requested):
This PR changes the wrong file unfortunately (because gradle-baseline actually uses gradle-baseline).

You need to modify the other intellij-java-palantir-style.xml.

robert3005:
Please reopen when you intend to fix it

PR #481, opened 2018-11-30
Suggest avoiding default case in switch statements

### LINKED ISSUE #480: Consider disallowing default cases in switch statements
## What happened?

If a switch statement has a default case, then the [MissingCasesInEnumSwitch](http://errorprone.info/bugpattern/MissingCasesInEnumSwitch.html) check is ineffective. This makes is easy to accidentally not handle new enum variants.

For example, suppose we have:
```
enum Colors {
    RED,
    GREEN
}

boolean isRedOrBlue(Color color) {
    switch (color) {
        case RED:
            return true;
        case GREEN:
            return false;
        default:
            throw new RuntimeException("Unknown color");
    }
}
```

And then a `BLUE` variant is added to the enum:
```
enum Colors {
    RED,
    GREEN,
    BLUE
}
```

There will not be a warning that the switch statement does not explicitly handle the `BLUE` case and that the method is now incorrect.

Each enum variant should be explicitly handled to ensure that we get a warning when new variants are added. To ensure the [MissingCasesInEnumSwitch](http://errorprone.info/bugpattern/MissingCasesInEnumSwitch.html) check always checks for missing cases, `default` cases should be disallowed.

The initial example would change to:
```
boolean isRedOrBlue(Color color) {
    switch (color) {
        case RED:
            return true;
        case GREEN:
            return false;
    }

    throw new RuntimeException("Unknown color");
}
```

pkoenig10:
Fixes #480

A severity of `WARNING` is probably too strong for this check.

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/SwitchStatementDefaultCase.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/SwitchStatementDefaultCase.java
@@ -35,7 +35,7 @@ import java.util.Objects;
         link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
         linkType = LinkType.CUSTOM,
         severity = SeverityLevel.SUGGESTION,
-        summary = "Avoid default cases in switch statements to correctly handle new enum variants")
+        summary = "Avoid default cases in switch statements to correctly handle new enum values")
 public final class SwitchStatementDefaultCase extends BugChecker implements BugChecker.SwitchTreeMatcher {
 
     @Override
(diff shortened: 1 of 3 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/SwitchStatementDefaultCase.java
```diff
@@ -0,0 +1,57 @@
...
+import com.google.errorprone.matchers.Description;
+import com.sun.source.tree.CaseTree;
+import com.sun.source.tree.SwitchTree;
+import java.util.Objects;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "SwitchStatementDefaultCase",
+        category = Category.ONE_OFF,
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
```
iamdanfox, line 35: Could you add a sentence or two to the README to flesh out the reasoning behind this a bit more (i.e. recommending to throw an unreachable exception)?

iamdanfox:
Thanks for implementing this @pkoenig10!

pkoenig10:
@iamdanfox do you have any opposition to making this a warning? If not now, maybe in the future?

iamdanfox (approved):

iamdanfox:
I think it's OK to run with `SUGGESTION` for now because people with `-Werror` will pick this up straight away anyway.

PR #482, opened 2018-12-04
Enable the class uniqueness plugin by default

carterkozak:


iamdanfox (changes requested):
I only want to merge this when it's working on template-wc, template-java, template-gradle-plugin and a few real-world repos as otherwise it might be too annoying for people to fix themselves leading to them stopping taking upgrades

iamdanfox:
So I think it might be worth excavating this plugin out to people so that we can iron out any remaining kinks.

PR #484, opened 2018-12-05
Only apply the baseline-eclipse plugin on Java projects

markelliot:
# Before this PR
All subprojects are hidden when importing from a repo where the `com.palantir.baseline` plugin is applied to the root project, even when the root project is not a Java project.

# After this PR
Only apply the baseline-eclipse (and the underlying Gradle eclipse plugin) on Java projects, else no-op; when root projects are not Java projects, the baseline-eclipse plugin has no effect.

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/Baseline.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/Baseline.java
@@ -39,13 +39,15 @@ public final class Baseline implements Plugin<Project> {
         rootProject.allprojects(proj -> {
             proj.getPluginManager().apply(BaselineCheckstyle.class);
             proj.getPluginManager().apply(BaselineScalastyle.class);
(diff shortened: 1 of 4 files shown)


markelliot:
It might be preferable to have the baseline plugin somehow decide if a particular project is a Java project, but I'm not sure I understand enough about Gradle lifecycle ordering that this would work correctly.

iamdanfox:
To detect if a project has the java plugin applied, you can do the following.

```gradle
project.getPlugins().withType(JavaPlugin.class, p -> {
	// only execute this on java plugins
});
```

I think doing this inside the eclipse plugin would be a neat fix :)

markelliot:
Does that check require plugins to be applied in a particular ordering?


________________________________

iamdanfox:
Nope, if the java plugin is applied later then your lambda will still be invoked.

markelliot:
@iamdanfox ready for another pass

iamdanfox (approved):

PR #493, opened 2019-01-02
[improvement] Resolves deprecation warning for use of DefaultTask.newInputFile()

schlosna:
## Before this PR

```
The DefaultTask.newInputFile() method has been deprecated.
This is scheduled to be removed in Gradle 6.0.
Please use the ObjectFactory.fileProperty() method instead.
2 usages
    CheckBomConflictTask.java:45
    CheckNoUnusedPinTask.java:45
```

## After this PR

Use non-deprecated file property factory method to avoid breakage when gradle 6.0 comes out.

See tritium build scan flagging deprecation: https://scans.gradle.com/s/pflkoibzlelue/deprecations#deprecation-0

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/versions/CheckBomConflictTask.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/versions/CheckBomConflictTask.java
@@ -42,7 +42,7 @@ import org.gradle.api.tasks.options.Option;
 public class CheckBomConflictTask extends DefaultTask {
 
     private final Property<Boolean> shouldFix = getProject().getObjects().property(Boolean.class);
-    private final RegularFileProperty propsFileProperty = newInputFile();
+    private final RegularFileProperty propsFileProperty = getProject().getObjects().fileProperty();
 
     public CheckBomConflictTask() {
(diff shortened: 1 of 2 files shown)


schlosna:
See https://github.com/palantir/gradle-baseline/pull/494 for bit rot build fix

ferozco:
👍

robert3005:
Just an FYI that this adds gradle 5.0 minimum version. Probably should update the readme after this merges

iamdanfox:
I'd prefer to present this as a '[break]' actually and update the README as part of the same PR, as our minimum gradle version is only real user commitment!

j-baker:
Might as well leave this until we have reasonable coverage of Gradle 5 internally? So this doesn't block the rollout train?

lycarter:
hey @iamdanfox @j-baker this is currently blocking an improvement to gradle-conjure: https://github.com/palantir/gradle-conjure/pull/116

What's the update on the Gradle 5 internal coverage? Any chance that could be prioritized?

iamdanfox:
Honestly I think it would be simpler to just stop using RegularFileProperty and just use a Property<File>, as we already say baseline requires at least Gradle 4.10.0.

PR #504, opened 2019-01-15
[fix] Disable certain errorprone checks for gradle plugins

### LINKED ISSUE #389: Automatically disable error-prone logging checks for gradle plugins
## What happened

We have lots of repos internally where we maintain gradle plugins, but the `Slf4jLogsafeArgs` and `Slf4jConstantLogMessage` checks are pretty redundant as they are intended for server-side or library usage, whereas Gradle plugins are used pretty much exclusively on local machines.

```
apply plugin: 'java-gradle-plugin'

dependencies {
    // ..
}

tasks.withType(JavaCompile) {
    options.errorprone.errorproneArgs += ['-Xep:Slf4jLogsafeArgs:OFF', '-Xep:Slf4jConstantLogMessage:OFF']
}
```

## What should have happened

baseline-error-prone should detect when the `java-gradle-plugin` plugin has been applied and automatically disable these custom logging checks.

dansanduleac:
Fixes #389 

## Before this PR

Gradle plugins would be subjected to certain error prone checks that are only really applicable to code that will run as part of deployed services:

* Slf4jLogsafeArgs
* PreferSafeLoggableExceptions

## After this PR

These checks are disabled for gradle plugin projects, i.e. projects that apply `java-gradle-plugin`.

cc @iamdanfox

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
@@ -31,7 +31,6 @@ import org.gradle.api.Plugin;
 import org.gradle.api.Project;
 import org.gradle.api.file.FileCollection;
 import org.gradle.api.plugins.ExtensionAware;
-import org.gradle.api.plugins.JavaPluginConvention;
 import org.gradle.api.tasks.compile.JavaCompile;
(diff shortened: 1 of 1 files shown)


iamdanfox:
:+1: should hopefully allow us to get rid of the common block: https://github.com/palantir/gradle-conjure/blob/develop/gradle-conjure/build.gradle#L46-L50

PR #505, opened 2019-01-15
[fix] Disable Slf4jConstantLogMessage for gradle plugins

iamdanfox:
Follow-up to https://github.com/palantir/gradle-baseline/pull/504


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
```diff
@@ -65,6 +65,7 @@ public void apply(Project project) {
                                 .configure(ErrorProneOptions.class, errorProneOptions -> {
                                     errorProneOptions.check("Slf4jLogsafeArgs", CheckSeverity.OFF);
                                     errorProneOptions.check("PreferSafeLoggableExceptions", CheckSeverity.OFF);
+                                    errorProneOptions.check("Slf4jConstantLogMessage", CheckSeverity.OFF);
```
iamdanfox, line 68: Arguably this is actually still valuable for perf, maybe it should just be a warning?

carterkozak: This one shouldn’t be disabled

PR #506, opened 2019-01-16
Split PreconditionsConstantMessage logsafe validation into LogSafePreconditionsConstantMessage

carterkozak:
This provides a more accurate failure message when logsafe
preconditions fail to use constant strings.
Suppressing guava preconditions constant string validation will
no longer also suppress validation for logsafe preconditions.


ferozco (commented):
Code change seems reasonable. Whats the motivation behind the change? It seems like splitting the validation would encourage people to override/customize, which i'm not sure is desirable

carterkozak:
My fear is that the guava preconditions check will be suppressed, which devs won't realize also disables safe-logging validation.

ferozco:
Perhaps the solution is to improve the documentation around the check? The current description is a little ambiguous https://github.com/palantir/gradle-baseline/pull/506/files#diff-62156c50f0b69ea0d47f88342836e1f5R40. Have you seen users disabling the check in practice?

carterkozak:
I have seen projects that disable this check entirely because it cannot be applied only to safe-logging preconditions.

ferozco:
👍

PR #514, opened 2019-01-19
[improvement] Allow disabling com.palantir.baseline-versions

dansanduleac:
## Before this PR

After #465, if `coreBomSupport` was turned on (`nebula.dependency-recommender` feature). then `BaselineVersions` would only apply dependency-recommender and do nothing else.

## After this PR

Always configure `checkVersionsProps` and `checkNoUnusedPin` regardless of whether nebula is being configured or not.

It's now possible to completely turn off (not applying) nebula.dependency-recommender by defining the project property:
```
com.palantir.baseline-versions.disable
```

e.g. 
```
./gradlew -Pcom.palantir.baseline-versions.disable
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -229,23 +229,19 @@ dependencyRecommendations {
 Adds the following tasks:
 
 - `checkVersionsProps` - A catch-all task to lint your versions.props file.
-- `checkBomConflict`<sup>†</sup> - Ensures your versions.props pins don't force the same version that is already recommended by a BOM.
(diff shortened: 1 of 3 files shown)


robert3005:
👍 - we could keep nebula if not disable-nebula but since we want to replace it long term I think this is fine

robert3005:
👍

ferozco:
could we update the documentation with the new property?

ferozco:
👍

PR #516, opened 2019-01-21
[improvement] Add separate file for custom suppressions, and an id for the shaded classes check.

JacekLach:
## Before this PR

Currently any project-specific overrides in checkstyle-suppressions.xml will be reset on a config update. Especially annoying if done automatically on every config release.

## After this PR
This adds a separate, optional, unmanaged suppressions file that avoids the problem.

Also adds an id property to IllegalImport checking for shaded classes, so it can be suppressed individually.


dansanduleac:
👍

Issue #507, opened 2019-01-16
FR: Add check that simple uses of Objects.requireNonNull and Preconditions.check* use safe-logging


carterkozak:
Uses of the following with constant-value strings (already validated by `PreconditionsConstantMessage`) and no parameters should be migrated to use safe-logging preconditions , which does not require changes beyond switching the method invocation:

`Objects.requireNonNull(Object)`,
`Objects.requireNonNull(Object, String)`,
`com.google.common.base.Preconditions.check*(Object)`,
`com.google.common.base.Preconditions.check*(Object, String)`

carterkozak:
Open question: do we want to error/warn on functions which do not provide a message? Moving to safe-logging won't provide additional data, but does provide some standardization.

carterkozak:
This was resolved by #517

PR #517, opened 2019-01-21
[improvement] Enforce guava and objects calls with no parameters and constant strings use logsafe

dsd987:
## Before this PR
There is no standardization in the use of basic calls to verify conditions like guava's Preconditions.check* or Objects.requireNonNull.

## After this PR
Simple use cases of guava Preconditions.check* and Objects.requireNonNull are now required to be changed to use logsafe variants. This will help to standardize calls where it is clear there can be a direct one-for-one replacement made with logsafe calls.

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,7 +142,7 @@ Safe Logging can be found at [github.com/palantir/safe-logging](https://github.c
 - `PreferSafeLoggingPreconditions`: Users should use the safe-logging versions of Precondition checks for standardization when there is equivalent functionality
     ```diff
     -com.google.common.base.Preconditions.checkNotNull(variable, "message");
-    -com.palantir.logsafe.Preconditions.checkNotNull(variable, "message"); // equivalent functionality is available in the safe-logging variant
+    +com.palantir.logsafe.Preconditions.checkNotNull(variable, "message"); // equivalent functionality is available in the safe-logging variant
     ```
 - `ShutdownHook`: Applications should not use `Runtime#addShutdownHook`.
 - `SwitchStatementDefaultCase`: Switch statements should avoid using default cases. Default cases prevent the [MissingCasesInEnumSwitch](http://errorprone.info/bugpattern/MissingCasesInEnumSwitch.html) check from detecting when an enum value is not explicitly handled. This check is important to help avoid incorrect behavior when new enum values are introduced.
--- a/README.md
+++ b/README.md
@@ -139,6 +139,11 @@ Safe Logging can be found at [github.com/palantir/safe-logging](https://github.c
     -throw new RuntimeException("explanation", e); // this message will be redacted when logs are collected
     +throw new SafeRuntimeException("explanation", e); // this message will be preserved (allowing easier debugging)
     ```
+- `PreferSafeLoggingPreconditions`: Users should use the safe-logging versions of Precondition checks for standardization when there is equivalent functionality
+    ```diff
+    -com.google.common.base.Preconditions.checkNotNull(variable, "message");
+    -com.palantir.logsafe.Preconditions.checkNotNull(variable, "message"); // equivalent functionality is available in the safe-logging variant
+    ```
 - `ShutdownHook`: Applications should not use `Runtime#addShutdownHook`.
 - `SwitchStatementDefaultCase`: Switch statements should avoid using default cases. Default cases prevent the [MissingCasesInEnumSwitch](http://errorprone.info/bugpattern/MissingCasesInEnumSwitch.html) check from detecting when an enum value is not explicitly handled. This check is important to help avoid incorrect behavior when new enum values are introduced.
 
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferSafeLoggingPreconditions.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferSafeLoggingPreconditions.java
@@ -25,10 +25,12 @@ import com.google.errorprone.fixes.SuggestedFixes;
 import com.google.errorprone.matchers.CompileTimeConstantExpressionMatcher;
 import com.google.errorprone.matchers.Description;
 import com.google.errorprone.matchers.Matcher;
+import com.google.errorprone.matchers.Matchers;
 import com.google.errorprone.matchers.method.MethodMatchers;
 import com.google.errorprone.util.ASTHelpers;
 import com.sun.source.tree.ExpressionTree;
 import com.sun.source.tree.MethodInvocationTree;
+import com.sun.tools.javac.code.Symbol.MethodSymbol;
 import java.util.List;
 import java.util.regex.Pattern;
 
@@ -48,17 +50,14 @@ public final class PreferSafeLoggingPreconditions extends BugChecker implements 
     private final Matcher<ExpressionTree> compileTimeConstExpressionMatcher =
             new CompileTimeConstantExpressionMatcher();
 
-    private static final String LOGSAFE_NAME = "com.palantir.logsafe.Preconditions";
-    private static final Matcher<ExpressionTree> PRECONDITIONS_MATCHER = MethodMatchers.staticMethod()
-            .onClass("com.google.common.base.Preconditions")
-            .withNameMatching(Pattern.compile("checkArgument|checkState|checkNotNull"));
-    private static final Matcher<ExpressionTree> OBJECTS_MATCHER = MethodMatchers.staticMethod()
-            .onClass("java.util.Objects")
-            .withNameMatching(Pattern.compile("requireNonNull"));
-
-    private static final String DESCRIPTION_MESSAGE =
-            "The call can be replaced with an equivalent one from com.palantir.logsafe.Preconditions for "
-                    + "standardization as the functionality is the same.";
+    private static final Matcher<ExpressionTree> METHOD_MATCHER =
+            Matchers.anyOf(
+                    MethodMatchers.staticMethod()
+                            .onClass("com.google.common.base.Preconditions")
+                            .withNameMatching(Pattern.compile("checkArgument|checkState|checkNotNull")),
+                    MethodMatchers.staticMethod()
+                            .onClass("java.util.Objects")
+                            .withNameMatching(Pattern.compile("requireNonNull")));
 
     @Override
     public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
@@ -66,10 +65,7 @@ public final class PreferSafeLoggingPreconditions extends BugChecker implements 
             return Description.NO_MATCH;
         }
 
-        boolean matchPreconditions = PRECONDITIONS_MATCHER.matches(tree, state);
-        boolean matchObjects = OBJECTS_MATCHER.matches(tree, state);
-
-        if (!matchPreconditions && !matchObjects) {
+        if (!METHOD_MATCHER.matches(tree, state)) {
             return Description.NO_MATCH;
         }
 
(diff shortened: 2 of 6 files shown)


dsd987:
To address #507

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/PreconditionsTests.java
```diff
@@ -62,8 +75,6 @@ public final void validGuava() throws Exception {
                 "class Test {",
                 "  private static final String compileTimeConstant = \"constant\";",
                 "  void f(boolean bArg, int iArg, Object oArg) {",
-                "    Preconditions.checkArgument(bArg);",
-                "    Preconditions.checkArgument(bArg, \"message\");",
```
carterkozak: I don't think this should be modified, it is valid for the existing test, just not the new check

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/GuavaPreconditionsEnforceLogSafeUse.java
```diff
@@ -0,0 +1,59 @@
...
+import com.google.errorprone.matchers.Matcher;
+import com.google.errorprone.matchers.Matchers;
+import com.google.errorprone.matchers.method.MethodMatchers;
+import com.sun.source.tree.ExpressionTree;
+import com.sun.source.tree.MethodInvocationTree;
+import java.util.regex.Pattern;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "GuavaPreconditionsEnforceLogSafeUse",
```
carterkozak: We're really checking one thing: functionality that can be replaced with safe-logging utilities without losing any information/utility. I'd prefer to have a single check for that.

Based on the existing `PreferSafeLoggableExceptions` check we should call it `PreferSafeLoggingPreconditions`.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/GuavaPreconditionsEnforceLogSafeUse.java
```diff
@@ -0,0 +1,59 @@
...
+        category = BugPattern.Category.ONE_OFF,
+        severity = BugPattern.SeverityLevel.ERROR,
```
carterkozak: Based on `PreferSafeLoggableExceptions` we should use `SeverityLevel.SUGGESTION` by default, in certain large internal codebases we will override it to ERROR. Longer term we should set the new check and PreferSafeLoggableExceptions to error, but I'd like to decouple that from this change.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/EnforceLogSafeUseFormat.java
```diff
@@ -0,0 +1,59 @@
...
+        }
+
+        List<? extends ExpressionTree> args = tree.getArguments();
+        if (args.size() > 2) {
+            return Description.NO_MATCH;
+        }
+
+        if (args.size() == 2) {
+            ExpressionTree messageArg = args.get(1);
+            if (!compileTimeConstExpressionMatcher.matches(messageArg, state)) {
```
carterkozak: This needs to check the type of args[1], otherwise we'll get false positives on `Objects.requireNonNull(obj, () -> calculateExpensiveMessage())` using the supplier overload.

carterkozak (changes requested):

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ObjectsEnforceLogSafeUse.java
```diff
@@ -0,0 +1,58 @@
...
+import com.sun.source.tree.MethodInvocationTree;
+import java.util.regex.Pattern;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "ObjectsEnforceLogSafeUse",
+        category = BugPattern.Category.ONE_OFF,
+        severity = BugPattern.SeverityLevel.ERROR,
+        summary = "Objects.requireNonNull(Object) and Objects.requireNonNull(Object, String) methods with a "
+                + "constant message and no parameters should use logsafe Preconditions.checkNotNull(Object, String)")
```
iamdanfox: might as well give the FQN: `com.palantir.logsafe.Preconditions#checkNotNull(Object, String)`

iamdanfox: Also I'd recommend giving a little hint as to why error-prone is recommending the change, e.g. the PreferSafeLoggableExceptions rule says "Throw SafeLoggable exceptions to ensure the exception message will not be redacted"

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/GuavaPreconditionsEnforceLogSafeUse.java
```diff
@@ -0,0 +1,59 @@
...
+                + "message and no parameters should use logsafe Preconditions.checkX(Object, String)")
+public final class GuavaPreconditionsEnforceLogSafeUse extends EnforceLogSafeUseFormat {
+
+    private static final long serialVersionUID = 1L;
+
+    private static final Matcher<ExpressionTree> PRECONDITIONS_METHOD =
+            Matchers.anyOf(
+                    MethodMatchers.staticMethod()
+                            .onClass("com.google.common.base.Preconditions")
+                            .withNameMatching(Pattern.compile("checkArgument|checkState|checkNotNull")));
```
schlosna: Probably a separate PR, but would be good to check Guava’s [`Verify.verify` & `Verify.verifyNotNull`](https://google.github.io/guava/releases/23.1-jre/api/docs/com/google/common/base/Verify.html) as well as [commons-lang `Validate.*`](https://commons.apache.org/proper/commons-lang/javadocs/api-3.1/org/apache/commons/lang3/Validate.html)?

carterkozak: That would be difficult -- guava Preconditions throws standard IllegalState/IllegalArg/NullPointer exceptions, but these methods throw library-specific exception types, so a migration _could_ break existing catches.

I think we should suggest against using them in favor of safe-logging preconditions, but I'd prefer to do so in a different check.

schlosna: 👍 Makes sense for Guava's `Verify` and I expect much lower usage of that.

The commons-lang `Validate` is much more heavily used in some places, and those should be straightforward replacements by equivalent `Safe` preconditions

carterkozak: I had though lang3 validate used custom exception types, thanks for the correction! In that case it makes sense to add common methods to this check. Let's add it in a separate PR.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferSafeLoggingPreconditions.java
```diff
@@ -0,0 +1,96 @@
...
+        } else if (OBJECTS_MATCHER.matches(tree, state)) {
+            return buildDescription(tree)
+                    .setMessage("This java.util.Objects call can be replaced with an equivalent "
+                            + "one from com.palantir.logsafe.Preconditions for standardization as the "
+                            + "functionality is the same")
+                    .addFix(SuggestedFix.builder()
+                            .replace(tree, String.format(
+                                    "com.palantir.logsafe.Preconditions.checkNotNull(%s)", tree.getArguments()))
+                            .build())
+                    .build();
```
carterkozak: Can we combine these? The message is different, but doesn't need to be because we output a suggested change which illustrates the current and proposed states. We can add a function which maps from a given method name to a safe-logging Preconditions method name instead.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferSafeLoggingPreconditions.java
```diff
@@ -0,0 +1,96 @@
...
+    @Override
+    public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
+        List<? extends ExpressionTree> args = tree.getArguments();
+        if (args.size() > 2) {
+            return Description.NO_MATCH;
+        }
+
+        if (args.size() == 2) {
+            ExpressionTree messageArg = args.get(1);
+            if (!messageArg.getKind().equals(Tree.Kind.STRING_LITERAL)
```
carterkozak: Do we actually want to test that this is a string literal? Will that catch references to compile-time constant methods?

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferSafeLoggingPreconditions.java
```diff
@@ -0,0 +1,96 @@
...
+                    || !compileTimeConstExpressionMatcher.matches(messageArg, state)) {
```
carterkozak: I think we should check the method matchers prior to this, I imagine the compile time constant matcher is more expensive than checking method names, and this will fire for all 2-arg methods with a string as the second argument.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferSafeLoggingPreconditions.java
dsd987: Good catch! Yeah, that won't work with compile-time constants. Will fix.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferSafeLoggingPreconditions.java
dsd987: 👍

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferSafeLoggingPreconditions.java
dsd987: 👍

carterkozak:
This check should be added to the list of exclusions for gradle plugins, since those don't use safe-logging

README.md
```diff
@@ -139,6 +139,11 @@ Safe Logging can be found at [github.com/palantir/safe-logging](https://github.c
     -throw new RuntimeException("explanation", e); // this message will be redacted when logs are collected
     +throw new SafeRuntimeException("explanation", e); // this message will be preserved (allowing easier debugging)
     ```
+- `PreferSafeLoggingPreconditions`: Users should use the safe-logging versions of Precondition checks for standardization when there is equivalent functionality
+    ```diff
+    -com.google.common.base.Preconditions.checkNotNull(variable, "message");
+    -com.palantir.logsafe.Preconditions.checkNotNull(variable, "message"); // equivalent functionality is available in the safe-logging variant
```
dsd987: Should be a "+"

carterkozak:
Please don't force-push, that makes it hard to review diffs!

iamdanfox (approved):
Looks solid to me :) I'm looking forward to actually exposing those suggested fixes!

carterkozak (approved):
lgtm, thanks @dsd987!

PR #521, opened 2019-01-24
[improvement] Enforce lang3 Validate calls with no parameters and constant strings use logsafe

dsd987:
## Before this PR
There is a lower degree of standardization in the use of basic calls to verify conditions, specifically as it pertains to the use of commons-lang3 Validate calls.

## After this PR
Simple use cases of commons-lang3 Validate.isTrue, Validate.notNull, Validate.validState are now required to be changed to use logsafe variants. This will help to standardize calls where it is clear there can be a direct one-for-one replacement made with logsafe calls.

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferSafeLoggingPreconditions.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferSafeLoggingPreconditions.java
@@ -17,6 +17,7 @@
 package com.palantir.baseline.errorprone;
 
 import com.google.auto.service.AutoService;
+import com.google.common.collect.ImmutableMap;
 import com.google.errorprone.BugPattern;
 import com.google.errorprone.VisitorState;
 import com.google.errorprone.bugpatterns.BugChecker;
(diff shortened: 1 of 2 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferSafeLoggingPreconditions.java
```diff
@@ -57,7 +59,16 @@
...
+                            .named("requireNonNull"),
+                    MethodMatchers.staticMethod()
+                            .onClass("org.apache.commons.lang3.Validate")
+                            .withNameMatching(Pattern.compile("isTrue|notNull|validState")));
+
+    private static final Map<String, String> METHOD_TRANSLATIONS = ImmutableMap.of(
+            "requireNonNull", "checkNotNull",
+            "isTrue", "checkArgument",
+            "notNull", "checkNotNull",
+            "validState", "checkState");
```
carterkozak: nit: would prefer to add comments with each entry describing which utility class the method comes from

carterkozak:
@schlosna fyi :-)

carterkozak (approved):
lgtm, thanks!

iamdanfox (approved):

PR #533, opened 2019-02-06
[fix] Correct prioritisation of versions.props to match nebula logic

dansanduleac:
## Before this PR

`versions.props` property names (matchers) were incorrectly being prioritised when deciding which `versions.props` line is most specific and matches a given dependency.

The current logic assumes that the last line to match a dependency wins.

## After this PR

Implemented the logic that `nebula.dependency-recommender` [actually uses](https://github.com/nebula-plugins/nebula-dependency-recommender-plugin/blob/dae66d3f5f9fb3b656ddf5348cd0527e000b4340/src/main/groovy/netflix/nebula/dependency/recommender/provider/FuzzyVersionResolver.java#L77), which is to weigh the matchers by "number of characters which are not globs, i.e. `*`".
The most specific matcher is the one with the highest weight.

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/versions/BaselineVersions.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/versions/BaselineVersions.java
@@ -16,10 +16,13 @@
 
 package com.palantir.baseline.plugins.versions;
 
+import com.google.common.base.Splitter;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Streams;
 import java.io.File;
 import java.io.IOException;
 import java.nio.file.Files;
+import java.util.Comparator;
 import java.util.Set;
 import java.util.stream.Collectors;
 import netflix.nebula.dependency.recommender.DependencyRecommendationsPlugin;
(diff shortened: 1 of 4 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/versions/BaselineVersions.java
```diff
@@ -162,4 +166,28 @@ private static File rootVersionsPropsFile(Project project) {
...
+            Comparator.comparing(BaselineVersions::versionsPropsMatcherWeight);
+
+    /**
+     * The weight of a matcher in {@code versions.props} according to the disambiguation logic defined in
+     * {@code nebula.dependency-recommender}.
+     *
+     * This matches the logic in {@link FuzzyVersionResolver}.
+     */
+    private static int versionsPropsMatcherWeight(String propertiesLine) {
+        return Streams.stream(Splitter.on("*").split(propertiesLine)).mapToInt(String::length).sum();
```
carterkozak: nit: `CharMatcher.isNot('*').countIn(propertiesLine)`

ferozco (approved):

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/versions/BaselineVersions.java
dansanduleac: your guava fu is too high

carterkozak (approved):

PR #537, opened 2019-02-08
[feature] Warn against .parallel() calls on Java streams

gm2211:
## Before this PR
As explained in detail here: https://stackoverflow.com/a/54581148/7182570, using `.parallel()` on Java streams might lead to unexpected system-wide performance issues down the road due to the work-stealing nature of the thread pool shared by all Java streams. 

## After this PR
We will warn against calling `.parallel()` on Java streams and recommend alternatives like using an `ExecutorCompletionService` or https://github.com/amaembo/streamex, which lets you mitigate the impact of work-stealing by allowing you to provide your own `ForkJoinPool` (so you can control exactly which parts of your system will share that pool).


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferSafeLoggingPreconditions.java
```diff
@@ -64,7 +63,7 @@
                             .onClass("org.apache.commons.lang3.Validate")
                             .withNameMatching(Pattern.compile("isTrue|notNull|validState")));
 
-    private static final Map<String, String> TRANSLATIONS_TO_LOGSAFE_PRECONDITIONS_METHODS = ImmutableMap.of(
+    private static final ImmutableMap<String, String> TRANSLATIONS_TO_LOGSAFE_PRECONDITIONS_METHODS = ImmutableMap.of(
```
gm2211, line 66: Unrelated change, but error prone was warning against this when running tests locally - kind of meta..

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/DangerousParallelStreamUsageTest.java
```diff
@@ -0,0 +1,252 @@
...
+                "import java.util.function.ToDoubleFunction;",
+                "import java.util.function.ToIntFunction;",
+                "import java.util.function.ToLongFunction;",
+                "import java.util.stream.Collector;",
+                "import java.util.stream.DoubleStream;",
+                "import java.util.stream.IntStream;",
+                "import java.util.stream.LongStream;",
+                "import java.util.stream.Stream;",
+                "class Test {",
+                "   public static class FooStream<T> implements Stream<T> {",
```
jkozlowski, line 73: nit: There must be a simpler way to create a stream? Stream.of(1)?

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DangerousParallelStreamUsage.java
```diff
@@ -0,0 +1,72 @@
...
+            + "does not allow you to provide your own pool. Fork/join pools implement work-stealing, where any thread "
+            + "might steal a task from a different thread's queue when blocked waiting for a subtask to complete. "
+            + "This might not seem like an issue at first glance, but if you use .parallel() for short tasks "
+            + "extensively throughout your codebase and later on you add one piece of code that uses .parallel() "
+            + "for long (e.g. I/O) tasks, the other parts of your codebase that use .parallel(), and that you'd "
+            + "expect to have consistent performance, might experience performance degradation for no apparent reason. "
+            + "The reason is work stealing.\n"
+            + "You can suppress this warning if you are certain that all your code will always only use .parallel() "
+            + "for short tasks, but even then, you have no real control over the level of parallelism, so you're "
+            + "still better off using MoreStreams (linked above)\n"
```
jkozlowski, line 55: nit: Not sure if specifying all of this background is necessary.

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/DangerousParallelStreamUsageTest.java
gm2211: what I'm testing for here is subclassing a stream to add a method with name `parallel` but more arguments and make sure we fail for that too - the idea here is also banning `StreamEx.of(1).parallel(new ForkJoinPool())` for the reasons explained by James on our email thread (i.e. poor control over parallelism since the `ForkJoinPool` takes the number of cores in a system as an input to estimate how many tasks to run in parallel vs taking the max number of concurrent tasks explicitly)

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DangerousParallelStreamUsage.java
gm2211: Do you think this much background is confusing? Any suggestion on what to cut? In general I dislike error prone checks that tell me not to do something without explanation. Happy to defer that to the stackoverflow link though

jkozlowski: Cool, if you think it gives useful detail that's fine.

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/DangerousParallelStreamUsageTest.java
jkozlowski: Cool, I'll defer to Dan if there's a better way to do this. Ideally your code would just say something like "when I call a method from this interface". This test is kinda asserting your understanding of this, but really it's testing if errorprone does the right thing.

jkozlowski:
@palantir/infrastructure could we get a review here?

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DangerousParallelStreamUsage.java
ferozco: I think its useful to have all the detail, it gives users more context being the rule and extra motivation for fixing the issue

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/DangerousParallelStreamUsageTest.java
ferozco: I agree its not ideal to have such a verbose test. I'm not sure of a much more succinct way of expressing this though

ferozco (approved):

jkozlowski:
and merge/release?

ferozco:
Sorry, used to blacklist bulldozer and autorelease...

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/DangerousParallelStreamUsageTest.java
gm2211: I agree this test is questionable, however, I wanted to document the ban of `StreamEx#parallel(ForkJoinPool)` (or any variant of `parallel`) explicitly in a test - that method is not part of the `Stream` api - so in theory one could argue that they could have their own, correct, implementation (in reality they don't, they just defer to Java's parallel streams), so erring on the side of caution

PR #539, opened 2019-02-12
[feature] New 'com.palantir.baseline-reproducibility' plugin

iamdanfox:
## Before this PR

I discovered that in most of our projects, running `./gradlew jar` twice on a completely clean checkout results in jars with different hashes. This defeats any kind of fancy gradle caching we might want to enable in the future.

## After this PR

Projects can `apply plugin: 'com.palantir.baseline-reproducibility'` to opt-in to more sensible behaviour.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,7 @@ _Baseline is a family of Gradle plugins for configuring Java projects with sensi
 | `com.palantir.baseline-circleci`         | [CircleCI](https://circleci.com/) integration using `$CIRCLE_ARTIFACTS` and `$CIRCLE_TEST_REPORTS` dirs
 | `com.palantir.baseline-versions`         | Source dependency versions from a `versions.props` file using [nebula dependency recommender](https://github.com/nebula-plugins/nebula-dependency-recommender-plugin)
 | `com.palantir.baseline-config`           | Config files for the above plugins
+| `com.palantir.baseline-reproducibility`  | Sensible defaults to ensure Jar, Tar and Zip tasks can be reproduced
 
 See also the [Baseline Java Style Guide and Best Practises](./docs).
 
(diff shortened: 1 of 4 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineReproducibility.java
```diff
@@ -0,0 +1,40 @@
...
+    @Override
+    public void apply(Project project) {
+        project.getTasks().withType(AbstractArchiveTask.class).configureEach(t -> {
+            t.setPreserveFileTimestamps(false);
+            t.setReproducibleFileOrder(true);
+            t.setDuplicatesStrategy(DuplicatesStrategy.WARN);
+        });
+
+        project.getPluginManager().withPlugin("nebula.info", p -> {
+            project.getLogger().warn("Please remove the 'nebula.info' plugin from '{}' as it breaks "
```
iamdanfox, line 36: I just removed this from our internal publish plugin.

README.md
```diff
@@ -289,3 +290,18 @@ spotless {
...
+## com.palantir.baseline-reproducibility
+
+This plugin is a shorthand for the following snippet, which opts-in to reproducible behaviour for all Gradle's Jar, Tar and Zip tasks. (Surprisingly, these tasks are not reproducible by default).
+
+```gradle
+tasks.withType(AbstractArchiveTask) {
+    preserveFileTimestamps = false
+    reproducibleFileOrder = true
+}
+```
```
iamdanfox, line 303: See https://dzone.com/articles/reproducible-builds-in-java for where I discovered this!

dansanduleac (approved):

PR #543, opened 2019-02-15
[improvement] BaselineIdea: Using targetCompatibility for jdkName (if available) instead of always using sourceCompatibility

gm2211:
## Before this PR

`sourceCompatibility` is used to set both `jdkName` (i.e. target jvm binary version) and `LANGUAGE_LEVEL` (i.e. what language features are allowed).

## After this PR
If `targetCompatibility` is specified, it will be used to set the `jdkName` used for compilation, while `sourceCompatibility` is still used to set the language level. If no `targetCompatibility` is specified, `sourceCompatibility` will be used for both (same as current behavior).


gm2211:
On second thought, I'm not sure we actually want to allow this though, since compiling for jdk11 even with language level 1.8 does not guarantee compatibility. On the other hand, though, we are now running tests in circle with jdk11 to test compatibility, so I'm not quite sure

Issue #176, opened 2017-06-02
ValidateConstantMessage check is buggy


darora:
Haven't played around with error prone too much, but from the looks of it we might need to split out the method on Validate.* based on the number of params they take before the error message. The current implementation (afaict) looks for the last string arg and takes that to be the pattern message.

An easy way to cause the current implementation to fail is using a string variable as the value to be interpolated into the error string. e.g. https://github.com/palantir/gradle-baseline/compare/develop...darora:develop

darora:
@ash211 to correct me if I overlooked something

gracew:
related: #175

iamdanfox:
Closing out stale issue

PR #546, opened 2019-02-20
Fix incorrect argument selection in ValidateConstantMessage

dsd987:
## Before this PR
The ValidateConstantMessage check would flag on the last argument that is a string rather than on the message string. Additionally, the grouping of test cases under the single positive test does not correctly test cases after the first one.

## After this PR
The ValidateConstantMessage check now knows the argument index of all message argument in all calls so it can be correctly identified and checked. Test cases are now all completely separated in their own tests.


dsd987:
Proposed fix for #176

iamdanfox (approved):
Thanks @dsd987, heroic work on the tests!

PR #547, opened 2019-02-21
[RFC] Fail for unused dependencies

gatesn:
Currently the task will:
* Fail if there are any jars referenced as a first-level dependency from which the `main` source set doesn't reference any classes.
* Suggest an alternative transitive dependency that contains one or more referenced classes.

This means if A -> B -> C, and a project takes a dependency on A but only uses classes from C, then the task will fail and recommend instead taking a dependency on C.

I wonder if a better/safer place to start would be to just throw for any dependency of which no classes are referenced from it or any of its transitives. i.e. a project takes a dependency on A -> B -> C, but no classes are referenced from any of A, B or C.


iamdanfox:
Moved to #548 so we can iterate

PR #549, opened 2019-02-24
[improvement] Disallow Multimap.keys() by default

mattsills:

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,6 +148,7 @@ Safe Logging can be found at [github.com/palantir/safe-logging](https://github.c
     ```
 - `ShutdownHook`: Applications should not use `Runtime#addShutdownHook`.
 - `SwitchStatementDefaultCase`: Switch statements should avoid using default cases. Default cases prevent the [MissingCasesInEnumSwitch](http://errorprone.info/bugpattern/MissingCasesInEnumSwitch.html) check from detecting when an enum value is not explicitly handled. This check is important to help avoid incorrect behavior when new enum values are introduced.
+- `MultimapKeys`: `Multimap.keys()` is usually an error, where the author meant to use `Multimap.keySet()`.
 
 ## com.palantir.baseline-checkstyle
 Checkstyle rules can be suppressed on a per-line or per-block basis. (It is good practice to first consider formatting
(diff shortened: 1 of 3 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/MultimapKeys.java
```diff
@@ -0,0 +1,60 @@
...
+        summary = "Multimap.keys() is usually an error. Did you mean to use Multimap.keySet()?")
+public final class MultimapKeys extends BugChecker implements BugChecker.MethodInvocationTreeMatcher {
+
+    private static final long serialVersionUID = 1L;
+    private static final String errorMsg =
+            "Multimap.keys() is usually an error. Did you mean to use Multimap.keySet()?";
+
+    private static final Matcher<ExpressionTree> multimapKeysMethod = MethodMatchers.instanceMethod()
+            .onClass(TypePredicates.isDescendantOf("com.google.common.collect.Multimap"))
+            .withNameMatching(Pattern.compile("keys"));
```
j-baker, line 47: given what this is, should probably be `.named("keys")` rather than invoking the regex version?

README.md
```diff
@@ -148,6 +148,7 @@ Safe Logging can be found at [github.com/palantir/safe-logging](https://github.c
     ```
 - `ShutdownHook`: Applications should not use `Runtime#addShutdownHook`.
 - `SwitchStatementDefaultCase`: Switch statements should avoid using default cases. Default cases prevent the [MissingCasesInEnumSwitch](http://errorprone.info/bugpattern/MissingCasesInEnumSwitch.html) check from detecting when an enum value is not explicitly handled. This check is important to help avoid incorrect behavior when new enum values are introduced.
+- `MultimapKeys`: `Multimap.keys()` is usually an error, where the author meant to use `Multimap.keySet()`.
```
iamdanfox, line 151: Don't think it's really good enough to say "foo is usually an error", as that is true by definition for all the checks - this sentence needs to explain to people _why_ we consider this a foot gun!

iamdanfox (changes requested):
@mattsills can you fix up the messaging here please? Would be great to get this into the next release

carterkozak:
@mattsills Are you still working on this? I'd like to take advantage of it when it's ready :-)

PR #550, opened 2019-02-25
[improvement] ./gradlew idea deletes redundant ipr files

### LINKED ISSUE #538: Something should fail if there are 2 .ipr files in the project root
## What happened?

Both members of my team have now spent >30 minutes dealing with an internal automatic refactor to change the project root name in support of gradle consistent versions.

What happens is that you keep using the old classpath, but future dependency updates don't apply to you.

Instead, I'd prefer that something throws.

According to @ferozco, the right place to throw is in here.

Externally, it'd be good if we could make the gradle-idea plugin throw if you have multiple .ipr files. Internally, we should be gating upgrading to consistent versions on having a version of baseline that includes this check. An alternative is that gradle-consistent-versions runs this check as a courtesy, and then there's no mandatory baseline upgrades.

That said, I think it's worth blocking future rollout of consistent versions on this; I view myself as being well above our bar for understanding how Gradle works, and if I can end up losing a good chunk of a day on this, then I expect that this will happen basically everywhere.

iamdanfox:
fixes https://github.com/palantir/gradle-baseline/issues/538

## Before this PR

While adopting gradle-consistent-versions (which sometimes requires renaming the root project), people have been confused by leftover *.ipr files.

## After this PR

Running ./gradlew idea will result in exactly on ipr file.

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
@@ -19,7 +19,9 @@ package com.palantir.baseline.plugins
 import groovy.xml.XmlUtil
 import java.nio.file.Files
 import java.nio.file.Paths
+import org.gradle.api.Action
 import org.gradle.api.Project
+import org.gradle.api.Task
 import org.gradle.api.tasks.compile.JavaCompile
 import org.gradle.plugins.ide.idea.IdeaPlugin
 import org.gradle.plugins.ide.idea.model.IdeaModel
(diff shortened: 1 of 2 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
```diff
@@ -56,6 +58,20 @@ class BaselineIdea extends AbstractBaselinePlugin {
...
         }
+
+        // leftover ipr files from a different rootProject.name are confusing - let's proactively
+        // clean them up. Intentionally using an Action<Task> to allow up-to-dateness
+        Action<Task> cleanup = new Action<Task>() {
+            void execute(Task t) {
+                project.delete(project.fileTree(
+                        dir: project.getProjectDir(),
+                        include: '*.ipr',
+                        exclude: "${rootProject.name}.ipr"
```
carterkozak: Any reason not to delete the *.iml and *.iws while we're at it?

iamdanfox:
Adding `do not merge` as this plugin gets applied to lots of subprojects as well as the root project, don't want to delete too much!

ferozco:
👍

ellisjoe:
This change seems to have broken me:

screen shot 2019-02-27 at 8 59 06 am

ellisjoe:
PR with a failing test here: https://github.com/palantir/gradle-baseline/pull/555

PR #555, opened 2019-02-27
[fix] Fix removal of subproject *.iml files

ellisjoe:
## Before this PR
Subproject *.iml files would be removed causing Intellij to fail to load the project

## After this PR
Subproject *.iml files are no longer removed

### DIFF
--- a/gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineIdeaIntegrationTest.groovy
+++ b/gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineIdeaIntegrationTest.groovy
@@ -254,4 +254,25 @@ class BaselineIdeaIntegrationTest extends AbstractPluginTest {
         real.exists()
         !iws.exists() && !iml.exists() && !ipr.exists()
     }
+
+    def 'does not delete subproject iml files'() {
+        when:
+        buildFile << standardBuildFile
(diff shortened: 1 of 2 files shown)


gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineIdeaIntegrationTest.groovy
```diff
@@ -254,4 +254,25 @@ class BaselineIdeaIntegrationTest extends AbstractPluginTest {
...
+        then:
+        !real.exists()
+        !subIml.exists()
+        oldSubIml.exists()
+
+        with('idea').build()
+
+        real.exists()
+        subIml.exists()
+        !oldSubIml.exists()
```
ellisjoe, line 276: interesting this is the part that fails

ellisjoe:
The issue ended up being that there was a subproject with the same name as the root project

iamdanfox:
@ellisjoe I think it's not really desirable for this task to delete stuff recursively... probably still consider this a bug, as baseline should not prevent you from having identically named projects!

ellisjoe:
well the issue as far as I could tell was that the duplicate name cause the project's *.iml file to be named `<project-name>-<project-name>.iml` which then cause this task to delete it. Not sure why the *.iml file would be named like that? but don't think this is actually recursively deleting things

PR #558, opened 2019-02-27
[fix] Special case for proj-proj.ipr files

### LINKED ISSUE #557: 0.51.0 breaks projects with root and one subproject name the same
I _think_ this is already being tracked by @iamdanfox per his comment here: https://github.com/palantir/gradle-baseline/pull/555#issuecomment-467837448, but filing formally.

iamdanfox:
## Before this PR

baseline-idea 0.51.0 broke workflows for any repo where the rootproject has the same name as a subproject.

## After this PR

Sensible behaviour is restored - files of pattern `name-name.ipr` are no longer deleted.

fixes https://github.com/palantir/gradle-baseline/issues/557

cc @gregakinman @ellisjoe


dansanduleac:
Can we do #559 instead?

PR #559, opened 2019-02-28
[fix] Fix cleanup of old idea project files

### LINKED ISSUE #557: 0.51.0 breaks projects with root and one subproject name the same

dansanduleac:
## Before this PR

Regression introduced in #550 means that `.iml` files might be removed if the idea extension were configured in such a way as for them to differ from the project name.
Even if not explicitly configured, if a subproject had the same name as the root project, the IDEA plugin would give it a name like `${project.name}-${project.name}.iml` as mentioned in https://github.com/palantir/gradle-baseline/pull/555#issuecomment-467851332, which differs from the name the cleanup task was instructed to exclude.

## After this PR

Decide based on the outputs of the actual tasks creating these ipr / iws / iml files what the name that they would use is, and exclude just that name, rather than assuming it's going to be e.g. `${project.name}.iml`.

Fixes #557.

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
@@ -64,7 +64,7 @@ class BaselineIdea extends AbstractBaselinePlugin {
         Action<Task> cleanup = new Action<Task>() {
             void execute(Task t) {
                 project.delete(project.fileTree(
-                        dir: project.getProjectDir(), include: '*.ipr', exclude: "${project.name}.ipr"));
+                        dir: project.getProjectDir(), include: '*.ipr', exclude: "${project.name}.ipr"))
                 project.delete(project.fileTree(
                         dir: project.getProjectDir(), include: '*.iml', exclude: "${project.name}.iml"))
                 project.delete(project.fileTree(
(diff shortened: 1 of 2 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
```diff
@@ -53,26 +60,36 @@ class BaselineIdea extends AbstractBaselinePlugin {
...
-            moveProjectReferencesToEnd(ideaModuleModel);
+            markResourcesDirs(ideaModuleModel)
+            moveProjectReferencesToEnd(ideaModuleModel)
         }
 
         // If someone renames a project, leftover {ipr,iml,ipr} files may still exist on disk and
         // confuse users, so we proactively clean them up. Intentionally using an Action<Task> to allow up-to-dateness.
         Action<Task> cleanup = new Action<Task>() {
             void execute(Task t) {
+                if (t.project.rootProject == t.project) {
```
iamdanfox, line 72: Why do we still need this if you're accurately tracking the outputs of the idea task?

dansanduleac: the ipr and iws tasks only exist on the root project

dansanduleac: remember that this plugin is applied on all projects when you apply `com.palantir.baseline`

iamdanfox:
:+1:

gregakinman:
thanks!

PR #561, opened 2019-03-01
[improvement] errorprone 2.3.2 -> 2.3.3

iamdanfox:
error-prone usually brings some welcome new static analysis.

Looks like we just needed to delete our Category annotations: https://github.com/google/error-prone/commit/1f2fed928c87a548588278fe4333688e40b31ffe

I'd prefer not to start using the new tags just yet so that anyone who happens to be forcing 2.3.2 somehow keeps working


carterkozak (approved):
lgtm, thanks Dan!

PR #564, opened 2019-03-05
[improvement] Split out circle style plugin into generic junit reports plugin

gatesn:
## Before this PR
The Baseline Circle style plugin performed all configuration for setting up junit reporting for test tasks, javac tasks, etc. In order to do the same, one must depend on all of gradle-baseline.

## After this PR
A new `com.palantir.junit-reports` plugin has been created that sets up junit XML reporting to a configurable reports directory. The Baseline circle style plugin then configures this with the Circle environment variables. This allows us to move away from the now deprecated gradle-circle-style repository.


robert3005 (approved):
I think this looks good. Would be nice to refactor baselineCircleci in similar style but that's fine to make a separate pr for it

PR #576, opened 2019-03-15
[fix] Avoid lambda to allow build caching of checkstyle results

jdm2212:
## Before this PR
Caching checkstyle results isn't possible due to same issue as in https://github.com/palantir/gradle-conjure/pull/116

I can share an internal build scan pointing to this as the culprit.

## After this PR
Caching checkstyleMain results should be possible.


carterkozak:
Is this only on gradle 5.1+ as well?

jdm2212:
probably? it's the same issue -- and we're on 5.2.1

iamdanfox:
:+1: As a long-term solution (and hopefully to prevent regressions on this), I started writing some gradle plugin specific error-prone rules here: https://github.com/iamdanfox/gradle-clippy.

Seems like actually passing a lambda do a doLast is a much more obviously wrong thing that the other lints I already had!

PR #581, opened 2019-03-22
[improvement] Prevent commit of ShutdownStrategy.SKIP

nackner:
Per https://github.com/palantir/docker-compose-rule/blob/a2adeae9ac12b7fc4c9f7a8d1bacaa7db97a430a/docker-compose-rule-core/src/main/java/com/palantir/docker/compose/configuration/ShutdownStrategy.java#L56, we do not want to commit docker-compose-rule's ShutdownStrategy.SKIP to avoid leaking resources during CI.


palantirtech:
Thanks for your interest in [palantir/gradle-baseline](https://github.com/palantir/gradle-baseline), @nackner! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

iamdanfox:
Alternatively, what about making a change to the DCR library so that SKIP detects when it's on CI and fails hard?

Don't really want to use checkstyle to enforce the ins-and-outs of other random libraries

nackner:
Agree with your counterproposal, closing out.

PR #582, opened 2019-03-23
[feature] New `com.palantir.baseline-release-compatibility` plugin

schlosna:
## Before this PR

When compiling with `targetCompatibility = 1.8` while running on JDK 9+, one may accidentally introduce binary incompatibilities in the generated byte code as it is implicitly linked against the building JDK's class files. This can lead to runtime issues, for example running on JDK 8 `java.lang.NoSuchMethodError: java.nio.ByteBuffer.rewind()Ljava/nio/ByteBuffer;` as seen in https://github.com/palantir/conjure-java/pull/293

## After this PR

When using JDK 9+ to compile with a `targetCompatibility` less than JDK 9, this plugin adds compiler arguments per [JEP 247](https://openjdk.java.net/jeps/247) to explicitly set the target JDK platform API to maintain binary compatibility.

See also [Gradle JDK release issue](https://github.com/gradle/gradle/issues/2510).


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineJdkRelease.java
```diff
@@ -0,0 +1,73 @@
...
+public final class BaselineJdkRelease extends AbstractBaselinePlugin {
+
+    @Override
+    public void apply(Project project) {
+        this.project = project;
+        project.getPluginManager().withPlugin("java", plugin -> {
+            project.getTasks().withType(JavaCompile.class).configureEach(javaCompile -> {
+                try {
+                    configureJavaCompiler(javaCompile);
+                } catch (Throwable throwable) {
```
ferozco: What errors are we attempt to catch? Do we need the try-catch?

ferozco (commented):
looks good!

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineJdkRelease.java
```diff
@@ -0,0 +1,73 @@
...
+                    project.getLogger().error("Failed to configure BaselineJdkRelease", throwable);
+                }
+            });
+        });
+    }
+
+    private void configureJavaCompiler(JavaCompile javaCompile) {
+        JavaVersion sourceCompatibility = JavaVersion.toVersion(javaCompile.getTargetCompatibility());
```
uschi2000: source vs target, is that on purpose? if so, please comment and explain

gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineJdkReleaseTest.groovy
```diff
@@ -0,0 +1,59 @@
...
+import spock.lang.Specification
+
+class BaselineJdkReleaseTest extends Specification {
+    private Project project
+
+    def setup() {
+        project = ProjectBuilder.builder().build()
+        project.plugins.apply 'java'
+        project.plugins.apply BaselineJdkRelease
+        project.getConvention().getPlugin(JavaPluginConvention.class).sourceCompatibility = 7
```
uschi2000: this line is unused in the test

uschi2000:
also needs some README love, please

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineJdkRelease.java
schlosna: I think we want use `targetCompatibility` in general (at least as default), though if we wanted to allow flexibility we could introduce a separate `releaseCompatibility` extension property.

* `sourceCompatibility`: The version of Java source code language features that are permitted by the compiler. This must be less than or equal to `target`
* `targetCompatibility`: The version of JVM bytecode class files that will be generated. This must be greater than or equal to `source`
* `releaseCompatibility`: The version of JDK API libraries to link against (equivalent to pre JDK 9 `bootclasspath`). This implicitly indicates that `source == target == release`

See also https://docs.oracle.com/en/java/javase/11/tools/javac.html#GUID-AEEC9F07-CB49-4E96-8BC7-BCC2C7F725C9 and https://docs.oracle.com/javase/8/docs/technotes/tools/windows/javac.html#BHCIJIEG 

Note that you can't use `source` and `target` with `release` for `javac`:
```
$ javac -version -source 6 -target 7 --release 8 Foo.java
javac 11.0.2
error: option -source cannot be used together with --release
error: option -target cannot be used together with --release
```

gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineJdkReleaseTest.groovy
schlosna: this is setting `sourceCompatibility` to ensure that the decision is *not* driven by source compatibility.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineJdkRelease.java
schlosna: I added this as I was debugging since this was my first gradle plugin so I'd have a place to set a debug breakpoint, and gradle was eating some exceptions, so wanted to make sure they were logged if one occurs

schlosna (commented):
Thanks Felipe and Rob for reviews. I will try to write up some README notes, but that probably won't happen today in my timezone, though happy if you or your teams want to contribute some docs and validation.

iamdanfox:
Before rolling this out to everyone, I'd like to confirm whether we're affected at all by https://github.com/google/error-prone/issues/1237

iamdanfox:
@schlosna I gave this a go locally and found we have to use afterEvaluate or providers, otherwise our plugin would try to set up the `--release` flag before the user's closure to set `targetCompatibility = 1.8` has actually been evaluated!

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineReleaseCompatibility.java
```diff
@@ -0,0 +1,91 @@
...
+                return Collections.emptyList();
+            }
+
+            return ImmutableList.of("--release", target.getMajorVersion());
+        });
+    }
+
+    // The --release flag was added in Java 9: https://openjdk.java.net/jeps/247
+    private static boolean supportsReleaseFlag(JavaVersion jdkVersion) {
+        return jdkVersion.compareTo(JavaVersion.VERSION_1_8) > 0;
```
dansanduleac: consider using isJava9Compatible

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineReleaseCompatibility.java
```diff
@@ -0,0 +1,91 @@
...
+    private static final Logger log = LoggerFactory.getLogger(BaselineReleaseCompatibility.class);
+
+    @Override
+    public void apply(Project project) {
+        this.project = project;
+
+        project.getTasks().withType(JavaCompile.class).configureEach(this::configureTask);
+    }
+
+    private void configureTask(JavaCompile javaCompile) {
```
dansanduleac: can we make this static?

dansanduleac (approved):

PR #583, opened 2019-03-27
[break] Enable running of unique class check on multiple configurations

raiju:
## Before this PR

For e.g. spark docker builds we need to be able to check the class uniqueness on two different configurations.

## After this PR

An optional extension is added to add different configurations to check. If nothing is specified, the build will continue function as normal.

## Break context

This changes gradle task names (`checkClassUniqueness` => `checkRuntimeClassUniqueness`). I can set up an dummy task to keep back-compat if preferable.


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/extensions/BaselineClassUniquenessExtension.java
```diff
@@ -0,0 +1,40 @@
...
+import javax.inject.Inject;
+import org.gradle.api.model.ObjectFactory;
+import org.gradle.api.provider.ListProperty;
+import org.gradle.api.provider.Provider;
+
+public class BaselineClassUniquenessExtension {
+    private final ListProperty<String> configurations;
+
+    @Inject
+    public BaselineClassUniquenessExtension(ObjectFactory objectFactory) {
```
iamdanfox: I think this imposes quite a high Gradle minimum version - would prefer to just pass in the whole Project as normal.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckClassUniquenessTask.java
```diff
@@ -23,55 +23,60 @@
...
-    public final Configuration getConfiguration() {
-        return configuration;
+    public final Provider<List<Configuration>> getConfigurations() {
+        return configurations;
     }
 
-    public final void setConfiguration(Configuration configuration) {
-        this.configuration = configuration;
+    public final void setConfigurations(Provider<List<Configuration>> configs) {
+        this.configurations = configs;
```
iamdanfox: I think the private `configurations` field should really be a ListProperty<Configuration> and call .set() here.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineClassUniquenessPlugin.java
```diff
@@ -30,9 +33,18 @@
 
     @Override
     public final void apply(Project project) {
+        BaselineClassUniquenessExtension extension = project.getExtensions()
+                .create("classUniqueness", BaselineClassUniquenessExtension.class);
+
         project.getPlugins().withId("java", plugin -> {
             project.getTasks().create("checkClassUniqueness", CheckClassUniquenessTask.class, task -> {
-                task.setConfiguration(project.getConfigurations().getByName("runtime"));
+                task.setConfigurations(extension.getConfigurations().map(confs -> {
```
iamdanfox: rather than making one task do more things, what about just creating more tasks?

e.g. perhaps we say 'runtimeClasspath' is the only configuration we check initially, and we'd create a `checkRuntimeClasspathUniqueness` task.

I think Gradle has a concept of Rules that might be helpful here...

raiju: Not very familiar with that type of gradle usage. Do you have a sample plugin that does that?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineClassUniquenessPlugin.java
```diff
@@ -27,14 +31,28 @@
...
     public final void apply(Project project) {
+        BaselineClassUniquenessExtension extension = project.getExtensions()
+                .create("classUniqueness", BaselineClassUniquenessExtension.class, project);
+
         project.getPlugins().withId("java", plugin -> {
-            project.getTasks().create("checkClassUniqueness", CheckClassUniquenessTask.class, task -> {
-                task.setConfiguration(project.getConfigurations().getByName("runtime"));
-                project.getTasks().getByName("check").dependsOn(task);
-            });
+            List<String> configurations = extension.getConfigurations().getOrElse(DEFAULT_CONFIGURATION);
```
raiju: @iamdanfox Concretely, how do I make this bit wait for the extension configuration to be passed in?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckClassUniquenessTask.java
```diff
@@ -46,14 +46,15 @@ public final Configuration getConfiguration() {
         return configuration;
     }
 
-    public final void setConfiguration(Configuration configuration) {
-        this.configuration = configuration;
+    public final void setConfiguration(Configuration configs) {
```
ferozco: think this can be reverted

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/rules/BaselineClassUniquenessRule.java
```diff
@@ -0,0 +1,68 @@
...
+
+            List<String> candidates = ImmutableList.of(altName, confName);
+
+            Optional<Configuration> configOpt = candidates.stream()
+                    .map(project.getConfigurations()::findByName)
+                    .filter(Objects::nonNull)
+                    .findFirst();
+
+            if (configOpt.isPresent()) {
+                project.getTasks().create(taskName, CheckClassUniquenessTask.class, task -> {
```
ferozco: `project.getTasks.register(...) allows you to lazily create the task

raiju:
If you want this to not be breaking, can apply:
```diff
diff --git a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineClassUniquenessPlugin.java b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineClassUniquenessPlugin.java
index e5a520e..46d8412 100644
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineClassUniquenessPlugin.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineClassUniquenessPlugin.java
@@ -32,7 +32,8 @@ public class BaselineClassUniquenessPlugin extends AbstractBaselinePlugin {
         project.getTasks().addRule(new BaselineClassUniquenessRule(project));
 
         project.getPlugins().withId("java", plugin -> {
-            project.getTasks().getByName("check").dependsOn("checkRuntimeClassUniqueness");
+            project.getTasks().register("checkClassUniqueness", task -> task.dependsOn("checkRuntimeClassUniqueness"));
+            project.getTasks().getByName("check").dependsOn("checkClassUniqueness");
         });
     }
 }
```

ferozco (approved):

iamdanfox:
This looks fantastic, thanks @raiju. No need to worry about keeping the exact same old task name - I think the few users of this plugin can get used to the new one!

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineClassUniquenessPlugin.java
```diff
@@ -27,14 +27,16 @@
...
+        BaselineClassUniquenessRule rule = new BaselineClassUniquenessRule(project);
+
+        project.getTasks().addRule(rule);
+
         project.getPlugins().withId("java", plugin -> {
-            project.getTasks().create("checkClassUniqueness", CheckClassUniquenessTask.class, task -> {
-                task.setConfiguration(project.getConfigurations().getByName("runtime"));
-                project.getTasks().getByName("check").dependsOn(task);
-            });
+            rule.apply("checkRuntimeClassUniqueness");
```
raiju, line 37: Needed to do ^ to get `register` in the BaselineClassUniquenessRule class to work (if I change `register` to `create`, I don't need the `rule.apply`)

raiju:
@iamdanfox @ferozco Tests pass now. A bit nervous about the "hack" I had to add... Seems register has slightly different behavior from create after all.

raiju:
What is the release frequency of this repository?

PR #601, opened 2019-04-09
[fix] checkImplicitDependencies shouldn't count ignored artifacts

dansanduleac:
## Before this PR

Ignored undeclared dependencies are only filtered out of the suggestion, but not the count of implicit dependencies.

```
-----------
* What went wrong:
Execution failed for task ':foo:checkImplicitDependencies'.
> Found 1 implicit dependencies - consider adding the following explicit dependencies to 'foo/build.gradle', or avoid using classes from these jars:
      dependencies {

      }
```

## After this PR

==COMMIT_MSG==
`checkImplicitDependencies` doesn't count ignored artifacts in the total number of implicit dependencies found.
==COMMIT_MSG==


iamdanfox:
:+1:

PR #605, opened 2019-04-10
[fix] BaselineReleaseCompatibility up-to-date checking of compile tasks

iamdanfox:
## Before this PR

In 0.54.0, we introduced the `com.palantir.baseline-release-compatibility` plugin to add the `--release` flag and make sure we don't rely on parts of the standard library that aren't actually available.

However, it seems I broke up-to-dateness of compile tasks 😭. From 0.54.0 onwards, running `./gradlew classes` twice is slow both times:

1. `./gradlew classes` (1m30)
1. `./gradlew classes` (1m00)

Screenshot 2019-04-10 at 22 42 32

## After this PR

Up-to-dateness checking works again, so the second compile is drastically faster:

1. `./gradlew classes` (1m30)
1. `./gradlew classes` (4s)

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineReleaseCompatibility.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineReleaseCompatibility.java
@@ -22,6 +22,7 @@ import java.util.Optional;
 import org.gradle.api.JavaVersion;
 import org.gradle.api.Project;
 import org.gradle.api.tasks.compile.JavaCompile;
+import org.gradle.process.CommandLineArgumentProvider;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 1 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineReleaseCompatibility.java
```diff
@@ -42,13 +45,22 @@
...
-        project.getTasks().withType(JavaCompile.class).configureEach(BaselineReleaseCompatibility::configureTask);
+        project.getTasks().withType(JavaCompile.class).configureEach(javaCompile -> {
+            javaCompile.getOptions().getCompilerArgumentProviders().add(new ReleaseFlagProvider(javaCompile));
+        });
     }
 
-    private static void configureTask(JavaCompile javaCompile) {
-        // using a lazy argument provider is crucial because otherwise we'd try to read sourceCompat / targetCompat
-        // before the user has even set it in their build.gradle!
-        javaCompile.getOptions().getCompilerArgumentProviders().add(() -> {
```
iamdanfox, line 51: making this a lambda was a crime according to gradle.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineReleaseCompatibility.java
```diff
@@ -42,13 +43,22 @@
...
+    // using a lazy argument provider is crucial because otherwise we'd try to read sourceCompat / targetCompat
+    // before the user has even set it in their build.gradle!
+    private static final class ReleaseFlagProvider implements CommandLineArgumentProvider {
```
ferozco, line 53: can this be an anonymous class for the sake of readability?

iamdanfox: I actually want this to be named so that if it ever appears in a stacktrace or warning message we'll know exactly where to look.  The `$0` message before was pretty hard to spot.

ferozco:
👍

PR #607, opened 2019-04-11
[fix] Exact dependency only for jar deps

gatesn:
## Before this PR
We had a dependency `<group>:<artifact>@pom` that caused the exact dependencies plugin to fail.

`Cannot accept visitor on URL: ...`

## After this PR
Only jars are visited


iamdanfox:
We should really start adding tests for this soon....

iamdanfox (approved):

gatesn:
Yeah, don't merge this. It breaks :(

PR #608, opened 2019-04-12
[feature] Gradle task action errorprone check

dansanduleac:
## Before this PR

Devs could accidentally pass a lambda to `Task.doFirst` or `Task.doLast` actions in gradle plugins. This makes the task non-cacheable, as pointed out in https://github.com/gradle/gradle/issues/5510.

## After this PR

==COMMIT_MSG=
Add an errorprone rule `GradleCacheableTaskAction` that prevents passing a lambda to `Task.doFirst` or `Task.doLast` when implementing gradle plugins.
==COMMIT_MSG==


ferozco:
👍

PR #622, opened 2019-04-30
[improvement] Replace Iterables.partition(List, int) with Lists.partition(List, int)

### LINKED ISSUE #621: FR: Prefer Lists.partition over Iterables.partition when arg is already a list
## What happened?

While trying to reduce memory allocation in Atlas, @diogoholanda found that calls to `Iterables.partition(list, number)` were actually allocating unnecessarily.

> Iterables.partition pre-allocates an Array of the desired batch size for every batch. This is generally a good perf optimization, but can lead to bad situations in extreme scenarios.
For instance, LockRefreshingLockService refreshes the locks in batches of 500K every 5 seconds. For a service containing just a few locks, this leads to allocation of arrays of size 500K every 5 seconds (2Mb every 5 seconds).
If the Iterable being partitioned is a List, Lists.partition is much more efficient and simply relies on the .sublist method.

errorprone should detect when users call Iterables.partition with a list as the first argument and tell people to use Lists.partition

schlosna:
## Before this PR

No warnings when invoking `Iterables.partition(Iterable, int)` with a first argument declared type of `List`.

## After this PR

The user is warned that one should prefer `Lists.partition(List, int)`.

Fixes https://github.com/palantir/gradle-baseline/issues/621

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DangerousIterablesPartitionUsage.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DangerousIterablesPartitionUsage.java
@@ -22,6 +22,7 @@ import com.google.errorprone.BugPattern.SeverityLevel;
 import com.google.errorprone.VisitorState;
 import com.google.errorprone.bugpatterns.BugChecker;
 import com.google.errorprone.fixes.SuggestedFix;
+import com.google.errorprone.fixes.SuggestedFixes;
 import com.google.errorprone.matchers.Description;
 import com.google.errorprone.matchers.Matcher;
 import com.google.errorprone.matchers.Matchers;
@@ -64,12 +65,12 @@ public final class DangerousIterablesPartitionUsage extends BugChecker
 
             if (LIST_MATCHER.matches(args.get(0), state)) {
                 // Fail on any 'Iterables.partition(List, int) invocation
+                SuggestedFix.Builder fix = SuggestedFix.builder();
+                String qualifiedType = SuggestedFixes.qualifyType(state, fix, "com.google.common.collect.Lists");
+                String method = qualifiedType + ".partition";
                 return buildDescription(tree)
                         .setMessage(ERROR_MESSAGE)
-                        .addFix(SuggestedFix.builder()
-                                .replace(tree, String.format("Lists.partition(%s, %s)", args.get(0), args.get(1)))
-                                .addImport("com.google.common.collect.Lists")
-                                .build())
+                        .addFix(fix.replace(tree.getMethodSelect(), method).build())
                         .build();
             }
         }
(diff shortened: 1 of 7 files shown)


baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/DangerousIterablesPartitionUsageTests.java
```diff
@@ -0,0 +1,66 @@
...
+        compilationHelper
+                .addSourceLines(
+                        "Test.java",
+                        "import com.google.common.collect.Iterables;",
+                        "import java.lang.Iterable;",
+                        "import java.util.List;",
+                        "class Test {",
+                        "  Iterable<?> f(List<?> list) {",
+                        "    // BUG: Diagnostic contains: Prefer Lists.partition",
+                        "    return Iterables.partition(list, 10);",
```
carterkozak: Mind adding a test for correct use of Iterables.partition on a non-list?

The safe preconditions check tests the suggested replacement as well, could we use that here?

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DangerousIterablesPartitionUsage.java
```diff
@@ -0,0 +1,80 @@
...
+
+            if (LIST_MATCHER.matches(args.get(0), state)) {
+                // Fail on any 'Iterables.partition(List, int) invocation
+                return buildDescription(tree)
+                        .setMessage(ERROR_MESSAGE)
+                        .addFix(SuggestedFix.builder()
+                                .replace(tree, String.format("Lists.partition(%s, %s)", args.get(0), args.get(1)))
+                                .addImport("com.google.common.collect.Lists")
+                                .build())
+                        .build();
```
carterkozak: May want to borrow from here for suggested-fix edge cases, it will add the import unless there's already an import for a `Lists` from another package, in which case it uses fqcn.

https://github.com/palantir/gradle-baseline/blob/24cae3c56c77e4bd2025aca71bd364d43f60f4bf/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferSafeLoggingPreconditions.java#L97-L107

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/DangerousIterablesPartitionUsageTests.java
carterkozak: Example: https://github.com/palantir/gradle-baseline/blob/24cae3c56c77e4bd2025aca71bd364d43f60f4bf/baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/PreferSafeLoggableExceptionsTest.java#L43-L61

carterkozak:
lgtm, thanks! 👍

markelliot:
this feels like a good candidate for pushing upstream to errorprone core?

schlosna:
@markelliot yeah, I can look at pushing this and the Collections2/Lists.transform upstream to error-prone and possibly Guava.

PR #623, opened 2019-05-01
[improvement] Prefer Lists or Collections2 transfrom over Iterables.transform

schlosna:
When using Guava's lazy transforms such as `Iterables.transform` to avoid full copying of collection via Java Streams, one should generally prefer `Lists.transform` or `Collections2.transform` where possible as the resulting lazy collection provides `O(1)` `size()` and generally no allocation for `isEmpty()`.

## Before this PR

No warnings when invoking `Iterables.transform(Iterable, Function)` with a first argument declared type of `List` or `Collection`.

## After this PR

The user is warned that [one should prefer `Lists.transform(List, Function)` or `Collections2.transform(Collection, Function)` per Guava JavaDoc](https://google.github.io/guava/releases/23.0/api/docs/com/google/common/collect/Iterables.html#transform-java.lang.Iterable-com.google.common.base.Function-).

Similar to #622


carterkozak:
Love it 👍
Feel free to merge when you're ready

schlosna:
@carterkozak I don’t have write/merge on this repo and can’t label “merge when ready” either.

PR #629, opened 2019-05-10
[fix] Add Javac Settings to uncheck "Use compiler from module target JDK when possible"

### LINKED ISSUE #618: Configure ideal ProjectRootManager to match project targetcompatibility
## What happened?
Generate a project using java 11 via `./gradlew idea` that has targetCompatibility of 1.8. That then fails to compile in intellij with javac compilation failure

The compilation should succeed. One way to make this work is to make ProjectRootManager refer to 1.8 of jdk, though it's confusing why using java 11 causes intellij compilation to fail.

diogoholanda:
Fixes #618

## Before this PR
"Use compiler from module target JDK when possible" is checked by default, which prevents compilation of projects in IntelliJ with an SDK11 and source compatibility 8.

## After this PR
"Use compiler from module target JDK when possible" is unchecked and compilation works


iamdanfox (approved):
I ran publishToMavenLocal and verified this works with IntelliJ CE 2019.1.

Screenshot 2019-05-13 at 11 51 08

Thanks for hunting down this option @diogoholanda!

PR #631, opened 2019-05-10
Discourage use of assert

aioobe:


docs/best-practices/java-coding-guidelines/readme.md
```diff
@@ -270,6 +270,13 @@ Preconditions.checkArgument(Pattern.matches(PASSWORD_REGEX, password),
 
 See *Effective Java, 2nd Edition, Item 38*
 
+### Prefer explicit `if` and `throw` or `Precondition` over `assert`
+
+An `assert` statement is only executed if the JVM is started with
+`--enableassertions`. This is not the case in production, which means
+that asserts can give a false impression of actually guarding against
+a condition.
```
schlosna: This one is probably a bit of a debate. There are ways to leverage `assert`s well to perform expensive state validations in test environments without incurring that expense in production environments. There are a lot of existing uses of asserts in projects to perform these expensive correctness checks, including in AtlasDB via https://github.com/palantir/atlasdb/blob/develop/atlasdb-commons/src/main/java/com/palantir/util/AssertUtils.java

aioobe: Interesting. How about we add something like "Exception: If alternatives such as `Preconditions` incur an unacceptable performance overhead, `assert` is still better than nothing."

Just to cover cases were people otherwise would think it's the recommended way simply because they have seen it elsewhere in pgdev.

uschi2000: I would like to avoid bifurcating the world into test and production behavior. I suspect that expensive *internal*  state validations at runtime are an antipattern; if some invariant is expected to hold, then the invariant should be "proven" through test coverage, and the remaining uncertainty must be handled at with runtime checks in production. As such, I am in favour of either discouraging or banning `assert`, or alternatively enabling it in production. The former is likely a better choice because it gives us more much freedom for defining the logging behavior in case an assertion fails.

aioobe: @uschi2000 I agree with your argument. However, I also think that in order to get a wide adoption and acceptance of this type of document, one should try to strike a balance between the ideal world situation and the current state of affairs.

There are currently 3470 asserts in non-test code in one of the large repositories and "forbidding" this might be met with "well that's not how we do it here" and poor adoption in general.

aioobe:
@schlosna updated wording slightly.

docs/best-practices/java-coding-guidelines/readme.md
uschi2000: Sorry, this fell off my radar, @aioobe. The written docs here are toothless compared to CI-time tooling; in many cases, they represent the aspiration rather than reality. I would thus prefer to keep the aspirational statements in such that they indicate the right direction, and gradually work on tooling adoption to converge reality.

schlosna: I'm supportive of writing an error-prone warning check to flag uses of `assert` to push more of the cheap checks to `Preconditions`/`Verify` checks, and requiring explicit suppression where we do complex & expensive validations that should truly only occur when running in tests & simulation environments and not eat that cost in production.

Could use https://github.com/google/error-prone/blob/master/core/src/main/java/com/google/errorprone/bugpatterns/AssertFalse.java / https://errorprone.info/bugpattern/AssertFalse as a starting point.

aioobe: So no exceptions in the docs. Recommendation is to always avoid `assert`. Happy to implement error-prone warnings, but if we can agree that the docs update is good in itself, it would be nice to get closure on this and then I'll open a separate PR for the slightly bigger error-prone part of the work when I get around to it.

uschi2000: +1. Thank you!

uschi2000:
@carterkozak are you cool with what is proposed here?
- docs to say `assert` should be avoided
- corresponding error-prone check to be added to baseline
- internal repos to be adapted accordingly (one by one)

carterkozak:
Yep, I support this. We've been bitten by several asserts that should have been preconditions. I agree that there are cases where it's helpful, but they're the exception and an error-prone suppression sounds reasonable for those situations.

uschi2000 (approved):

PR #638, opened 2019-05-16
[fix] class uniqueness rule must have a config

jmcampanini:
## Before this PR
The rule would match against `checkClassUniqueness`, which would fail since upper-casing an empty string ends up in an index out of bounds exception.

## After this PR
Doesn't match against `checkClassUniqueness` since it now requires at least 1 character in the regex.


ferozco:
👍

PR #641, opened 2019-05-21
[improvement] Spotless to remove blank lines at start of methods

gatesn:
## Before this PR
Methods starting with blank lines are annoying

## After this PR
Methods cannot start with blank lines

### DIFF
--- a/gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineFormatIntegrationTest.groovy
+++ b/gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineFormatIntegrationTest.groovy
@@ -36,8 +36,18 @@ class BaselineFormatIntegrationTest extends AbstractPluginTest {
 
     def validJavaFile = '''
     package test;
-    import com.java.tests;
-    public class Test { void test() {} }
+    public class Test { void test() {
+        int x = 1;
+    } }
+    '''.stripIndent()
+
+    def invalidJavaFile = '''
+    package test;
+    public class Test { void test() {
+    
+    
+        int x = 1;
+    } }
     '''.stripIndent()
 
     def 'can apply plugin'() {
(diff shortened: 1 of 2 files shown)


gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineFormatIntegrationTest.groovy
```diff
@@ -59,16 +69,13 @@ class BaselineFormatIntegrationTest extends AbstractPluginTest {
...
 
         then:
         BuildResult result = with('format').build();
         result.task(":format").outcome == TaskOutcome.SUCCESS
         result.task(":spotlessApply").outcome == TaskOutcome.SUCCESS
-        file('src/main/java/test/Test.java').text == '''
-            package test;
-            public class Test { void test() {} }
-        '''.stripIndent()
+        file('src/main/java/test/Test.java').text == validJavaFile
```
ellisjoe, line 79: think this test used to verify that the unused `import com.java.tests` was removed, so probably want to keep that around

gatesn: Added it back as `import com.java.unused;` to be explicit

dansanduleac (approved):

gregakinman:
I think it's helpful to have blank lines at the beginning of methods if the method signature is multi-line and the method implementation is long enough to be broken up into multiple logical groups by blank lines. A bit contrived of an example, but I think:

```
private Foo computeFoo(String arg1,
        String arg2,
        String arg3) {

    doAThing();
    doAThing();
    doAThing();

    doAnotherThing();
    doAnotherThing();
    doAnotherThing();
    
    return result;
```

is more readable than:

```
private Foo computeFoo(String arg1,
        String arg2,
        String arg3) {
    doAThing();
    doAThing();
    doAThing();

    doAnotherThing();
    doAnotherThing();
    doAnotherThing();
    
    return result;
```

For me, at least, it helps improve readability by clearly demarcating between "end of method signature" and "beginning of implementation" in a way that goes further than just the indentation difference that is present. I also think this is subjective enough not to enforce one over the other (I haven't seen anything like this in a style guide somewhere).

carterkozak:
Thanks for pushing this through, makes things much cleaner :-)

fmance:
+1 to @gregakinman's comment

iamdanfox:
@fmance @gregakinman if you can implement something that enforces (and auto-fixes) your desired formatting, I'd happily review a PR!

fmance:
@iamdanfox Could we just revert this PR ? Maybe baseline shouldn't really have an opinion on this matter of blank lines at the start of methods.

fmance:
Alternatively, it could be turned into a checkstyle check than can be disabled

iamdanfox:
As a rule of thumb, I usually want baseline to help people make their repos look more and more consistent and homogenous, especially when there's an autoformatter.  

This means that by making the autoformatter better, we can improve every single repo.  If lots of teams don't like this one, then sure we can revert, but so far I think people have generally just merged it.

PR #649, opened 2019-06-04
[improvement] New PreferBuiltInConcurrentKeySet suggestion

iamdanfox:
This is not really that important - just a minor followup from https://github.com/palantir/gradle-baseline/pull/633.

It's only a SUGGESTION, so shouldn't break anyone's builds.

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferBuiltInConcurrentKeySet.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferBuiltInConcurrentKeySet.java
@@ -1,0 +1,63 @@
+/*
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.baseline.errorprone;
+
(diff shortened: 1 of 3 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferBuiltInConcurrentKeySet.java
```diff
@@ -0,0 +1,63 @@
...
+import com.sun.source.tree.MethodInvocationTree;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "PreferBuiltInConcurrentKeySet",
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
+        severity = BugPattern.SeverityLevel.SUGGESTION,
+        summary = "Prefer Java's built-in Concurrent Set implementation over Guava's ConcurrentHashSet, as it does "
+                + "roughly the same thing without relying on Guava")
```
carterkozak: nit: I'd remove "roughly", it does the same thing with (theoretically) less overhead since it avoids the wrapper type.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferBuiltInConcurrentKeySet.java
```diff
@@ -0,0 +1,63 @@
...
+public final class PreferBuiltInConcurrentKeySet extends BugChecker implements BugChecker.MethodInvocationTreeMatcher {
+
+    private static final Matcher<ExpressionTree> MATCHER =
+            MethodMatchers.staticMethod()
+                    .onClass("com.google.common.collect.Sets")
+                    .named("newConcurrentHashSet");
```
carterkozak: We should specify the zero-parameter constructor, `newConcurrentHashSet` can take an iterable, which isn't supported by `ConcurrentHashMap.newKeySet()`. Probably worth a test ;-)

iamdanfox: :+1: test added

carterkozak:
Thanks 👍

PR #654, opened 2019-06-17
[improvement] Error-prone check to help prevent logging AuthHeader and BearerToken

dsd987:
## Before this PR
Authentication information contained in AuthHeader and BearerToken could be accidentally logged resulting in a potential security risk.

## After this PR
A new error-prone check will now flag logging of AuthHeader and BearerToken objects via the use of slf4j and Safe/Unsafe args as compile-time errors when enabled.

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreventTokenLogging.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreventTokenLogging.java
@@ -1,0 +1,78 @@
+/*
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.baseline.errorprone;
+
+import com.google.auto.service.AutoService;
+import com.google.errorprone.BugPattern;
+import com.google.errorprone.VisitorState;
+import com.google.errorprone.bugpatterns.BugChecker;
+import com.google.errorprone.matchers.Description;
+import com.google.errorprone.matchers.Matcher;
+import com.google.errorprone.matchers.Matchers;
+import com.google.errorprone.matchers.method.MethodMatchers;
+import com.sun.source.tree.ExpressionTree;
+import com.sun.source.tree.MethodInvocationTree;
+import com.sun.source.tree.Tree;
+import java.util.List;
+import java.util.regex.Pattern;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "PreventTokenLogging",
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
+        severity = BugPattern.SeverityLevel.ERROR,
+        summary = "Authentication token information should never be logged as it poses a security risk. Prevents "
+                + "AuthHeader and BearerToken information from being passed to common logging calls.")
+public class PreventTokenLogging extends BugChecker implements BugChecker.MethodInvocationTreeMatcher {
+
+    private static final Matcher<ExpressionTree> METHOD_MATCHER =
+            Matchers.anyOf(
+                    MethodMatchers.instanceMethod()
+                            .onExactClass("org.slf4j.Logger")
+                            .withNameMatching(Pattern.compile("trace|debug|info|warn|error")),
+                    MethodMatchers.staticMethod()
+                            .onClass("com.palantir.logsafe.SafeArg")
+                            .named("of"),
+                    MethodMatchers.staticMethod()
+                            .onClass("com.palantir.logsafe.UnsafeArg")
+                            .named("of"));
+
+    private static final Matcher<Tree> AUTH_MATCHER =
+            Matchers.anyOf(
+                    Matchers.isSubtypeOf("com.palantir.tokens.auth.AuthHeader"),
+                    Matchers.isSubtypeOf("com.palantir.tokens.auth.BearerToken"));
+
+    @Override
+    public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
+        if (METHOD_MATCHER.matches(tree, state)) {
+            List<? extends ExpressionTree> args = tree.getArguments();
+            if (args.size() < 2) {
+                return Description.NO_MATCH;
+            }
+
+            for (Tree arg : args.subList(1, args.size())) {
+                if (AUTH_MATCHER.matches(arg, state)) {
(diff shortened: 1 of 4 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreventTokenLogging.java
```diff
@@ -0,0 +1,78 @@
...
+        linkType = BugPattern.LinkType.CUSTOM,
+        severity = BugPattern.SeverityLevel.ERROR,
+        summary = "Authentication token information should never be logged as it poses a security risk. Prevents "
+                + "AuthHeader and BearerToken information from being passed to common logging calls.")
+public class PreventTokenLogging extends BugChecker implements BugChecker.MethodInvocationTreeMatcher {
+
+    private static final Matcher<ExpressionTree> METHOD_MATCHER =
+            Matchers.anyOf(
+                    MethodMatchers.instanceMethod()
+                            .onExactClass("org.slf4j.Logger")
```
carterkozak: I think we want to match on types extending `org.slf4j.Logger`, `org.slf4j.ext.XLogger` for instance. `onDescendantOf`?

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreventTokenLogging.java
```diff
@@ -0,0 +1,78 @@
...
+                            .withNameMatching(Pattern.compile("trace|debug|info|warn|error")),
```
carterkozak: I don't think there are _any_ methods on these things where we would want to allow passing credential material, can we remove this line?

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreventTokenLogging.java
```diff
@@ -0,0 +1,78 @@
...
+                    MethodMatchers.staticMethod()
+                            .onClass("com.palantir.logsafe.SafeArg")
```
carterkozak: nit: could use `onClassAny("com.palantir.logsafe.SafeArg", "com.palantir.logsafe.UnsafeArg")`

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreventTokenLogging.java
```diff
@@ -0,0 +1,78 @@
...
+                            .named("of"),
+                    MethodMatchers.staticMethod()
+                            .onClass("com.palantir.logsafe.UnsafeArg")
+                            .named("of"));
+
+    private static final Matcher<Tree> AUTH_MATCHER =
+            Matchers.anyOf(
+                    Matchers.isSubtypeOf("com.palantir.tokens.auth.AuthHeader"),
+                    Matchers.isSubtypeOf("com.palantir.tokens.auth.BearerToken"));
```
carterkozak, line 52: Do we have these on the classpath? We might as well reference the .class directly if so

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreventTokenLogging.java
```diff
@@ -0,0 +1,78 @@
...
+
+    @Override
+    public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
+        if (METHOD_MATCHER.matches(tree, state)) {
+            List<? extends ExpressionTree> args = tree.getArguments();
+            if (args.size() < 2) {
+                return Description.NO_MATCH;
+            }
```
carterkozak: I'd prefer to remove this check for simplicity, we can check all arguments, it should be impossible for the first arg to be a token in many cases.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreventTokenLogging.java
iamdanfox: These both inherit from Arg: https://github.com/palantir/safe-logging/blob/develop/safe-logging/src/main/java/com/palantir/logsafe/Arg.java

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreventTokenLogging.java
```diff
@@ -0,0 +1,78 @@
...
+
+            for (Tree arg : args.subList(1, args.size())) {
+                if (AUTH_MATCHER.matches(arg, state)) {
+                    return buildDescription(arg)
+                            .setMessage("Authentication information is not allowed to be logged.")
+                            .build();
```
iamdanfox, line 61: It would be awesome if we could also implement a suggestedfix that just deletes the relevant argument???

carterkozak: In some cases folks may prefer to extract a jwt, in others a deletion is best. I don't think we can pick the right one consistently enough.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreventTokenLogging.java
carterkozak: To my knowledge (I may very well be mistaken) you can't match a static method on a descendant. We may be able to build a util, not sure if it's worth doing though.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreventTokenLogging.java
dsd987: 👍

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreventTokenLogging.java
dsd987: Doesn't look like it, unfortunately.

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/PreventTokenLoggingTests.java
```diff
@@ -0,0 +1,300 @@
...
+        compilationHelper.addSourceLines(
+                "Test.java",
+                "import org.slf4j.Logger;",
+                "import org.slf4j.LoggerFactory;",
+                "import com.palantir.tokens.auth.AuthHeader;",
+                "import com.palantir.tokens.auth.BearerToken;",
+                "class Test {",
+                "  private static final Logger log = LoggerFactory.getLogger(Test.class);",
+                "  void f(AuthHeader authHeader, BearerToken bearerToken, String message, Object arg1, Object arg2) {",
+                "    " + precondition,
```
carterkozak: This isn’t a precondition. Statement?

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/PreventTokenLoggingTests.java
```diff
@@ -0,0 +1,300 @@
...
+    private void passLogSafe(String precondition) {
+        compilationHelper.addSourceLines(
+                "Test.java",
+                "import com.palantir.logsafe.SafeArg;",
+                "import com.palantir.logsafe.UnsafeArg;",
+                "import com.palantir.tokens.auth.AuthHeader;",
+                "import com.palantir.tokens.auth.BearerToken;",
+                "class Test {",
+                "  void f(AuthHeader authHeader, BearerToken bearerToken, String name, Object value) {",
+                "    " + precondition,
```
carterkozak: Here as well

iamdanfox (approved):

PR #655, opened 2019-06-19
[improvement] Ensure Optional#orElse argument is not method invocation

gatesn:
## Before this PR
Run into `Optional#orElse` code three times now that was intended to be lazy.

## After this PR
Argument to `Optional#orElse` must not include a method invocation. Otherwise a fix is suggested to convert to `Optional#orElseGet(Supplier)`

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/OptionalOrElseConstant.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/OptionalOrElseConstant.java
@@ -23,12 +23,13 @@ import com.google.errorprone.VisitorState;
 import com.google.errorprone.bugpatterns.BugChecker;
 import com.google.errorprone.bugpatterns.BugChecker.MethodInvocationTreeMatcher;
 import com.google.errorprone.fixes.SuggestedFix;
-import com.google.errorprone.matchers.CompileTimeConstantExpressionMatcher;
 import com.google.errorprone.matchers.Description;
 import com.google.errorprone.matchers.Matcher;
+import com.google.errorprone.matchers.Matchers;
 import com.google.errorprone.matchers.method.MethodMatchers;
 import com.sun.source.tree.ExpressionTree;
 import com.sun.source.tree.MethodInvocationTree;
+import com.sun.source.tree.Tree;
 
 @AutoService(BugChecker.class)
 @BugPattern(
@@ -45,8 +46,8 @@ public final class OptionalOrElseConstant extends BugChecker implements MethodIn
             .onExactClass("java.util.Optional")
             .named("orElse");
 
-    private final Matcher<ExpressionTree> compileTimeConstExpressionMatcher =
-            new CompileTimeConstantExpressionMatcher();
+    private final Matcher<Tree> nonMethodExpressionMatcher =
+            Matchers.contains(ExpressionTree.class, MethodMatchers.anyMethod());
 
     @Override
     public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
@@ -56,7 +57,7 @@ public final class OptionalOrElseConstant extends BugChecker implements MethodIn
 
         ExpressionTree orElseArg = tree.getArguments().get(0);
 
-        if (compileTimeConstExpressionMatcher.matches(orElseArg, state)) {
+        if (!nonMethodExpressionMatcher.matches(orElseArg, state)) {
             return Description.NO_MATCH;
         }
 
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/OptionalOrElseConstant.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/OptionalOrElseConstant.java
@@ -1,0 +1,72 @@
+/*
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.baseline.errorprone;
+
+import com.google.auto.service.AutoService;
+import com.google.errorprone.BugPattern;
+import com.google.errorprone.BugPattern.SeverityLevel;
+import com.google.errorprone.VisitorState;
+import com.google.errorprone.bugpatterns.BugChecker;
+import com.google.errorprone.bugpatterns.BugChecker.MethodInvocationTreeMatcher;
+import com.google.errorprone.fixes.SuggestedFix;
+import com.google.errorprone.matchers.CompileTimeConstantExpressionMatcher;
+import com.google.errorprone.matchers.Description;
+import com.google.errorprone.matchers.Matcher;
+import com.google.errorprone.matchers.method.MethodMatchers;
+import com.sun.source.tree.ExpressionTree;
+import com.sun.source.tree.MethodInvocationTree;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "OptionalOrElseConstant",
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
+        severity = SeverityLevel.ERROR,
+        summary = "Allow passing only compile-time constants to Optional#orElse.")
+public final class OptionalOrElseConstant extends BugChecker implements MethodInvocationTreeMatcher {
+
+    private static final long serialVersionUID = 1L;
+
+    private static final Matcher<ExpressionTree> OR_ELSE_METHOD = MethodMatchers.instanceMethod()
+            .onExactClass("java.util.Optional")
+            .named("orElse");
+
+    private final Matcher<ExpressionTree> compileTimeConstExpressionMatcher =
+            new CompileTimeConstantExpressionMatcher();
+
+    @Override
+    public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
+        if (!OR_ELSE_METHOD.matches(tree, state)) {
+            return Description.NO_MATCH;
+        }
+
+        ExpressionTree orElseArg = tree.getArguments().get(0);
+
+        if (compileTimeConstExpressionMatcher.matches(orElseArg, state)) {
+            return Description.NO_MATCH;
+        }
+
+        return buildDescription(tree)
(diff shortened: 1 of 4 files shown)


markelliot:
I think compile time constant is too aggressive. We should just enforce it’s not an expression (a variable further up the function is totally appropriate as an argument to orElse).

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/OptionalOrElseConstant.java
```diff
@@ -0,0 +1,73 @@
...
+import com.sun.source.tree.MethodInvocationTree;
+import com.sun.source.tree.Tree;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "OptionalOrElseConstant",
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
+        severity = SeverityLevel.ERROR,
+        summary = "Allow passing only compile-time constants to Optional#orElse.")
```
ferozco: can you update this as well?

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/OptionalOrElseConstant.java
```diff
@@ -0,0 +1,73 @@
...
+import com.google.errorprone.matchers.Matcher;
+import com.google.errorprone.matchers.Matchers;
+import com.google.errorprone.matchers.method.MethodMatchers;
+import com.sun.source.tree.ExpressionTree;
+import com.sun.source.tree.MethodInvocationTree;
+import com.sun.source.tree.Tree;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "OptionalOrElseConstant",
```
ferozco: I think this is a little stale, perhaps `OptionalOrElseMethodInvocation`, `OptionalOrElseValue`?

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/OptionalOrElseConstant.java
```diff
@@ -0,0 +1,73 @@
...
+
+        if (!nonMethodExpressionMatcher.matches(orElseArg, state)) {
+            return Description.NO_MATCH;
+        }
+
+        return buildDescription(tree)
+                .setMessage("Optional#orElse uses a non-constant expression")
+                .addFix(SuggestedFix.builder()
+                        .postfixWith(tree.getMethodSelect(), "Get")
+                        .prefixWith(orElseArg, "() -> ")
```
ferozco: Could we pass a method reference instead of allocating a new lamba? i.e. `.orElseGet(this::foo)` or `orElseGet(foo)`

gatesn: It's not so easy to detect these things, and if we could, they should probably be done as their own check.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/OptionalOrElseMethodInvocation.java
```diff
@@ -0,0 +1,73 @@
...
+import com.sun.source.tree.MethodInvocationTree;
+import com.sun.source.tree.Tree;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "OptionalOrElseMethodInvocation",
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
+        severity = SeverityLevel.ERROR,
+        summary = "Ensure Optional#orElse argument does not invoke a ny methods.")
```
carterkozak: `s/a ny/any`

It's fairly obvious in this case, but it may be helpful to spell out exactly why orElseGet is better in the summary

carterkozak:
Very excited about this check. Thanks for implementing the suggested fix, they've been invaluable for a certain large internal project!

gatesn:
Addressed all comments, should be good to go

markelliot (commented):
Could we add a test that shows we prevent `foo + bar` as an argument to orElse?

iamdanfox:
Probably worth mentioning this on the README with the other error-prone checks too pls! https://github.com/palantir/gradle-baseline#baseline-error-prone-checks

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/OptionalOrElseMethodInvocationTests.java
```diff
@@ -0,0 +1,89 @@
...
+                        "Test.java",
+                        "import java.util.Optional;",
+                        "class Test {",
+                        "  private static final String compileTimeConstant = \"constant\";",
+                        "  String f() { return \"hello\"; }",
+                        "  void test() {",
+                        "    Optional.of(\"hello\").orElse(\"constant\");",
+                        "    Optional.of(\"hello\").orElse(compileTimeConstant);",
+                        "    String string = f();",
+                        "    Optional.of(\"hello\").orElse(string);",
```
uschi2000, line 91: this is missing test coverage for `Optional.of(foo).orElseGet(() -> bar);`

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/OptionalOrElseMethodInvocationTests.java
```diff
@@ -0,0 +1,89 @@
...
+
+    @Test
+    public void testNonCompileTimeConstantExpression() {
+        compilationHelper
+                .addSourceLines(
+                        "Test.java",
+                        "import java.util.Optional;",
+                        "class Test {",
+                        "  String f() { return \"hello\"; }",
+                        "  // BUG: Diagnostic contains: non-constant expression",
```
uschi2000: what does "non-constant" mean? suspect that's not quite the right word. what you mean here is "method invocation"

gatesn: Well it is, but then as Mark says this doesn't cover operator expressions. Not sure what the right word is?

uschi2000: If I understand the `Matchers.contains(ExpressionTree.class, MethodMatchers.anyMethod())` thing correctly, then it errors iff a method is called anywhere inside the `orElse(...)`. A reasonable description would be, for instance, `Expression passed to Optional#orElse must not invoke a method`, or positively (not sure what the right one is), `Expression passed to Optional#orElse invokes a method`.

uschi2000: (or `contains a method invocation`, if you prefer). To double-check: `foo.orElse(bar + baz)` is allowed, ja?

gatesn: Well, it is currently but should it be allowed? Are operators ever sufficiently expensive for us to care about? If yes, then we need to think again about our matcher. Perhaps compileTimeConstant OR identifier.

carterkozak: String concatenation using operators should be required to execute in a lambda. I don't mind the lambda allocation for math because the optional is already boxing numerical results with similar (avoidable) overhead to creating a lambda.

uschi2000: .. then you'll likely see
```
String bar = boom + baz;
return foo.orElse(bar);  // can't concat in here.
```

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/OptionalOrElseMethodInvocationTests.java
gatesn: Not sure what you mean, you want this check to revert it to `orElse(bar)`?

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/OptionalOrElseMethodInvocationTests.java
uschi2000: I don't really mind the exact version of the check, as long as you make the text reflect the check. (that's what I was commenting on initially, sorry if this was unclear.)

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/OptionalOrElseMethodInvocationTests.java
uschi2000: I don't see a test verifying that `Optional.of(foo).orElseGet(() -> bar)` is an acceptable thing to write

uschi2000: sorry, I mean `Optional.of(foo).orElseGet(() -> f());`

gatesn: Ah, I've done that now

gatesn:
Is this good to go now?

markelliot (approved):
May want to consider in a separate PR the opposite of this — if you’re calling orElseGet when you could be calling orElse it’d be good to warn (IntelliJ has that as a refactor suggestion), but not worth blocking on

iamdanfox:
Didn't update the README 😢

ferozco:
@gatesn I think that this check may be overly aggressive since accessing any field on an immutable is considered a method invocation. In general, I feel like this check has a very large number of false positives so we may want to lower the severity to warn or reconsider the check entirely.

gatesn:
@ferozco can you / @carterkozak enumerate the false positives in case we want to pick this back up? 

There are a few I disagree with, for instance I'm not convinced there's a difference in accessing a property of an immutable as `orElse(foo.bar())` or `orElseGet(foo::bar)`. If its purely style, I'd strongly suggest we prefer the latter.

A viable opposition would be for example `Collections.emptyList()`. However it would be useful to understand all the cases you hit so we can attempt to refine the check.

markelliot:
It might be that we need an automated refactor of method calls into suppliers in order to make this work. Seems reasonable that you'd `orElseGet(Collections::emptyList)`, too.

gatesn:
If that's the case, then I imagine it shouldn't be too hard to implement the logic as either in the check itself, or as part of a [refaster rule](https://errorprone.info/docs/refaster)

gatesn:
So, we can now auto-apply error-prone fixes: https://github.com/palantir/gradle-baseline/pull/660/files

Is this sufficient? Or do we want to block on having the `() -> foo()` -> `this::foo` rewrite

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/OptionalOrElseConstant.java
carterkozak: https://github.com/palantir/gradle-baseline/pull/763

PR #659, opened 2019-06-21
Revert "[improvement] Ensure Optional#orElse argument is not method invocation"

ferozco:
Reverts palantir/gradle-baseline#655.

After testing the check on a few internal projects, @carterkozak and I found it to result in too many false positives to be worthwhile which is shame since I quite like the idea behind the check


carterkozak:
thanks 👍

markelliot:
What are some of the examples of false positives?

ferozco:
You get a false positive if you pass any field from an immutable or conjure-defined object or use any empty collection helper method i.e. `Collections.emptyList()`

gatesn:
@ferozco time to revert this?

PR #661, opened 2019-06-24
[improvement] Refaster rule compilation

gatesn:
## After this PR
Passing `-PrefasterApply` to the `compileJava` tasks will apply any refaster rules discovered in the `refaster` configuration.

Includes some default baseline refaster rules.


baseline-refaster-plugin/src/main/java/com/palantir/baseline/refaster/BaselineRefasterCompiler.java
```diff
@@ -0,0 +1,54 @@
...
+        return getClass().getSimpleName();
+    }
+
+    @Override
+    public void init(JavacTask task, String... args) {
+        Iterator<String> argIterator = Arrays.asList(args).iterator();
+        String path = null;
+        while (argIterator.hasNext()) {
+            if (argIterator.next().equals("--out")) {
+                path = argIterator.next();
```
carterkozak: nit: next without hasNext check, this can throw prior to the checkArgument below without a helpful message. Can we operate directly on the List instead of Iterator and use `List.indexOf("--out")`?

baseline-refaster-plugin/src/main/java/com/palantir/baseline/refaster/BaselineRefasterCompiler.java
```diff
@@ -0,0 +1,54 @@
...
+                break;
+            }
+        }
+        Preconditions.checkArgument(path != null, "No --out specified");
+
+        task.addTaskListener(new BaselineRefasterCompilerAnalyzer(
+                ((BasicJavacTask) task).getContext(), FileSystems.getDefault().getPath(path)));
```
carterkozak: Mind adding a precondition to validate that `task instanceof BasicJavacTask`?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/RefasterCompileTask.java
```diff
@@ -0,0 +1,78 @@
...
+import org.gradle.api.file.ConfigurableFileCollection;
+import org.gradle.api.file.RegularFileProperty;
+import org.gradle.api.tasks.InputFiles;
+import org.gradle.api.tasks.OutputFile;
+import org.gradle.api.tasks.compile.JavaCompile;
+
+public class RefasterCompileTask extends JavaCompile {
+
+    private final ConfigurableFileCollection refasterSources = getProject().files();
+    private final RegularFileProperty refasterRulesFile = newInputFile();
```
iamdanfox: I think using newInputFile() results in deprecation warnings... can we use `project().objects().property(File.class)` instead?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/RefasterCompileTask.java
```diff
@@ -0,0 +1,78 @@
...
+        // Ensure we hit the non-incremental code-path since we override it
+        getOptions().setIncremental(false);
+    }
+
+    @Override
+    protected final void compile() {
+        // Clear out the default error-prone providers
+        getOptions().getCompilerArgumentProviders().clear();
+        getOptions().setCompilerArgs(ImmutableList.of(
+                "-Xplugin:BaselineRefasterCompiler --out " + refasterRulesFile.get().getAsFile().getAbsolutePath()));
```
iamdanfox: I'm surprised a string containing spaces works here... shouldn't it be a list with three elements, `Xplugin:BaselineRefasterCompiler`, `--out` etc?

gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineErrorProneRefasterIntegrationTest.groovy
```diff
@@ -0,0 +1,113 @@
...
+        '''.stripIndent()
+
+        then:
+        BuildResult result = with('compileJava', '-i', '-PrefasterApply').build()
+        result.task(":compileJava").outcome == TaskOutcome.SUCCESS
+        file('src/main/java/test/Test.java').text == '''
+        package test;
+        import java.util.ArrayList;
+        public class Test {
+            boolean empty = new ArrayList<>().isEmpty();
```
iamdanfox, line 58: this is kinda mind blowing 🎉🎉🎉

settings.gradle
```diff
@@ -2,6 +2,8 @@ rootProject.name = "gradle-baseline"
 enableFeaturePreview("STABLE_PUBLISHING")
 
 include "baseline-error-prone"
+include "baseline-refaster-plugin"
```
iamdanfox: I'd like to rename this to `baseline-refaster-javac-plugin` so it's clear this is not a gradlep plugin but an actual _compiler_ plugin!

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
```diff
@@ -31,30 +32,63 @@
...
+                    .create("refasterSources", configuration -> {
+                        configuration.extendsFrom(refasterConfiguration);
+                        configuration.setTransitive(false);
+                    });
+            Configuration refasterCompilerConfiguration = project.getConfigurations()
+                    .create("refasterCompiler", configuration -> configuration.extendsFrom(refasterConfiguration));
+
+            project.getDependencies().add(
+                    REFASTER_CONFIGURATION,
+                    "com.palantir.baseline:baseline-refaster-rules:" + version + ":sources");
```
iamdanfox, line 75: I think it might not actually be necessary to ship this to everyone as the _vast_ majority of builds will not use `-PrefasterApply`, still fine to use it in tests tho

iamdanfox:
This is amazing - v. excited to try it out!

gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/RefasterCompileTask.java
gatesn: Nope, this is what tells it to pass the args to the javac plugin, instead of the compiler itself.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
gatesn: If they don't use refasterApply, then no harm in adding it to the refaster configuration right?

gatesn:
Here's a whole suite of templates we could add into baseline if we wanted to: https://github.com/google/error-prone/tree/master/core/src/test/java/com/google/errorprone/refaster/testdata/template

iamdanfox:
Those are cool, but I don't think we'd actually want to blanket apply them because they seem to be just test data to exercise refaster itself... e.g. we don't necessarily want to change all LinkedList -> ArrayList, as occasionally people have a real justification for it: https://github.com/google/error-prone/blob/master/core/src/test/java/com/google/errorprone/refaster/testdata/template/DiamondTemplate.java

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
iamdanfox: Fair enough - we've stopped using `resolveConfigurations` in our CircleCI templates, so should be fine!

iamdanfox (approved):

PR #666, opened 2019-06-27
[improvement] JUnit 5 boilerplate

iamdanfox:
## Before this PR

We are proliferating boilerplate with PRs like this: https://github.com/palantir/gradle-consistent-versions/pull/160

## After this PR

baseline contains a few more strong opinions, and will give you maximum parallelization.

cc @ferozco 

## Possible downsides

- the parallelization thing is quite a strong opinion, and people will have to opt-out in quite a verbose way if they need serial tests.

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineTesting.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineTesting.java
@@ -20,6 +20,7 @@ import java.util.concurrent.atomic.AtomicBoolean;
 import org.gradle.api.Plugin;
 import org.gradle.api.Project;
 import org.gradle.api.artifacts.Configuration;
+import org.gradle.api.plugins.JavaPlugin;
 import org.gradle.api.tasks.testing.Test;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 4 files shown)


CRogers (approved):

iamdanfox:
cc @robert3005 hopefully this means you can delete some boilerplate 🎉🎉

robert3005:
nice, FWIW even if you enable parallelization you can still force certain tests to be serial

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineTesting.java
```diff
@@ -18,14 +18,53 @@
...
     public void apply(Project project) {
         project.getTasks().withType(Test.class).all(task -> {
             task.jvmArgs("-XX:+HeapDumpOnOutOfMemoryError", "-XX:+CrashOnOutOfMemoryError");
         });
+
+        project.getPlugins().withType(JavaPlugin.class, p -> {
+
+            // afterEvaluate necessary because the junit-jupiter dep might be added further down the build.gradle
+            project.afterEvaluate(unused -> {
+                project.getConfigurations()
```
robert3005, line 40: Ideally this would match transitive dependencies and not just in project ones

PR #669, opened 2019-06-28
[fix] fix potential NPE in com.palantir.baseline-testing

ferozco:
## Before this PR
We would fail to configure some project with dependencies that did not have a group.

## After this PR
Correctly configure test dependencies

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineTesting.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineTesting.java
@@ -16,6 +16,7 @@
 
 package com.palantir.baseline.plugins;
 
+import java.util.Objects;
 import org.gradle.api.Plugin;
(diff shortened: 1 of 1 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineTesting.java
```diff
@@ -40,7 +40,7 @@ public void apply(Project project) {
                 project.getConfigurations()
                         .getByName(JavaPlugin.TEST_RUNTIME_CLASSPATH_CONFIGURATION_NAME)
                         .getAllDependencies()
-                        .matching(dep -> dep.getGroup().equals("org.junit.jupiter")
+                        .matching(dep -> dep.getGroup() != null && dep.getGroup().equals("org.junit.jupiter")
```
iamdanfox: Objects.equals?

iamdanfox (approved):

PR #673, opened 2019-07-01
Ignore DesignForExtension for ParameterizedTest

ashrayjain:
## Before this PR

Tests with the `@ParameterizedTest` JUnit 5 annotation would fail DesignForExtension checkstyle
## After this PR

No more failures 🎉


dansanduleac:
👍

PR #674, opened 2019-07-01
[fix] Prevent null from flagging as authentication information in PreventTokenLogging check

dsd987:
## Before this PR
Passing null into SLF4J and Safe/Unsafe Arg functions would incorrectly trigger false positives in the PreventTokenLogging error-prone check.

## After this PR
The PreventTokenLogging error-prone check will now correctly handle null use in SLF4J and Safe/Unsafe Arg functions.

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreventTokenLogging.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreventTokenLogging.java
@@ -55,7 +55,7 @@ public final class PreventTokenLogging extends BugChecker implements BugChecker.
     public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
         if (METHOD_MATCHER.matches(tree, state)) {
             for (Tree arg : tree.getArguments()) {
-                if (AUTH_MATCHER.matches(arg, state)) {
+                if (AUTH_MATCHER.matches(arg, state) && arg.getKind() != Tree.Kind.NULL_LITERAL) {
                     return buildDescription(arg)
                             .setMessage("Authentication information is not allowed to be logged.")
                             .build();
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreventTokenLogging.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreventTokenLogging.java
@@ -46,16 +46,19 @@ public final class PreventTokenLogging extends BugChecker implements BugChecker.
                             .onClassAny("com.palantir.logsafe.SafeArg", "com.palantir.logsafe.UnsafeArg")
                             .named("of"));
 
-    private static final Matcher<Tree> AUTH_MATCHER =
-            Matchers.anyOf(
-                    Matchers.isSubtypeOf("com.palantir.tokens.auth.AuthHeader"),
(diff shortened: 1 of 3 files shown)


baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/PreventTokenLoggingTests.java
```diff
@@ -144,6 +144,16 @@ public void testSlf4jTraceNoArgs() {
         passSlf4j("log.trace(message);");
     }
 
+    @Test
+    public void testSlf4jTraceNullMessageNoArgs() {
+        passSlf4j("log.trace(null);");
```
uschi2000, line 149: what's the use-case for this?

carterkozak: I don't think there's a use case for this, perhaps it should fail, but I don't think the failure should be caused by `PreventTokenLogging`.

dsd987: Yeah, no use case that I could think of, just a technically allowable case that I didn't think made sense to be enforced by this check.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreventTokenLogging.java
```diff
@@ -55,7 +55,7 @@
     public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
         if (METHOD_MATCHER.matches(tree, state)) {
             for (Tree arg : tree.getArguments()) {
-                if (AUTH_MATCHER.matches(arg, state)) {
+                if (AUTH_MATCHER.matches(arg, state) && arg.getKind() != Tree.Kind.NULL_LITERAL) {
```
ellisjoe: this should probably be added as part of the `AUTH_MATCHER` directly

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreventTokenLogging.java
```diff
@@ -46,16 +46,19 @@
...
-            Matchers.anyOf(
-                    Matchers.isSubtypeOf("com.palantir.tokens.auth.AuthHeader"),
-                    Matchers.isSubtypeOf("com.palantir.tokens.auth.BearerToken"));
+    private static final Matcher<ExpressionTree> AUTH_MATCHER =
+            Matchers.allOf(
+                    Matchers.anyOf(
+                            Matchers.isSubtypeOf("com.palantir.tokens.auth.AuthHeader"),
+                            Matchers.isSubtypeOf("com.palantir.tokens.auth.BearerToken")),
+                    Matchers.not(
+                            Matchers.kindIs(Tree.Kind.NULL_LITERAL)));
```
ellisjoe, line 55: there's also a `Matchers.isNonNull()`

dsd987: From what I remember about Matchers.isNonNull(), is that it only returns true when the expression is provably non-null via a null analysis to make its determination and could sometimes return false when we need true since I believe we are only concerned with the use of literal null.

carterkozak (approved):
thanks

PR #677, opened 2019-07-02
[fix] Prefer short names for Preconditions over fully-qualified names

gatesn:
## Before this PR
The suggested fixes for safe logging Preconditions would always render the fully-qualified class name.

## After this PR
The suggested fixes render short names, removing the Guava import, and adding the safe-logging import.

## Potential Downsides
We don't remove the commons Validate import since there may be other instances of this class being used. However, this is the same behaviour as before so..


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferSafeLoggingPreconditions.java
```diff
@@ -94,10 +93,12 @@ public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState
...
 
-        SuggestedFix.Builder fix = SuggestedFix.builder();
-        String logSafeQualifiedClassName = SuggestedFixes.qualifyType(state, fix, "com.palantir.logsafe.Preconditions");
+        SuggestedFix.Builder fix = SuggestedFix.builder()
+                .addImport("com.palantir.logsafe.Preconditions")
+                .removeImport("com.google.common.base.Preconditions");
+
         String logSafeMethodName = getLogSafeMethodName(ASTHelpers.getSymbol(tree));
-        String replacement = String.format("%s.%s", logSafeQualifiedClassName, logSafeMethodName);
+        String replacement = String.format("Preconditions.%s", logSafeMethodName);
```
carterkozak, line 101: This breaks compilation for files that use some replaceable guava preconditions, and others with interpolated parameters, which we cannot programmatically replace.

carterkozak (changes requested):

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferSafeLoggingPreconditions.java
gatesn: Not sure what you mean by interpolated parameters?

carterkozak:
```java
com.google.common.base.Preconditions;

Preconditions.checkState(value != 5, "Value %s was not 5", value);
```

pkoenig10:
As part of this change can we also update this to fix static imports as well? If I'm using static imports with guava preconditions, I probably also want to use static imports with safe-logging preconditions.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferSafeLoggingPreconditions.java
```diff
@@ -94,10 +93,12 @@ public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState
             return Description.NO_MATCH;
         }
 
-        SuggestedFix.Builder fix = SuggestedFix.builder();
-        String logSafeQualifiedClassName = SuggestedFixes.qualifyType(state, fix, "com.palantir.logsafe.Preconditions");
+        SuggestedFix.Builder fix = SuggestedFix.builder()
+                .addImport("com.palantir.logsafe.Preconditions")
+                .removeImport("com.google.common.base.Preconditions");
```
iamdanfox, line 98: For our repos `removeImport` should be unnecessary because we run `spotlessApply` inside excavator which actually gets rid of unused imports automatically.

dansanduleac:
@gatesn  wanna fix up this PR, or close it if it doesn't make sense anymore?

felixdesouza:
😢 Will some modified version of this make a comeback? Or might this make more sense as a refaster rule? 🤷‍♂ 

Reason being, most interpolated code will break out parameters over many lines, so changing to fully qualified names won't go over 120 character line limit. Whereas the constant message with no params, is usually a line, and fully qualifying takes it over the threshold.

Even if we did formatting, it would then make simple one-liners kinda horrible, and still requires a manual fix up. I would also guess that there are more one liner preconditions *without* args than there are *with* args.

PR #679, opened 2019-07-02
add refaster rule to migrate away from orElse(supplier.get())

ferozco:
## Before this PR
In support of bringing back #655.

## After this PR
==COMMIT_MSG==
Add refaster rule to migrate away from optional.orElse(supplier.get())
==COMMIT_MSG==

## Possible downsides?
Unfortunately, this rule only seems to match cases where the parameter is explicitly a supplier and ignores cases such as a field access on an immutable.

@gatesn for SA

### DIFF
--- a/baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/OptionalOrElseSupplier.java
+++ b/baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/OptionalOrElseSupplier.java
@@ -1,0 +1,34 @@
+/*
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


gatesn (commented):
How often does this occur in practice?

carterkozak:
Curious if we can write tests for these? Doesn't look like we have any yet.

ferozco:
@gatesn I think that these occur relatively rarely, but in those cases this seems desirable. The original motivation behind this work was to find a mechanism to cleanly correct the issues identified by #655. I wasn't able to find a solution, so this PR was opened to at least capture some value from the time spent investigating.

gatesn:
There are integration tests, we might want to build some test utils that don’t run Gradle though, particularly for libraries writing their own.

https://github.com/palantir/gradle-baseline/blob/develop/gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineErrorProneRefasterIntegrationTest.groovy

dansanduleac (approved):

PR #680, opened 2019-07-02
Revert "Revert "[improvement] Ensure Optional#orElse argument is not method invocation""

ferozco:
Reverts palantir/gradle-baseline#659

Now that excavator automatically applies error prone fixes when it updates baseline I think that re-introducing this rule is safe.

@carterkozak @gatesn for SA

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/OptionalOrElseMethodInvocation.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/OptionalOrElseMethodInvocation.java
@@ -1,0 +1,73 @@
+/*
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


carterkozak:
Should we include a refaster rule to rewrite `() -> methodInvocation()` into `this::methodInvocation`?

ferozco:
@carterkozak I had a similar idea on the original PR. According to @gatesn it wasn't possible to express this as an error prone fix, and after playing around with refaster I wasn't able to find a soluton there  either. I've opened #679 to handle some of the more basic cases

dansanduleac:
@carterkozak rewriting lambdas as method references sometimes breaks at runtime, so not sure if it's safe to make this conversion globally.

carterkozak:
I imagine the compiler has enough data to do it safely where possible, but I don't know how.  I'm fine with this.

dansanduleac (approved):

pkoenig10:
Can we please add exceptions for methods that return known constant values? I frequently use things like `.orElse(ImmutableSet.of())` and I would think `.orElseGet(ImmutableSet::of)` is more expensive.

PR #687, opened 2019-07-12
Disallow dangling parenthesis

gatesn:
## Before this PR
Generally our style has always been to disallow dangling parenthesis, but nothing applies this rule.

The aim of this PR, and any style PR, is to increase consistency of coding style and reduce time spent in CR commenting with stylistic nit picks.

```java
Preconditions.checkArgument(
    myExpression,
    "My expression isn't as true as it should be"
);
```

## After this PR
```java
Preconditions.checkArgument(
    myExpression,
    "My expression isn't as true as it should be");
```

### DIFF
--- a/changelog/@unreleased/pr-687.v2.yml
+++ b/changelog/@unreleased/pr-687.v2.yml
@@ -1,25 +1,6 @@
 type: improvement
 improvement:
   description: |-
-    Disallow dangling parenthesis
-
-    ## Before this PR
-    Generally our style has always been to disallow dangling parenthesis, but nothing applies this rule.
-
-    The aim of this PR, and any style PR, is to increase consistency of coding style and reduce time spent in CR commenting with stylistic nit picks.
-
-    ```java
-    Preconditions.checkArgument(
-        myExpression,
-        "My expression isn't as true as it should be"
-    );
-    ```
-
-    ## After this PR
-    ```java
-    Preconditions.checkArgument(
-        myExpression,
-        "My expression isn't as true as it should be");
-    ```
+    Spotless check for disallowing dangling parenthesis.
   links:
   - https://github.com/palantir/gradle-baseline/pull/687
--- a/changelog/@unreleased/pr-687.v2.yml
+++ b/changelog/@unreleased/pr-687.v2.yml
@@ -1,0 +1,25 @@
+type: improvement
+improvement:
+  description: |-
+    Disallow dangling parenthesis
+
+    ## Before this PR
+    Generally our style has always been to disallow dangling parenthesis, but nothing applies this rule.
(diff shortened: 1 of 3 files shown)


CRogers:
I'm in favour, provided it will fix the parens automatically in the excavator.

gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineFormatIntegrationTest.groovy
```diff
@@ -48,6 +50,9 @@ class BaselineFormatIntegrationTest extends AbstractPluginTest {
     
     
         int x = 1;
+        System.out.println(
+            "Hello"
+        );
```
CRogers, line 58: Maybe a test as well to ensure that

```
foo(() -> {
    //...
})
```
is not counted as a dangling paren?

changelog/@unreleased/pr-687.v2.yml
```diff
@@ -0,0 +1,25 @@
+type: improvement
+improvement:
+  description: |-
+    Disallow dangling parenthesis
```
CRogers: This changelog is probably too long

gatesn:
Fixed all comments, should be good to go

ferozco:
@gatesn could you add the test case callum asked for?

ferozco (approved):

a-k-g:
This check just rolled out to our repo. Was the intention here to also enforce the following change?

From:

        Identifier transactionIdentifier =
                catalogService.startTransaction(
                        authHeader,
                        param,
                        StartTransactionRequest.builder()
                                .anotherParam(anotherParam)
                                .build()
                ).getIdentifier();

To:

        Identifier transactionIdentifier =
                service.runTransaction(
                        authHeader,
                        param,
                        StartTransactionRequest.builder()
                                .anotherParam(anotherParam)
                                .build()).getIdentifier();

To me first one seems easier to read.

gatesn:
Whenever methods chain in this sort of fluent style, I'd recommend keeping the chained method calls aligned with each other for the best readability, e.g.

```java
Identifier transactionIdentifier = service
        .runTransaction(
                authHeader,
                param,
                StartTransactionRequest.builder()
                        .anotherParam(anotherParam)
                        .build())
        .getIdentifier();
```

bkrieger:
@gatesn would be awesome if we could enforce that as well!

JacekLach:
You can't do it if the first one is a function call though. Best you get:

```
result = foo(
      arg1,
      arg2,
      bunch,
      of,
      args)
      .methodThatCompletelyBlendsInWithArgs()
```

PR #690, opened 2019-07-12
Autofix orElseGet

gatesn:
## Before this PR
OptionalOrElseMethodInvocation is not auto-fixed by `-PerrorProneApply`.

## After this PR
It does.

Will follow-up with a PR to auto-detect all baseline checks.


ferozco:
👍

PR #691, opened 2019-07-12
Allow collection constants in orElse

pkoenig10:
## Before this PR
Using known constant collection values in `orElse` is forbidden. The current check requires changing this to use `orElseGet` which is less performant and a bit more confusing.

## After this PR
Using known constant collection values is allowed in `orElse`.

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/OptionalOrElseMethodInvocation.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/OptionalOrElseMethodInvocation.java
@@ -30,6 +30,7 @@ import com.google.errorprone.matchers.method.MethodMatchers;
 import com.sun.source.tree.ExpressionTree;
 import com.sun.source.tree.MethodInvocationTree;
 import com.sun.source.tree.Tree;
+import java.util.regex.Pattern;
 
 @AutoService(BugChecker.class)
 @BugPattern(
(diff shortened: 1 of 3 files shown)


ferozco:
I think that we want to avoid special casing which method calls are allowed and which ones aren't. Totally agree that having to manually migrate from  `orElse(ImmutableList.of())` to `orElseGet(ImmutableList::of)` is cubersome and happy to spend some time to try and make it more automatic. However, I do think that this rule prevents serious bugs and is worth any potential perf impact

pkoenig10:
What's the motivation for avoiding special casing here? This whitelist doesn't affect the ability of this rule to prevent bugs. 

The cases on this whitelist are known to be acceptable and probably should be preferred because we can avoid instantiating a supplier. I don't think it's reasonable to be required to take an unnecessary perf hit when it is easily avoidable.

gatesn:
@ferozco I don't know how possible it is to make this more automatic. Errorprone doesn't give you access to the AST of classes that aren't in your source set, so it's not quite the right tool for figuring out if internally a method returns a compile-time constant.

ellisjoe:
@pkoenig10 I think the general argument against special casing is that it leads to a system that's harder to describe. In this case we'd be going from saying "If you have a compile time constant use `orElse`, otherwise use `orElseGet`" to saying that plus "unless you're calling one of these 20 methods, and maybe we'll add some more methods in the future. And we're pretty sure we won't take any of those away."

Also, I don't think there's a perf hit when using `orElseGet(ImmutableSet::of)` rather than `orElseGet(ImmutableSet.of())`. The benchmarks I ran seems to show they end up being equivalent:

Code:
```
@BenchmarkMode(Mode.AverageTime)
@OutputTimeUnit(TimeUnit.NANOSECONDS)
@Measurement(iterations = 100, time = 100, timeUnit = TimeUnit.MILLISECONDS)
@Warmup(iterations = 5, time = 1)
@Fork(value = 1)
public class OptionalBenchmark {
    @Benchmark
    public final Set<String> methodCall() {
        return Optional.<Set<String>>empty().orElse(ImmutableSet.of());
    }

    @Benchmark
    public final Set<String> methodReference() {
        return Optional.<Set<String>>empty().orElseGet(ImmutableSet::of);
    }
}
```

Results:
```
Benchmark                          Mode  Cnt  Score   Error  Units
OptionalBenchmark.methodCall       avgt  100  2.365 ± 0.055  ns/op
OptionalBenchmark.methodReference  avgt  100  2.268 ± 0.011  ns/op
```

pkoenig10:
Alright, happy to close this then.

PR #694, opened 2019-07-16
Don't cache test tasks

dansanduleac:
## Before this PR

Enabling the gradle build cache could cause test tasks (particularly, integration and ete tests) to appear up to date and not run again, even though details about the other services being spun up in docker might have changed (such as their versions or configuration).

## After this PR
==COMMIT_MSG==
Don't cache test tasks in the build cache

This is to remediate the situation where the inputs of test tasks aren't properly configured. Specifically, the tasks might be relying on external files (such as `docker-compose.yml`) without gradle being aware of this, and so when only those files change, gradle doesn't know that the test tasks should be re-run and considers them up-to-date.
==COMMIT_MSG==

## Possible downsides?
All test tasks (`test`, `integrationTest` etc) now run unconditionally in CI.

### DIFF
--- a/changelog/@unreleased/pr-694.v2.yml
+++ b/changelog/@unreleased/pr-694.v2.yml
@@ -1,0 +1,7 @@
+type: improvement
+improvement:
+  description: |
+    Don't cache test tasks in the build cache by default.
+    It's possible to restore caching by adding `com.palantir.baseline.restore-test-cache = true` to your `gradle.properties`.
+  links:
+  - https://github.com/palantir/gradle-baseline/pull/694
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineTesting.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineTesting.java
@@ -30,8 +30,11 @@ public final class BaselineTesting implements Plugin<Project> {
 
     @Override
     public void apply(Project project) {
-        project.getTasks().withType(Test.class).all(task -> {
+        project.getTasks().withType(Test.class).configureEach(task -> {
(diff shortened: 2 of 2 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineTesting.java
```diff
@@ -47,6 +47,9 @@ public void apply(Project project) {
                         .findAny()
                         .ifPresent(ignored -> enableJUnit5ForAllTestTasks(project));
             });
+
+            // Never cache test tasks, until we work out the correct inputs for ETE / integration tests
+            project.getTasks().withType(Test.class).configureEach(test -> test.getOutputs().cacheIf(task -> false));
```
ferozco: There were some requests internally to allow users to opt out of this behaviour. We could expose a property that allows people to toggle ti

dansanduleac: `com.palantir.baseline.restore-test-cache = true` ?

changelog/@unreleased/pr-694.v2.yml
```diff
@@ -0,0 +1,7 @@
+type: improvement
+improvement:
+  description: |
+    Don't cache test tasks in the build cache by default.
+    It's possible to restore caching by adding `com.palantir.baseline.restore-test-cache = true` to your `gradle.properties`.
```
ferozco, line 5: can you update the readme too

ferozco:
👍

changelog/@unreleased/pr-694.v2.yml
dansanduleac: sorry, didn't see in time :)

PR #701, opened 2019-07-18
Correctly configure Junit5 for non-default test sets

ferozco:
## Before this PR
We would only configure tests to use junitPlatform if they had a `junit-jupiter` dependency to their test runtime classpath. This was confusing since projects which only had integration/ETE tests would have to add an unecessary dependency

## After this PR
==COMMIT_MSG==
Determine whether to use junitPlatform on a per source set basis
==COMMIT_MSG==

## Possible downsides?
N/A

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineTesting.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineTesting.java
@@ -19,7 +19,11 @@ package com.palantir.baseline.plugins;
 import java.util.Objects;
 import org.gradle.api.Plugin;
 import org.gradle.api.Project;
+import org.gradle.api.artifacts.Dependency;
 import org.gradle.api.plugins.JavaPlugin;
+import org.gradle.api.plugins.JavaPluginConvention;
+import org.gradle.api.specs.Spec;
+import org.gradle.api.tasks.SourceSet;
 import org.gradle.api.tasks.testing.Test;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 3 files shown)


gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineTestingIntegrationTest.groovy
```diff
@@ -79,4 +78,30 @@ class BaselineTestingIntegrationTest extends AbstractPluginTest {
...
+        '''.stripIndent()
+        buildFile << standardBuildFile
+        buildFile << '''
+
+        testSets {
+            integrationTest
+        }
+        
+        dependencies {
+            integrationTestImplementation "org.junit.jupiter:junit-jupiter:5.4.2"
```
iamdanfox, line 97: I just double checked and this should keep the `junit-jupiter-engine` as runtime only. https://mvnrepository.com/artifact/org.junit.jupiter/junit-jupiter/5.5.1

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineTesting.java
```diff
@@ -39,38 +43,60 @@ public void apply(Project project) {
...
+                            if (testTask == null) {
+                                // Fall back to the source set name, since that is what gradle-testsets-plugin does
+                                testTask = (Test) proj.getTasks().findByName(ss.getName());
+                                if (testTask == null) {
+                                    log.warn(
+                                            "Detected 'org:junit.jupiter:junit-jupiter', but unable to find test task");
+                                    return;
+                                }
+                            }
+                            enableJunit5ForTestTask(testTask);
```
iamdanfox, line 67: Wanna move the `log.info` line from 54 to just above here, then we can also include which task name we're enabling it for?

iamdanfox (approved):
This looks great, I was super concerned with our old approach that we might get vacuously passing integration tests, but I think this will now more precisely switch on `useJUnitPlatform` only when it's actually necessary.

PR #702, opened 2019-07-18
Separate positive test cases

pkoenig10:
## Before this PR
Using the same `CompilationTestHelper` to test a positive test case followed by a false positive test case (when we expect an error that isn't emitted) may lead to the false positive test case incorrectly succeeding.

Our tests are incorrect in `OptionalOrElseMethodInvocationTests#test`. It appears we verify that string concatenation results in a error, but an error is not actually emitted (because string concatenation is not a method invocation). Instead the error from the previous positive test case is satisfying the expected error.

## After this PR
All positive test cases are run in separate test methods.

I've opened a PR to fix the issue in error-prone in https://github.com/google/error-prone/pull/1313.

I've also updated the `OptionalOrElseMethodInvocation` to check for constructor invocations (which was my original intent before I discovered this bug).

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/OptionalOrElseMethodInvocation.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/OptionalOrElseMethodInvocation.java
@@ -29,7 +29,6 @@ import com.google.errorprone.matchers.Matchers;
 import com.google.errorprone.matchers.method.MethodMatchers;
 import com.sun.source.tree.ExpressionTree;
 import com.sun.source.tree.MethodInvocationTree;
-import com.sun.source.tree.Tree;
 
 @AutoService(BugChecker.class)
 @BugPattern(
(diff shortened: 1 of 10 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/OptionalOrElseMethodInvocation.java
```diff
@@ -46,8 +45,13 @@
...
             .named("orElse");
 
-    private static final Matcher<Tree> METHOD_INVOCATIONS =
-            Matchers.contains(ExpressionTree.class, MethodMatchers.anyMethod());
+    private static final Matcher<ExpressionTree> METHOD_OR_CONSTRUCTOR = Matchers.anyOf(
+            MethodMatchers.anyMethod(),
+            MethodMatchers.constructor());
+
+    private static final Matcher<ExpressionTree> METHOD_INVOCATIONS = Matchers.anyOf(
+            METHOD_OR_CONSTRUCTOR,
```
pkoenig10, line 53: This is required because while the symbol for a method invocation is a `MethodSymbol`, the symbol for a constructor invocation is just a `TypeSymbol`. Thus `Matchers.contains(...)` won't detect a expression that is just a constructor invocation.

iamdanfox (approved):

PR #704, opened 2019-07-18
Promote safe logging checks from SUGGESTION -> WARNING

pkoenig10:
## Before this PR
The severity of `PreferSafeLoggableExceptions` and `PreferSafeLoggingPreconditions` is `SUGGESTION`.

## After this PR
The severity of `PreferSafeLoggableExceptions` and `PreferSafeLoggingPreconditions` is `WARNING`.

Reasons:
  1. It seems like we wanted to change these at some point in the future:
    - https://github.com/palantir/gradle-baseline/pull/440#issue-225172777
    - https://github.com/palantir/gradle-baseline/pull/517#discussion_r249577418
  2. We provide a suggested fix for `PreferSafeLoggingPreconditions`.
  3. `WARNING` seems like a better _default_ severity. Downgrading the severity for either check is as easy as adding an [error-prone option](http://errorprone.info/docs/flags) like `-Xep:PreferSafeLoggableExceptions:SUGGESTION`.


iamdanfox (approved):
LGTM, using `-PerrorProneApply` should give people the instant fix!

PR #709, opened 2019-07-25
Allow method references in Optional orElse

pkoenig10:
## Before this PR
Using a method reference in an `Optional.orElse(...)` call resulted in a false positive.

## After this PR
Using a method reference in an `Optional.orElse(...)` call is allowed.

The `MethodMatchers` matchers only look at symbols and check whether a given symbol refers to a method. This means they are unable to distinguish between a method invocation and a method reference.

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/OptionalOrElseMethodInvocation.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/OptionalOrElseMethodInvocation.java
@@ -45,13 +45,9 @@ public final class OptionalOrElseMethodInvocation extends BugChecker implements 
             .onExactClass("java.util.Optional")
             .named("orElse");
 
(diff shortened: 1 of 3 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/OptionalOrElseMethodInvocation.java
```diff
@@ -74,4 +70,21 @@ public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState
                 .build();
     }
 
+    private enum MethodInvocationMatcher implements Matcher<ExpressionTree> {
+
+        INSTANCE;
+
+        @Override
+        public boolean matches(ExpressionTree tree, VisitorState state) {
+            switch (tree.getKind()) {
```
ferozco: perhaps an if statement instead of a switch?

ferozco:
👍

dxiao (approved):
works for me! Thanks

PR #711, opened 2019-07-29
Allow static import of spark sql functions

rahij:
## Before this PR

If we static import spark sql functions such as `lit`, `sum` etc, checkstyle fails even though this is the recommended way to use them for code like this:

```
df.agg(sum("col1")).withColumn("idx", lit("5")))
```

## After this PR
==COMMIT_MSG==
Allows static importing spark sql functions
==COMMIT_MSG==

### DIFF
--- a/gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
+++ b/gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
@@ -88,7 +88,8 @@
                 java.util.stream.Collectors.*,
                 com.palantir.logsafe.Preconditions.*,
                 com.google.common.base.Preconditions.*,
-                org.apache.commons.lang3.Validate.*"/>
+                org.apache.commons.lang3.Validate.*",
+                "org.apache.spark.sql.functions.*"/>
         </module>
         <module name="ClassTypeParameterName"> <!-- Java Style Guide: Type variable names -->
             <property name="format" value="(^[A-Z][0-9]?)$|([A-Z][a-zA-Z0-9]*[T]$)"/>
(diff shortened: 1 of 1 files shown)


iamdanfox:
The alternative is this right?

```java
df.agg(Functions.sum("col1")).withColumn("idx", Functions.lit("5")))
```

rahij:
I believe you need to do `org.apache.spark.sql.functions.sum` since it comes from scala

rahij:
@iamdanfox friendly ping. I can add a changelog if you agree with the PR.

iamdanfox (approved):
Eh looks OK to me.  Could you try it out and make sure this is sufficient to actually pass? Are there any other spark things you'd need to statically import?

rahij:
Yep, I have tried it out on an internal repo. I believe it's just the functions "class"

gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
```diff
@@ -88,7 +88,8 @@
                 java.util.stream.Collectors.*,
                 com.palantir.logsafe.Preconditions.*,
                 com.google.common.base.Preconditions.*,
-                org.apache.commons.lang3.Validate.*"/>
+                org.apache.commons.lang3.Validate.*",
+                "org.apache.spark.sql.functions.*"/>
```
carterkozak: This is not correctly formatted.

All the other exclusions are core utilities we use across nearly every product, spark doesn't fit that pattern, and I'm not sure we should open the door for more static imports.

iamdanfox: Oh good spot - the quotes are wrong!

rahij: oops, fixed. serves me right for copy pasting.

iamdanfox:
@rahij have you considered just using the `custom-suppressions.xml` file to disable this on just your repo?

rahij:
I can, it's just that it only lets me suppress the AvoidStaticImport check altogether instead of the specific spark package. If you're opposed to this PR, that's probably what I'll end up doing.

PR #712, opened 2019-07-30
Do not overwrite junit5 test config

ferozco:
## Before this PR
Baseline would overwrite any options that a user provided to the junit5 test framework

## After this PR
==COMMIT_MSG==
Do not overwrite user provided test configure when using junit5
==COMMIT_MSG==

## Possible downsides?
N/A


dansanduleac:
👍

PR #714, opened 2019-07-31
add check to ensure junit5 tests are not used with junit4 Rule/ClassRule

ferozco:
## Before this PR
Users could use Junit4 `Rule` and `ClassRule` annotations within junit5 tests and have the rules silently not execute

## After this PR
==COMMIT_MSG==
Add errorprone check to ensure junit5 tests are not used with junit4 Rule/ClassRule
==COMMIT_MSG==

## Possible downsides?
N/A

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/Junit5RuleUsage.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/Junit5RuleUsage.java
@@ -1,0 +1,70 @@
+/*
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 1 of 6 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/Junit5RuleUsage.java
```diff
@@ -0,0 +1,70 @@
...
+import com.google.errorprone.matchers.Matcher;
+import com.google.errorprone.matchers.Matchers;
+import com.google.errorprone.util.ASTHelpers;
+import com.sun.source.tree.ClassTree;
+import com.sun.source.tree.VariableTree;
+import com.sun.tools.javac.code.Symbol;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "Junit5RuleUsage",
```
iamdanfox: JUnit5RuleUsage

iamdanfox (approved):
Maybe mention in the README too?

PR #716, opened 2019-07-31
Wrap long parameterized types where necessary

gatesn:
## Before this PR
Long type references aren't split and can cause lines to overflow 120 characters.

## After this PR
Type references are "wrapped when necessary"


ferozco:
@gatesn could you slap a changelog on this and merge in develop to get the `markdown` fixed?

iamdanfox (approved):

PR #718, opened 2019-07-31
Remove custom idea JDK configuration

pkoenig10:
## Before this PR
- The `baseline-idea` plugin explicitly sets the JDK version for each module instead of inheriting the JDK version from the idea project.
  - This makes it cumbersome to change the JDK version for an entire idea project because it requires changing it for each module individually.
  - It seems a bit strange to allow different modules in the same project to use different JDK versions.
  - This is broken if the user does not have all the necessary JDK versions installed. Users should only need a single JDK version, as long as it is compatible with the desired language levels.
- The `baseline-idea` plugin configures the project language level when applying the plugin to each module (even if modules have different language levels).

## After this PR
- The `baseline-idea` plugin configures modules to inherit the JDK version from the idea project (this is the [default behavior of the idea plugin](https://docs.gradle.org/current/dsl/org.gradle.plugins.ide.idea.model.IdeaModule.html#org.gradle.plugins.ide.idea.model.IdeaModule:jdkName)).
    - This means that all modules in a project will now use the same JDK version. Note that this is different than the language level, which determines what features are allowed in a module.
- The `baseline-idea` plugin no longer sets the project language level to the source compatibility. In fact, the [documentation specifically recommends](https://docs.gradle.org/current/dsl/org.gradle.plugins.ide.idea.model.IdeaProject.html#org.gradle.plugins.ide.idea.model.IdeaProject:languageLevel) _not_ setting this value.

    > Generally, it isn't recommended to change this value. Instead, you are encouraged to set sourceCompatibility and targetCompatibility for your Gradle projects which allows you to have full control over language levels in Gradle projects, and means that Gradle and IDEA will use the same settings when compiling.
    >
    > When not explicitly set, this is calculated as the maximum language level for the Idea modules of this Idea project.

<br>

As always, these options can be overridden by modifying the idea plugin config (for example, if a project truly needs to use different JDK versions in different modules). But I think this change results in a better default idea configuration.

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
@@ -61,7 +61,6 @@ class BaselineIdea extends AbstractBaselinePlugin {
             }
 
             // Configure Idea module
-            addJdkVersion(ideaModuleModel)
             markResourcesDirs(ideaModuleModel)
             moveProjectReferencesToEnd(ideaModuleModel)
         }
@@ -226,29 +225,6 @@ class BaselineIdea extends AbstractBaselinePlugin {
         }
 
         return base.appendNode(name, attributes + defaults)
-    }
-
-    /**
-     * Configures JDK and Java language level of the given IdeaModel according to the sourceCompatibility property.
-     */
-    private void addJdkVersion(IdeaModel ideaModel) {
-        def compileJavaTask = (JavaCompile) project.tasks.findByName('compileJava')
-        if (compileJavaTask) {
-            def javaVersion = compileJavaTask.sourceCompatibility
-            def jdkVersion = 'JDK_' + javaVersion.replaceAll('\\.', '_')
-            project.logger.debug("BaselineIdea: Configuring IDEA Module for Java version: " + javaVersion)
-
-            if (ideaModel.project != null) {
-                ideaModel.project.languageLevel = javaVersion
-            }
-
-            ideaModel.module.jdkName = javaVersion
-            ideaModel.module.iml.withXml {
-                it.asNode().component.find { it.@name == 'NewModuleRootManager' }.@LANGUAGE_LEVEL = jdkVersion
-            }
-        } else {
-            project.logger.debug("BaselineIdea: No Java version found in sourceCompatibility property.")
-        }
     }
 
     /**
--- a/gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineIdeaIntegrationTest.groovy
+++ b/gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineIdeaIntegrationTest.groovy
@@ -70,36 +70,6 @@ class BaselineIdeaIntegrationTest extends AbstractPluginTest {
         with('idea').build()
     }
 
-    def 'Modules for subprojects pick up the correct sourceCompatibility'() {
-        when:
-        buildFile << standardBuildFile
-        buildFile << '''
-            sourceCompatibility = 1.6
-        '''.stripIndent()
-        def subproject = multiProject.create(["subproject1", "subproject2"])
-        subproject["subproject1"].buildGradle << '''
-            apply plugin: 'java'
-            apply plugin: 'com.palantir.baseline-idea'
-            sourceCompatibility = 1.7
-        '''.stripIndent()
-        subproject["subproject2"].buildGradle << '''
-            apply plugin: 'java'
-            apply plugin: 'com.palantir.baseline-idea'
-            sourceCompatibility = 1.8
-        '''.stripIndent()
-
-        then:
-        with('idea').build()
-        def rootIml = Files.asCharSource(new File(projectDir, projectDir.name + ".iml"), Charsets.UTF_8).read()
-        rootIml ==~ /(?s).*orderEntry[^\\n]*jdkName="1.6".*/
-        def subproject1Iml = Files.asCharSource(new File(projectDir, "subproject1/subproject1.iml"),
-                Charsets.UTF_8).read()
-        subproject1Iml ==~ /(?s).*orderEntry[^\\n]*jdkName="1.7".*/
-        def subproject2Iml = Files.asCharSource(new File(projectDir, "subproject2/subproject2.iml"),
-                Charsets.UTF_8).read()
-        subproject2Iml ==~ /(?s).*orderEntry[^\\n]*jdkName="1.8".*/
-    }
-
     def 'Idea project has copyright configuration'() {
         when:
         buildFile << standardBuildFile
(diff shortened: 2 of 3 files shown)


pkoenig10:
Here is the full diff for the removed test case as a result of this change:
```
diff -r before/project.iml after/project.iml
4a5
>     <orderEntry type="inheritedJdk"/>
10d10
<     <orderEntry type="jdk" jdkName="1.6" jdkType="JavaSDK"/>

diff -r before/project.ipr after/project.ipr
101c101
<   <component name="ProjectRootManager" version="2" languageLevel="JDK_1_6" assert-keyword="true" jdk-15="true" project-jdk-type="JavaSDK" assert-jdk-15="true" project-jdk-name="1.8">
---
>   <component name="ProjectRootManager" version="2" languageLevel="JDK_1_8" assert-keyword="true" jdk-15="true" project-jdk-type="JavaSDK" assert-jdk-15="true" project-jdk-name="1.8">

diff -r before/subproject1/subproject1.iml after/subproject1/subproject1.iml
4a5
>     <orderEntry type="inheritedJdk"/>
10d10
<     <orderEntry type="jdk" jdkName="1.7" jdkType="JavaSDK"/>

diff -r before/subproject2/subproject2.iml after/subproject2/subproject2.iml
3c3
<   <component name="NewModuleRootManager" inherit-compiler-output="true" LANGUAGE_LEVEL="JDK_1_8">
---
>   <component name="NewModuleRootManager" inherit-compiler-output="true">
4a5
>     <orderEntry type="inheritedJdk"/>
10d10
<     <orderEntry type="jdk" jdkName="1.8" jdkType="JavaSDK"/>
```

Explaining the changes:
- The 3 `*.iml` `type="inheritedJdk"` changes are the intention of this PR.
- The `project.ipr` `languageLevel` change is because we are no longer explicitly setting the idea project language level (as recommended by the docs). Instead Gradle is calculating the maximum language level of all modules in the project. Explicitly setting a JDK version of 1.6 here would now be wrong because submodules which require a newer JDK version are inheriting the project JDK version.
- The `subproject2.iml` `LANGUAGE_LEVEL` change is because we no longer need to explicitly set a language level for this module since it uses to project default. This is equivalent to selecting the "Project default" option in the module language level settings in Intellij. This logic can be seen in Gradle [here](https://github.com/gradle/gradle/blob/v4.10.0/subprojects/ide/src/main/java/org/gradle/plugins/ide/idea/IdeaPlugin.java#L425).

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
```diff
@@ -228,29 +227,6 @@ class BaselineIdea extends AbstractBaselinePlugin {
...
-     */
-    private void addJdkVersion(IdeaModel ideaModel) {
-        def compileJavaTask = (JavaCompile) project.tasks.findByName('compileJava')
-        if (compileJavaTask) {
-            def javaVersion = compileJavaTask.sourceCompatibility
-            def jdkVersion = 'JDK_' + javaVersion.replaceAll('\\.', '_')
-            project.logger.debug("BaselineIdea: Configuring IDEA Module for Java version: " + javaVersion)
-
-            if (ideaModel.project != null) {
-                ideaModel.project.languageLevel = javaVersion
```
pkoenig10, line 242: It's probably wrong to set project values here since this is called when applying this plugin to each subproject. If subprojects specify different source compatibilities, then we have a race condition (since I don't think Gradle provides guarantees about the ordering plugins are applied to subprojects).

gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineIdeaIntegrationTest.groovy
```diff
@@ -70,36 +70,6 @@ class BaselineIdeaIntegrationTest extends AbstractPluginTest {
...
-        def subproject = multiProject.create(["subproject1", "subproject2"])
-        subproject["subproject1"].buildGradle << '''
-            apply plugin: 'java'
-            apply plugin: 'com.palantir.baseline-idea'
-            sourceCompatibility = 1.7
-        '''.stripIndent()
-        subproject["subproject2"].buildGradle << '''
-            apply plugin: 'java'
-            apply plugin: 'com.palantir.baseline-idea'
-            sourceCompatibility = 1.8
```
uschi2000, line 88: after this change, what's the resulting config in the iml xml?

pkoenig10:
Here is the test case showing the config before and after this change: [gradle-baseline-pr718.zip](https://github.com/palantir/gradle-baseline/files/3457483/gradle-baseline-pr718.zip)

pkoenig10:
Here is a summary of the changes. The easiest way to see the changes is to just import the projects I attached above into Intellij and look at the project/module settings.

|                        	| Before                                	| After                                                                    	|
|------------------------	|---------------------------------------	|--------------------------------------------------------------------------	|
| Project JDK            	| System default                        	| System default                                                           	|
| Project lang level     	| 6 (this might be non-deterministic)                                  	| 8 (computed by Gradle as maximum of module lang levels)                	|
| Root module JDK        	| 1.6 (broken if 1.6 JDK not installed) 	| Project JDK                                                              	|
| Root module lang level 	| 6                                     	| 6                                                                        	|
| Subproject1 module JDK 	| 1.7 (broken if 1.7 JDK not installed) 	| Project JDK                                                              	|
| Subproject1 module lang level 	| 7                                     	| 7                                                                        	|
| Subproject2 module JDK 	| 1.8 (broken if 1.8 JDK not installed) 	| Project JDK                                                              	|
| Subproject2 module lang level 	| 8                                     	| Project lang level ([Gradle prefers this](https://github.com/gradle/gradle/blob/v4.10.0/subprojects/ide/src/main/java/org/gradle/plugins/ide/idea/IdeaPlugin.java#L425) over an explicit value when possible) 	|

gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineIdeaIntegrationTest.groovy
iamdanfox: I ran it on our internal foundry email-service and it gave me this `idea.iml` file:
```
<?xml version="1.0" encoding="UTF-8"?>
<module relativePaths="true" type="JAVA_MODULE" version="4">
  <component name="NewModuleRootManager" inherit-compiler-output="true">
    <exclude-output />
    <content url="file://$MODULE_DIR$">
      <excludeFolder url="file://$MODULE_DIR$/.gradle" />
      <excludeFolder url="file://$MODULE_DIR$/build" />
    </content>
    <orderEntry type="inheritedJdk" />
    <orderEntry type="sourceFolder" forTests="false" />
  </component>
</module>
```

iamdanfox:
Before we merge this, I'm wanna double check the scenario where a repo uses Java11 for its main server code and Java8 for its published conjure API, cos i think we're gonna need to push people towards this soon!

EDIT confirmed looks like it works :)

image


image

iamdanfox (approved):
Thanks @pkoenig10!

PR #722, opened 2019-08-01
Enforce spotless format

ferozco:
## Before this PR
We did not enforce spotless format. This was OK since the checkstyle would catch the format violations. This breaks down when a repo opts into using the eclipse formatter since nothing ensures that your code is well formatted

## After this PR
==COMMIT_MSG==
Enforce that all source passes spotless format
==COMMIT_MSG==

## Possible downsides?
There may be some red PRs if users had previously disabled checkstyle. In those cases, users can always opt out through the spotless extension


iamdanfox:
Are you sure about this? I'm pretty sure I've seen spotless violations cause red builds already?

iamdanfox:
Closing out as seems to be workign already

PR #724, opened 2019-08-01
Fix spotless apply with eclipse config

ferozco:
## Before this PR
There was a race when running spotlessApply with eclipse formatting enabled for the first time since `:baselineUpdateConfig` and `spotlessApply` would run concurrently

## After this PR
==COMMIT_MSG==
Correctly set dependency between spotlessApply and baselineUpdateConfig to prevent a race
==COMMIT_MSG==

## Possible downsides?
N/A

@diogoholanda for SA

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineFormat.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineFormat.java
@@ -65,10 +65,11 @@ class BaselineFormat extends AbstractBaselinePlugin {
             Task formatTask = project.task("format");
             project.afterEvaluate(p -> {
                 Task spotlessJava = project.getTasks().getByName("spotlessJava");
+                Task spotlessApply = project.getTasks().getByName("spotlessApply");
                 if (eclipseFormattingEnabled(project) && !Files.exists(eclipseXml)) {
                     spotlessJava.dependsOn(project.getTasks().findByPath(":baselineUpdateConfig"));
                 }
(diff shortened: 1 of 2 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineFormat.java
```diff
@@ -63,13 +63,13 @@ public void apply(Project project) {
...
             // necessary because SpotlessPlugin creates tasks in an afterEvaluate block
             Task formatTask = project.task("format");
-            if (eclipseFormattingEnabled(project) && !Files.exists(eclipseXml)) {
-                formatTask.dependsOn(project.getTasks().findByPath(":baselineUpdateConfig"));
-            }
             project.afterEvaluate(p -> {
-                Task spotlessApply = project.getTasks().getByName("spotlessApply");
-                formatTask.dependsOn(spotlessApply);
-                project.getTasks().withType(JavaCompile.class).configureEach(spotlessApply::mustRunAfter);
+                Task spotlessJava = project.getTasks().getByName("spotlessJava");
```
ferozco, line 67: There was also another slight bug, where we referenced the wrong spotless task. Turns out that the task doing all of the work is `spotlessJava` and that it uses the presence of other tasks in the task graph to determine how to behave https://github.com/diffplug/spotless/blob/master/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessPlugin.java#L99

iamdanfox (approved):
Ah nice this is much more precise!

PR #727, opened 2019-08-01
Allow suppression of TODO checkstyle check

jdm2212:
## Before this PR
Allow suppression of the `TODO` checkstyle check by giving it an ID. Clarify its comment to allow `// TODO(username): ...`

## After this PR
==COMMIT_MSG==
No more unhelpful checkstyle demand for a ticket associated with each TODO.
==COMMIT_MSG==

### DIFF
--- a/.baseline/checkstyle/checkstyle.xml
+++ b/.baseline/checkstyle/checkstyle.xml
@@ -374,6 +374,11 @@
             <property name="message" value="Redundant ''static'' modifier."/>
         </module>
         <module name="RegexpSinglelineJava">
+            <property name="id" value="TodoFormat"/>
+            <property name="format" value="\/\/TODO|\/\/ TODO(?!\([^()\s]+\): )"/>
+            <property name="message" value="TODO format: '// TODO(#issue): explanation' or '// TODO(username): explanation'"/>
+        </module>
+        <module name="RegexpSinglelineJava">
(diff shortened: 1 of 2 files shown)


jdm2212:
I can't click the "generate changelog" button 😞

jdm2212:
I just checked and even Google's style guide is less restrictive than this check, allowing `TODO(username)`, `TODO(user@domain.com)`, and `TODO(ticket #)`:  https://google.github.io/styleguide/cppguide.html#TODO_Comments

uschi2000:
I'm not in favour of removing this check altogether.

iamdanfox:
Mmm I think I disagree with the premise that `The TODO checkstyle check is unhelpful and widely disobeyed`. Pretty much all Foundry repos comply with this without too much trouble.  Valid lines are:

```
// TODO(dfox): expand this section to include foo
// TODO(#123): delete this when x becomes true
```

I think they're pretty helpful because it makes it more likely that future readers of TODO lines will be able to find a bit more context about why the comment was added, or at least go and ask the right person. If the messaging isn't clear about this, I'd prefer to improve the messaging.

If you want to suppress this particular one, I think we could add an id so you can suppress just this one without taking out the other 31 RegexpSinglelineJava.

```
            <property name="id" value="TodoFormat"/>
```

jdm2212:
@iamdanfox I added the ID, also added `TODO(username): ...` to the comment and update the PR description. changelog-bot seems unhappy though and I'm not sure how to give it access to my fork.

iamdanfox (approved):

PR #733, opened 2019-08-05
print out suggestion for module dependencies inclusion in useful format

esword:
using "project(':project-name')" rather than treating it as a jar.

## Before this PR
Suggestions for adding implicit dependency to another module in the same project were not useful because they displayed as artifacts (like jars)

## After this PR
==COMMIT_MSG==
print out suggestion for module dependencies inclusion in useful format
using "project(':project-name')" rather than treating it as a jar.
==COMMIT_MSG==

## Possible downsides?
none

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckImplicitDependenciesTask.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckImplicitDependenciesTask.java
@@ -19,20 +19,13 @@ package com.palantir.baseline.tasks;
 import com.google.common.collect.Sets;
 import com.google.common.collect.Streams;
 import com.palantir.baseline.plugins.BaselineExactDependencies;
-import java.nio.file.Path;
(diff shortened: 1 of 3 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckImplicitDependenciesTask.java
```diff
@@ -97,6 +92,30 @@ public final void checkImplicitDependencies() {
         }
     }
 
+    private String getSuggestionString(ResolvedArtifact artifact) {
+        String artifactNameString;
```
iamdanfox: I think this would be an OK use of a ternary operator - seems less gross than the `artifactNameString` thingy

esword: done

iamdanfox (approved):
Thanks!

PR #734, opened 2019-08-05
add flag to not fail if implicit deps are found.  

esword:
This is so the task can be used in workflow where its outputs are used.

## Before this PR
The set of implicit dependencies is not accessible to other tasks, such as one that may attempt to automatically update the gradle config files for you.  This is the first step in letting the task be used in such a way.

An alternative approach would be to separate out all the logic to an "Analyzer" class like is done in for the CheckClassUniquenessTask/ClassUniquenessAnalyzer, but that would involve a lot of duplication of properties, so this seemed cleaner.

## After this PR
==COMMIT_MSG==
Task can be used without stopping the build on failure.
==COMMIT_MSG==

## Possible downsides?
None

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckImplicitDependenciesTask.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckImplicitDependenciesTask.java
@@ -35,6 +35,7 @@ import org.gradle.api.artifacts.ResolvedArtifact;
 import org.gradle.api.artifacts.ResolvedDependency;
 import org.gradle.api.artifacts.component.ProjectComponentIdentifier;
 import org.gradle.api.file.FileCollection;
+import org.gradle.api.model.ObjectFactory;
 import org.gradle.api.provider.ListProperty;
 import org.gradle.api.provider.Property;
 import org.gradle.api.provider.Provider;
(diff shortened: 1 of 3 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckImplicitDependenciesTask.java
```diff
@@ -48,15 +49,19 @@
     private final ListProperty<Configuration> dependenciesConfigurations;
     private final Property<FileCollection> sourceClasses;
     private final SetProperty<String> ignore;
+    private final Property<Boolean> failIfImplicitDependencies;
 
     public CheckImplicitDependenciesTask() {
         setGroup("Verification");
         setDescription("Ensures all dependencies are explicitly declared, not just transitively provided");
-        dependenciesConfigurations = getProject().getObjects().listProperty(Configuration.class);
+        final ObjectFactory factory = getProject().getObjects();
```
iamdanfox, line 57: I think these final keywords are pretty much redundant in modern Java - could we omit them just  to minimise visual clutter?

iamdanfox:
Hang on why doesn't the built-in `--continue` flag do what you want here?

esword:
It can, but I thought that still failed the build even if it let other tasks run?

iamdanfox:
Maybe that's acceptable if the workflow is just run locally?

esword:
Unfortunately not easily.  I need to write a task which depends on checkImplicitDependencies, and if you have a task which depends on another task that fails, the downstream task won't be run.  If you really don't like this flag, I can refactor out stuff and create an independent analyzer.  I was just hoping to avoid that for now.

esword:
Abandoning this - pulling out an analyzer class so easier to reuse

Issue #567, opened 2019-03-07
Exact dependencies check recommends circular module self-reference


schlosna:
## What happened?

I enabled the `com.palantir.baseline-exact-dependencies` plugin added to gradle-baseline in 0.52.0 for [tritium](https://github.com/palantir/tritium) 

Steps to repro:
* Check out `ds/checkDeps` branch from [tritium](https://github.com/palantir/tritium) (see https://github.com/palantir/tritium/pull/200 )
* Run `./gradlew checkImplicitDependencies` and receive the following error message that the `tritium-api` module should declare `implementation 'com.palantir.tritium:tritium-api'` dependency

```
FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':tritium-api:checkImplicitDependencies'.
> Found 1 implicit dependencies - consider adding the following explicit dependencies to 'tritium-api/build.gradle', or avoid using classes from these jars:
      dependencies {
          implementation 'com.palantir.tritium:tritium-api'
      }
```

## What did you want to happen?

The check should not recommend modules adding implementation dependency on themselves.

schlosna:
Interestingly I don't get the module self-reference recommendation in circle build https://circleci.com/gh/palantir/tritium/924 but I do locally.

schlosna:
After [addressing a couple legit implicit transitive dependencies](https://github.com/palantir/tritium/pull/200/commits/f323b7d64fcf159b675b8a2f6f84cab72c66f931), the [jdk11 build](https://circleci.com/gh/palantir/tritium/931) exhibits the self-reference issue, while the [jdk8 build](https://circleci.com/gh/palantir/tritium/tree/ds%2FcheckDeps) passed:

```
org.gradle.api.GradleException: Found 1 implicit dependencies - consider adding the following explicit dependencies to 'tritium-tracing/build.gradle', or avoid using classes from these jars:
    dependencies {
        implementation 'com.palantir.tritium:tritium-tracing'
    }
	at com.palantir.baseline.tasks.CheckImplicitDependenciesTask.checkImplicitDependencies(CheckImplicitDependenciesTask.java:92)
```

iamdanfox:
Just hit this same thing on palantir/tracing-java

esword:
fixed by https://github.com/palantir/gradle-baseline/pull/736

PR #736, opened 2019-08-05
detect if dependency is actually the current project.

esword:
## Before this PR
Described here: https://github.com/palantir/gradle-baseline/issues/567

This can happen if you have a multimodule build.  If you run the implicit deps task on module A that depends o module B and then run the task on module B, the artifact for module B will already be in the artifact cache and so will be detected and returned.

## After this PR
==COMMIT_MSG==
We filter out artifacts from the current project
==COMMIT_MSG==

## Possible downsides?
None

### DIFF
--- a/changelog/@unreleased/pr-736.v2.yml
+++ b/changelog/@unreleased/pr-736.v2.yml
@@ -1,6 +1,6 @@
 type: fix
 fix:
-  description: detect if dependency is actually the current project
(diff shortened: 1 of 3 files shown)


iamdanfox (approved):

iamdanfox:
Wanna merge in develop to pick up #739?

dansanduleac:
👍

Issue #614, opened 2019-04-16
FR: help people spot *almost* unused dependencies


iamdanfox:
## What happened?

We recently found someone importing all of hadoop just for the Time class.

## What did you want to happen?

It would great if we could somehow warn people that they're taking a huge dependency just for one class.

esword:
This can be done if switch to using the jdeps dependency analyzer (https://docs.oracle.com/javase/9/tools/jdeps.htm#JSWOR690).  Can also use the --apionly option with that analyzer to be able to report what dependencies should be api vs implementation.

PR #738, opened 2019-08-05
pull out implicit deps analysis logic to independent class so can use with other tasks

esword:
## Before this PR
pull out implicit deps analysis logic to independent class so can use with other tasks

## After this PR
==COMMIT_MSG==
Analyzer class let's us reuse the output so can do things like auto-update gralde config files
==COMMIT_MSG==

## Possible downsides?
none


esword:
This is not the most elegant way of invoking separated logic, but could be a good basis for additional tasks, like https://github.com/palantir/gradle-baseline/issues/614

PR #740, opened 2019-08-06
fix #737: Implement DangerousCompletableFutureUsage errorprone check

### LINKED ISSUE #737: CompletableFuture *async methods should require an executor
## What happened?
The overload which doesn't take an executor uses the fork join pool. Contention from misbehaving blocking work blocks execution in ways that we don't expect. The fork-join-pool doesn't report metrics we can use to diagnost problems.

We should require an executor to be passed to all methods which can take one. `supplyAsync`, `runAsync`, `thenRunAsync`, etc. In the unlikely case we do really want to use the fork join pool, we can pass `ForkJoinPool.commonPool()`.

carterkozak:
## Before this PR
We could easily accidentally enqueue work onto the ForkJoinPool.
See #737

## After this PR
==COMMIT_MSG==
Implement DangerousCompletableFutureUsage errorprone check
==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DangerousCompletableFutureUsage.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DangerousCompletableFutureUsage.java
@@ -88,16 +88,16 @@ public final class DangerousCompletableFutureUsage
     }
 
     private static boolean isCompletionStageAsyncMethodWithoutExecutor(MethodInvocationTree tree, VisitorState state) {
-        if (COMPLETION_STAGE_ASYNC_INVOCATION.matches(tree, state)) {
-            Symbol.MethodSymbol symbol = ASTHelpers.getSymbol(tree);
-            // Should only happen if there are errors in the AST
-            if (symbol != null) {
-                List<Type> parameterTypes = symbol.type.getParameterTypes();
-                Type lastParameterType = parameterTypes.get(parameterTypes.size() - 1);
-                return !ASTHelpers.isSameType(
-                        lastParameterType, state.getTypeFromString(Executor.class.getName()), state);
-            }
+        if (!COMPLETION_STAGE_ASYNC_INVOCATION.matches(tree, state)) {
+            return false;
         }
-        return false;
+        Symbol.MethodSymbol symbol = ASTHelpers.getSymbol(tree);
+        if (symbol == null) {
+            // Errors in the AST, allow the build to continue and provide a more helpful compile error.
+            return false;
+        }
+        List<Type> parameterTypes = symbol.type.getParameterTypes();
+        Type lastParameterType = parameterTypes.get(parameterTypes.size() - 1);
+        return !ASTHelpers.isSameType(lastParameterType, state.getTypeFromString(Executor.class.getName()), state);
     }
 }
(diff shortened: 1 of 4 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DangerousCompletableFutureUsage.java
```diff
@@ -0,0 +1,103 @@
...
+            + "and specify the desired parallelism (i.e. the max number of concurrent tasks that will be submitted).\n"
+            + "By default, CompletableFuture uses the globally shared ForkJoinPool. Fork/join pools implement "
+            + "work-stealing, where any thread might steal a task from a different thread's queue when blocked "
+            + "waiting for a subtask to complete. This might not seem like an issue at first glance, but if you rely"
+            + "on the ForkJoinPool for short tasks extensively throughout your codebase and later on you add one "
+            + "piece of code that uses the pool for long (e.g. I/O) tasks, the other parts of your codebase that"
+            + "you'd expect to have consistent performance might experience performance degradation for no apparent "
+            + "reason.\n"
+            + "If you're absolutely certain that the ForkJoinPool is correct, please pass ForkJoinPool.commonPool()"
+            + "directly, ideally with a comment explaining why it is ideal.";
```
carterkozak, line 62: Slightly modified message from `DangerousParallelStreamUsage.ERROR_MESSAGE`

CRogers: This is great, love the explanation.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DangerousCompletableFutureUsage.java
```diff
@@ -0,0 +1,103 @@
...
+            return buildDescription(tree).setMessage(ERROR_MESSAGE).build();
+        }
+        return Description.NO_MATCH;
+    }
+
+    private static boolean isCompletionStageAsyncMethodWithoutExecutor(MethodInvocationTree tree, VisitorState state) {
+        if (COMPLETION_STAGE_ASYNC_INVOCATION.matches(tree, state)) {
+            Symbol.MethodSymbol symbol = ASTHelpers.getSymbol(tree);
+            // Should only happen if there are errors in the AST
+            if (symbol != null) {
```
CRogers: Perhaps extract the comment to a variable (ie make a var that is `boolean noErrorsInAst = symbol != null`)?

CRogers: Maybe also invert the `if` so it returns early to reduce nesting?

CRogers (approved):

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DangerousCompletableFutureUsage.java
carterkozak: I can't make that flow very well, I think the comment inside the (inverted) conditional provides similar readability.

carterkozak: 👍

PR #743, opened 2019-08-07
Rewrite addAll to add for arrays

gatesn:
After this PR
Refactor `addAll(Arrays.asList(<array>))` to `add(array)`. No point paying the conversion overheads.

## Possible downsides?
Need to make sure we run spotlessApply after the refaster rules, but I think we do that in general.


baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/AddAllArrayToBuilder.java
```diff
@@ -0,0 +1,44 @@
...
+import com.google.errorprone.refaster.annotation.AfterTemplate;
+import com.google.errorprone.refaster.annotation.BeforeTemplate;
+import java.util.Arrays;
+
+public final class AddAllArrayToBuilder<E> {
+
+    @BeforeTemplate
+    ImmutableCollection.Builder<E> addAllAsList(
+            ImmutableCollection.Builder<E> builder, E[] elements) {
+        return builder.addAll(Refaster.anyOf(
```
iamdanfox, line 32: `Refaster.anyOf` is v. slick

iamdanfox (approved):

gatesn:
@iamdanfox any ideas why this might fail on Java 11?

dansanduleac:
Wait, so there's no way to get this test to work on java 11?
Will it actually work if run in gradle under a java 11 compiler?

iamdanfox:
Hmmm don't think we can really merge something that just straight up doesn't work on a Java11 JDK :-1:

```
assumeThat(System.getProperty("java.specification.version")).isEqualTo("1.8");
```

iamdanfox (approved):
We've actually got a precedent for this "if java8" stuff already, reasoning that some code-rewriting is better than no code rewriting.

PR #751, opened 2019-08-12
Include more checks to be patched

pkoenig10:
## Before this PR
`MutableConstantField`, `UnusedMethod`, and `UnusedVariable` suggested fixes are not automatically applied.

## After this PR
`MutableConstantField`, `UnusedMethod`, and `UnusedVariable` suggested fixes are automatically applied.

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineErrorProneExtension.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineErrorProneExtension.java
@@ -34,7 +34,6 @@ public class BaselineErrorProneExtension {
             // Built-in checks
             "ArrayEquals",
             "MissingOverride",
-            "MutableConstantField",
             "UnusedMethod",
             "UnusedVariable");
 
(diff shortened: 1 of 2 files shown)


markelliot:
is there a programmatic way to keep this list up-to-date?

pkoenig10:
Not easily/currently.

See this comment:
https://github.com/palantir/gradle-baseline/blob/1.11.0/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java#L124-L130

And the related error-prone PR:
https://github.com/google/error-prone/pull/947

gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineErrorProneExtension.java
```diff
@@ -33,7 +33,10 @@
 
             // Built-in checks
             "ArrayEquals",
-            "MissingOverride");
+            "MissingOverride",
+            "MutableConstantField",
```
iamdanfox: Annoyingly the auto-fix for this actually results in quite a lot of lines going over the 120 char limit we have...

Given that this means more manual actions to merge one of these PRs (and might block baseline upgrades), could we actually just drop this one for now and keep the others?

I tried it out locally by adding:

```gradle
    plugins.withId('com.palantir.baseline-error-prone') {
        plugins.withId('java') {
            baselineErrorProne {
                patchChecks.add('MutableConstantField')
            }
        }
    }
```

And then running `./gradlew classes -PerrorProneApply`

iamdanfox (approved):

carterkozak:
🌶🌶🌶 This deletes method invocations that have side-effects!

iamdanfox:
Yep. I've seen this delete fields annotated with `@JsonProperty` (which is dangerous because other systems might be relying on the serialized form), and also deleted side-effecty setup code from tests.

PR #753, opened 2019-08-15
Error prone validation that Stream.sort is invoked on comparable streams

carterkozak:
## Before this PR
ClassCastException was thrown at runtime for a problem we can detect at compile time.

## After this PR
==COMMIT_MSG==
Error prone validation that Stream.sort is invoked on comparable streams
==COMMIT_MSG==


iamdanfox (approved):
Haven't seen this in the wild, but +1 if you think this will catch real badness!

carterkozak:
> Haven't seen this in the wild, but +1 if you think this will catch real badness!

Oh yes, yes it will :-)

PR #754, opened 2019-08-15
Fix #656 `DangerousStringInternUsage`: Disallow String.intern() invocations

### LINKED ISSUE #656: Disallow invocations of java.lang.String.intern()
## What happened?

String.intern sounds great, the [javadoc](https://docs.oracle.com/javase/9/docs/api/java/lang/String.html#intern--) describes exactly what we want: a canonical representation of a string value.
Unfortunately it's too good to be true, and has serious performance penalties, described here: https://shipilev.net/jvm/anatomy-quarks/10-string-intern/

We have several alternatives for similar functionality, including [Guava Interners](https://guava.dev/releases/27.0.1-jre/api/docs/com/google/common/collect/Interners.html), java9+ compact strings [JEP 254](https://openjdk.java.net/jeps/254), and G1 string deduplication [JEP 192](https://openjdk.java.net/jeps/192)

Given arcane sharp edges of String.intern, and benefits of the alternatives, we should not allow String.intern.

carterkozak:
## After this PR
==COMMIT_MSG==
`DangerousStringInternUsage`: Disallow String.intern() invocations
==COMMIT_MSG==


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DangerousStringInternUsage.java
```diff
@@ -0,0 +1,69 @@
...
+import com.google.errorprone.matchers.Description;
+import com.google.errorprone.matchers.Matcher;
+import com.google.errorprone.matchers.method.MethodMatchers;
+import com.sun.source.tree.ExpressionTree;
+import com.sun.source.tree.MethodInvocationTree;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "DangerousStringInternUsage",
+        severity = SeverityLevel.WARNING,
```
carterkozak, line 33: This is at WARNING severity for now, I'd feel more comfortable if it was an ERROR, but that would be severe for the initial rollout.

schlosna (approved):
👍 fixes https://github.com/palantir/gradle-baseline/issues/656

Honestly I’m surprised there’s no existing error prone rule for this, might be worth considering pushing upstream.

robert3005 (approved):
All for it - I think it makes sense to have users think about it

PR #759, opened 2019-08-20
errorprone `OptionalOrElseThrowThrows` to prevent throwing in orElseThrow

carterkozak:
==COMMIT_MSG==
Implement `OptionalOrElseThrowThrows` to prevent throwing from orElseThrow
==COMMIT_MSG==

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,6 +163,7 @@ Safe Logging can be found at [github.com/palantir/safe-logging](https://github.c
 - `DangerousCompletableFutureUsage`: Disallow CompletableFuture asynchronous operations without an Executor.
 - `NonComparableStreamSort`: Stream.sorted() should only be called on streams of Comparable types.
 - `DangerousStringInternUsage`: Disallow String.intern() invocations in favor of more predictable, scalable alternatives.
+- `OptionalOrElseThrowThrows`: Optional.orElseThrow argument must return an exception, not throw one.
 
 ## com.palantir.baseline-checkstyle
 Checkstyle rules can be suppressed on a per-line or per-block basis. (It is good practice to first consider formatting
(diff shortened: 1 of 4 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/OptionalOrElseThrowThrows.java
```diff
@@ -0,0 +1,89 @@
...
+import com.sun.source.util.SimpleTreeVisitor;
+import java.util.Optional;
+import java.util.function.Supplier;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "OptionalOrElseThrowThrows",
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
+        severity = BugPattern.SeverityLevel.WARNING,
```
carterkozak, line 48: Using `SeverityLevel.WARNING` for rollout, assuming nothing breaks I'll push it to ERROR later.

carterkozak:
I wrote an embarrassing bug earlier today, this will prevent me from doing it again!

asipser:
The exact error that shows up in intellij when compiling looks like this!
`Error: java: unreported exception java.lang.Throwable; must be caught or declared to be thrown`

README.md
```diff
@@ -163,6 +163,7 @@ Safe Logging can be found at [github.com/palantir/safe-logging](https://github.c
 - `DangerousCompletableFutureUsage`: Disallow CompletableFuture asynchronous operations without an Executor.
 - `NonComparableStreamSort`: Stream.sorted() should only be called on streams of Comparable types.
 - `DangerousStringInternUsage`: Disallow String.intern() invocations in favor of more predictable, scalable alternatives.
+- `OptionalOrElseThrowThrows`: Optional.orElseThrow argument must return an exception, not throw one.
```
ferozco, line 166: could we be more precise and say this only supports lambda arguments?

ferozco:
Thanks Carter! I'm sure i've written the same bug before too, so this is great 👍

README.md
carterkozak: I think that's an implementation detail, it's the most common case, and the one that I implemented this to support, but it wouldn't be unreasonable to eventually add support for anonymous classes (though I don't think it's worthwhile).

PR #761, opened 2019-08-20
LogSafePreconditionsMessageFormat disallows slf4j-style format characters

carterkozak:
## Before this PR
SafeLoggable exceptions thrown with slf4j style placeholders, despite not supporting any interpolation on safe-loggable exceptions. This has come up on several CRs and should be automated away by either disallowing substitution syntax, or supporting interpolation. I would prefer not to support interpolation.

## After this PR
==COMMIT_MSG==
LogSafePreconditionsMessageFormat disallows slf4j-style format characters
==COMMIT_MSG==

## Possible downsides?
Our tests and documentation previously suggested using slf4j placeholders despite resulting in substandard output.
see: https://github.com/palantir/safe-logging/pull/194

Adding `do not merge` for now, I imagine this would break several projects which followed the readme example from safe-logging. https://github.com/palantir/safe-logging/pull/194 should be resolved before the do-not-merge label is removed here.


iamdanfox (approved):
Looks good to me, especially given that we just weren't interpolating this at all!

PR #763, opened 2019-08-22
Error Prone LambdaMethodReference check

carterkozak:
This check uses `SeverityLevel.SUGGESTION`, and doesn't block
compilation, but is used by default with  `errorProneApply`.
This is meant to pair with `OptionalOrElseMethodInvocation` to
produce safe, legible code.

## Before this PR
```java
optional.orElseGet(() -> ImmutableList.of())
```

## After this PR
```java
optional.orElseGet(ImmutableList::of)
```
==COMMIT_MSG==
Error Prone LambdaMethodReference check
==COMMIT_MSG==

## Possible downsides?
It's possible I missed a case that could cause compile breaks.

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/LambdaMethodReference.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/LambdaMethodReference.java
@@ -1,0 +1,123 @@
+/*
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.baseline.errorprone;
+
+import com.google.auto.service.AutoService;
+import com.google.errorprone.BugPattern;
+import com.google.errorprone.VisitorState;
+import com.google.errorprone.bugpatterns.BugChecker;
+import com.google.errorprone.fixes.SuggestedFix;
+import com.google.errorprone.fixes.SuggestedFixes;
+import com.google.errorprone.matchers.Description;
+import com.google.errorprone.util.ASTHelpers;
+import com.sun.source.tree.BlockTree;
+import com.sun.source.tree.ExpressionTree;
+import com.sun.source.tree.LambdaExpressionTree;
+import com.sun.source.tree.MethodInvocationTree;
+import com.sun.source.tree.ReturnTree;
+import com.sun.source.tree.StatementTree;
+import com.sun.source.tree.Tree;
+import com.sun.tools.javac.code.Symbol;
+import java.util.Optional;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "LambdaMethodReference",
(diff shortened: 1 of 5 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/LambdaMethodReference.java
```diff
@@ -0,0 +1,123 @@
...
+                .filter(Symbol::isStatic)
+                .flatMap(symbol -> {
+                    SuggestedFix.Builder builder = SuggestedFix.builder();
+                    return toMethodReference(SuggestedFixes.qualifyType(state, builder, symbol))
+                            .map(qualified -> builder.replace(root, qualified).build());
+                });
+    }
+
+    private static Optional<String> toMethodReference(String qualifiedMethodName) {
+        int index = qualifiedMethodName.lastIndexOf('.');
```
ferozco, line 116: This feels a little brittle, but its probably alright

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/LambdaMethodReference.java
```diff
@@ -0,0 +1,123 @@
...
+        // simplifying map.forEach((k, v) -> func(k, v)) to map.forEach(this::func)
+        if (!tree.getParameters().isEmpty()) {
+            return Description.NO_MATCH;
+        }
+        LambdaExpressionTree.BodyKind bodyKind = tree.getBodyKind();
+        Tree body = tree.getBody();
+        // n.b. These checks are meant to avoid any and all cleverness. The goal is to be confident
+        // that we can fix the most basic problems correctly, not to take risks and produce code
+        // that may not compile.
+        switch (bodyKind) {
```
ferozco, line 60: could we use a visitor instead of reflection here?
```suggestion
body.accept(new SimpleTreeVisitor<Description, Object>(Description.NO_MATCH) {
	@Override
    public Description visitMethodInvocation(MethodInvocationTree node, Object o) {
    	...
    }
    
    @Override
    public Description visitBlock(BlockTree node, Object o) {
		...
    }
}, null)
```

gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineErrorProneExtension.java
```diff
@@ -24,12 +24,13 @@
 
     private static final ImmutableList<String> DEFAULT_PATCH_CHECKS = ImmutableList.of(
             // Baseline checks
+            "LambdaMethodReference",
```
ferozco, line 27: If these are applied in order, should `LambdaMethoReference` come after `OptionalOrElseMethodInvocation`

ferozco (commented):
Thanks Carter!

gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineErrorProneExtension.java
carterkozak: order doesn't appear to make a difference, either way it takes two runs to migrate to orElseGet + method reference.

ferozco: ok, good to know!

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/LambdaMethodReference.java
carterkozak: It would be cleaner, but I'm worried that this check will run a lot. We've had issues with errorprone performance in the past, so I avoided allocation until we've discovered a failure.

carterkozak: I'm not opposed to trying it, and walking back if we see degraded build times

ferozco:
👍

PR #764, opened 2019-08-22
failing test: OptionalOrElseMethodInvocation suggestion fails compilation

carterkozak:
The method target and parameters must be effectively final.

## After this PR
==COMMIT_MSG==
OptionalOrElseMethodInvocation suggestion fails compilation
==COMMIT_MSG==


carterkozak:
Another failure case:
orElse(callThrowsChecked()) -> orElseGet(() -> callThrowsChecked()) does not compile.

In this case it's almost definitely something we should fail in one way or another because exceptions tend to be thrown by things that do work.

PR #765, opened 2019-08-22
Eliminate some warnings/deprecations

iamdanfox:
We have one `import org.gradle.api.plugins.quality.FindBugs` line which makes baseline unusable in Gradle 6.0.  Interestingly, this has been deprecated for a while, but we didn't notice because we haven't been running with `-Werror` or `-Xlint:deprecation` (?!?!)


gradle-junit-reports/src/main/java/com/palantir/gradle/junit/JunitReportsFinalizer.java
```diff
@@ -50,18 +48,18 @@ public static void registerFinalizer(
...
+                new File(dir, task.getProject().getName() + "-" + task.getName() + ".xml")));
+        finalizer.reportDir.set(reportDir);
 
         task.finalizedBy(finalizer);
     }
 
     private Task styleTask;
     private TaskTimer taskTimer;
     private FailuresSupplier failuresSupplier;
-    private final RegularFileProperty targetFile = getProject().getLayout().fileProperty();
```
robert3005: Why not RegularFileProperty anymore?

ferozco:
👍

PR #771, opened 2019-08-23
LambdaMethodReference avoids suggestions for non-static methods

carterkozak:
No change to fixes that are automatically applied. Now the check
avoids making suggestions without suggested fixes to eliminate
noise.

## After this PR
==COMMIT_MSG==
LambdaMethodReference avoids suggestions for non-static methods
==COMMIT_MSG==

## Possible downsides?
None


ferozco:
👍

PR #773, opened 2019-08-24
Fix: Filter out pom only artifacts from dependnecies in checkUnusedDependencies

### LINKED ISSUE #594: BaselineExactDependencies should filter for only jar artifacts
We use the jar analyzer, meaning it breaks when taking a dependency on a pom file for example

robert3005:
## Before this PR
With pom only dependencies you would see #594 

## After this PR
==COMMIT_MSG==
Remove pom only dependencies from analysis in checkUnusedDependencies
==COMMIT_MSG==
fixes #594 

## Possible downsides?
We only take jar and project dependencies into consideration. There could be others that the maven dependency analyzer handles but we would be ignoring them

### DIFF
--- a/changelog/@unreleased/pr-773.v2.yml
+++ b/changelog/@unreleased/pr-773.v2.yml
@@ -1,0 +1,5 @@
+type: fix
+fix:
+  description: Remove pom only dependencies from analysis in checkUnusedDependencies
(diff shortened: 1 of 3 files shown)


ferozco:
👍

PR #775, opened 2019-08-28
baseline-circleci: Ensures root project name is specified

ferozco:
## Before this PR
We published to the wrong coordinates because we had not specified the root project name. 

## After this PR
==COMMIT_MSG==
Enforce that the root project name is specified
==COMMIT_MSG==

## Possible downsides?
N/A


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/Baseline.java
```diff
@@ -33,6 +34,8 @@ public void apply(Project project) {
                     "com.palantir.baseline should be applied to the root project only, not '{}'",
                     project.getName());
         }
+        Preconditions.checkState(!project.getName().equals(project.getProjectDir().getName()),
+                "Root project name must specified in settings.gradle");
```
carterkozak: I don't think this is quite right, it will match locally in most cases

iamdanfox: I'd prefer to just ban the specific name `"project"` because this is what we hit with the Circle1 -> Circle2 conversion

iamdanfox (approved):

ferozco (commented):
👍

dansanduleac (approved):

PR #784, opened 2019-09-02
Add error-prone strict mode

schlosna:
## Before this PR

Promoting baseline provided error-prone checks & compiler lint warnings or suggestions to errors requires a [bunch of `build.gradle` mess (e.g. in tritium)](https://github.com/palantir/tritium/blob/f6fc389/build.gradle#L81-L124).

## After this PR
==COMMIT_MSG==
Setting the following will promote all upstream & baseline provided
error-prone checks to error, enable compiler lint checks (`-Werror`,
`-Xlint:deprecation`, `-Xlint:unchecked`), and also promote all compiler
warnings to error.

```
baselineErrorProne {
    strict true
}
```
==COMMIT_MSG==

## Possible downsides?

Baseline tends to avoid configuration of any kind, though it balances that with not enabling overly aggressive default changes that would cause major breaks in consumer projects. Allowing configuration of strict mode maintains the current defaults, and the ideal here would be that new projects would default to strict mode enabled, and legacy projects could enable strict mode as they are cleaned up.

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineErrorProneExtension.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineErrorProneExtension.java
@@ -17,8 +17,14 @@
 package com.palantir.baseline.extensions;
 
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
+import com.google.errorprone.BugCheckerInfo;
 import org.gradle.api.Project;
 import org.gradle.api.provider.ListProperty;
+import org.gradle.api.provider.Property;
+import org.gradle.api.provider.Provider;
+import org.gradle.api.provider.SetProperty;
+import org.gradle.api.tasks.options.Option;
 
 public class BaselineErrorProneExtension {
 
@@ -36,14 +42,55 @@ public class BaselineErrorProneExtension {
             "ArrayEquals",
             "MissingOverride");
 
+    private static final ImmutableSet<String> DEFAULT_DISABLED_CHECKS = ImmutableSet.of(
+            "AndroidJdkLibsChecker", // ignore Android
+            "Java7ApiChecker", // we require JDK8+
+            "StaticOrDefaultInterfaceMethod", // Android specific
+            "Var" // high noise, low signal
+    );
+
     private final ListProperty<String> patchChecks;
+
+    private final SetProperty<String> disabledChecks;
+
+    private final Property<Boolean> isStrict;
 
     public BaselineErrorProneExtension(Project project) {
         patchChecks = project.getObjects().listProperty(String.class);
         patchChecks.set(DEFAULT_PATCH_CHECKS);
+        disabledChecks = project.getObjects().setProperty(String.class);
+        disabledChecks.set(DEFAULT_DISABLED_CHECKS);
+        isStrict = project.getObjects().property(Boolean.class);
+        isStrict.set(false);
     }
 
     public final ListProperty<String> getPatchChecks() {
         return patchChecks;
+    }
+
+    public final boolean isStrict() {
+        return Boolean.TRUE.equals(isStrict.get());
+    }
+
+    @Option(option = "strict", description = "Whether to apply strict compilation checks")
+    public final void strict(boolean shouldEnableStrict) {
+        this.isStrict.set(shouldEnableStrict);
+    }
+
+    public final void strict(Provider<Boolean> shouldEnableStrict) {
+        this.isStrict.set(shouldEnableStrict);
+    }
+
+    public final boolean isCheckEnabled(BugCheckerInfo check) {
+        return isCheckEnabled(check.canonicalName());
+    }
+
+    public final boolean isCheckEnabled(String check) {
+        return !disabledChecks.get().contains(check);
+    }
+
+    @Option(option = "disableCheck", description = "Disable specific error-prone check")
+    public final void disableCheck(String check) {
+        this.disabledChecks.add(check);
     }
 }
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineErrorProneExtension.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineErrorProneExtension.java
@@ -72,7 +72,6 @@ public class BaselineErrorProneExtension {
         return Boolean.TRUE.equals(isStrict.get());
     }
 
-    @Option(option = "strict", description = "Whether to apply strict compilation checks")
     public final void strict(boolean shouldEnableStrict) {
         this.isStrict.set(shouldEnableStrict);
     }
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
@@ -233,7 +233,7 @@ public final class BaselineErrorProne implements Plugin<Project> {
         });
     }
 
-    private Action<ErrorProneOptions> configureErrorProne(
+    private static Action<ErrorProneOptions> configureErrorProne(
             Project project,
             JavaCompile javaCompile,
             JavaVersion jdkVersion,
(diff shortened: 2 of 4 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
```diff
@@ -46,11 +82,50 @@
...
     public static final String REFASTER_CONFIGURATION = "refaster";
     public static final String EXTENSION_NAME = "baselineErrorProne";
+    public static final String PROP_STRICT = "com.palantir.baseline-error-prone.strict";
 
     private static final String ERROR_PRONE_JAVAC_VERSION = "9+181-r4173-1";
     private static final String PROP_ERROR_PRONE_APPLY = "errorProneApply";
     private static final String PROP_REFASTER_APPLY = "refasterApply";
 
+    private static final ImmutableSet<BugCheckerInfo> allBaselineChecks =
+            BuiltInCheckerSuppliers.getSuppliers(ImmutableList.of(
```
schlosna: error-prone maintains collections of their `ENABLED_ERRORS `, `ENABLED_WARNINGS`, `DISABLED_CHECKS`, see https://github.com/google/error-prone/blob/master/core/src/main/java/com/google/errorprone/scanner/BuiltInCheckerSuppliers.java

Thoughts on exposing the baseline versions of these to consumers via `public static` method?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
```diff
@@ -90,55 +165,16 @@ public void apply(Project project) {
...
+                        .configure(
+                                ErrorProneOptions.class,
+                                configureErrorProne(
+                                        project,
+                                        javaCompile,
+                                        jdkVersion,
+                                        errorProneExtension,
+                                        refasterRulesFile,
+                                        compileRefaster
+                                ));
```
schlosna, line 115: extracted this method as it was bumped over the cyclomatic complexity limit

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
```diff
@@ -195,6 +232,76 @@ public void apply(Project project) {
...
+            errorProneOptions.setDisableWarningsInGeneratedCode(true);
+            errorProneOptions.check("EqualsHashCode", CheckSeverity.ERROR);
+            errorProneOptions.check("EqualsIncompatibleType", CheckSeverity.ERROR);
+            errorProneOptions.check("StreamResourceLeak", CheckSeverity.ERROR);
+            errorProneOptions.check("InputStreamSlowMultibyteRead", CheckSeverity.ERROR);
+            errorProneOptions.check("JavaDurationGetSecondsGetNano", CheckSeverity.ERROR);
+            errorProneOptions.check("URLEqualsHashCode", CheckSeverity.ERROR);
+
+            if (isStrict(project) && !isRefactoring(project)) {
+                Set<String> compilerArgs = ImmutableSet.of("-Werror", "-Xlint:deprecation", "-Xlint:unchecked");
```
schlosna: thoughts on enabling these in strict mode?

schlosna (commented):
Wanted to throw this PR up for folks thoughts. I've run across a few different relatively new projects lately that I'd like to easily fully enable strict compilation & checks without copy/pasting a bunch of gradle across projects.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
robert3005: We already bootstrap new projects with `"-Werror", "-Xlint:deprecation"` so I am all for it

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
carterkozak: That would be helpful, I know of a large internal project which would benefit from the list.

dansanduleac:
You should merge develop to get the new circleci config with the expected checks.

Also, instead of `ext.'com.palantir.baseline-error-prone.strict' = true`, can we make this a typed property on a new `baselineErrorProne` extension?
Otherwise, it's easy to mess this up or manually mistype it and you'd have no feedback from gradle that it's wrong.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
pkoenig10: Can repos still disable certain checks? In particular I'm worried about `SwitchStatementDefaultCase` as there are probably some valid uses of default cases, most notably with non-API enum with many variants.

schlosna: You can `@SuppressWarnings("SwitchStatementDefaultCase")` for targeted classes or methods today.

We could potentially allow suppressing specific checks as part of baseline gradle config, but defer to baseline infra folks if that’s something we want to allow.

pkoenig10: I'm referring to modifying the error-prone options, which you can do today and which I assume will still work after this change.

schlosna: ^^^ would be good test cases, I don’t know offhand how the ordering in current PR would handle project adjusted error prone check args.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
```diff
@@ -46,11 +82,50 @@
...
+            PreferListsPartition.class,
+            PreferSafeLoggableExceptions.class,
+            PreferSafeLoggingPreconditions.class,
+            PreventTokenLogging.class,
+            ShutdownHook.class,
+            Slf4jConstantLogMessage.class,
+            Slf4jLogsafeArgs.class,
+            SwitchStatementDefaultCase.class,
+            ValidateConstantMessage.class
+    ));
```
iamdanfox: Could we use reflection to build this list? I'm concerned that we'll add a new class and forget to add it here.

iamdanfox:
I'm in favour of something like this - I think we've found that java libraries usually need to be held to a slightly higher bar than backend servers because there's so much more potential for breakage. Having one convenient switch to flip sounds like a good shout.

I wonder if we'd want to also flip on things like nullaway as part of this?

schlosna:
@iamdanfox I'm supportive of including NullAway as it has been valuable in my experience, but might suggest it being opt-in for now as it can be a bit of lift to annotate everything appropriately.

I'll try address some of the comments and get an updated PR shortly given travel.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
schlosna: Updated to use `ServiceLoader.load(BugChecker.class, BaselineErrorProne.class.getClassLoader())` filtered down to `com.palantir.**` classes. Thoughts on this approach?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
schlosna: @pkoenig10 added the ability to explicitly disable a specific check in strict mode (open to feedback on structure (as is this just leaves the specific check at its default severity level instead of promoting it to error -- do we actually want more control over severity?)

```
baselineErrorProne {
    strict true
    disableCheck 'SwitchStatementDefaultCase'
}
```

One can still use something along the lines of `options.errorprone.errorproneArgs << '-Xep:NullAway:ERROR'` as [`net.ltgt.gradle.errorprone.ErrorProneOptions#toString()`](https://github.com/tbroyer/gradle-errorprone-plugin/blob/master/src/main/kotlin/net/ltgt/gradle/errorprone/ErrorProneOptions.kt#L66-L84) is used to output all the compiler args and those will be output after the `checks` per the following sequence :

```
        return (
            sequenceOf(
                "-XepDisableAllChecks".takeIf { disableAllChecks },
                "-XepAllErrorsAsWarnings".takeIf { allErrorsAsWarnings },
                "-XepAllDisabledChecksAsWarnings".takeIf { allDisabledChecksAsWarnings },
                "-XepDisableWarningsInGeneratedCode".takeIf { disableWarningsInGeneratedCode },
                "-XepIgnoreUnknownCheckNames".takeIf { ignoreUnknownCheckNames },
                "-XepCompilingTestOnlyCode".takeIf { isCompilingTestOnlyCode },
                "-XepExcludedPaths:$excludedPaths".takeUnless { excludedPaths.isNullOrEmpty() }
            ).filterNotNull() +
                checks.asSequence().onEach(::validateName).map { (name, severity) -> "-Xep:$name${severity.asArg}" } +
                checkOptions.asSequence().map { (name, value) -> "-XepOpt:$name=$value" } +
                errorproneArgs +
                errorproneArgumentProviders.asSequence().flatMap { it.asArguments().asSequence() }
            ).onEach(::validate)
            .joinToString(separator = " ")

```

pkoenig10: Do we actually need this `disableCheck`? Is it not sufficient to do:
```
tasks.withType(JavaCompile) {
    options.errorprone.errorproneArgs += ['-Xep:SwitchStatementDefaultCase:OFF']
}
```

gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineErrorProneExtension.java
```diff
@@ -36,14 +42,55 @@
...
 
     public final ListProperty<String> getPatchChecks() {
         return patchChecks;
     }
+
+    public final boolean isStrict() {
+        return Boolean.TRUE.equals(isStrict.get());
+    }
+
+    @Option(option = "strict", description = "Whether to apply strict compilation checks")
```
iamdanfox: I don't think the @Option annotation makes sense on this extension - it's just for tasks so that you can apply `./foo --bar 123`.

schlosna:
not dead yet

PR #791, opened 2019-09-04
Break: Add gradle 6.0-20190904072820+0000 compatibility. Compile with -Werror and -Xlint:deprecation

robert3005:
## Before this PR
Compilation would fail if baseline was used with gradle 6 snapshots due to deprecations being removed

## After this PR
==COMMIT_MSG==
Add gradle 6.0-20190904072820+0000 compatibility. This raises minimum required version of gradle for plugins from this repo to 5.0. Additionally add '-Werror' and '-Xlint:deprecation' so we can detect compatibility issues early
==COMMIT_MSG==

## Possible downsides?
This pr is naive and just requires min gradle version to be 5 instead of 4.10

### DIFF
--- a/changelog/@unreleased/pr-791.v2.yml
+++ b/changelog/@unreleased/pr-791.v2.yml
@@ -1,0 +1,8 @@
+type: improvement
+improvement:
+  description: Compile with `-Werror` and `-Xlint:deprecation` to catch deprecations
+    when they're introduced instead of when they become compilation error. This allows
+    us to future proof early before regular users are hit with those issues in their
+    projects
+  links:
+  - https://github.com/palantir/gradle-baseline/pull/791
--- a/changelog/@unreleased/pr-791.v2.yml
+++ b/changelog/@unreleased/pr-791.v2.yml
@@ -1,8 +1,5 @@
-type: improvement
-improvement:
-  description: Compile with `-Werror` and `-Xlint:deprecation` to catch deprecations
-    when they're introduced instead of when they become compilation error. This allows
-    us to future proof early before regular users are hit with those issues in their
-    projects
+type: break
+break:
+  description: Add gradle 6.0-20190904072820+0000 compatibiltiy. This raises minimum required version of gradle for plugins from this repo to 5.0. Additionally add '-Werror' and '-Xlint:deprecation' so we can detect compatiblilty issues early
   links:
   - https://github.com/palantir/gradle-baseline/pull/791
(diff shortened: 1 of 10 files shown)


changelog/@unreleased/pr-791.v2.yml
```diff
@@ -0,0 +1,8 @@
+type: improvement
+improvement:
+  description: Compile with `-Werror` and `-Xlint:deprecation` to catch deprecations
```
ferozco: Could you describe the change in a way that would be helpful for consumers of the plugin? something like: `gradle-baseline should now be compatible with Gradle  6.0. gradle-baseline now requires at least Gradle 5.0 and is no longer compatible with 4.X.`

robert3005: I am not sure we are compatible with gradle 6 though since it's not released. I could say it's compatible with gradle snapshot that I tested with.

robert3005: Also if I was just making it compatible with gradle 6 I wouldn't be adding the compiler flags

changelog/@unreleased/pr-791.v2.yml
```diff
@@ -0,0 +1,8 @@
+type: improvement
+improvement:
```
schlosna: we should flag [this as a break](https://github.com/palantir/gradle-baseline/pull/493#issuecomment-451897295) so any folks lagging on gradle < 5.x upgrade

robert3005: As I mentioned in the pr description it's possible to not make it a break so I leave it to the maintainers to decide what they prefer

gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineIntegrationTest.groovy
```diff
@@ -43,6 +43,6 @@ class BaselineIntegrationTest extends AbstractPluginTest {
         with().withArguments('-s').withGradleVersion(gradleVersion).build()
 
         where:
-        gradleVersion << ['4.10.2', '5.0-milestone-1']
+        gradleVersion << ['5.0']
```
iamdanfox: could we add `6.0-20190904072820+0000` here?

changelog/@unreleased/pr-791.v2.yml
iamdanfox: Yeah I'd just phrase this as a break and declare compatibility with `6.0-20190904072820+0000`.  Seems like we've got rid of most gradle4 users in the foundry org:

Screenshot 2019-09-04 at 15 39 47

iamdanfox: Mmmm looks like there are more stragglers outside Foundry, but I think this is probably OK:

image

gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineIntegrationTest.groovy
robert3005: sure

iamdanfox (approved):
Ok this will cause gradle-baseline 2.0.  Anything else we want to squeeze in before we cut this?

PR #793, opened 2019-09-06
Fix: don't apply error prone patches for disabled checks

dansanduleac:
## Before this PR

`-PapplyErrorProne` will rewrite code that fails all of the patch checks in the `baselineErrorProne` extension, whether they are enabled or not on a project.

## After this PR
==COMMIT_MSG==
Stop applying error prone patches for checks that have been turned off.
==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferSafeLoggableExceptions.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferSafeLoggableExceptions.java
@@ -41,6 +41,7 @@ import java.util.stream.Collectors;
         linkType = BugPattern.LinkType.CUSTOM,
         severity = BugPattern.SeverityLevel.WARNING,
         summary = "Throw SafeLoggable exceptions to ensure the exception message will not be redacted")
+@SuppressWarnings("PreferSafeLoggableExceptions")
 public final class PreferSafeLoggableExceptions extends BugChecker implements BugChecker.NewClassTreeMatcher {
 
     private static final long serialVersionUID = 1L;
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
@@ -24,8 +24,10 @@ import com.palantir.baseline.tasks.RefasterCompileTask;
 import java.io.File;
 import java.util.AbstractList;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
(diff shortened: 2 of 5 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferSafeLoggableExceptions.java
```diff
@@ -41,6 +41,7 @@
         linkType = BugPattern.LinkType.CUSTOM,
         severity = BugPattern.SeverityLevel.WARNING,
         summary = "Throw SafeLoggable exceptions to ensure the exception message will not be redacted")
+@SuppressWarnings("PreferSafeLoggableExceptions")
```
dansanduleac, line 44: this was undeterministically flaking compilation locally from gradle.. :(

ferozco: I think we should disable the check on the whole project

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
```diff
@@ -132,10 +138,25 @@ public void apply(Project project) {
...
-                                            "-XepPatchLocation:IN_PLACE"));
+                                    errorProneOptions.getErrorproneArgumentProviders().add(() -> {
+                                        // Don't apply checks that have been explicitly disabled
+                                        Stream<String> errorProneChecks = errorProneExtension
+                                                .getPatchChecks()
+                                                .get()
+                                                .stream()
+                                                .filter(check -> {
+                                                    if (checkExplicitlyDisabled(errorProneOptions, check)) {
+                                                        log.warn("Task {}: not applying errorprone check {} because it "
```
ferozco: this log line seems a little overkill

dansanduleac: alright, can make it info

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferSafeLoggableExceptions.java
dansanduleac: it is already off, the baseline plugin itself turns it off for projects applying `java-gradle-plugin`.. I verified this with a snippet like so:
```
tasks.withType(JavaCompile) {
    println "${it.path}: ${options.errorprone.checks['PreferSafeLoggableExceptions']}"
}
```
and they are all OFF, however it doesn't seem to always take effect somehow

ferozco (approved):

PR #794, opened 2019-09-06
baseline-idea: configures IntelliJ EclipseCodeFormatter plugin

ferozco:
## Before this PR
There was no IDE integration when using the eclipse formatter locally.

## After this PR
==COMMIT_MSG==
Automatically configure the [Intellij Eclipse format plugin](https://plugins.jetbrains.com/plugin/6546-eclipse-code-formatter) to use the eclipse formatter
==COMMIT_MSG==

Users will get a little prompt in the bottom-right of their IDE:

Screenshot 2019-09-09 at 10 50 32

Clicking 'Install required plugins' will get the right formatter set up.

Screenshot 2019-09-09 at 10 50 40

### DIFF
--- a/changelog/@unreleased/pr-794.v2.yml
+++ b/changelog/@unreleased/pr-794.v2.yml
@@ -1,0 +1,6 @@
+type: feature
+feature:
+  description: Automatically configure the [Intellij Eclipse format plugin](https://plugins.jetbrains.com/plugin/6546-eclipse-code-formatter)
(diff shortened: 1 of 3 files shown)


iamdanfox (approved):
LGTM, this will only affect repos that have `com.palantir.baseline-format.eclipse=true` in their gradle.properties, so should be low risk!

ferozco:
👍

PR #807, opened 2019-09-11
Fix: eclipse formatter multicatch alignment passes checkstyle 

dansanduleac:
## Before this PR

Formatter aligns a multicatch block in a way that fails checkstyle.

https://github.com/palantir/gradle-consistent-versions/pull/228/commits/7c68b834c41f34951a7d9ca7ebd57ae96dc97cf2#diff-3f11fbb18c48f096735ae7340c349471R61

## After this PR
==COMMIT_MSG==
Eclipse formatter now aligns multicatch so that it passes checkstyle.
==COMMIT_MSG==


carterkozak:
I'm curious if we have a minimal checkstyle configuration that we guarantee will pass after running the formatter regardless of what code-soup we pour in?

iamdanfox:
I think my hope is that we can actually just say the output of the formatter is correct by definition and just kill checkstyle.

We'd likely still need some kind of DesignForExtension / FinalClass thing in errorprone to maintain our current class structure enforcement.

iamdanfox (approved):

PR #810, opened 2019-09-11
Fix: classpath uniqueness checks runtimeClasspath by default

dansanduleac:
## Before this PR

It would try to resolve the `runtime` configuration by default, which is deprecated, and doensn't even contain all (perhaps any) of the actual runtime dependencies.

## After this PR
==COMMIT_MSG==
ClassUniquenessPlugin now checks the `runtimeClasspath` configuration by default.
==COMMIT_MSG==


ferozco:
👍

PR #815, opened 2019-09-12
SafeLoggingExceptionMessageFormat disallows {} in safelog exception messages

carterkozak:
Validates safelog exception creation the same way as safe preconditions

## Before this PR
A few times a week I provide this feedback on PRs, or run into places that have merged this bug.

## After this PR
==COMMIT_MSG==
SafeLoggingExceptionMessageFormat disallows `{}` in safelog exception messages
==COMMIT_MSG==

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,6 +165,7 @@ Safe Logging can be found at [github.com/palantir/safe-logging](https://github.c
 - `DangerousStringInternUsage`: Disallow String.intern() invocations in favor of more predictable, scalable alternatives.
 - `OptionalOrElseThrowThrows`: Optional.orElseThrow argument must return an exception, not throw one.
 - `LambdaMethodReference`: Lambda should use a method reference.
+- `SafeLoggingExceptionMessageFormat`: SafeLoggable exceptions do not interpolate parameters.
 
 ## com.palantir.baseline-checkstyle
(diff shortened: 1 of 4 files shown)


iamdanfox (approved):
Always in favour of automating out code reviews!

lycarter:
What is the preferred format? For example, how would you adjust this snippet?
```
log.error("New product release change for {} was not added to catalog successfully. Failed change "
      + "is {}", SafeArg.of("coordinate", changeCoordinate),
      SafeArg.of("productReleaseChange", changeWithContext), e);
```

carterkozak:
@lycarter The snippit you posted is correct, this change is only relevant to Safe*Exception messages, which do not do interpolation.

carterkozak:
@lycarter To clarify though, in many projects we do not use interpolation at all, because the [witchcraft logging api](https://github.com/palantir/witchcraft-api/blob/80f5ccf7368079313259d97d28fb998a5fd1a11e/witchcraft-logging-api/src/main/conjure/witchcraft-logging-api.yml#L30-L84) doesn't interpolate, instead we provide a description message, and key/value pairs describing state

e.g.

```java
log.error("New product release change was not added to catalog successfully",
    SafeArg.of("changeCoordinate", changeCoordinate),
    SafeArg.of("productReleaseChange", changeWithContext),
    e);
```

PR #819, opened 2019-09-12
Strict unused variable check

### LINKED ISSUE #818: Unused parameters should not compile
## What happened?

Currently, errorprone's exist UnusedVariable check will prevent you declaring _local_ variables or fields but never using them:

```
> Task :log-receiver:compileJava FAILED
/Volumes/git/log-receiver/log-receiver/src/main/java/com/palantir/logreceiver/LogReceiverServer.java:37: warning: [UnusedVariable] The local variable 'whoops' is never read.
        String whoops = "alsdjld";
               ^
    (see https://errorprone.info/bugpattern/UnusedVariable)
  Did you mean to remove this line?
```

HOWEVER, an arguably more pernicious problem is unused arguments to a method will not be flagged:

```java
public void deleteThingy(AuthHeader authHeader) { // 🌶 auth never checked!
    store.deleteThingy();
}
```

Baseline should prevent this foot gun. It seems that Eclipse and IntelliJ can flag mistakes like this, and it seems that [sonarlint](https://rules.sonarsource.com/java/tag/unused/RSPEC-1172) has a similar thing ([source code](https://github.com/SonarSource/sonar-java/blob/master/java-checks/src/main/java/org/sonar/java/checks/unused/UnusedMethodParameterCheck.java))... [gradle plugin here](https://gitlab.com/remal/gradle-plugins/blob/master/src%2Fmain%2Fkotlin%2Fname%2Fremal%2Fgradle_plugins%2Fplugins%2Fcode_quality%2Fsonar%2FSonarLintPlugin.kt)

ferozco:
Closes #818 
## After this PR
==COMMIT_MSG==
Correctly identify unused parameters in public methods
==COMMIT_MSG==

## Possible downsides?
N/A

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StrictUnusedVariable.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StrictUnusedVariable.java
@@ -38,7 +38,6 @@ import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.ListMultimap;
 import com.google.common.collect.Multimaps;
-import com.google.common.collect.Sets;
 import com.google.errorprone.BugPattern;
 import com.google.errorprone.VisitorState;
 import com.google.errorprone.bugpatterns.BugChecker;
@@ -97,7 +96,7 @@ import javax.lang.model.element.Name;
 
 /**
  * Copy from {@Link UnusedVariable } except we consider all parameter for unused analysis.
- * We modified the `onlyCheckForReassignments` filter to exclude abstract methods
+ * We modified the `onlyCheckForReassignments` filter to exclude abstract methods, to check loggers
  * */
 @AutoService(BugChecker.class)
 @BugPattern(
@@ -126,9 +125,7 @@ public final class StrictUnusedVariable extends BugChecker implements BugChecker
                     "org.openqa.selenium.support.FindBys");
 
     /** The set of types exempting a type that is extending or implementing them. */
-    private static final ImmutableSet<String> EXEMPTING_SUPER_TYPES =
-            ImmutableSet.of(
-            );
+    private static final ImmutableSet<String> EXEMPTING_SUPER_TYPES = ImmutableSet.of();
 
     /** The set of types exempting a field of type extending them. */
     private static final ImmutableSet<String> EXEMPTING_FIELD_SUPER_TYPES =
@@ -139,13 +136,6 @@ public final class StrictUnusedVariable extends BugChecker implements BugChecker
                     "serialVersionUID",
                     // TAG fields are used by convention in Android apps.
                     "TAG");
-
-    private static final ImmutableSet<Modifier> LOGGER_REQUIRED_MODIFIERS =
-            Sets.immutableEnumSet(Modifier.PRIVATE, Modifier.STATIC, Modifier.FINAL);
-
-    private static final ImmutableSet<String> LOGGER_TYPE_NAME = ImmutableSet.of("GoogleLogger");
-
-    private static final ImmutableSet<String> LOGGER_VAR_NAME = ImmutableSet.of("logger");
 
     @Override
     public Description matchCompilationUnit(CompilationUnitTree tree, VisitorState state) {
@@ -580,14 +570,7 @@ public final class StrictUnusedVariable extends BugChecker implements BugChecker
 
         private boolean isFieldEligibleForChecking(VariableTree variableTree, Symbol.VarSymbol symbol) {
             return variableTree.getModifiers().getFlags().contains(Modifier.PRIVATE)
-                    && !SPECIAL_FIELDS.contains(symbol.getSimpleName().toString())
-                    && !isLoggerField(variableTree);
-        }
-
-        private boolean isLoggerField(VariableTree variableTree) {
-            return variableTree.getModifiers().getFlags().containsAll(LOGGER_REQUIRED_MODIFIERS)
-                    && LOGGER_TYPE_NAME.contains(variableTree.getType().toString())
-                    && LOGGER_VAR_NAME.contains(variableTree.getName().toString());
+                    && !SPECIAL_FIELDS.contains(symbol.getSimpleName().toString());
         }
 
         /** Returns whether {@code sym} can be removed without updating call sites in other files. */
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StrictUnusedVariable.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StrictUnusedVariable.java
@@ -23,7 +23,7 @@ import static com.google.common.collect.ImmutableList.toImmutableList;
 import static com.google.common.collect.Iterables.getLast;
 import static com.google.common.collect.Iterables.getOnlyElement;
 import static com.google.errorprone.BugPattern.ProvidesFix.REQUIRES_HUMAN_ATTENTION;
-import static com.google.errorprone.BugPattern.SeverityLevel.ERROR;
+import static com.google.errorprone.BugPattern.SeverityLevel.WARNING;
 import static com.google.errorprone.util.ASTHelpers.*;
 import static com.google.errorprone.util.SideEffectAnalysis.hasSideEffect;
 import static com.sun.source.tree.Tree.Kind.*;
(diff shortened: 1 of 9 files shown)


ferozco:
👍

iamdanfox:
Problems encountered so far:

- interfaces
- -XepDisableWarningsInGeneratedCode doesn't stop _errors_ in generated code, so you get a ton of complaints about immutables.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/UnusedParameter.java
```diff
@@ -0,0 +1,943 @@
...
+import java.util.Set;
+import java.util.concurrent.atomic.AtomicBoolean;
+import javax.annotation.Nullable;
+import javax.lang.model.element.ElementKind;
+import javax.lang.model.element.Modifier;
+import javax.lang.model.element.Name;
+
+/**
+ * Copy from {@Link UnusedVariable } except we consider all parameter for unused analysis.
+ * We removed the `onlyCheckForReassignments` filter
```
carterkozak: Should this be named `StrictUnusedVariable`, if it checks variables beyond parameters?

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/UnusedParameter.java
```diff
@@ -0,0 +1,939 @@
...
+                    "org.openqa.selenium.support.FindBys");
+
+    /** The set of types exempting a type that is extending or implementing them. */
+    private static final ImmutableSet<String> EXEMPTING_SUPER_TYPES =
+            ImmutableSet.of(
+            );
+
+    /** The set of types exempting a field of type extending them. */
+    private static final ImmutableSet<String> EXEMPTING_FIELD_SUPER_TYPES =
+            ImmutableSet.of("org.junit.rules.TestRule");
```
carterkozak: This should include java serialization magic methods, It's common to disallow deserialization by implementing `readObject` as a no-op.
https://docs.oracle.com/javase/8/docs/api/java/io/Serializable.html

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/UnusedParameter.java
```diff
@@ -0,0 +1,939 @@
...
+                    "serialVersionUID",
+                    // TAG fields are used by convention in Android apps.
+                    "TAG");
+
+    private static final ImmutableSet<Modifier> LOGGER_REQUIRED_MODIFIERS =
+            Sets.immutableEnumSet(Modifier.PRIVATE, Modifier.STATIC, Modifier.FINAL);
+
+    private static final ImmutableSet<String> LOGGER_TYPE_NAME = ImmutableSet.of("GoogleLogger");
+
+    private static final ImmutableSet<String> LOGGER_VAR_NAME = ImmutableSet.of("logger");
```
carterkozak: include `log`?

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/UnusedParameterTest.java
```diff
@@ -0,0 +1,56 @@
...
+    public void ohno() {
+        compilationHelper.addSourceLines(
+                "Test.java",
+                "import java.util.Optional;",
+                "class Test {",
+                "  // BUG: Diagnostic contains: Unused",
+                "  public static void foo(String buggy) {",
+                "  }",
+                "}").doTest();
+    }
```
carterkozak: Can we test that unused parameters of type `Void` don't trigger the check?

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/UnusedParameter.java
ferozco: Actually, I think we want to remove this special case entirely since you'd like to avoid having unused loggers sitting around

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StrictUnusedVariable.java
```diff
@@ -0,0 +1,926 @@
...
+ * Copy from {@Link UnusedVariable } except we consider all parameter for unused analysis.
+ * We modified the `onlyCheckForReassignments` filter to exclude abstract methods, to check loggers
+ * */
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "StrictUnusedVariable",
+        altNames = {"unused", "StrictUnusedVariable"},
+        summary = "Unused.",
+        providesFix = REQUIRES_HUMAN_ATTENTION,
+        severity = ERROR,
```
iamdanfox: I think we gotta put this at WARN otherwise we wreck immutables

iamdanfox (approved):

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/UnusedParameter.java
ferozco: The rule will ignore any parameter prefixed with `unused` so you should be able to easily avoid getting hit.

In general, I'd like to avoid adding special casing so that its easy to keep the rule up to date

schlosna:
Is this intended to handle the cause where an implementation of interface or overridden method does not use a method parameter?

This is failing on tritium in https://circleci.com/gh/palantir/tritium/3813 which at least some feel like false positives that I would not want to have to `@SuppressWarnings("StrictUnusedVariable")`

```
> Task :tritium-core:compileJava
/Volumes/git/tritium/tritium-core/src/main/java/com/palantir/tritium/event/NoOpInvocationEventHandler.java:46: warning: [StrictUnusedVariable] The parameter 'context' is never read.
    public void onFailure(@Nullable InvocationContext context, @Nonnull Throwable cause) {
                                                      ^
    (see https://errorprone.info/bugpattern/StrictUnusedVariable)
  Did you mean 'public void onFailure(@Nonnull Throwable cause) {'?
/Volumes/git/tritium/tritium-core/src/main/java/com/palantir/tritium/event/NoOpInvocationEventHandler.java:41: warning: [StrictUnusedVariable] The parameter 'result' is never read.
    public void onSuccess(@Nullable InvocationContext context, @Nullable Object result) {
                                                                                ^
    (see https://errorprone.info/bugpattern/StrictUnusedVariable)
  Did you mean 'public void onSuccess(@Nullable InvocationContext context) {'?
/Volumes/git/tritium/tritium-core/src/main/java/com/palantir/tritium/event/NoOpInvocationEventHandler.java:41: warning: [StrictUnusedVariable] The parameter 'context' is never read.
    public void onSuccess(@Nullable InvocationContext context, @Nullable Object result) {
                                                      ^
    (see https://errorprone.info/bugpattern/StrictUnusedVariable)
  Did you mean 'public void onSuccess(@Nullable Object result) {'?
/Volumes/git/tritium/tritium-core/src/main/java/com/palantir/tritium/event/NoOpInvocationEventHandler.java:46: warning: [StrictUnusedVariable] The parameter 'cause' is never read.
    public void onFailure(@Nullable InvocationContext context, @Nonnull Throwable cause) {
                                                                                  ^
    (see https://errorprone.info/bugpattern/StrictUnusedVariable)
  Did you mean 'public void onFailure(@Nullable InvocationContext context) {'?
1 error
4 warnings

> Task :tritium-core:compileJava FAILED

> Task :tritium-test:compileJava
/Volumes/git/tritium/tritium-test/src/main/java/com/palantir/tritium/test/TestImplementation.java:43: warning: [StrictUnusedVariable] The parameter 'string' is never read.
    public void multiArgumentMethod(String string, int count, Collection<String> foo) {
                                           ^
    (see https://errorprone.info/bugpattern/StrictUnusedVariable)
  Did you mean 'public void multiArgumentMethod(int count, Collection<String> foo) {'?
/Volumes/git/tritium/tritium-test/src/main/java/com/palantir/tritium/test/TestImplementation.java:43: warning: [StrictUnusedVariable] The parameter 'count' is never read.
    public void multiArgumentMethod(String string, int count, Collection<String> foo) {
                                                       ^
    (see https://errorprone.info/bugpattern/StrictUnusedVariable)
  Did you mean 'public void multiArgumentMethod(String string, Collection<String> foo) {'?
/Volumes/git/tritium/tritium-test/src/main/java/com/palantir/tritium/test/TestImplementation.java:43: warning: [StrictUnusedVariable] The parameter 'foo' is never read.
    public void multiArgumentMethod(String string, int count, Collection<String> foo) {
                                                                                 ^
    (see https://errorprone.info/bugpattern/StrictUnusedVariable)
  Did you mean 'public void multiArgumentMethod(String string, int count) {'?
1 error
3 warnings

> Task :tritium-test:compileJava FAILED

FAILURE: Build completed with 2 failures.
```

iamdanfox:
Yes, we do want to flag those (as a particular goal is to catch people implementing conjure-defined interfaces that specify an AuthHeader, but the implementation fails to use it)

The recommended suppression in this case is to just rename the variables to have an `unused` prefix, as this is a bit less verbose than the whole @SuppressWarnings thing.

```diff
-    public void onFailure(@Nullable InvocationContext context, @Nonnull Throwable cause) {
+    public void onFailure(@Nullable InvocationContext unusedContext, @Nonnull Throwable unusedCause) {
```

PR #827, opened 2019-09-13
baseline-refaster-rules can be overridden

iamdanfox:
## Before this PR

An excavator PR to upgrade _witchcraft_ was also applying WC refaster rules, but _also_ applying baseline ones.  The baseline ones were actually problematic in this case (i.e. needed human attention and blocked merging), especially because there was already an open PR that applied the refaster rules.

## After this PR
==COMMIT_MSG==
Excavator PRs that apply other refaster rules (e.g. Witchcraft ones) will not also apply baseline refaster rules.
==COMMIT_MSG==

## Possible downsides?
~I haven't written any tests for this, so it might regress :/~


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
```diff
@@ -67,13 +66,15 @@ public void apply(Project project) {
             String version = Optional.ofNullable(getClass().getPackage().getImplementationVersion())
                     .orElse("latest.release");
 
-            Configuration refasterConfiguration = project.getConfigurations().create(REFASTER_CONFIGURATION);
+            Configuration refasterConfiguration = project.getConfigurations().create("refaster", conf -> {
```
ferozco: Could we add a quick test for this? All we have to do is add any dependency to the refaster configuration and validate that compile was a no-op

iamdanfox: Annoyingly I can't just add any-old dependency (we validate too hard).  Needs to be a real refaster source dep.

```

* What went wrong:
Execution failed for task ':compileJava'.
> Can't load Refaster rule from /private/var/folders/nf/lxxlj0y55yn5p9tt4yhw6rpx081fjc/T/junit9097390508034524153/build/refaster/rules.refaster

```

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
```diff
@@ -129,9 +134,14 @@ public void apply(Project project) {
...
-                                    errorProneOptions.getErrorproneArgumentProviders().add(() -> ImmutableList.of(
-                                            "-XepPatchChecks:refaster:" + refasterRulesFile.get().getAbsolutePath(),
-                                            "-XepPatchLocation:IN_PLACE"));
+                                    errorProneOptions.getErrorproneArgumentProviders().add(() -> {
+                                        String file = refasterRulesFile.get().getAbsolutePath();
+                                        return new File(file).exists()
+                                                ? ImmutableList.of(
+                                                "-XepPatchChecks:refaster:" + file,
+                                                "-XepPatchLocation:IN_PLACE")
+                                                : Collections.emptyList();
```
iamdanfox: @ferozco this is what I had to do to make the test pass...

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
```diff
@@ -48,166 +49,185 @@
...
-
-            Task compileRefaster = project.getTasks().create("compileRefaster", RefasterCompileTask.class, task -> {
-                task.setSource(refasterConfiguration);
-                task.getRefasterSources().set(refasterConfiguration);
-                task.setClasspath(refasterCompilerConfiguration);
-                task.getRefasterRulesFile().set(refasterRulesFile);
-            });
+    public void apply(Project proj) {
+        proj.getPluginManager().withPlugin("java", plugin -> {
+            applyToJavaProject(proj);
```
iamdanfox: I had to do this refactor because cyclomatic complexity dinged us. Happy to put it in a separate PR if you like, otherwise it's just the last commit to read.

CRogers (approved):

PR #828, opened 2019-09-13
Fix compilation of baseline-error-prone to be deterministic

dansanduleac:
## Before this PR

The current errorprone project ends up using errorprone rules from _its own output directory_ leading to all sorts of confusion when running locally, especially after compiling from different branches where you might have remaining class files lying around. 

This occurrs when the local checkout's version is higher (according to gradle) than the baseline version used in buildscript. This is almost always true unless checking out an old branch.

The reason for this is gradle's aggressive replacing of external dependencies with a project dependency if the `group:name` matches.

## After this PR
==COMMIT_MSG==
Make baseline-error-prone pick up buildscript baseline's errorprone rules correctly so compilation is deterministic.

This avoids the confusing feedback loop where this project's own compilation output would be used to determine errorprone rules in future compilations of itself.

Additionally, disabled some checks globally which make no sense in projects that are not services.
==COMMIT_MSG==

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,3 +1,5 @@
+import net.ltgt.gradle.errorprone.CheckSeverity
+
 buildscript {
     repositories {
         jcenter()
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -47,6 +49,12 @@ allprojects {
         sourceCompatibility = 1.8
         tasks.withType(JavaCompile) {
             options.compilerArgs += ['-Werror', '-Xlint:deprecation']
+            options.errorprone {
+                check("Slf4jLogsafeArgs", CheckSeverity.OFF)
```
robert3005, line 53: `disable` instead of off

robert3005 (approved):

build.gradle
dansanduleac: sadly this needs [errorprone plugin 0.8](https://github.com/tbroyer/gradle-errorprone-plugin/releases/tag/v0.8) which would require gradle 5.2

build.gradle
```diff
@@ -47,6 +49,12 @@ allprojects {
...
+                check("PreferSafeLoggableExceptions", CheckSeverity.OFF)
+                check("PreferSafeLoggingPreconditions", CheckSeverity.OFF)
+                check("PreconditionsConstantMessage", CheckSeverity.OFF)
```
carterkozak, line 56: Not sure this makes sense to disable, string concatenation isn’t great.

robert3005: not sure this matters that much for a gradle plugin

PR #830, opened 2019-09-13
Fix: Checkstyle no longer verifies resources

dansanduleac:
## Before this PR

Checkstyle was going into the resources dirs, with seemingly no way to turn it off.
It can sometimes crash on these files which makes it almost impossible to exclude them.

## After this PR
==COMMIT_MSG==
Checkstyle tasks only check their own source set and only actual java sources. They don't look in your `src/*/resources` directory anymore.
==COMMIT_MSG==

### DIFF
--- a/gradle-junit-reports/src/main/java/com/palantir/gradle/junit/XmlReportFailuresSupplier.java
+++ b/gradle-junit-reports/src/main/java/com/palantir/gradle/junit/XmlReportFailuresSupplier.java
@@ -55,7 +55,11 @@ public final class XmlReportFailuresSupplier implements FailuresSupplier {
     @Override
     public List<Failure> getFailures() throws IOException {
         File sourceReport = reporting.getReports().findByName("xml").getDestination();
-        return XmlUtils.parseXml(reportHandler, new FileInputStream(sourceReport)).failures();
+        try {
+            return XmlUtils.parseXml(reportHandler, new FileInputStream(sourceReport)).failures();
+        } catch (Exception e) {
+            throw new RuntimeException(String.format("Failed to parse failures XML: %s", sourceReport), e);
+        }
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


gradle-junit-reports/src/main/java/com/palantir/gradle/junit/XmlReportFailuresSupplier.java
```diff
@@ -55,7 +55,11 @@ private XmlReportFailuresSupplier(
     @Override
     public List<Failure> getFailures() throws IOException {
         File sourceReport = reporting.getReports().findByName("xml").getDestination();
-        return XmlUtils.parseXml(reportHandler, new FileInputStream(sourceReport)).failures();
+        try {
+            return XmlUtils.parseXml(reportHandler, new FileInputStream(sourceReport)).failures();
+        } catch (Exception e) {
+            throw new RuntimeException(String.format("Failed to parse failures XML: %s", sourceReport), e);
+        }
```
dansanduleac, line 62: I snuck this one in as well as checkstyle crashing (on files from resources) caused an unfinished xml to be written, and it wasn't obvious where that XML was

robert3005 (approved):

gradle-junit-reports/src/main/java/com/palantir/gradle/junit/XmlReportFailuresSupplier.java
```diff
@@ -55,7 +55,11 @@ private XmlReportFailuresSupplier(
     @Override
     public List<Failure> getFailures() throws IOException {
         File sourceReport = reporting.getReports().findByName("xml").getDestination();
-        return XmlUtils.parseXml(reportHandler, new FileInputStream(sourceReport)).failures();
+        try {
+            return XmlUtils.parseXml(reportHandler, new FileInputStream(sourceReport)).failures();
```
schlosna, line 59: using try-with-resources may also have shown useful error on cleanup
```
        try (InputStream input = new FileInputStream(sourceReport)) {
            return XmlUtils.parseXml(reportHandler, input).failures();
        } catch (Exception e) {
            throw new RuntimeException(String.format("Failed to parse failures XML: %s", sourceReport), e);
        }
```

PR #832, opened 2019-09-15
Fix #831: Error Prone StringBuilderConstantParameters

### LINKED ISSUE #831: StringBuilder with constant number of args can be replaced with concatenation
## What happened?

StringBuilder usage with a constant number of arguments can be replaced with simple string concatenation (`+` operator).

See the discussion on https://github.com/palantir/conjure-java/pull/555

Compiling for a java 8 target results in the same bytecode, and applications which target java 9+ take advantage of [JEP 280](https://openjdk.java.net/jeps/280) for a more efficient codepath.

```java
String value = new StringBuilder().append("foo").append(bar).toString();
```

Can be replaced with

```java
String value = "foo" + bar;
```

We can build an errorprone check with either a suggested fix, or a refaster rule to automatically replace these in existing code.

carterkozak:
StringBuilder with a constant number of parameters should be replaced by simple concatenation.
The Java compiler (jdk8) replaces concatenation of a constant number of arguments with a StringBuilder, while jdk 9+ take advantage of JEP 280 (https://openjdk.java.net/jeps/280) to efficiently pre-size the result for better performance than a StringBuilder.

==COMMIT_MSG==
Error Prone StringBuilderConstantParameters
==COMMIT_MSG==


iamdanfox (approved):

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StringBuilderConstantParameters.java
```diff
@@ -0,0 +1,164 @@
...
+    private static final class StringBuilderVisitor
+            extends SimpleTreeVisitor<Optional<List<ExpressionTree>>, VisitorState> {
+        private static final StringBuilderVisitor INSTANCE = new StringBuilderVisitor();
+
+        private StringBuilderVisitor() {
+            super(Optional.empty());
+        }
+
+        @Override
+        public Optional<List<ExpressionTree>> visitNewClass(NewClassTree node, VisitorState state) {
```
ferozco, line 121: What is the semantic difference between an absent list and present empty list?

carterkozak: Documented here: https://github.com/palantir/gradle-baseline/pull/832/files#diff-515ce9d314ac0796483b739fc5035b83R104-R110

Optional of empty list will result from:
```java
return new StringBuilder().toString();
```

Which should be replaced with:
```java
return "";
```

Where empty optional tells us this invocation doesn't match, and we should return `Description.NO_MATCH`.

PR #835, opened 2019-09-16
Fix: Long try-with-resources adheres to palantir style

dansanduleac:
## Before this PR

Long try-with-resources statements would be aligned such that the first statement would be on a new line.

## After this PR
==COMMIT_MSG==
Long try-with-resources statements are now aligned such that the first assignment stays on the first line.
==COMMIT_MSG==

FYI how to construct an aligment mask: https://github.com/eclipse/eclipse.jdt.core/blob/master/org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/DefaultCodeFormatterOptions.java#L41-L95

### DIFF
--- a/changelog/@unreleased/pr-835.v2.yml
+++ b/changelog/@unreleased/pr-835.v2.yml
@@ -1,0 +1,6 @@
+type: fix
+fix:
+  description: Long try-with-resources statements are now aligned such that the first
+    assignment stays on the first line.
+  links:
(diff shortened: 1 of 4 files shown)


iamdanfox (approved):
🎉🎉🎉🎉

carterkozak:
very nice!

ferozco:
👍

PR #841, opened 2019-09-17
ErrorProne PreferAssertj migration to AssertJ from org.junit.Assert

carterkozak:
## After this PR
==COMMIT_MSG==
ErrorProne PreferAssertj migration to AssertJ from org.junit.Assert
==COMMIT_MSG==

## Possible downsides?
There are a few cases this doesn't cover, `Assert.assertThat(T, Matcher<T>)` being the most common. I did not enable this check by default because it doesn't add a dependency on assertj, and I want to avoid blocking baseline upgrades in modules which don't already depend on assertj.
This might be a candidate for a separate library or module since it's mostly a one-time migration, however it's also helpful if we can fail when folks attempt to use `org.junit.Assert` methods again.

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferAssertj.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferAssertj.java
@@ -57,8 +57,6 @@ import java.util.function.BiConsumer;
         summary = "Prefer AssertJ fluent assertions")
 public final class PreferAssertj extends BugChecker implements BugChecker.MethodInvocationTreeMatcher {
 
-    private static final String MESSAGE = "Prefer AssertJ fluent assertions";
-
     private static final String ASSERTJ_ASSERTIONS = "org.assertj.core.api.Assertions";
     private static final String ASSERT_THAT = "assertThat";
     private static final String ASSERTJ_ASSERT_THAT = ASSERTJ_ASSERTIONS + '.' + ASSERT_THAT;
@@ -267,7 +265,6 @@ public final class PreferAssertj extends BugChecker implements BugChecker.Method
         }
         if (FAIL_DESCRIPTION.matches(tree, state) || FAIL.matches(tree, state)) {
             return buildDescription(tree)
-                    .setMessage(MESSAGE)
                     .addFix(SuggestedFix.builder()
                             .removeStaticImport("org.junit.Assert.fail")
                             .addStaticImport(ASSERTJ_ASSERTIONS + ".fail")
@@ -321,9 +318,7 @@ public final class PreferAssertj extends BugChecker implements BugChecker.Method
             } else {
                 // Does not fix assertArrayEquals(double[], double[], double)
                 // or assertArrayEquals(float[], float[], float)
-                return buildDescription(tree)
-                        .setMessage(MESSAGE)
-                        .build();
+                return describeMatch(tree);
             }
         }
         if (ASSERT_NOT_EQUALS_CATCHALL.matches(tree, state)) {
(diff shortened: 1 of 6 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferAssertj.java
```diff
@@ -0,0 +1,406 @@
...
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "PreferAssertj",
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
+        severity = BugPattern.SeverityLevel.SUGGESTION,
+        summary = "Prefer AssertJ fluent assertions")
+public final class PreferAssertj extends BugChecker implements BugChecker.MethodInvocationTreeMatcher {
+
+    private static final String MESSAGE = "Prefer AssertJ fluent assertions";
```
dansanduleac: not sure if this message is necessary if it's the same as the BugPattern's summary, the fix should inherit that one.

dansanduleac: setMessage

> Set a custom error message for this {@code Description}. The custom message will be used
> instead of the summary field as the text for the diagnostic message.

carterkozak: Oh very nice!

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferAssertj.java
```diff
@@ -0,0 +1,398 @@
...
+ * existing assertions to AssertJ without losing any information. It's an explicit non-goal to take advantage
+ * of better assertions provided by AssertJ, because those should be implemented in such a way to improve poor
+ * uses of AssertJ as well, which may run after the suggested fixes provided by this checker.
+ *
+ * Remaining work: This doesn't handle Assert.assertThat(value, matcher). We should be able to handle common
+ * cases relatively easily.
+ */
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "PreferAssertj",
```
iamdanfox, line 50: Do we need to add this to BaselineErrorProneExtension's `DEFAULT_PATCH_CHECKS`  to get the excavator PRs?

carterkozak: Yes, if we want this applied by default. I noted in the description that I didn't add this to the default patch checks because it doesn't add a dependency on assertj, however the more I think about this, the more I like the idea of running this by default and pushing maintainers to add standard test dependencies to their projects.
Thoughts?

iamdanfox: I'd just whack it on.

I think in the vast majority of cases it would merge fine because most people have assertj dependencies already... if a repo has a genuine reason why they can't use assertj, they can always turn off the check??

ferozco (commented):
Did you try to write refaster rules to handle the migration? This seems like an ideal use case

dansanduleac:
@ferozco indeed those can probably do the job, but I think the issue with them is that they don't prevent you from accidentally regressing.
An errorprone rule on the other hand will actually fail compilation.

carterkozak:
They're also a bit less expressive, so it's harder to catch specific cases. assertEquals, assertNotEquals, and assertArrayEquals, for example, have a lot of overloads, so it was more difficult.

ferozco:
I think when rewriting code we should to prefer refaster rules vs. errorprone fixes whenever possible since they are simpler and easier to maintain. We can then have short errorprone rules to ensure repos don't regress. In this case however, it may be more trouble than its worth

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferAssertj.java
carterkozak: Done :-)

ferozco:
👍

PR #837, opened 2019-09-16
New checkJUnitDependencies task

iamdanfox:
## Before this PR

It's possible to get into a state where you have some JUnit5 and some JUnit4 tests in the same project. In this state, it's easy to accidentally misconfigure dependencies, leading to a state where tests may _silently_ not run)!!

This actually happened [on gradle-revapi](https://github.com/palantir/gradle-revapi/pull/9/commits/1c29a04c1b2ad727c95124f1df62e84080c6155e), leading to none of our JUnit5 tests actually running 🔥🔥🔥

## After this PR
==COMMIT_MSG==
A new checkJUnitDependencies detects misconfigured JUnit set-ups which could result in some tests silently not running.
==COMMIT_MSG==

BEFORE MERGING
- [x] actually implement the detecting bit of this thing
- [x] tests

## Possible downsides?
- not sure how to cache this guy

PR #843, opened 2019-09-17
new JUnit5SuiteMisuse error-prone check

iamdanfox:
## Before this PR

I started building some defensiveness in another PR (https://github.com/palantir/gradle-baseline/pull/837) but it turns out it didn't actually catch the problem that @delta003 noticed.

His case was that he still had a floating `@RunWith(Suite.class)` class (Junit4) but it _referenced_ classes containing JUnit5 tests, leading to tests _silently not running_ 🔥🔥🔥

## After this PR
==COMMIT_MSG==
ErrorProne will now detect dangerous usage of `@RunWith(Suite.class)` that references JUnit5 classes, as this can cause tests to silently not run!
==COMMIT_MSG==

## Possible downsides?
- I wonder if accumulating things in a hashmap will end up OOMing??


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/JUnit5SuiteMisuse.java
```diff
@@ -0,0 +1,117 @@
...
+        severity = BugPattern.SeverityLevel.ERROR,
+        summary = "Referencing JUnit5 tests from JUnit4 Suites will silently not work")
+public final class JUnit5SuiteMisuse extends BugChecker
+        implements BugChecker.ClassTreeMatcher, BugChecker.AnnotationTreeMatcher {
+
+    private static final long serialVersionUID = 1L;
+
+    // We remember classes and validate them later because error-prone doesn't let us arbitrarily explore classes we
+    // discover when reading the @SuiteClasses annotation.
+    private static final Set<Type.ClassType> knownJUnit5TestClasses = new HashSet<>();
```
ferozco, line 52: This seems pretty weird, but I can't think of another alternative.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/JUnit5SuiteMisuse.java
```diff
@@ -0,0 +1,117 @@
...
+                return buildDescription(tree)
+                        .setMessage("Don't reference JUnit5 test classes from JUnit4 SuiteClasses annotation")
+                        .build();
+            }
+        }
+
+        return Description.NO_MATCH;
+    }
+
+    private static List<JCTree.JCFieldAccess> getReferencedClasses(AnnotationTree tree, VisitorState state) {
```
ferozco: `List<ClassType>` instead?

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StringBuilderConstantParameters.java
```diff
@@ -45,6 +45,8 @@
 @AutoService(BugChecker.class)
 @BugPattern(
         name = "StringBuilderConstantParameters",
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
```
ferozco, line 48: unrelated but nice

ferozco (commented):
Could we test this out against some repos just to double check everythings fine

ferozco:
👍

iamdanfox:
Verified on scheduler, this would catch the specific problem that @delta003 had:

```
./gradlew scheduler:integrationTestClasses

> Task :scheduler:compileIntegrationTestJava
/Volumes/git/scheduler/scheduler/src/integrationTest/java/com/palantir/scheduler/resource/SchedulerResourceIntegrationTest.java:46: error: [JUnit5SuiteMisuse] Class uses JUnit5 tests but is referenced by a JUnit4 SuiteClasses
public final class SchedulerResourceIntegrationTest extends SchedulerBaseIntegrationTest {
             ^
1 error

> Task :scheduler:compileIntegrationTestJava FAILED
```

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/JUnit5SuiteMisuse.java
iamdanfox: Agreed - I tried a bunch of things before landing on this. You can traverse up your class hierarchy nice and happily, and you can do whatever you like inside a file, I think they enforce that you can't go spelunking around arbitrary classes to ensure things are performant.

PR #850, opened 2019-09-18
PreferAssertj provides better hamcrest replacements

carterkozak:
Avoids using the less ergonomic HamcrestCondition in common
cases.

## Before this PR
```java
assertThat(value).describedAs("desc").is(new HamcrestCondition<>(is("str")));
```

## After this PR
```java
assertThat(value).isEqualTo("str");
```
==COMMIT_MSG==
PreferAssertj provides better replacements fixes
==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferAssertj.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferAssertj.java
@@ -513,10 +513,15 @@ public final class PreferAssertj extends BugChecker implements BugChecker.Method
                         .named("notNullValue")
                         .withParameters("java.lang.Class"));
 
-        private static final Matcher<ExpressionTree> HAS_ITEM = MethodMatchers.staticMethod()
-                .onClassAny(MATCHERS)
-                .namedAnyOf("hasItem", "hasItemInArray")
-                .withParameters(Object.class.getName());
+        private static final Matcher<ExpressionTree> CONTAINS = Matchers.anyOf(
+                MethodMatchers.staticMethod()
+                        .onClassAny(MATCHERS)
+                        .namedAnyOf("hasItem", "hasItemInArray")
+                        .withParameters(Object.class.getName()),
+                MethodMatchers.staticMethod()
+                        .onClassAny(MATCHERS)
+                        .named("containsString")
+                        .withParameters(String.class.getName()));
 
         // Note: cannot match array/vararg arguments
         private static final Matcher<ExpressionTree> HAS_ITEMS = MethodMatchers.staticMethod()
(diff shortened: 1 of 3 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferAssertj.java
```diff
@@ -286,42 +289,61 @@ public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState
         if (ASSERT_EQUALS_FLOATING.matches(tree, state)) {
             return withAssertThat(tree, state, (assertThat, fix) -> fix
                     .addStaticImport("org.assertj.core.api.Assertions.within")
-                    .replace(tree, assertThat + "(" + argSource(tree, state, 1) + ").isCloseTo("
-                            + argSource(tree, state, 0) + ", within(" + argSource(tree, state, 2) + "))"));
+                    .replace(tree, assertThat + "(" + argSource(tree, state, 1)
```
ferozco: I think these would be much easier to read as format strings

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/PreferAssertjTests.java
```diff
@@ -988,63 +841,335 @@ public void fix_assertThatIntDescription() {
...
                         "Test.java",
                         "import static org.assertj.core.api.Assertions.assertThat;",
-                        "import static org.hamcrest.Matchers.is;",
+                        "import static org.hamcrest.Matchers.startsWith;",
                         "",
                         "import org.assertj.core.api.HamcrestCondition;",
                         "class Test {",
                         "  void foo(String value) {",
-                        "    assertThat(value).is(new HamcrestCondition<>(is(\"str\")));",
+                        "    assertThat(value).is(new HamcrestCondition<>(startsWith(\"str\")));",
```
ferozco: It would be great to automatically handle startsWith and contains. i.e. `assertThat(value).startsWith("str")`

iamdanfox: Orrr we could just do a refaster layer that will turn these hamcresty things into idomatic assertJ?

carterkozak: If it works, however there are a few places where the HamcrestCondition results in code that doesn't compile: https://github.com/joel-costigliola/assertj-core/issues/1607

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferAssertj.java
carterkozak: Good call, updated.

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/PreferAssertjTests.java
carterkozak: Updated to support startsWith, endsWIth, and contains.

ferozco:
@carterkozak could you try out the new replacements on tritium just to double check they work as expected? Then happy to merge this in

carterkozak:
Tritium is 100% assertj already, I'll find another project.

ferozco:
👍

PR #851, opened 2019-09-18
Some low-effort AssertJ refasters

iamdanfox:
## Before this PR

People can end up writing assertions that are _correct_, but don't give you as much debugging info as possible when they fail.

cc @carterkozak 

## After this PR
==COMMIT_MSG==
Some AssertJ assertions can now be automatically replaced with more idiomatic ones using refaster.
==COMMIT_MSG==

## Possible downsides?
- if nobody every actually makes these mistakes then we have some pointless code in the repo, and will spent excavator CPU cycles matching against them...

### DIFF
--- a/baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/AssertjCollectionHasSizeExactly.java
+++ b/baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/AssertjCollectionHasSizeExactly.java
@@ -21,7 +21,6 @@ import static org.assertj.core.api.Assertions.assertThat;
 import com.google.errorprone.refaster.annotation.AfterTemplate;
 import com.google.errorprone.refaster.annotation.BeforeTemplate;
 import java.util.Collection;
-import java.util.Optional;
 
 public final class AssertjCollectionHasSizeExactly<T> {
 
--- a/baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/AssertjCollectionHasSizeExactly.java
+++ b/baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/AssertjCollectionHasSizeExactly.java
@@ -23,7 +23,7 @@ import com.google.errorprone.refaster.annotation.BeforeTemplate;
 import java.util.Collection;
 import java.util.Optional;
 
-public final class AssertjCollectionHasSize<T> {
+public final class AssertjCollectionHasSizeExactly<T> {
 
     @BeforeTemplate
     void bad1(Collection<T> things, int size) {
--- a/baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/AssertjCollectionIsEmpty.java
+++ b/baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/AssertjCollectionIsEmpty.java
@@ -34,26 +34,31 @@ public final class AssertjCollectionIsEmpty<T> {
 
     @BeforeTemplate
     void bad2(Collection<T> things) {
-        assertThat(things.size()).isEqualTo(0);
+        assertThat(things.isEmpty()).isTrue();
     }
 
     @BeforeTemplate
     void bad3(Collection<T> things) {
-        assertThat(things).isEqualTo(Collections.emptyList());
+        assertThat(things.size() == 0).isTrue();
     }
 
     @BeforeTemplate
     void bad4(Collection<T> things) {
-        assertThat(things).isEqualTo(Collections.emptySet());
+        assertThat(things).isEqualTo(Collections.emptyList());
     }
 
     @BeforeTemplate
     void bad5(Collection<T> things) {
-        assertThat(things).isEqualTo(ImmutableList.of());
+        assertThat(things).isEqualTo(Collections.emptySet());
     }
 
     @BeforeTemplate
     void bad6(Collection<T> things) {
+        assertThat(things).isEqualTo(ImmutableList.of());
+    }
+
+    @BeforeTemplate
+    void bad7(Collection<T> things) {
         assertThat(things).isEqualTo(ImmutableSet.of());
     }
 
--- a/baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/AssertjCollectionIsEmpty.java
+++ b/baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/AssertjCollectionIsEmpty.java
@@ -18,9 +18,12 @@ package com.palantir.baseline.refaster;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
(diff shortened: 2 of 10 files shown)


baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/AssertjCollectionIsEmpty.java
```diff
@@ -0,0 +1,64 @@
...
+import com.google.errorprone.refaster.annotation.AfterTemplate;
+import com.google.errorprone.refaster.annotation.BeforeTemplate;
+import java.util.Collection;
+import java.util.Collections;
+
+public final class AssertjCollectionIsEmpty<T> {
+
+    @BeforeTemplate
+    void bad1(Collection<T> things) {
+        assertThat(things.size() == 0).isTrue();
```
carterkozak, line 32: Maybe one more case for `assertThat(things.isEmpty()).isTrue();

baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/AssertjCollectionIsNotEmpty.java
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    @BeforeTemplate
+    void bad2(Collection<T> things) {
+        assertThat(things.size() == 0).isFalse();
+    }
+
+    @BeforeTemplate
+    void bad3(Collection<T> things) {
+        assertThat(things.size()).isNotEqualTo(0);
```
carterkozak, line 42: Maybe another case for `assertThat(things.isEmpty()).isFalse();` and `assertThat(!things.isEmpty()).isTrue();`

baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/AssertjOptionalIsPresent.java
```diff
@@ -0,0 +1,36 @@
...
+
+package com.palantir.baseline.refaster;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import com.google.errorprone.refaster.annotation.AfterTemplate;
+import com.google.errorprone.refaster.annotation.BeforeTemplate;
+import java.util.Optional;
+
+public final class AssertjOptionalIsPresent<T> {
```
ferozco, line 25: add equality checks? `assertThat(foo).isEqualTo(Optional.of("foo")` -> `asserThat(foo).hasValue("foo")`

iamdanfox: I actually find the verbosity of the `.isEqualTo(Optional.of("foo"))` kinda reassuring... I think I could easily misinterpret what `hasValue` actually means...

baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/AssertjCollectionHasSizeExactly.java
```diff
@@ -0,0 +1,41 @@
...
+
+import com.google.errorprone.refaster.annotation.AfterTemplate;
+import com.google.errorprone.refaster.annotation.BeforeTemplate;
+import java.util.Collection;
+
+public final class AssertjCollectionHasSizeExactly<T> {
+
+    @BeforeTemplate
+    void bad1(Collection<T> things, int size) {
+        assertThat(things.size() == size).isTrue();
```
gatesn, line 29: How does this play with https://github.com/palantir/gradle-baseline/blob/develop/baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/CollectionsIsEmpty.java

baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/AssertjCollectionIsEmpty.java
```diff
@@ -0,0 +1,69 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
+import com.google.errorprone.refaster.annotation.AfterTemplate;
+import com.google.errorprone.refaster.annotation.BeforeTemplate;
+import java.util.Collection;
+import java.util.Collections;
+
+public final class AssertjCollectionIsEmpty<T> {
```
gatesn, line 28: Might be cleaner to make all these inner classes of an AssertJ refaster class. Keep the verbose but related stuff together

baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/AssertjCollectionHasSizeExactly.java
iamdanfox: I _think_ if this one refasters it to hasSize(0), the other one will get it on the next pass, so it'll be eventually consistent!

baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/AssertjCollectionIsEmpty.java
iamdanfox: Eh, all the files have the same prefix, I think they're pretty easy to navigate!

baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/AssertjCollectionHasSizeExactly.java
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    @BeforeTemplate
+    void bad2(Collection<T> things, int size) {
+        assertThat(things.size()).isEqualTo(size);
+    }
+
+    @AfterTemplate
+    void after(Collection<T> things, int size) {
+        assertThat(things).hasSize(size);
```
carterkozak, line 39: is there any way we can make this work with assertions that have descriptions without duplicating each refaster rule?

Example:
```java
assertThat(things.size()).describedAs("size was not updated").isEqualTo(size);
```
I would like that to be updated.

ferozco: I think what we want to do in all of these cases is to match against the return type of `assertThat(T)` instead so that chained calls are also fixed.
i.e.
```
@AfterTemplate
void after(ListAssert<T> assert, int size) {
	assert.hasSize(size);
}
```

iamdanfox:
@ferozco can you clarify the `do not merge` pls?

ferozco:
Spoke offline, happy to ship 👍

iamdanfox:
Thanks Felipe! Just wanted this to be a MVP approach - can always put more time into getting the describedAs / chains to work if we see them in the wild!

PR #852, opened 2019-09-18
Allow static imports of AssertJ and Mockito

pkoenig10:
## Before this PR
It's common to have a test utility project which is a test dependency in other projects. We currently allow static imports in tests (#240) but that does not apply to test utility projects since they are main sources.

This paper cut is aggravated by the `PreferAssertJ` check (#841) which static imports AssertJ methods. Doing this causes automated baseline upgrades to require manual intervention to successfully compile.

## After this PR
We allow static imports of AssertJ and Mockito methods.

### DIFF
--- a/changelog/@unreleased/pr-852.v2.yml
+++ b/changelog/@unreleased/pr-852.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Baseline now allows static imports of AssertJ and Mockito methods.
+  links:
+  - https://github.com/palantir/gradle-baseline/pull/852
--- a/gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
+++ b/gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
@@ -84,11 +84,13 @@
         <module name="AvoidStarImport"/> <!-- Java Style Guide: No wildcard imports -->
         <module name="AvoidStaticImport"> <!-- Java Style Guide: No static imports -->
             <property name="excludes" value="
+                com.google.common.base.Preconditions.*,
+                com.palantir.logsafe.Preconditions.*,
                 java.util.Collections.*,
                 java.util.stream.Collectors.*,
-                com.palantir.logsafe.Preconditions.*,
-                com.google.common.base.Preconditions.*,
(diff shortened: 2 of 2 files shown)


ferozco:
Given that the preferAsserJ migration is a onetime cost, I'm not sure if its worth widening the set of allowed static imports

pkoenig10:
This isn't just for the AssertJ migration. It's cumbersome to not be able to use static imports in test utility classes just because they are part of the main source set in a test utility project.

These classes aren't going to be imported in non-test code so what's the downside of allowing them to be statically imported? These classes are almost universally statically imported in test code currently.

carterkozak:
We allow static imports in test classes:
https://github.com/palantir/gradle-baseline/blob/develop/gradle-baseline-java-config/resources/checkstyle/checkstyle-suppressions.xml#L13

schlosna:
@pkoenig10 can you use [Gradle 5.6+ test fixtures](https://docs.gradle.org/5.6/release-notes.html#test-fixtures-for-java-projects)?

pkoenig10:
@carterkozak this is utility class used in tests, but it is not in a test source set. It is in a separate project that is imported as a test dependency in multiple other projects.

@schlosna test fixtures sound like exactly what I'm looking for, thanks for pointing that out!

I still think there is value in allowing these classes to be statically imported. Not everyone is going to have the ability or resources to migrate to test fixtures and the downside of allowing these classes to be imported is low.

iamdanfox (approved):
I just found an excavator assertj PR which was blocked by this checkstyle rule, because they'd defined a TestRule in their foo-ete project, in the src/main/java directory.

I think whitelisting the assertThat one is low regret

pkoenig10:
@dfox I edited the changelog-app comment but it doesn't appear to be generating a changelog. Maybe you can do it?

iamdanfox:
I think it's because this PR came from your personal fork, and changelog-app's repo-specific doesn't have perms on your fork sadly.  We've recently granted all palantirians write access to our OSS repos to workaround this.

carterkozak:
changelog bot... bump?

pkoenig10:
Moved to #915

PR #854, opened 2019-09-18
strictUnusedVariable only support _ suppression prefix

ferozco:
## Before this PR
We allowed users to suppress `StrictUnusedVariable` with `unused` or `_` prefix.

## After this PR
==COMMIT_MSG==
StrictUnusedVariable can only be suppressed with `_` prefix
==COMMIT_MSG==

We also will automatically migrate previous supressions with `unused` to `_`

## Possible downsides?
N/A


iamdanfox (approved):
Yep much nicer to be completely consistent

PR #855, opened 2019-09-18
Promote StrictUnusedVariable to error by default

ferozco:
## After this PR
==COMMIT_MSG==
StrictUnusedVariable is now an error by default
==COMMIT_MSG==

## Possible downsides?
We have a bunch of automation around this rule, so turning it into an error should just prevent people from shooting themselves in the foot


iamdanfox:
~When I tried this before, the compiler started spewing errors from Immutables-generated classes, because they have a constructor with an unused param. This is annoying because users can't actually fix them.~

```java
    private ImmutableFailure(ImmutableFailure original, String node, String message) {
                          // ^ unused!
        this.node = node;
        this.message = message;
    }
```

~Seems like our -XepDisableWarningsInGeneratedCode doesn't disable _errors_ in generated code.~

EDIT nvm looks like you've taken care of this already! https://github.com/palantir/gradle-baseline/pull/853

iamdanfox (approved):

PR #861, opened 2019-09-19
People have to opt-in to PreferAssertj patching

iamdanfox:
## Before this PR

Some repos are failing to merge 'latest baseline' refactoring PRs because it has huge assertj changes (like +2500lines or so).

I've pulled this out to a separate excavator PR that explicitly opts-in to the PreferAssertj check, so I think we can flip this to be opt-in.

## After this PR
==COMMIT_MSG==
The PreferAssertj refactoring will only be applied if you have explicitly opted in (e.g. using `baselineErrorProne { patchChecks += 'PreferAssertj' }`
==COMMIT_MSG==

## Possible downsides?
- more PRs to people's repos?

### DIFF
--- a/changelog/@unreleased/pr-861.v2.yml
+++ b/changelog/@unreleased/pr-861.v2.yml
@@ -1,0 +1,6 @@
+type: fix
+fix:
+  description: The PreferAssertj refactoring will only be applied if you have explicitly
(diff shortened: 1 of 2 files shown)


ferozco:
👍

carterkozak:
Thanks :+1:

PR #863, opened 2019-09-19
Formatter preserves NLS comments in multiline string constants

carterkozak:
==COMMIT_MSG==
Failing test for multiline string constant with NLS comments
==COMMIT_MSG==


iamdanfox:
Would it be better to check in the current borked behaviour with a comment explaining why it's bad? Then it's more convenient to iterate on when people branch off develop to tinker with the formatter?

carterkozak:
Removing the `<setting id="org.eclipse.jdt.core.formatter.join_wrapped_lines" value="false"/>` line makes it more sane in some ways, but not great for code that already has an operator in the right place.

iamdanfox (approved):

carterkozak:
:+1:

carterkozak:
bulldozer nudge

PR #864, opened 2019-09-19
Errorprone ignores all generated code

ferozco:
## Before this PR
We only ignored code generated in the build directory. This caused issues when interacting with conjure and other code generators

## After this PR
==COMMIT_MSG==
Error prone will correctly ignore all source files in the build directory and in any generated source directory
==COMMIT_MSG==

## Possible downsides?
N/A

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
@@ -178,7 +178,8 @@ public final class BaselineErrorProne implements Plugin<Project> {
 
         errorProneOptions.setEnabled(true);
         errorProneOptions.setDisableWarningsInGeneratedCode(true);
-        errorProneOptions.setExcludedPaths(project.getBuildDir().getAbsolutePath() + "/.*");
+        errorProneOptions.setExcludedPaths(
+                String.format("%s/(build|src/generated.*)/.*", project.getProjectDir().getAbsolutePath()));
         errorProneOptions.check("EqualsHashCode", CheckSeverity.ERROR);
         errorProneOptions.check("EqualsIncompatibleType", CheckSeverity.ERROR);
         errorProneOptions.check("StreamResourceLeak", CheckSeverity.ERROR);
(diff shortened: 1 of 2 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
```diff
@@ -178,7 +178,8 @@ private static void configureErrorProneOptions(
 
         errorProneOptions.setEnabled(true);
         errorProneOptions.setDisableWarningsInGeneratedCode(true);
-        errorProneOptions.setExcludedPaths(project.getBuildDir().getAbsolutePath() + "/.*");
+        errorProneOptions.setExcludedPaths(
+                String.format("%s/(build|src/generated.*)/.*", project.getProjectDir().getAbsolutePath()));
```
iamdanfox: I wonder if this absolute path is gonna affect the build-cache relocatability of these compile tasks??

iamdanfox:
Can you run a pTML and verify this behaves on that red template-witchcraft PR?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
carterkozak: Thoughts on including `.*/generated/.*`? It's possible applications have packages named `generated`, but that doesn't make a great deal of sense for non-generated code.

ferozco:
I tested locally against template-witchcraft and codex-indexer and it seemed to work

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
ferozco: good call, updating it to be relative path

ferozco: I'd prefer to be more strict to begin with since being overly lenient could lead to false negatives.

ferozco:
Unrelated, but looks like we'll need to roll out the changelog-app config to all our OSS repos too

iamdanfox:
Could we verify this also doesn't freak out for jmh generated code (in sls-logging) and also assertj generated code? (in the foundry-audit project)

iamdanfox (approved):

ferozco:
Confirmed to work with both!

PR #869, opened 2019-09-19
Update string concatenation format from `80` to `16`

carterkozak:
This is 'compact' mode, so the formatter will not aggressively
split lines unless line length is reached. This counteracts
the option to disable joining wrapped lines which would otherwise
re-merge wrapped lines in a better format.

## After this PR
==COMMIT_MSG==
Update string concatenation format from `80` to `16`
==COMMIT_MSG==


iamdanfox (approved):
Nice!

PR #870, opened 2019-09-19
StrictUnusedVariable maintains side effects

ferozco:
## Before this PR
The original Error prone rule maintained two seperate suggested fixes for preserving side effects and removing them, but would always prefer removing side effects.

## After this PR
==COMMIT_MSG==
When removing unused variables, `StrictUnusedVariable` will no longer delete method calls as these may have side effects
==COMMIT_MSG==

## Possible downsides?
The way that side effects are detected is very conservative and this change may leave behind unnecessary initializers

### DIFF
--- a/baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/StrictUnusedVariableTest.java
+++ b/baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/StrictUnusedVariableTest.java
@@ -146,6 +146,31 @@ public class StrictUnusedVariableTest {
     }
 
     @Test
+    public void side_effects_are_preserved() {
+        refactoringTestHelper
+                .addInputLines(
+                        "Test.java",
+                        "class Test {",
+                        "  private static int _field = 1;",
+                        "  public static void privateMethod() {",
+                        "    Object foo = someMethod();",
+                        "  }",
+                        "  private static Object someMethod() { return null; }",
+                        "}")
+                .addOutputLines(
+                        "Test.java",
+                        "class Test {",
+                        "  private static int field = 1;",
+                        "  public static void privateMethod() {",
+                        "    someMethod();",
+                        "  }",
+                        "  private static Object someMethod() { return null; }",
+                        "  }",
+                        "}"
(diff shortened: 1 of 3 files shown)


dansanduleac (approved):

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/StrictUnusedVariableTest.java
```diff
@@ -145,6 +145,31 @@ public void fails_suppressed_but_used_variables() {
...
+                        "  private static Object someMethod() { return null; }",
+                        "}")
+                .addOutputLines(
+                        "Test.java",
+                        "class Test {",
+                        "  private static int field = 1;",
+                        "  public static void privateMethod() {",
+                        "    someMethod();",
+                        "  }",
+                        "  private static Object someMethod() { return null; }",
```
dansanduleac, line 167: seems like the test is rewriting this method to be multi-line for some reason...?
or maybe it's comparing AST _first_ and failing on that before comparing text

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/StrictUnusedVariableTest.java
```diff
@@ -145,6 +145,31 @@ public void fails_suppressed_but_used_variables() {
...
+                        "  private static int _field = 1;",
+                        "  public static void privateMethod() {",
+                        "    Object foo = someMethod();",
+                        "  }",
+                        "  private static Object someMethod() { return null; }",
+                        "}")
+                .addOutputLines(
+                        "Test.java",
+                        "class Test {",
+                        "  private static int field = 1;",
```
dansanduleac: this should remain `_field`

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/StrictUnusedVariableTest.java
dansanduleac: nvm, hadn't noticed the extra `}`

iamdanfox (approved):
Extra approved

PR #874, opened 2019-09-20
Avoid ambiguity in assertions on iterable maps

carterkozak:
==COMMIT_MSG==
Avoid ambiguity in assertions on iterable maps
==COMMIT_MSG==


ferozco:
👍

PR #876, opened 2019-09-22
Improve AssertjCollectionIsEmpty and add test cases

carterkozak:
the `isEqualTo(Collections.emptyList())` and `ImmutableList.of()` cases
do not appear to work.

## Before this PR
no tests. fewer fix cases.

## After this PR
==COMMIT_MSG==
Improve AssertjCollectionIsEmpty and add test cases
==COMMIT_MSG==


baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/AssertjCollectionIsEmpty.java
```diff
@@ -39,36 +41,42 @@ void bad2(Collection<T> things) {
 
     @BeforeTemplate
     void bad3(Collection<T> things) {
-        assertThat(things.size() == 0).isTrue();
```
carterkozak, line 42: exact duplicate of `bad1`

iamdanfox (approved):
Mmm looks like I really should have just written the test before merging! Thanks @carterkozak

PR #880, opened 2019-09-23
Less strict junit dependency check

ferozco:
## Before this PR
We would fail builds if we found junit dependencies on the testCompileClasspath without any corresponding tests. This was a little aggressive and blocked roll out of baseline.

We also ran into some issues with how we found test task and JMH since JMH creates a task named after the source set

## After this PR
==COMMIT_MSG==
Allow junit4 dependencies to exist without junit4 tests
==COMMIT_MSG==

## Possible downsides?
Some users could continue to write junit4. However this would only pass CI if the project still had the vintage engine, so the test would still be executed

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckJUnitDependencies.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckJUnitDependencies.java
@@ -109,16 +109,7 @@ public class CheckJUnitDependencies extends DefaultTask {
                                 + "'org.junit.jupiter:junit-jupiter' dependency "
                                 + "because tests use JUnit4 and useJUnitPlatform() is not enabled.");
             }
-        } else {
-            String compileClasspath = ss.getCompileClasspathConfigurationName();
-            boolean compilingAgainstOldJunit = hasDep(compileClasspath, CheckJUnitDependencies::isJunit4);
-            Preconditions.checkState(
-                    !compilingAgainstOldJunit,
-                    "Extraneous dependency on JUnit4 (no test mentions JUnit4 classes). Please exclude "
-                            + "this from compilation to ensure developers don't accidentally re-introduce it, e.g.\n\n"
-                            + "    configurations." + compileClasspath + ".exclude module: 'junit'\n\n");
         }
-
         // sourcesets might also contain Spock classes, but we don't have any special validation for these.
     }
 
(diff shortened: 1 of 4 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckJUnitDependencies.java
```diff
@@ -109,16 +109,7 @@ private void validateSourceSet(SourceSet ss, Test task) {
                                 + "'org.junit.jupiter:junit-jupiter' dependency "
                                 + "because tests use JUnit4 and useJUnitPlatform() is not enabled.");
             }
-        } else {
-            String compileClasspath = ss.getCompileClasspathConfigurationName();
-            boolean compilingAgainstOldJunit = hasDep(compileClasspath, CheckJUnitDependencies::isJunit4);
-            Preconditions.checkState(
```
iamdanfox, line 115: Rather than deleting this entirely, could we just emit it as a `warn` or `info` log line?

ferozco: I agree you don't want devs to write junit4 tests, but I don't think thats solved by enforcing that junit and junit-jupiter don't appear on the same classpath without the vintage engine. The earlier condition to check that there are no junit4 tests without the vintage engine is sufficient

ferozco: For example, dropwizard-testing contains both junit4 rules and junit5 extensions in the same JAR

iamdanfox: This is going for a slightly different goal though, this is trying to say once you think you've fully migrated to JUnit5, there should be a nice ratchet so that new devs don't start adding junit4 stuff back agian.

note that this only excludes it from _compilation_, it can still exist at runtime.

ferozco: right, conceivably once you've fully migrated to junit5 you would no longer have the vintage engine and so if a user added a junit4 test the check would fail with `Tests may be silently not running! Some tests still use JUnit4, but Gradle has...`

iamdanfox: I was intending this to be the thing that encourages people to delete their `vintage engine` dependency though, otherwise I think these will litter our projects forever!

ferozco: I've updated the PR to log the message at info

iamdanfox (approved):
Thanks @ferozco

PR #883, opened 2019-09-24
PreferAssertj supports migration of zero-delta floating point array asserts

carterkozak:
==COMMIT_MSG==
PreferAssertj supports migration of zero-delta floating point array asserts
==COMMIT_MSG==


ferozco:
👍

PR #887, opened 2019-09-24
AssertJ Refaster fixes use static `assertThat` imports

carterkozak:
==COMMIT_MSG==
AssertJ Refaster fixes use static `assertThat` imports
==COMMIT_MSG==


ferozco:
👍

iamdanfox (approved):

PR #881, opened 2019-09-23
Refaster for AssertJ isZero/isNotZero/isOne and collections

carterkozak:
Examples using specific AssertJ asserts without `assertThat`.

## After this PR
==COMMIT_MSG==
Refaster for AssertJ isZero/isNotZer/isOne and collections
==COMMIT_MSG==

PR #889, opened 2019-09-24
One more assertj refaster

iamdanfox:
## Before this PR
I saw a codebase with exactly this before state, which can easily be replaced with a version that produces a nicer error message!

## After this PR
==COMMIT_MSG==
refaster replacement for assertj hasSize(foo.size) -> hasSameSizeAs
==COMMIT_MSG==

## Possible downsides?
n/a

### DIFF
--- a/baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/AssertjCollectionHasSameSizeAs.java
+++ b/baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/AssertjCollectionHasSameSizeAs.java
@@ -1,0 +1,38 @@
+/*
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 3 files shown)


baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/AssertjCollectionHasSameSizeAs.java
```diff
@@ -0,0 +1,38 @@
...
+import com.google.errorprone.refaster.annotation.AfterTemplate;
+import com.google.errorprone.refaster.annotation.BeforeTemplate;
+import com.google.errorprone.refaster.annotation.UseImportPolicy;
+import java.util.List;
+import org.assertj.core.api.IterableAssert;
+
+public final class AssertjCollectionHasSameSizeAs<T,U> {
+
+    @BeforeTemplate
+    IterableAssert<T> before(IterableAssert<T> assertInProgress, List<U> expected) {
```
carterkozak: List can be Collection

iamdanfox: Iterable doesn't have the .size() method, so I used Collection initially, but I can't even get my test to pass with this :/

Currently dying in RefasterScanner - think I may have borked my IDE a bit. Might try again tomorrow!

image

carterkozak: I think refaster requires an exact match on the assert type returned by AssertThat:

This should work:

```java
@BeforeTemplate
ListAssert<T> before(ListAssert<T> assertInProgress, Collection<U> expected) {
    return assertInProgress.hasSize(expected.size());
}
```

However it won't apply to collections or iterable targerrs

carterkozak: Based on #881 it may or may not work in java 11...

iamdanfox:
Closing out because I can't get this to work and it doesn't seem _that_ high value.

PR #891, opened 2019-09-25
AssertJ refaster migrations support string descriptions

carterkozak:
==COMMIT_MSG==
AssertJ refaster migrations support string descriptions
==COMMIT_MSG==


iamdanfox:
I wonder if we should break out a standalone repo of 'assertj best practises / auto-refactoring' or something?? I feel like single-purpose things (e.g. nullaway) are often easier for other OSS repos to adopt rather than our conglomeration of baseline plugins...

iamdanfox (approved):

PR #896, opened 2019-09-26
BaselineFormat task exclude generated code on Windows

carterkozak:
## Before this PR
Running the `format` task on windows resulted in modification to generated files.

## After this PR
==COMMIT_MSG==
BaselineFormat task exclude generated code on Windows
==COMMIT_MSG==

Difficult to test here.


ferozco:
nice 👍

PR #899, opened 2019-09-27
Don't apply com.palantir.baseline-versions automatically

iamdanfox:
## Before this PR

We have been discouraging people from using nebula-dependency recommender in favour of GCV for a long time. Most repos now have a `com.palantir.baseline-versions.disable=true` line in their versions.props, and occasionally people still have trouble with this.

## After this PR
==COMMIT_MSG==
nebula-dependency-recommender is no longer applied by default. If you have not yet migrated to GCV, you need to explicitly depend on `com.netflix.nebula:nebula-dependency-recommender` and `apply plugin: 'com.palantir.baseline-versions'`.
==COMMIT_MSG==

## Possible downsides?

- anyone who was relying on nebula-dependency-recommender for some other purpose might stop taking baseline upgrades until they manually pull it in. Seems like that's about 60 repos:

image

### DIFF
--- a/gradle-baseline-java/build.gradle
+++ b/gradle-baseline-java/build.gradle
@@ -10,7 +10,7 @@ dependencies {
     compile 'com.diffplug.spotless:spotless-plugin-gradle'
     compile 'com.google.errorprone:error_prone_refaster'
     compile 'com.google.guava:guava'
-    compile 'com.netflix.nebula:nebula-dependency-recommender'
+    compileOnly 'com.netflix.nebula:nebula-dependency-recommender'
     compile 'com.palantir.configurationresolver:gradle-configuration-resolver-plugin'
     compile 'net.ltgt.gradle:gradle-errorprone-plugin'
     compile 'org.apache.maven.shared:maven-dependency-analyzer'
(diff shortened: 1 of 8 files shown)


iamdanfox:
@ferozco before we cut a major rev of gradle-baseline, I'd like to get this one in too!

gradle-baseline-java/build.gradle
```diff
@@ -10,7 +10,7 @@ dependencies {
     compile 'com.diffplug.spotless:spotless-plugin-gradle'
     compile 'com.google.errorprone:error_prone_refaster'
     compile 'com.google.guava:guava'
-    compile 'com.netflix.nebula:nebula-dependency-recommender'
+    compileOnly 'com.netflix.nebula:nebula-dependency-recommender'
```
ferozco, line 13: you may want to update the readme so that users know to bring their own nebula-dependency-recommender

ferozco:
👍

ferozco:
@iamdanfox with #906 we were able to avoid raising the minimum version of gradle so we don't need to cut a major rev. Perhaps we put this PR on hold  until we get to a situation that does require a break

PR #902, opened 2019-09-27
keep spotless from eagerly configuring all tasks.

esword:
## Before this PR
Including the baseline-format plugin would trigger eager configuration of tons of tasks due to a problem with spotless (https://github.com/diffplug/spotless/issues/444)

## After this PR
This applies the recommended fix until such time as the spotless plugin fixes the source issue.


carterkozak:
Any way we can test for this in `BaselineFormatTest`?

iamdanfox:
Good idea - you could use `register` a task that throws during construction? I think that would fail if spotless does everything eagerly and pass when this PR merges!

iamdanfox (approved):

PR #903, opened 2019-09-27
Revert gradle upgrade until we can fix the `compileRefaster` task

carterkozak:
Revert "Excavator: Upgrades gradle wrapper to the latest version (#804)"

This reverts commit 831e40d67482467fb75e137637d55af5ecfb3be9.

## Before this PR
```
-----------
* What went wrong:
Execution failed for task ':module:compileRefaster'.
> no source files
```

## After this PR
==COMMIT_MSG==
Revert gradle upgrade until we can fix the `compileRefaster` task
==COMMIT_MSG==

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -76,12 +76,11 @@ jobs:
 
   unit-test:
     docker: [{ image: 'circleci/openjdk:8u222-stretch-node' }]
-    resource_class: large
     environment:
       CIRCLE_TEST_REPORTS: /home/circleci/junit
       CIRCLE_ARTIFACTS: /home/circleci/artifacts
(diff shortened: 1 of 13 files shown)


robert3005:
That change was to make compilerefaster work on newer gradle. If you’re failing it means you’re likely on gradle older than 5.4

carterkozak:
@robert3005 WC is on 5.6.2 and produces this failure.

ferozco:
How did you see this error given that we haven't released the changes made in #804 yet

carterkozak:
`./gradlew publishToMavenLocal`. I test most error-prone/refaster changes on real projects before they're released for additional testing.

ferozco:
Ok, We're planning on releasing a major rev in part because of this change so I'd prefer to spend some time trying to fix the issue rather than reverting. I'll do some investigation and echo back with my finding/a fix

carterkozak:
Sounds good to me, thanks @ferozco!

PR #905, opened 2019-09-30
Add documentation about running refasterApply

mglazer:
## Before this PR

Docs about errorproneApply and refasterApply were hidden.

## After this PR
==COMMIT_MSG==
Added documentation for `errorProneApply` and `refasterApply`
==COMMIT_MSG==

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,6 +171,14 @@ Safe Logging can be found at [github.com/palantir/safe-logging](https://github.c
 - `JUnit5SuiteMisuse`: When migrating from JUnit4 -> JUnit5, classes annotated with `@RunWith(Suite.class)` are dangerous because if they reference any JUnit5 test classes, these tests will silently not run!
 - `PreferAssertj`: Prefer AssertJ fluent assertions.
 
+### Programmatic Application
+
+There exist a number of programmatic code modifiations available via [refaster](https://errorprone.info/docs/refaster). You can run these on your code to apply some refactorings automatically:
+
+```
+./gradlew compileJava compileTestJava -PrefasterApply -PerrorProneApply
+```
+
 ## com.palantir.baseline-checkstyle
 Checkstyle rules can be suppressed on a per-line or per-block basis. (It is good practice to first consider formatting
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -171,6 +171,14 @@ Safe Logging can be found at [github.com/palantir/safe-logging](https://github.c
 - `JUnit5SuiteMisuse`: When migrating from JUnit4 -> JUnit5, classes annotated with `@RunWith(Suite.class)` are dangerous because if they reference any JUnit5 test classes, these tests will silently not run!
 - `PreferAssertj`: Prefer AssertJ fluent assertions.
 
+### Programmatic Application
+
+There exist a number of programmatic code modifiations available via [refaster](https://errorprone.info/docs/refaster). You can run these on your code to apply some refactorings automatically:
+
+```
+./gradlew compileJava compileTestJava -PrefasterApply -PerrorProneApply
```
carterkozak, line 179: Are you sure this applies both? I thought they were mutually exclusive.

mglazer: Based off reading the code here:

https://github.com/palantir/gradle-baseline/blob/831e40d67482467fb75e137637d55af5ecfb3be9/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java#L212

I see nothing that makes them such.

iamdanfox: Internally we actually ended up making a `compileEverything` task which compiles all possible source sets, because running just `compileJava` and `compileTestJava` wouldn't get any jmh or integTest source sets... I wonder if we should just upstream this for convenience?

mglazer: I just tried `compileEverything` and it didn't work. Am I missing something?

iamdanfox: that's expected - we'd been hackily defining the `compileEverything` task just-in-time in an init.gradle file.

I'm suggesting that it might be time to do it in a non-hacky way as part of this plugin so that users just have a simple task to run!

iamdanfox (approved):
Happy to merge as is because some docs is better than no docs! thanks @mglazer

PR #906, opened 2019-09-30
Fix refaster compilation on gradle 5.x

ferozco:
## Before this PR
We were overriding a compilation method that only exists in gradle 6.0, causing usage from gradle 5.x to not work 

## After this PR
Correctly override a compilation method available in 5.0 and 6.0.

This actually means we don't need to raise our minimum version of gradle

## Possible downsides?
N/A

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CompileRefasterTask.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CompileRefasterTask.java
@@ -27,7 +27,7 @@ import org.gradle.api.provider.Property;
 import org.gradle.api.tasks.InputFiles;
 import org.gradle.api.tasks.OutputFile;
 import org.gradle.api.tasks.compile.JavaCompile;
-import org.gradle.work.InputChanges;
+import org.gradle.api.tasks.incremental.IncrementalTaskInputs;
 
 public class CompileRefasterTask extends JavaCompile {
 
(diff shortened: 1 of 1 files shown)


carterkozak (approved):
Manually verified this resolves the issue, thanks again Felipe!

gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CompileRefasterTask.java
```diff
@@ -38,12 +38,15 @@ public CompileRefasterTask() {
...
+        // Ensure we hit the incremental code-path since we override it
+        getOptions().setIncremental(true);
     }
 
     @Override
-    protected final void compile(InputChanges inputs) {
+    // TODO(forozco): override compile(InputChanges inputs) once we can raise our minimum version 6.0
+    @SuppressWarnings("deprecated")
+    protected final void compile(IncrementalTaskInputs inputs) {
+        getProject().getLogger().warn("Doing compilation - Felipe");
```
iamdanfox, line 49: DOING COMPILATION FELIPE

ferozco: 🤦‍♂

ferozco: cancelled publish to avoid people accidentally getting spammed. I'll try to be better going forward

carterkozak: ... I don't know how I missed this 🤦‍♂

ferozco: https://github.com/palantir/gradle-baseline/pull/908

PR #907, opened 2019-09-30
update changelog

ferozco:
@carterkozak #906 merged a little too quickly


carterkozak:
Should we cut a release on this since it's not a major version rev anymore?

carterkozak:
👍

PR #910, opened 2019-09-30
Added Refaster Rule for Optional 'hasValue' Assert

bmarcaur:
// cc @carterkozak 

## After this PR
Added refaster rule for optionals `hasValue` assert, replaces `isEqualsTo` with `hasValue`

### DIFF
--- a/baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/AssertjOptionalHasValue.java
+++ b/baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/AssertjOptionalHasValue.java
@@ -1,0 +1,39 @@
+/*
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.baseline.refaster;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import com.google.errorprone.refaster.ImportPolicy;
+import com.google.errorprone.refaster.annotation.AfterTemplate;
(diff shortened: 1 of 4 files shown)


baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/AssertjOptionalHasValue.java
```diff
@@ -0,0 +1,39 @@
...
+import com.google.errorprone.refaster.annotation.AfterTemplate;
+import com.google.errorprone.refaster.annotation.BeforeTemplate;
+import com.google.errorprone.refaster.annotation.UseImportPolicy;
+import java.util.Optional;
+
+public class AssertjOptionalHasValue<T> {
+
+    @BeforeTemplate
+    void before(Optional<T> thing, T expected) {
+        assertThat(thing.get()).isEqualTo(expected);
```
iamdanfox, line 31: I think you need to match on `Optional<T> expected`

baseline-refaster-rules/src/test/java/com/palantir/baseline/refaster/AssertjOptionalHasValueTest.java
```diff
@@ -0,0 +1,68 @@
...
+    @Test
+    public void test() {
+        RefasterTestHelper
+                .forRefactoring(AssertjOptionalHasValue.class)
+                .withInputLines(
+                        "Test",
+                        "import static org.assertj.core.api.Assertions.assertThat;",
+                        "import java.util.Optional;",
+                        "public class Test<T> {",
+                        "  void f(Optional<T> in, T out) {",
```
carterkozak: nit: test would be easier to follow without the generic, e.g. `Optional<String> in, T expected`

baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/AssertjOptionalHasValue.java
carterkozak: Not sure what you mean @iamdanfox.

bmarcaur: The rewrite is not for cases where 2 optionals are equal it is instead for cases where you are testing the wrapped value inside the optional is equal to something.

iamdanfox: Ah yep sorry I didn't spot the `.get()` and thought it was testing the whole optional<T> against an expected optional<T> value.

This looks fine!

iamdanfox (approved):

carterkozak:
:+1:

PR #911, opened 2019-09-30
Add refaster rule to simplify empty optional asserts

carterkozak:
==COMMIT_MSG==
Add refaster rule to simplify empty optional asserts
==COMMIT_MSG==

### DIFF
--- a/baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/AssertjOptionalIsNotPresent.java
+++ b/baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/AssertjOptionalIsNotPresent.java
@@ -16,13 +16,13 @@
 
 package com.palantir.baseline.refaster;
 
+import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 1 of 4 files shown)


ferozco:
Could we also add some rules to handle `assertThat(foo).isEqualTo(Optional.of(bar)` -> `assertThat(foo).hasValue(bar)`

carterkozak:
Sure, mind if that lands in a separate rule?

iamdanfox:
I actually don't want to auto refactor that `assertThat(foo).hasValue(bar)`, because I think it doesn't actually improve the debuggability, while actually slightly reducing the readability (I didn't on first glance spot that `hasValue()` means the arg is an optional).

I know we have lots of refactoring power, just don't think that one is unambiguously net positive.

ferozco:
👍

ferozco:
@iamdanfox i think its a personal preference since to me removing the extra wrapping makes it clearer. Don't feel strongly enough to push for it if you disagree

PR #912, opened 2019-09-30
AssertJ refasters for some integer inequalities

iamdanfox:
## Before this PR

Excavator did this to the timelock repo:

Screenshot 2019-09-30 at 19 30 29

## After this PR
==COMMIT_MSG==
Refaster rules will automatically convert inequalities like `assertThat(a < b).isTrue()` to the easier to debug form: `assertThat(a).isLessThan(b)`
==COMMIT_MSG==

## Possible downsides?

- there is a lot of code and it still doesn't cover longs, floats, doubles. starting to feel like refaster might not be the right tool for the job here.


ferozco:
👍 probably want a changelog on this bad boy

iamdanfox:
So I actually think I'd like to split out all the assertj stuff from this repo, and this PR won't actually affect how baseline affects users (as excavator makes the assertj prs independently), so I think no changelog is actually ok

PR #913, opened 2019-09-30
Assertj refasters for array equality

iamdanfox:
## Before this PR

Excavator did this to timelock:

image

## After this PR
==COMMIT_MSG==
AssertJ refasters for array equality
==COMMIT_MSG==

## Possible downsides?
N/A

### DIFF
--- a/baseline-refaster-rules/src/test/java/com/palantir/baseline/refaster/AssertjArrayEqualsTest.java
+++ b/baseline-refaster-rules/src/test/java/com/palantir/baseline/refaster/AssertjArrayEqualsTest.java
@@ -180,4 +180,23 @@ public class AssertjArrayEqualsTest {
                         "  void f(boolean[] a, boolean[] b) { assertThat(a).isEqualTo(b); }",
                         "}");
     }
+
+    @Test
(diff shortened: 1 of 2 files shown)


baseline-refaster-rules/src/test/java/com/palantir/baseline/refaster/AssertjArrayEqualsTest.java
```diff
@@ -0,0 +1,202 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.junit.Test;
+
+public class AssertjArrayEqualsTest {
+    @Test
+    public void sanity_check() {
+        byte[] bytes = {1, 2, 3};
+        // both of these work, but I think isEqualTo reads a bit more nicely
```
carterkozak, line 27: +1. containsExactly makes more sense if we deconstruct array creation into varargs. e.g. `assertThat(bytes).containsExactly(1, 2, 3);` but I think that's a bit extreme for automation without providing significantly better failure information.

carterkozak:
👍

PR #914, opened 2019-09-30
More refasters for AssertJ optional hasValue method

iamdanfox:
Inspired by more silly looking code I discovered:

image


carterkozak:
👍

PR #915, opened 2019-09-30
Allow static imports of AssertJ and Mockito

pkoenig10:
## Before this PR
It's common to have a test utility project which is a test dependency in other projects. We currently allow static imports in tests (#240) but that does not apply to test utility projects since they are main sources.

This paper cut is aggravated by the `PreferAssertJ` check (#841) which static imports AssertJ methods. Doing this causes automated baseline upgrades to require manual intervention to successfully compile.

## After this PR
We allow static imports of AssertJ and Mockito methods.


carterkozak:
👍

PR #917, opened 2019-10-01
Remove refaster AssertjIsOne, which isn't worthwhile

carterkozak:
This refactor can sometimes refactor
`assertThat(collection.size()).isEqualTo(1)`
into `assertThat(collection.size()).isOne()` instead of the
preferred `assertThat(collection).hasSize(1)`. isZero/isNotZero
provide clean code and tend not to interfere with other rules.
Collections can refactor from isZero to isEmpty, for instance.

==COMMIT_MSG==
Remove refaster AssertjIsOne
==COMMIT_MSG==


iamdanfox (approved):
Yep this refaster gave me a chuckle earlier today when a line of asserts was `isEqualTo(3)`, `isEqualTo(5)`, `isEqualTo(2)`, `isOne()`!

PR #918, opened 2019-10-01
extra AssertjOptionalHasValue case

iamdanfox:
I saw some of our internal repos actually do this two-stage assert on optionals. Taking it down to one line gives you better error messages (and I think is more readable).

### DIFF
--- a/baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/AssertjOptionalHasValue.java
+++ b/baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/AssertjOptionalHasValue.java
@@ -36,6 +36,12 @@ public final class AssertjOptionalHasValue<T> {
         assertThat(optional.isPresent() && optional.get().equals(innerValue)).isTrue();
     }
 
+    @BeforeTemplate
+    void before3(Optional<T> optional, T innerValue) {
+        assertThat(optional.isPresent()).isTrue();
+        assertThat(optional.get().equals(innerValue)).isTrue();
+    }
+
     @AfterTemplate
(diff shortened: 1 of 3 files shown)


baseline-refaster-rules/src/test/java/com/palantir/baseline/refaster/AssertjOptionalHasValueTest.java
```diff
@@ -46,6 +48,7 @@ public void test() {
                         "  void f(Optional<String> in, String out) {",
                         "    assertThat(in).hasValue(out);",
                         "    assertThat(in).hasValue(out);",
+                        "    assertThat(in).hasValue(out);",
```
carterkozak, line 55: looks like these expect a trailing blank line from the removed line.

carterkozak (approved):
lgtm once the test is happy.

baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/AssertjOptionalHasValue.java
```diff
@@ -36,6 +36,12 @@ void before2(Optional<T> optional, T innerValue) {
         assertThat(optional.isPresent() && optional.get().equals(innerValue)).isTrue();
     }
 
+    @BeforeTemplate
+    void redundantAssertion(Optional<T> optional, T innerValue) {
+        assertThat(optional).isPresent();
+        assertThat(optional).hasValue(innerValue);
```
iamdanfox, line 42: I realised that lots of different cases will actually reduce down to this form eventually (using other refaster rules), so we just need this one reduction.

PR #924, opened 2019-10-02
Use Mockito verifyNoInteractions over deprecated verifyZeroInteractions

schlosna:
## Before this PR
Mockito deprecated `verifyZeroInteractions` in favor of `verifyNoInteractions` in Mockito 3.0.1 via https://github.com/mockito/mockito/pull/1733

## After this PR
==COMMIT_MSG==
Use Mockito verifyNoInteractions over deprecated verifyZeroInteractions
==COMMIT_MSG==

## Possible downsides?
The after refactor requires Mockito 3.0.1+ to succeed

### DIFF
--- a/baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/MockitoVerifyNoInteractions.java
+++ b/baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/MockitoVerifyNoInteractions.java
@@ -1,0 +1,42 @@
+/*
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.baseline.refaster;
+
+import static org.mockito.Mockito.verifyNoInteractions;
+import static org.mockito.Mockito.verifyZeroInteractions;
+
+import com.google.errorprone.refaster.ImportPolicy;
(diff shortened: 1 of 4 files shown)


iamdanfox:
Sooo I think might need to be careful here as @CRogers spotted that the two methods actually have different behaviour:

https://static.javadoc.io/org.mockito/mockito-core/3.1.0/org/mockito/Mockito.html#verifyZeroInteractions-java.lang.Object...-

it sounds like `verifyNoInteractions` at the end of a @Test is pretty much pointless...

pkoenig10:
It's definitely not pointless, it's useful to verify that no more interactions with your mock occurred. Granted, you probably should be using `verifyNoMoreInteractions` instead in this case.

The question here is about intent when using `verifyZeroInteractions`. If `verifyNoInteractions` behavior was intended, then any failures caused by the change are bugs. If `verifyNoMoreInteractions` behavior was intended, then we've just created extra work for people.

I think most people were not aware of this distinction and most uses fall into the former case. It seems preferable to catch these bugs at the risk of creating some extra work.

iamdanfox (approved):

baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/MockitoVerifyNoInteractions.java
```diff
@@ -0,0 +1,42 @@
...
+    @SuppressWarnings("deprecation")
+    // must use @Repeated for varargs per https://github.com/google/error-prone/issues/568
+    void before(@Repeated Object varargs) {
+        verifyZeroInteractions(varargs);
+    }
+
+    @AfterTemplate
+    @UseImportPolicy(ImportPolicy.STATIC_IMPORT_ALWAYS)
+    void after(@Repeated Object varargs) {
+        verifyNoInteractions(varargs);
```
jkozlowski, line 40: As per javadoc on this stuff, it should be replacing it with verifyNoMoreInteractions. Granted that those types of tests are probably a bit iffy, but this replacement probably means people's excavators will back up.

jkozlowski: Althought that might be an argument for this change though, to make people look at those tests again.

iamdanfox: @dansanduleac fixed this yesterday: https://github.com/palantir/gradle-baseline/pull/975/files

PR #927, opened 2019-10-02
Rule to refacter out String.format from describedAs

carterkozak:
AssertjDescribedAsFormat
```
- assertThat(obj).describedAs(String.format("desc %s", "arg")).isEqualTo("foo");
+ assertThat(obj).describedAs("desc %s", "arg").isEqualTo("foo");
```

==COMMIT_MSG==
Rule to refacter out String.format from describedAs
==COMMIT_MSG==

### DIFF
--- a/baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/AssertjDescribedAsFormat.java
+++ b/baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/AssertjDescribedAsFormat.java
@@ -1,3 +1,19 @@
+/*
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ *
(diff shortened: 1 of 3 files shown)


ferozco (approved):

carterkozak:
changelog bot, everything okay buddy?

ferozco:
Looks like we got wrecked by the changelog-app major rev. Need to add an empty changelog.yml file to every project 🤦‍♂ 

@nmiyake for SA

nmiyake:
@ferozco just checked, and it looks like, for GH.com projects, the app is manually enabled on 34 repos, so those are the ones for which the file should be added/managed (and after that is done, app should be enabled on entire org rather than per-project).

I can open PRs to do this.

PR #931, opened 2019-10-02
Sort output of checkUnusedDependencies

pkoenig10:
## Before this PR
- The unused dependencies are only sorted by artifact ID, not the full name that is displayed
- The suggestions are unsorted

## After this PR
- The unused dependencies are sorted by the full name that is displayed
- The suggestions are sorted by the full name that is displayed

This makes it easier to determine which suggested replacement should be used, since there are often multiple suggestions with the same group.

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckUnusedDependenciesTask.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckUnusedDependenciesTask.java
@@ -87,10 +87,11 @@ public class CheckUnusedDependenciesTask extends DefaultTask {
         getLogger().debug("Possibly unused dependencies: {}",
                 possiblyUnused.stream()
                         .map(BaselineExactDependencies::asString)
+                        .sorted()
                         .collect(Collectors.toList()));
         List<ResolvedArtifact> declaredButUnused = possiblyUnused.stream()
                 .filter(artifact -> !shouldIgnore(artifact))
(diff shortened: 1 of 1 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckUnusedDependenciesTask.java
```diff
@@ -117,9 +118,12 @@ public final void checkUnusedDependencies() {
...
                     sb.append("\t\tDid you mean:\n");
-                    didYouMean.forEach(transitive -> sb.append("\t\t\timplementation '")
-                            .append(asString(transitive))
-                            .append("\' \n"));
+                    didYouMean.stream()
+                            .map(BaselineExactDependencies::asString)
+                            .sorted()
+                            .forEach(transitive -> sb.append("\t\t\timplementation '")
+                                    .append(transitive)
+                                    .append("\' \n"));
```
CRogers, line 126: I think you can use `Collectors.joining` instead of manually building up a `StringBuilder` here.

pkoenig10: This is using the existing string builder for the entire output. I can collect to a string and then append to builder if you prefer since it's a bit cleaner, but just note that it is slightly less efficient.

iamdanfox: I'm fine with keeping this as is tbh

iamdanfox (approved):

PR #935, opened 2019-10-03
Refaster rules for checks that maps do not contain a specific key

carterkozak:
Implemented new rules for `doesNotContainKey`
```diff
- assertThat(in.keySet().contains("foo")).isFalse();
- assertThat(in.containsKey("foo")).isFalse();
- assertThat(in.get("foo")).isNull();
+ assertThat(in).doesNotContainKey("foo");
```

Added a new case to the containsKey refaster
```diff
- assertThat(in.get("foo")).isNotNull();
+ assertThat(in).containsKey("foo");
```

## After this PR
==COMMIT_MSG==
Refaster rules for checks that maps do not contain a specific key
==COMMIT_MSG==

### DIFF
--- a/baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/AssertjBooleanNegationIsFalse.java
+++ b/baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/AssertjBooleanNegationIsFalse.java
@@ -22,29 +22,18 @@ import com.google.errorprone.refaster.ImportPolicy;
 import com.google.errorprone.refaster.annotation.AfterTemplate;
 import com.google.errorprone.refaster.annotation.BeforeTemplate;
 import com.google.errorprone.refaster.annotation.UseImportPolicy;
-import java.util.Map;
 
-public final class AssertjMapDoesNotContainKey<K, V> {
-
-    @BeforeTemplate
-    void before1(Map<K, V> things, K key) {
-        assertThat(things.containsKey(key)).isFalse();
-    }
+public final class AssertjBooleanNegationIsFalse {
 
     @BeforeTemplate
-    void before2(Map<K, V> things, K key) {
-        assertThat(things.get(key)).isNull();
-    }
-
-    @BeforeTemplate
-    @SuppressWarnings("RedundantCollectionOperation") // It's what we're fixing
-    void before3(Map<K, V> things, K key) {
-        assertThat(things.keySet().contains(key)).isFalse();
+    void before(boolean input) {
+        assertThat(!input).isFalse();
     }
 
     @AfterTemplate
     @UseImportPolicy(ImportPolicy.STATIC_IMPORT_ALWAYS)
-    void after(Map<K, V> things, K key) {
-        assertThat(things).doesNotContainKey(key);
+    void after(boolean input) {
+        assertThat(input).isTrue();
     }
+
 }
--- a/baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/AssertjMapDoesNotContainKey.java
+++ b/baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/AssertjMapDoesNotContainKey.java
@@ -24,27 +24,38 @@ import com.google.errorprone.refaster.annotation.BeforeTemplate;
 import com.google.errorprone.refaster.annotation.UseImportPolicy;
 import java.util.Map;
 
-public final class AssertjMapContainsKey<K, V> {
+public final class AssertjMapDoesNotContainKey<K, V> {
 
     @BeforeTemplate
     void before1(Map<K, V> things, K key) {
-        assertThat(things.containsKey(key)).isTrue();
+        assertThat(things.containsKey(key)).isFalse();
     }
 
     @BeforeTemplate
-    @SuppressWarnings("RedundantCollectionOperation") // It's what we're fixing
     void before2(Map<K, V> things, K key) {
-        assertThat(things.keySet().contains(key)).isTrue();
+        assertThat(!things.containsKey(key)).isTrue();
     }
 
     @BeforeTemplate
     void before3(Map<K, V> things, K key) {
-        assertThat(things.get(key)).isNotNull();
+        assertThat(things.get(key)).isNull();
+    }
+
+    @BeforeTemplate
+    @SuppressWarnings("RedundantCollectionOperation") // It's what we're fixing
+    void before4(Map<K, V> things, K key) {
+        assertThat(things.keySet().contains(key)).isFalse();
+    }
+
+    @BeforeTemplate
+    @SuppressWarnings("RedundantCollectionOperation") // It's what we're fixing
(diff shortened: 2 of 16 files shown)


baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/AssertjMapDoesNotContainKey.java
```diff
@@ -0,0 +1,61 @@
...
+    @BeforeTemplate
+    @SuppressWarnings("RedundantCollectionOperation") // It's what we're fixing
+    void before4(Map<K, V> things, K key) {
+        assertThat(things.keySet().contains(key)).isFalse();
+    }
+
+    @BeforeTemplate
+    @SuppressWarnings("RedundantCollectionOperation") // It's what we're fixing
+    void before5(Map<K, V> things, K key) {
+        assertThat(!things.keySet().contains(key)).isTrue();
```
iamdanfox: If we had a standalone refaster that changed `assertThat(!arbitaryBoolean).isFalse()` to `assertThat(arbitaryBoolean).isTrue()`, we wouldn't need this case right?

carterkozak: Yep, I can add that

iamdanfox: If we're doing silly ones, might want `assertThat(foo == null).isTrue()` -> `.isNull()` too?

baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/AssertjMapDoesNotContainKeyWithDescription.java
```diff
@@ -0,0 +1,62 @@
...
+    void before5(Map<K, V> things, K key, String description, @Repeated Object descriptionArgs) {
+        assertThat(!things.keySet().contains(key)).describedAs(description, descriptionArgs).isTrue();
+    }
+
+    @AfterTemplate
+    @UseImportPolicy(ImportPolicy.STATIC_IMPORT_ALWAYS)
+    void after(Map<K, V> things, K key, String description, @Repeated Object descriptionArgs) {
+        assertThat(things).describedAs(description, descriptionArgs).doesNotContainKey(key);
+    }
+}
```
iamdanfox, line 51: Do you think it's worth writing a test that could use reflection to go through each file in the refaster package and verify there is a matching Foo and FooWithDescription class, each with the same number of methods?

I feel like someone is going to forget at somepoint...

iamdanfox (approved):

baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/AssertjMapDoesNotContainKeyWithDescription.java
carterkozak: Perhaps, there are enough edge cases where the check with a description differs from the non-description version that it could be tricky (e.g. combining the optional isPresent and hasValue checks). I don't think we're at the point that it's worthwhile yet, especially considering the failure mode is that we don't automatically make something better.

baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/AssertjMapDoesNotContainKey.java
carterkozak: Removed the negation checks from this class and added the general rule.

baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/AssertjBooleanNegationIsFalse.java
```diff
@@ -0,0 +1,39 @@
...
+import com.google.errorprone.refaster.ImportPolicy;
+import com.google.errorprone.refaster.annotation.AfterTemplate;
+import com.google.errorprone.refaster.annotation.BeforeTemplate;
+import com.google.errorprone.refaster.annotation.UseImportPolicy;
+
+public final class AssertjBooleanNegationIsFalse {
+
+    @BeforeTemplate
+    void before(boolean input) {
+        assertThat(!input).isFalse();
```
iamdanfox, line 30: I am genuinely so curious to see if this catches anything!

baseline-refaster-testing/src/main/java/com/palantir/baseline/refaster/RefasterTestHelper.java
```diff
@@ -49,14 +51,30 @@ public static RefasterTestHelper forRefactoring(Class<?> refasterRuleClass) {
         return new RefasterTestHelper(refasterRuleClass);
     }
 
-    private RefasterTestHelper(Class<?> refasterRuleClass) {
+    /**
+     * The source code of the given refaster rules should exist in {@code src/main/java}.
+     */
+    public static RefasterTestHelper forRefactoring(Class<?>... refasterRuleClasses) {
```
iamdanfox, line 57: nice

carterkozak:
Thanks for such a quick review @iamdanfox!

baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/AssertjMapDoesNotContainKey.java
```diff
@@ -0,0 +1,50 @@
...
+public final class AssertjMapDoesNotContainKey<K, V> {
+
+    @BeforeTemplate
+    void before1(Map<K, V> things, K key) {
+        assertThat(things.containsKey(key)).isFalse();
+    }
+
+    @BeforeTemplate
+    void before2(Map<K, V> things, K key) {
+        assertThat(things.get(key)).isNull();
```
schlosna, line 36: I'm not sure we necessarily want to refactor this one to `assertThat(things).doesNotContainKey(key)` as while this is true for some implementations of map (e.g. `ConcurrentHashMap`, `ImmutableMap`), but not necessarily for others such as `HashMap` or `TreeMap` where a key may be mapped to `null` (i.e. `things.get(key) == null && things.containsKey(key) == true`)

schlosna: To be fair, this did correctly refactor several places in tritium where we can properly use `doesNotContainKey` in https://github.com/palantir/tritium/pull/454

carterkozak: I've found one place in practice where it broke a test, but in all other cases it produced better asserts. When this is hit, we either want to refactor to:
`assertThat(map).doesNotContainKey(key)`
or
`assertThat(map).containsEntry(key, null)`

The two are semantically different, and it's uncommon that test authors actually mean to test that either the value is null or the map does not contain a key.

PR #936, opened 2019-10-03
New feature flag: com.palantir.baseline-format.palantir-java-format=true

iamdanfox:
## Before this PR

- Baseline didn't do any interesting formatting out of the box, because we don't have strong conviction that any of the existing formatter are good enough (g-j-f is veeery different from palantir style, eclipse is close, but not quite there)
- we exposed one feature flag allowing people to opt in to the eclipse formatter (and have been excavating this out)

Demo'd https://github.com/palantir/gradle-consistent-versions/pull/283

## After this PR
==COMMIT_MSG==
Users can opt-in to format their files using our fork of google-java-format (palantir-java-format)
==COMMIT_MSG==

## Possible downsides?
- more feature flags = more complexity


gradle-baseline-java/build.gradle
```diff
@@ -15,6 +15,7 @@ dependencies {
     compile 'net.ltgt.gradle:gradle-errorprone-plugin'
     compile 'org.apache.maven.shared:maven-dependency-analyzer'
     compile 'org.github.ngbinh.scalastyle:gradle-scalastyle-plugin_2.11'
+    compile 'com.google.googlejavaformat:google-java-format'
```
iamdanfox: this is kinda interesting... I'd prefer to actually put this in a configuration, and resolve it only when the `format` task is called, as this would reduce the amount of stuff necessary for configuration.  HOWEVER, I think that would require using a JavaExec task, which I don't want to do as it incurs JVM startup time and heap size sadness.

dansanduleac: Actually it won't, Spotless allows you to load a classpath and run it in a different classloader. We could do that instead

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineFormat.java
```diff
@@ -95,7 +111,21 @@ static boolean eclipseFormattingEnabled(Project project) {
...
+        return project.hasProperty(PJF_PROPERTY);
+    }
+
     static Path eclipseConfigFile(Project project) {
         return project.getRootDir().toPath().resolve(".baseline/spotless/eclipse.xml");
     }
+
+    private static class PalantirJavaFormatterFunc implements FormatterFunc {
+        private final Formatter palantirJavaFormatter =
+                new Formatter(JavaFormatterOptions.builder().style(Style.PALANTIR).build());
```
iamdanfox: I've made this constructor private, so let's switch to service-loading!

dansanduleac: yeah, sadly we need to access more internals to do the string -> string conversion, but luckily the other internals are still public :)

dansanduleac:
👍

ferozco (approved):

Issue #895, opened 2019-09-26
Drop / disallow use of Lists.newArrayList et al


aioobe:
Drop / disallow use of obsolete static methods such as

- `Lists.newArrayList`
- `Lists.newLinkedList`
- `Maps.newConcurrentMap`
- `Maps.newEnumMap`
- `Maps.newHashMap`
- `Maps.newIdentityHashMap`
- `Maps.newLinkedHashMap`
- `Maps.newTreeMap`

and siblings (`...WithExpectedSize`).

Related discussion: https://github.com/palantir/gradle-baseline/pull/633

iamdanfox:
@aioobe would you like to make the PR for this? I think a refaster rule should do the trick, e.g.: 

- https://github.com/palantir/gradle-baseline/blob/develop/baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/SortedFirst.java
- https://github.com/palantir/gradle-baseline/blob/develop/baseline-refaster-rules/src/test/java/com/palantir/baseline/refaster/SortedFirstTest.java

This wouldn't necessarily disallow them as such, it would just have excavator make the PR to fix them.

aioobe:
Sure

carterkozak:
If we want the ability to enforce this pre-merge, I don't think a refaster rule is sufficient.

aioobe:
What's a good example to look at if we want it to be able to be enforced during pre-merge?

carterkozak:
There’s a check to prefer safe logging preconditions over guava, that might be a good starting point

PR #941, opened 2019-10-04
Added errorprone checks for Guavas static factory methods.

aioobe:
Adds errorprone checks for Guavas static factory methods. See https://github.com/palantir/gradle-baseline/issues/895


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferCollectionConstructors.java
```diff
@@ -0,0 +1,293 @@
...
+
+    private static final Matcher<ExpressionTree> NEW_COPY_ON_WRITE_ARRAY_LIST =
+            MethodMatchers.staticMethod()
+                    .onClass("com.google.common.collect.Lists")
+                    .named("newCopyOnWriteArrayList");
+
+    private static final Matcher<ExpressionTree> NEW_ARRAY_LIST_WITH_CAPACITY =
+            MethodMatchers.staticMethod()
+                    .onClass("com.google.common.collect.Lists")
+                    .namedAnyOf("newArrayListWithCapacity", "newArrayListWithExpectedSize")
```
carterkozak: `newArrayListWithExpectedSize` has slightly different (though often unexpectetd) behavior adding padding to the size, where `newArrayListWithCapacity` uses the input.

Based on the [docs](https://github.com/google/guava/blob/5496c37d4d904869297c2ced1f0d20e6f1507eaa/guava/src/com/google/common/collect/Lists.java#L182-L183) I'm okay with forcing this change.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferCollectionConstructors.java
```diff
@@ -0,0 +1,293 @@
...
+import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.CopyOnWriteArraySet;
+import java.util.stream.Collectors;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "PreferCollectionConstructors",
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
+        severity = BugPattern.SeverityLevel.WARNING,
```
carterkozak: Might want to start this at informational level and increase after we've validated it in a few repos.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferCollectionConstructors.java
```diff
@@ -0,0 +1,293 @@
...
+        SuggestedFix.Builder fixBuilder = SuggestedFix.builder();
+        String collectionType = SuggestedFixes.qualifyType(state, fixBuilder, collectionClass.getName());
+        String typeArgs = tree.getTypeArguments()
+                .stream()
+                .map(state::getSourceForNode)
+                .collect(Collectors.joining(", "));
+        String arg = argTree == null ? "" : state.getSourceForNode(argTree);
+        String replacement = "new " + collectionType + "<" + typeArgs + ">(" + arg + ")";
+        return buildDescription(tree)
+                .setMessage("The factory method call should be replaced with a constructor call.")
```
carterkozak: Can we add a link to baseline docs here and provide more information describing why this is preferable over guava factories?

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferCollectionConstructors.java
```diff
@@ -0,0 +1,293 @@
...
+        summary = "Since Java 7 the standard collection constructors should be used instead of the static factory "
+                + "methods provided by Guava.")
+public final class PreferCollectionConstructors extends BugChecker implements BugChecker.MethodInvocationTreeMatcher {
+
+    private static final long serialVersionUID = 1L;
+
+    private static final Matcher<ExpressionTree> NEW_ARRAY_LIST =
+            MethodMatchers.staticMethod()
+                    .onClass("com.google.common.collect.Lists")
+                    .named("newArrayList");
```
carterkozak: could we duplicate matchers with specific parameter types and reduce logic below?

aioobe: It seems like the MethodMatchers matches on the argument types rather than the formal parameter types. This means that something like `...withParameters("java.lang.Iterable")` matches something like `Lists.<Iterable<String>>newArrayList(someIterable)` even though this is an invocation of `Lists.newArrayList(E... elements)`.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferCollectionConstructors.java
aioobe: Do we have a suitable link already, or do you suggest we add a section in docs/java-style-guide/readme.md and link to that?

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferCollectionConstructors.java
carterkozak: I don't think that is correct, I don't believe the captured argument types are used by the matcher.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferCollectionConstructors.java
carterkozak: Didn't you contribute this? ;-)
https://github.com/palantir/gradle-baseline/blob/develop/docs/best-practices/java-coding-guidelines/readme.md#avoid-generics-clutter-where-possible

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferCollectionConstructors.java
aioobe: You are right indeed. Pushing update within short.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferCollectionConstructors.java
aioobe: Huh, look at that. I didn't think that got merged :-) Adding a link.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferCollectionConstructors.java
```diff
@@ -0,0 +1,345 @@
...
+            NEW_COPY_ON_WRITE_ARRAY_LIST_WITH_ITERABLE,
+            NEW_COPY_ON_WRITE_ARRAY_SET_WITH_ITERABLE,
+            NEW_LINKED_HASH_SET_WITH_ITERABLE,
+            NEW_TREE_SET_WITH_ITERABLE,
+            NEW_HASH_SET_WITH_ITERABLE);
+
+    @Override
+    public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
+
+        Class<?> collectionClass = findCollectionClassToUse(state, tree);
```
carterkozak, line 277: Might want to create a static fast check using Matchers.anyOf(classMap.keySet()) to avoid creating an iterable for every method invocation.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferCollectionConstructors.java
```diff
@@ -0,0 +1,345 @@
...
+                        + "See https://git.io/JeCT6 for more information.")
+                .addFix(fixBuilder.replace(tree, replacement).build())
+                .build();
+    }
+
+    private Class<?> findCollectionClassToUse(VisitorState state, ExpressionTree tree) {
+        boolean firstArgIsCollection = isFirstArgCollection(state, tree);
+        for (Map.Entry<Matcher<ExpressionTree>, Class<?>> entry : classMap.entrySet()) {
+            Matcher<ExpressionTree> matcher = entry.getKey();
+            if (matcher.matches(tree, state) && (!requiresCollectionArg.contains(matcher) || firstArgIsCollection)) {
```
carterkozak: If matcher.matches is true and the second check is false, I think null should be returned. Are there any cases where we would want to check more cases?

schlosna:
Would these be clearer as refaster templates than error-prone checks?

Curious if it would be possible to upstream these to Guava as [they have a some refaster refactorings](https://github.com/google/guava/wiki/Refactorings) https://github.com/google/guava/tree/master/refactorings for removed APIs.

carterkozak:
> Would these be clearer as refaster templates than error-prone checks?

I recommended against refaster for this because it cannot validate pre-merge. The code would be simpler though.

schlosna:
> > Would these be clearer as refaster templates than error-prone checks?
> 
> I recommended a

:+1: Makes sense, I missed the same discussion on the issue.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferCollectionConstructors.java
aioobe: The `ImmutableMap` seems to internally only create a key set once anyway, right? And the `anyOff(...)` matcher still just loops over the matchers. Maybe I'm missing something, but it seems like it wouldn't make a difference?

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferCollectionConstructors.java
aioobe: Right. As of now, all matchers happen to be mutually exclusive, so that should be a safe optimization. I'll update.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferCollectionConstructors.java
carterkozak: Ah right, I thought matchers.anyOf did something cleverer :-/

aioobe:
Next steps?

carterkozak:
I'll give this another pass tomorrow. I'd like to give the other infra folks a chance to review as well, they'll be back on Thursday.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferCollectionConstructors.java
```diff
@@ -0,0 +1,348 @@
...
+    private static final Matcher<ExpressionTree> NEW_ARRAY_LIST_WITH_ITERABLE =
+            MethodMatchers.staticMethod()
+                    .onClass("com.google.common.collect.Lists")
+                    .named("newArrayList")
+                    .withParameters("java.lang.Iterable");
+
+    private static final Matcher<ExpressionTree> NEW_ARRAY_LIST_WITH_CAPACITY =
+            MethodMatchers.staticMethod()
+                    .onClass("com.google.common.collect.Lists")
+                    .namedAnyOf("newArrayListWithCapacity", "newArrayListWithExpectedSize")
```
schlosna: This is likely uncommon, but I know there are a few places that intentionally use the `newArrayListWithExpectedSize` to presize with extra buffer to avoid growing & copy overhead when only a few extra are (conditionally) needed.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferCollectionConstructors.java
```diff
@@ -0,0 +1,348 @@
...
+        String collectionType = SuggestedFixes.qualifyType(state, fixBuilder, collectionClass.getName());
+        String typeArgs = tree.getTypeArguments()
+                .stream()
+                .map(state::getSourceForNode)
+                .collect(Collectors.joining(", "));
+        String arg = tree.getArguments().isEmpty() ? "" : state.getSourceForNode(tree.getArguments().get(0));
+        String replacement = "new " + collectionType + "<" + typeArgs + ">(" + arg + ")";
+        return buildDescription(tree)
+                .setMessage("The factory method call should be replaced with a constructor call. "
+                        + "See https://git.io/JeCT6 for more information.")
```
schlosna: curious why opaque shortened link?

```suggestion
                        + "See https://github.com/palantir/gradle-baseline/blob/develop/docs/best-practices/java-coding-guidelines/readme.md#avoid-generics-clutter-where-possible for more information.")
```

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferCollectionConstructors.java
carterkozak: newArrayListWithExpectedSize docs suggest it will be deprecated soon, and recommend pre-sizing ourselves. I don't have a strong opinion either way on this one, especially considering `newArrayListWithExpectedSize` doesn't make guarantees about how much extra space it provides, I'd be okay with leaving it out of this PR so we can approach it separately with more careful rollout.
I'm a little worried that if we ban `newArrayListWithCapacity(int)` and not `newArrayListWithExpectedSize(int)` people may use the latter instead of `new ArrayList(int)`

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferCollectionConstructors.java
```diff
@@ -0,0 +1,348 @@
...
+import java.util.TreeMap;
+import java.util.TreeSet;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.CopyOnWriteArraySet;
+import java.util.stream.Collectors;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "PreferCollectionConstructors",
```
carterkozak, line 55: I think we should add this to `BaselineErrorProneExtension.DEFAULT_PATCH_CHECKS` so baseline upgrades apply suggested fixes

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferCollectionConstructors.java
```diff
@@ -0,0 +1,348 @@
...
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
```
carterkozak, line 57: nit: mind adding `providesFix = BugPattern.ProvidesFix.REQUIRES_HUMAN_ATTENTION,`?

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferCollectionConstructors.java
aioobe: In a big repo I happened to have checked out there were 161 calls to this method. I looked through about 30 of them and 100% of those seemed like they should actually have been calls to `...WithCapacity` (or, rather `new ArrayList(int)`), i.e. none exceeded the given value.

I'd vote for pushing this through. If people prefer, I can look through the remaining calls.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferCollectionConstructors.java
aioobe: Done. What effect does this have?

carterkozak: Just docs/discoverability

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferCollectionConstructors.java
```diff
@@ -0,0 +1,349 @@
...
+        summary = "Since Java 7 the standard collection constructors should be used instead of the static factory "
+                + "methods provided by Guava.")
+public final class PreferCollectionConstructors extends BugChecker implements BugChecker.MethodInvocationTreeMatcher {
+
+    private static final long serialVersionUID = 1L;
+
+    private static final Matcher<ExpressionTree> NEW_ARRAY_LIST =
+            MethodMatchers.staticMethod()
+                    .onClass("com.google.common.collect.Lists")
+                    .named("newArrayList")
```
iamdanfox, line 69: :+1: method body is literally:

```java
  public static <E> ArrayList<E> newArrayList() {
    return new ArrayList<>();
  }
```

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferCollectionConstructors.java
iamdanfox: Personally I think I'd prefer to separate this one/leave it out so that we can ship an initial version of this errorprone rule which we know is an exact no-op, it seems a bit sneaky to have like 30 replacements which have exactly the same behaviour and one which gives you slightly different allocation/resizing behaviour.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferCollectionConstructors.java
```diff
@@ -0,0 +1,349 @@
...
+        String collectionType = SuggestedFixes.qualifyType(state, fixBuilder, collectionClass.getName());
+        String typeArgs = tree.getTypeArguments()
+                .stream()
+                .map(state::getSourceForNode)
+                .collect(Collectors.joining(", "));
+        String arg = tree.getArguments().isEmpty() ? "" : state.getSourceForNode(tree.getArguments().get(0));
+        String replacement = "new " + collectionType + "<" + typeArgs + ">(" + arg + ")";
+        return buildDescription(tree)
+                .setMessage("The factory method call should be replaced with a constructor call. "
+                        + "See https://github.com/palantir/gradle-baseline/blob/develop/docs/best-practices/java-coding-guidelines/readme.md#avoid-generics-clutter-where-possible for more information.")
```
iamdanfox, line 292: Given that there's already a first-clast 'link' field in the annotation, shall we move this up there?

iamdanfox:
Looks fine to me - I haven't really experienced any annoyance at people using these guava ones, but if you have then I think it's always nicer to encode this stuff in auto-fixable automation rather than needing to fill code-reviews with 'nit' comments every time!

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferCollectionConstructors.java
aioobe: Fair enough. I'll move this to a separate commit. :)

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferCollectionConstructors.java
aioobe: I have no opinion about this. (I don't know where these links end up actually.) @carterkozak, thoughts?

carterkozak: I don't have a strong opinion either way

carterkozak:
This is going to cause a fair amount of code churn -- I'm happy with the current state, but I'd like to hold off until we've rolled out the ThrowsError and Slf4jLevelCheck checks/fixes first.

We can either hold off merging this for a few days, or remove it from the list of automatic fixes and merge, so we have it available to test on repos without publishing to maven local. Thoughts?

iamdanfox:
Yeah I'd be happy to get the new static analysis onto develop, just not switch on the autofix straight away.

carterkozak:
👍 I think this will need a +1 from you as well @iamdanfox since I updated the extension

ferozco:
👍

PR #944, opened 2019-10-06
Add `providesFix = REQUIRES_HUMAN_ATTENTION` to relevant checks

carterkozak:
The default is NO_FIX which is misleading.

## After this PR
==COMMIT_MSG==
Add `providesFix = REQUIRES_HUMAN_ATTENTION` to relevant checks
==COMMIT_MSG==


iamdanfox (approved):
Seems harmless, but are we using this information anywhere? How did you notice this?

carterkozak:
I saw it used in a few google checks as well as StrictUnusedVariable. I don’t think it’s used for anything beyond documentation, but it makes it clear which checks are supposed to fix themselves.

PR #948, opened 2019-10-08
Improve checkUnusedDependencies message

pkoenig10:
## Before this PR
The `checkUnusedDependencies` error message has two deficiencies:
1. When an unused dependency has a classifier, it is not possible to distinguish between that dependency and a dependency without the classifier.
1. When depending on a project whose `default` configuration contains multiple sources sets and only some are unused, it is not possible to determine which source set is unused.

## After this PR
When displaying unused dependencies
1. Display the classifier in the resolved artifact string.
1. Display the `ModuleVersionIdentifier#getDisplayName` of the resolved artifact. The implementation isn't ideal because it appears as the following, but it's the only way to show the source set name without using internal classes.
    ```
    project(':my-other-project') (mySourceSet (project :my-other-project))
    ```

This PR also improves the display strings of project dependencies (both when unused and in suggestions) to appear as the following rather than artifact coordinates:
```
project(':my-other-project')
```

## Possible downsides?
The display of the source sets is verbose and can be confusing. If we find it too confusing I'm happy to remove that bit from this PR. But it was a real issue I ran into so I just wanted to put this up as a proposal.

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineExactDependencies.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineExactDependencies.java
@@ -39,6 +39,8 @@ import org.gradle.api.artifacts.Configuration;
 import org.gradle.api.artifacts.ModuleVersionIdentifier;
 import org.gradle.api.artifacts.ResolvedArtifact;
 import org.gradle.api.artifacts.ResolvedDependency;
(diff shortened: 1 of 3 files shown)


pkoenig10:
This should stay open

iamdanfox (approved):
Thanks @pkoenig10, I think displaying the `project(':foo')` thing will be much more helpful for people!

CRogers (approved):

PR #951, opened 2019-10-10
checkJUnitDependencies also detects spock (via nebula-test)

iamdanfox:
## Before this PR

We had the scary realization recently that no tests were running on develop for one of our gradle plugins because it used spock and junit5 without the vintage engine (see [fix PR](https://github.com/palantir/gradle-revapi/pull/48/files#diff-c197962302397baf3a4cc36463dce5eaR60))

## After this PR
==COMMIT_MSG==
checkJUnitDependencies detects a possible misconfiguration with spock and JUnit5 which could lead to tests silently not running.
==COMMIT_MSG==

## Possible downsides?
- the checkJUnitDependencies task is not cacheable, so might be slow on big projects.  Seems worth it to guarantee that people don't silently not run tests though!

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckJUnitDependencies.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckJUnitDependencies.java
@@ -73,7 +73,7 @@ public class CheckJUnitDependencies extends DefaultTask {
         boolean junitJupiterIsPresent = hasDep(deps, CheckJUnitDependencies::isJunitJupiter);
         boolean vintageEngineExists = hasDep(deps, CheckJUnitDependencies::isVintageEngine);
         boolean spockDependency = hasDep(deps, CheckJUnitDependencies::isSpock);
-        String testRuntimeOnly = ss.getRuntimeConfigurationName() + "Only";
+        String testRuntimeOnly = ss.getRuntimeOnlyConfigurationName();
         boolean junitPlatformEnabled = BaselineTesting.useJUnitPlatformEnabled(task);
 
         // If some testing library happens to provide the junit-jupiter-api, then users might start using the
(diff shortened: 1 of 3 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckJUnitDependencies.java
```diff
@@ -63,16 +65,24 @@ public final void validateDependencies() {
...
-                ss.getRuntimeClasspathConfigurationName(), CheckJUnitDependencies::isJunitJupiter);
+        Set<ResolvedComponentResult> deps = getProject().getConfigurations()
+                .getByName(ss.getRuntimeClasspathConfigurationName())
+                .getIncoming()
+                .getResolutionResult()
+                .getAllComponents();
+        boolean junitJupiterIsPresent = hasDep(deps, CheckJUnitDependencies::isJunitJupiter);
+        boolean vintageEngineExists = hasDep(deps, CheckJUnitDependencies::isVintageEngine);
+        boolean spockDependency = hasDep(deps, CheckJUnitDependencies::isSpock);
+        String testRuntimeOnly = ss.getRuntimeConfigurationName() + "Only";
```
dansanduleac: `ss.getRuntimeOnlyConfigurationName()`

iamdanfox:
I did a pTML and sanity checked this on gradle-revapi 1.0.0, and got the desired warning:

```
> Task :checkJUnitDependencies FAILED

* What went wrong:
Execution failed for task ':checkJUnitDependencies'.
> Tests may be silently not running! Spock dependency detected (which uses a JUnit4 Runner under the hood). Please add the following:

      testRuntimeOnly 'org.junit.vintage:junit-vintage-engine'
```

dansanduleac (approved):

PR #952, opened 2019-10-10
Fix error-prone `UnnecessaryParentheses` by default

carterkozak:
I've used this fix frequently on our largest internal projects
without any negative consequences. Fixing it by default will
resolve some compiler warnings and improve compilation log SNR.

## After this PR
==COMMIT_MSG==
Fix error-prone `UnnecessaryParentheses` by default
==COMMIT_MSG==

## Possible downsides?
- code changes stop excavator baseline upgrades auto-merging, which may hamper getting other baseline functionality onto develop.


iamdanfox (approved):
LGTM, have just added a possible downsides I can think of - still net positive!

PR #957, opened 2019-10-11
Implement Error Prone `ThrowError` to discourage throwing Errors in production code

### LINKED ISSUE #956: Disallow throwing AssertionError
## What happened?

Some code threw an AssertionError. It's not immediately obvious to all developers that neither `} catch (RuntimeException e) {` nor `} catch (Exception e) {` catch `Error`, so it bypassed handling code, resulting in a broken application state.

We shouldn't allow AssertionError to be thrown, not in main sources at least. Ideally we would take this a step farther and disallow all Error implementations from being thrown because it's relatively common that libraries fail to handle them properly.

### Automated Fix?

We could replace AssertionError with IllegalStateException/SafeIllegalStateException depending on whether or not the message is a compile-time constant.

carterkozak:
fix #956
==COMMIT_MSG==
Errors are often handled poorly by libraries resulting in unexpected
behavior and resource leaks. It's not obvious that 'catch (Exception e)'
does not catch Error.

This check  is intended to be advisory - it's fine to
`@SuppressWarnings("ThrowError")` in certain cases, but is usually not
recommended unless you are writing a testing library that throws
AssertionError.
==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ThrowError.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ThrowError.java
@@ -1,0 +1,66 @@
+/*
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.baseline.errorprone;
+
+import com.google.auto.service.AutoService;
+import com.google.errorprone.BugPattern;
+import com.google.errorprone.VisitorState;
+import com.google.errorprone.bugpatterns.BugChecker;
+import com.google.errorprone.matchers.Description;
+import com.google.errorprone.util.ASTHelpers;
+import com.sun.source.tree.ExpressionTree;
+import com.sun.source.tree.NewClassTree;
+import com.sun.source.tree.ThrowTree;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "ThrowError",
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
+        severity = BugPattern.SeverityLevel.WARNING,
+        summary = "Prefer throwing a RuntimeException rather than Error. Errors are often handled poorly by libraries "
+                + "resulting in unexpected behavior and resource leaks. It's not obvious that "
+                + "'catch (Exception e)' does not catch Error.\n"
+                + "Errors are normally thrown by the JVM when the system, not just the application, "
+                + "is in a bad state. For example, LinkageError is thrown by the JVM when it encounters "
+                + "incompatible classes, and OutOfMemoryError when allocations fail. These should be "
+                + "less common and handled differently from application failures.\n"
+                + "This check  is intended to be advisory - it's fine to @SuppressWarnings(\"ThrowError\") "
+                + "in certain cases, but is usually not recommended unless you are writing a testing library "
+                + "that throws AssertionError.")
+public final class ThrowError extends BugChecker implements BugChecker.ThrowTreeMatcher {
+
+    private static final String ERROR_NAME = Error.class.getName();
+
+    @Override
+    public Description matchThrow(ThrowTree tree, VisitorState state) {
+        ExpressionTree expression = tree.getExpression();
+        if (expression instanceof NewClassTree) {
+            NewClassTree newClassTree = (NewClassTree) expression;
+            if (ASTHelpers.isCastable(
+                    ASTHelpers.getType(newClassTree.getIdentifier()),
+                    state.getTypeFromString(ERROR_NAME),
+                    state)
+                    // Don't discourage developers from testing edge cases involving Errors.
+                    // It's also fine for tests throw AssertionError internally in test objects.
+                    && !TestCheckUtils.isTestCode(state)) {
+                return describeMatch(tree);
+            }
+        }
+        return Description.NO_MATCH;
+    }
+}
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ThrowError.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ThrowError.java
@@ -20,11 +20,18 @@ import com.google.auto.service.AutoService;
 import com.google.errorprone.BugPattern;
 import com.google.errorprone.VisitorState;
 import com.google.errorprone.bugpatterns.BugChecker;
+import com.google.errorprone.fixes.SuggestedFix;
+import com.google.errorprone.fixes.SuggestedFixes;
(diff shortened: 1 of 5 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ThrowError.java
```diff
@@ -0,0 +1,66 @@
...
+        if (expression instanceof NewClassTree) {
+            NewClassTree newClassTree = (NewClassTree) expression;
+            if (ASTHelpers.isCastable(
+                    ASTHelpers.getType(newClassTree.getIdentifier()),
+                    state.getTypeFromString(ERROR_NAME),
+                    state)
+                    // Don't discourage developers from testing edge cases involving Errors.
+                    // It's also fine for tests throw AssertionError internally in test objects.
+                    && !TestCheckUtils.isTestCode(state)) {
+                return describeMatch(tree);
```
carterkozak: Do we want to provide a suggested fix migrating specifically AssertionError to SafeIllegalStateException/IllegalStateException?

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ThrowError.java
```diff
@@ -0,0 +1,110 @@
...
+        }
+        NewClassTree newClassTree = (NewClassTree) expression;
+        Type throwableType = ASTHelpers.getType(newClassTree.getIdentifier());
+        if (!ASTHelpers.isCastable(
+                throwableType,
+                state.getTypeFromString(ERROR_NAME),
+                state)
+                // Don't discourage developers from testing edge cases involving Errors.
+                // It's also fine for tests throw AssertionError internally in test objects.
+                || TestCheckUtils.isTestCode(state)) {
```
carterkozak, line 72: I'd like to improve `isTestCode` to iterate over imports and return true if we import anything from junit/assertj/mockito. Several test utilities have non-test imports so we can avoid more potential noise.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ThrowError.java
```diff
@@ -0,0 +1,110 @@
...
+        name = "ThrowError",
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
+        severity = BugPattern.SeverityLevel.WARNING,
+        summary = "Prefer throwing a RuntimeException rather than Error. Errors are often handled poorly by libraries "
+                + "resulting in unexpected behavior and resource leaks. It's not obvious that "
+                + "'catch (Exception e)' does not catch Error.\n"
+                + "Errors are normally thrown by the JVM when the system, not just the application, "
+                + "is in a bad state. For example, LinkageError is thrown by the JVM when it encounters "
+                + "incompatible classes, and OutOfMemoryError when allocations fail. These should be "
```
iamdanfox: So we actually set `-XX:+CrashOnOutOfMemoryError` to try and avoid people trying to catch these and leaving their service in a weird state - shall we use another example (like NoSuchMethodError / NoClassDefFoundError)?

https://github.com/palantir/sls-packaging/blob/develop/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.java#L59

carterkozak: Sure thing -- note that there are some types of ooms that do not actually trigger crash-on-oom (iirc direct memory ooms)

gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineErrorProneExtension.java
```diff
@@ -34,6 +34,7 @@
             "PreferSafeLoggingPreconditions",
             "StrictUnusedVariable",
             "StringBuilderConstantParameters",
+            "ThrowError",
```
iamdanfox, line 37: I'm a little worried promoting this straight away may lead to unexpected/unintended code rewriting - could we do a trial run on a few repos first?

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/ThrowErrorTest.java
```diff
@@ -0,0 +1,134 @@
...
+
+package com.palantir.baseline.errorprone;
+
+import com.google.errorprone.BugCheckerRefactoringTestHelper;
+import com.google.errorprone.CompilationTestHelper;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.parallel.Execution;
+import org.junit.jupiter.api.parallel.ExecutionMode;
+
+@Execution(ExecutionMode.CONCURRENT)
```
iamdanfox, line 25: 😍

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ThrowError.java
```diff
@@ -0,0 +1,110 @@
...
+                throwableType,
+                state.getTypeFromString(ERROR_NAME),
+                state)
+                // Don't discourage developers from testing edge cases involving Errors.
+                // It's also fine for tests throw AssertionError internally in test objects.
+                || TestCheckUtils.isTestCode(state)) {
+            return Description.NO_MATCH;
+        }
+        return buildDescription(tree)
+                .addFix(generateFix(newClassTree, state))
```
iamdanfox, line 76: I'm a tiny bit worried this auto-fix is gonna funnel states which would previously crash-the-server into other existing error-handling flows (e.g. if a codebases catches RuntimeException further up), leading to a server limping along when it should really die quickly?

```diff
-throw new AssertionError("Unknown classification operator");
+throw new IllegalStateException("Unknown classification operator");
```

carterkozak: None of our servers really do that -- when we throw error in a conjure service we allow the server to log it and return an empty 500 response. We should probably log it and return 500 instead, adding an errorId, but the server doesn't react differently otherwise.

iamdanfox (approved):
For greenfield codebases, making people think twice about throwing AssertionErrors seems unambiguously better. I _think_ the `*Error` -> `IllegalStateException` auto-fix is gonna be net positive in existing codebases, but it definitely comes with a bit more risk

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ThrowError.java
carterkozak: I do think we want developers to review these migration without mindlessly merging, but there are a lot of cases where error isn't handled properly (we had two p0s last week from this)

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ThrowError.java
```diff
@@ -0,0 +1,110 @@
...
+        if (!ASTHelpers.isSameType(
+                throwableType,
+                state.getTypeFromString(AssertionError.class.getName()),
+                state)) {
+            return Optional.empty();
+        }
+        List<? extends ExpressionTree> arguments = newClassTree.getArguments();
+        if (arguments.isEmpty()) {
+            SuggestedFix.Builder fix = SuggestedFix.builder();
+            String qualifiedName = SuggestedFixes.qualifyType(state, fix, IllegalStateException.class.getName());
```
markelliot, line 93: should we use SafeIllegalStateException and maybe hand it a message of "assertion failure"?

carterkozak: I matched semantics of the PreferSafeLoggablExceptions check which uses the non-logsafe exception type when there is no message. I don't have a strong preference, but I'm not sure using the safe type provides any additional information.

PR #960, opened 2019-10-13
Implement Error Prone `Slf4jLevelCheck`

### LINKED ISSUE #947: error-prone check that slf4j level checks agree with logging
## What happened?

Over time, refactoring often results in level checks that do not agree with levels:

```java
if (log.isWarnEnabled()) {
    // Expensive value computation occurs even though trace logging is not enabled.
    log.trace("value is: {}", getExpensiveValue());
}
```

```java
if (log.isInfoEnabled()) {
    // Warn logging is missed entirely unless info level logging is enabled
    log.warn("uh oh!");
}
```

We should detect mismatches and fail compilation. Tests for logging are uncommon, and it's difficult to discover when logging is missing without looking for a precise line, making static analysis the ideal place to validate.

carterkozak:
fix #947: Validate that slf4j level checks agree with contained logging.

## Before this PR
Logging may not appear when levels are changed. It's easy to break these when we refactor code, and not somewhere we should spend time in code-review because automation is capable of catching it.

## After this PR
==COMMIT_MSG==
Implement Error Prone `Slf4jLevelCheck` to validate that slf4j level checks agree with contained logging.
==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/Slf4jLevelCheck.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/Slf4jLevelCheck.java
@@ -1,0 +1,183 @@
+/*
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.baseline.errorprone;
+
+import com.google.auto.service.AutoService;
+import com.google.common.base.CaseFormat;
+import com.google.errorprone.BugPattern;
+import com.google.errorprone.BugPattern.SeverityLevel;
+import com.google.errorprone.VisitorState;
+import com.google.errorprone.bugpatterns.BugChecker;
+import com.google.errorprone.bugpatterns.BugChecker.IfTreeMatcher;
+import com.google.errorprone.fixes.SuggestedFixes;
+import com.google.errorprone.matchers.Description;
+import com.google.errorprone.matchers.Matcher;
+import com.google.errorprone.matchers.method.MethodMatchers;
+import com.sun.source.tree.ExpressionTree;
+import com.sun.source.tree.IfTree;
+import com.sun.source.tree.MethodInvocationTree;
+import com.sun.source.tree.ParenthesizedTree;
+import com.sun.source.util.SimpleTreeVisitor;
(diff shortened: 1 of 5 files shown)


carterkozak:
I tested this on a large internal project, two changes I’d like to make before merging:
1. Ignore conditionals with else blocks, they tend to be complicated and I’d want to have stronger conviction before requesting that they make changes. While in most cases I think they could be written better, there aren’t many of them, nor have I encountered many that are incorrect
2. Ignore levels from logging statements in catch blocks. I’ve encountered a few places that catch exceptions while generating trace data and log those failures at warn/error.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/Slf4jLevelCheck.java
```diff
@@ -0,0 +1,183 @@
...
+        return buildDescription(tree)
+                .addFix(SuggestedFixes.renameMethodInvocation(
+                        levelCheckInvocation, mostSevere.levelCheckMethodName(), state))
+                .build();
+    }
+
+    @SuppressWarnings("PreferSafeLoggableExceptions")
+    private static LogLevel levelCheckLogLevel(MethodInvocationTree tree, VisitorState state) {
+        for (LogLevel level : LogLevel.values()) {
+            if (level.matchesLevelCheck(tree, state)) {
```
iamdanfox, line 88: 😮A ckozak snippet where we allocate an iterator and loop instead of pre-building a composed function! _(just teasing)_

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/Slf4jLevelCheck.java
```diff
@@ -0,0 +1,183 @@
...
+import java.util.Locale;
+import java.util.Optional;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "Slf4jLevelCheck",
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
+        severity = SeverityLevel.ERROR,
+        summary = "Slf4j logger.is[Level]Enabled level must match the most severe log statement")
```
iamdanfox, line 45: `s/logger.is[Level]Enabled/log.is[Level]Enabled/` just so people don't get the wrong ideas about what we recommend as best-practise for naming those variables

iamdanfox:
Interesting I haven't really encountered this problem in any Foundry repos - I appreciate gotham is a bigger codebase and has different performance constraints, just wondering if it would even be reasonable to discourage the `isFooEnabled()` method calls entirely in foundry code??

iamdanfox (approved):

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/Slf4jLevelCheck.java
carterkozak: Good call, updating shortly

carterkozak:
erm, hit merge when ready on the wrong one, sorry. update incoming.

> Interesting I haven't really encountered this problem in any Foundry repos.

It's hard to detect, because you have to be looking for a specific log line. I think it's better to lean on automation in this case. discouraging isFooEnabled is fine in most cases, unless we're in a tight loop, so I don't want to be too prescriptive.

PR #962, opened 2019-10-14
Slf4jLevelCheck additional work

carterkozak:
==COMMIT_MSG==
Slf4jLevelCheck additional work
==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/Slf4jLevelCheck.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/Slf4jLevelCheck.java
@@ -27,6 +27,7 @@ import com.google.errorprone.fixes.SuggestedFixes;
 import com.google.errorprone.matchers.Description;
 import com.google.errorprone.matchers.Matcher;
 import com.google.errorprone.matchers.method.MethodMatchers;
+import com.sun.source.tree.CatchTree;
 import com.sun.source.tree.ExpressionTree;
 import com.sun.source.tree.IfTree;
 import com.sun.source.tree.MethodInvocationTree;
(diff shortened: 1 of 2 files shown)


carterkozak:
should have finished my coffee before adding merge-when-ready... I had meant to merge the test utils fix, but that already went in :man_facepalming:

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/Slf4jLevelCheck.java
```diff
@@ -118,6 +122,23 @@ private ConditionVisitor() {
...
+        // is the most common case.
+        @Override
+        public Optional<MethodInvocationTree> visitBinary(BinaryTree node, VisitorState state) {
+            if (node.getKind() != Tree.Kind.CONDITIONAL_AND) {
+                return Optional.empty();
+            }
+            Optional<MethodInvocationTree> lhs = node.getLeftOperand().accept(this, state);
+            Optional<MethodInvocationTree> rhs = node.getRightOperand().accept(this, state);
+            // If there are level checks on both sides, bail
+            if (lhs.isPresent() && rhs.isPresent()) {
```
iamdanfox, line 137: Are these `.isPresent()` calls definitely safe even when the visitCatch on 160 returns `null`? Does it always make a reduce call under the hood??

iamdanfox (approved):

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/Slf4jLevelCheck.java
carterkozak: 160 is a method on the MostSevereLogStatementScanner TreeScanner, this is the ConditionVisitor (SimpleVisitor).

I wanted to use Optional.empty in the treescanner, but it's a bit more difficult because it doesn't allow a configurable default value.

carterkozak:
running this on good ol' large-internal before merging.

PR #970, opened 2019-10-15
fix #969: Implement error prone ReverseDnsLookup for unexpected reverse dns lookups

### LINKED ISSUE #969: error prone check to discourage InetSocketAddress.getHostName
## What happened?
InetSocketAddress.getHostName may do a reverse DNS lookup, which is a potentially unexpected network call.

We should provide a warning to discourage use which can be suppressed when when this behavior is acknowledged and desired.

carterkozak:
==COMMIT_MSG==
Implement error prone ReverseDnsLookup for unexpected reverse dns lookups

Calling address.getHostName may result in a DNS lookup which is a network request,
making the invocation significantly more expensive than expected depending on the
environment.
This check  is intended to be advisory - it's fine to
@SuppressWarnings("ReverseDnsLookup") in certain cases, but is usually not
recommended.
==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ReverseDnsLookup.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ReverseDnsLookup.java
@@ -35,8 +35,8 @@ import java.net.InetSocketAddress;
         link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
         linkType = BugPattern.LinkType.CUSTOM,
         severity = BugPattern.SeverityLevel.WARNING,
-        summary = "Calling address.getHostName may result in a reverse DNS lookup which is a network request, making the "
-                + "invocation significantly more expensive than expected depending on the environment.\n"
+        summary = "Calling address.getHostName may result in a reverse DNS lookup which is a network request, making "
+                + "the invocation significantly more expensive than expected depending on the environment.\n"
                 + "This check is intended to be advisory - it's fine to @SuppressWarnings(\"ReverseDnsLookup\") "
                 + "in certain cases, but is usually not recommended.")
 public final class ReverseDnsLookup extends BugChecker implements BugChecker.MethodInvocationTreeMatcher {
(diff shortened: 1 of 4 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ReverseDnsLookup.java
```diff
@@ -0,0 +1,72 @@
...
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "ReverseDnsLookup",
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
+        severity = BugPattern.SeverityLevel.WARNING,
+        summary = "Calling address.getHostName may result in a DNS lookup which is a network request, making the "
+                + "invocation significantly more expensive than expected depending on the environment.\n"
+                + "This check  is intended to be advisory - it's fine to @SuppressWarnings(\"ReverseDnsLookup\") "
```
raiju: Nit: double-space `check  is`

schlosna:
```suggestion
                + "This check is intended to be advisory - it's fine to @SuppressWarnings(\"ReverseDnsLookup\") "
```

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ReverseDnsLookup.java
```diff
@@ -0,0 +1,72 @@
...
+import java.net.InetAddress;
+import java.net.InetSocketAddress;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "ReverseDnsLookup",
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
+        severity = BugPattern.SeverityLevel.WARNING,
+        summary = "Calling address.getHostName may result in a DNS lookup which is a network request, making the "
```
iamdanfox:
```suggestion
        summary = "Calling address.getHostName may result in a reverse DNS lookup which is a network request, making the "
```

iamdanfox (approved):

PR #975, opened 2019-10-16
Correctly rewrite verifyZeroInteractions

dansanduleac:
## Before this PR

We were rewriting `verifyZeroInteractions` to `verifyNoInteractions`, which has different behaviour.
It expects that there were never any interactions, ignoring previously acknowledged interactions.

## After this PR

==COMMIT_MSG==
The deprecated `verifyZeroInteractions` now gets rewritten to `verifyNoMoreInteractions`, which has the same behaviour.
==COMMIT_MSG==


iamdanfox (approved):
Yep - seen some funky test failures because of this. Also confusing that their own javadoc is inconsistent!

pkoenig10:
Didn't we explicitly make the decision to replace `verifyZeroInteractions` with `verifyNoInteractions`? See https://github.com/palantir/gradle-baseline/pull/924#issuecomment-537682331.

PR #976, opened 2019-10-16
Add checkstyle rule for double boolean negation

dtobin:
## Before this PR
Use of !! is almost certainly a typo. This typo was the cause of a bug that took a while to track down in one of our repos.

## After this PR
==COMMIT_MSG==
Add a checkstyle rule to prevent double boolean negation (!!).
==COMMIT_MSG==

## Possible downsides?
RegexpSinglelineJava doesn't exclude comments, so if someone gets a bit over-excited in a comment this rule may hit their excessive exclamation marks.

### DIFF
--- a/.baseline/checkstyle/checkstyle.xml
+++ b/.baseline/checkstyle/checkstyle.xml
@@ -383,6 +383,10 @@
             <property name="format" value="(void setUp\(\))|(void setup\(\))|(void setupStatic\(\))|(void setUpStatic\(\))|(void beforeTest\(\))|(void teardown\(\))|(void tearDown\(\))|(void beforeStatic\(\))|(void afterStatic\(\))"/>
             <property name="message" value="Test setup/teardown methods are called before(), beforeClass(), after(), afterClass(), but not setUp, teardown, etc."/>
         </module>
+        <module name="RegexpSinglelineJava">
+            <property name="format" value="!!"/>
+            <property name="message" value="Unnecessary double negation (!!)."/>
+        </module>
         <module name="RightCurly"> <!-- Java Style Guide: Nonempty blocks: K & R style -->
             <property name="option" value="same"/>
             <property name="tokens" value="LITERAL_TRY, LITERAL_CATCH, LITERAL_FINALLY, LITERAL_IF, LITERAL_ELSE, LITERAL_DO"/>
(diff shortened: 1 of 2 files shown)


iamdanfox:
I'd actually prefer not to use checkstyle for this kind of thing because if there are any occurrences out in the wild, a human will have to go and fix them.  Can we make this either a refaster/error-prone check instead?

dtobin:
Ah that's interesting. I can think of two cases where you'd have `!!` in source code - an intentional double negation that we should refactor to remove entirely, and an unintentional double negation that was supposed to a single `!` but wasn't caught by tests. The second case is mainly the one that I wanted to catch, and in that case you need a human to rethink the code and verify a fix. Is there any room in the ecosystem for a check like that?

.baseline/checkstyle/checkstyle.xml
```diff
@@ -383,6 +383,10 @@
             <property name="format" value="(void setUp\(\))|(void setup\(\))|(void setupStatic\(\))|(void setUpStatic\(\))|(void beforeTest\(\))|(void teardown\(\))|(void tearDown\(\))|(void beforeStatic\(\))|(void afterStatic\(\))"/>
             <property name="message" value="Test setup/teardown methods are called before(), beforeClass(), after(), afterClass(), but not setUp, teardown, etc."/>
         </module>
+        <module name="RegexpSinglelineJava">
+            <property name="format" value="!!"/>
```
iamdanfox, line 387: I actually just grepped my home directory for java files containing `!!` and got a ton of matches, all of which I think are harmless (because they're inside strings or comments), e.g.:

```
assertNotNull("TEST_PROJECT should not be null!!!", TEST_PROJECT);
```
```
/* !!!!!!!!!! WARNING !!!!!!!!!!
```

```
return message; // Safe to log!!
```
```
if (!reordered.startsWith("!!")) {
```
```
pqlAssertEquals("2", "(to-lazy-list [1,2,3]) !! 1");
```

iamdanfox (changes requested):
I think this is going to have too many false positives, sorry.

I'd be OK with an auto-fixing error-prone rule that changes `!!foo` -> `foo` (so preserving the original semantics)

dtobin:
No worries, let's close then.

PR #978, opened 2019-10-16
ReadReturnValueIgnored: Check that read operation results are not ignored

carterkozak:
==COMMIT_MSG==
ReadReturnValueIgnored: Check that read operation results are not ignored
==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ReadReturnValueIgnored.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ReadReturnValueIgnored.java
@@ -1,0 +1,145 @@
+/*
+ * (c) Copyright 2018 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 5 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ReadReturnValueIgnored.java
```diff
@@ -0,0 +1,145 @@
...
+import com.sun.source.tree.MethodInvocationTree;
+import com.sun.source.tree.Tree;
+import java.io.InputStream;
+import java.io.RandomAccessFile;
+import java.util.List;
+import java.util.Optional;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "ReadReturnValueIgnored",
```
carterkozak, line 45: Open to suggestions for a better name. I considered using separate checks for InputStream and RandomAccessFile or read skip operations, but decided against it because these are the same bug: you're not consuming as much as you think.

nathklei: Maybe "Byte(s)ReadReturnValueIgnored"?

carterkozak: I could see applying this to Reader as well, which deals with chars rather than bytes

ferozco: I think the name is reasonable

ferozco (approved):
Seems like a nice check to have. My only concern is that the fix forces you to have a guava dependency since I think this is the first check to do so. Most of our projects already have a dependency so its probably not a big deal, but I thought I'd just flag

carterkozak:
That's fair. Since the check is applied at ERROR level, your case results in a different kind of compilation failure in that edge case (failure to find the guava class instead of `ReadReturnValueIgnored`) but either way we need a human to fix the bug.

PR #981, opened 2019-10-17
Restrict migration to safe-logging to source sets that have the library

dansanduleac:
## Before this PR

Baseline keeps opening PRs that migrate people's preconditions and exceptions to safe-logging, but this breaks when those libraries actually aren't on the classpath of the respective source set being migrated.

## After this PR
==COMMIT_MSG==
Stop migrating source sets to safe-logging, unless they already have the requisite library (`com.palantir.safe-logging:preconditions`).
==COMMIT_MSG==

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
@@ -19,6 +19,8 @@ package com.palantir.baseline.plugins;
 import com.google.common.base.Joiner;
 import com.google.common.base.Splitter;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Iterables;
+import com.google.common.collect.MoreCollectors;
 import com.palantir.baseline.extensions.BaselineErrorProneExtension;
 import com.palantir.baseline.tasks.CompileRefasterTask;
 import java.io.File;
@@ -37,12 +39,16 @@ import org.gradle.api.JavaVersion;
 import org.gradle.api.Plugin;
 import org.gradle.api.Project;
 import org.gradle.api.artifacts.Configuration;
+import org.gradle.api.artifacts.component.ModuleComponentIdentifier;
 import org.gradle.api.file.FileCollection;
 import org.gradle.api.file.RegularFile;
 import org.gradle.api.logging.Logger;
 import org.gradle.api.logging.Logging;
 import org.gradle.api.plugins.ExtensionAware;
+import org.gradle.api.plugins.JavaPluginConvention;
 import org.gradle.api.provider.Provider;
+import org.gradle.api.specs.Spec;
+import org.gradle.api.tasks.SourceSet;
 import org.gradle.api.tasks.compile.JavaCompile;
 import org.gradle.api.tasks.javadoc.Javadoc;
 import org.gradle.api.tasks.testing.Test;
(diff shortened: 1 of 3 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
```diff
@@ -250,10 +254,37 @@ private static void configureErrorProneOptions(
...
+                        && !compileClasspath.filter(file -> file.getName().startsWith("preconditions-")).isEmpty();
+        return check -> {
+            if (!hasPreconditions && check.equals("PreferSafeLoggingPreconditions")) {
+                log.info(
+                        "Disabling check PreferSafeLoggingPreconditions as 'com.palantir.safe-logging:safe-logging' "
+                                + "missing from source set for {}",
+                        compileClasspath);
+                return false;
+            }
+            if (!hasSafeLogging && check.equals("PreferSafeLoggableExceptions")) {
```
carterkozak: These live in the preconditions library, we only need the `hasPreconditions` check

iamdanfox: Rather than hard coding these deep in a method, could we have a map from checkName -> onlyIf predicate?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
```diff
@@ -250,10 +254,37 @@ private static void configureErrorProneOptions(
...
                 return false;
             }
-            return true;
+            return filterOutPreconditions.test(check);
         });
     }
 
+    /** Filters out preconditions checks if the required libraries are not on the classpath. */
+    public static Predicate<String> filterOutPreconditions(FileCollection compileClasspath) {
+        boolean hasSafeLogging = !compileClasspath.filter(file -> file.getName().startsWith("safe-logging-")).isEmpty();
```
iamdanfox: can we not get the configuration (containing raw dependencies) instead of having to work on files?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
dansanduleac: ah fair, but i'd still like to check for both since I don't trust trust just the check on a jar starting with `preconditions-`

dansanduleac: Well, the predicate would have to take into account the classpath. I'm not sure if it's worth abstracting this out just yet

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
dansanduleac: We can guess it but I couldn't find a direct way to extract this via the API if we start with the JavaCompile task

dansanduleac: nvm, I could do a pass through the source sets and find the one matching the `compileJava` task name

iamdanfox (approved):
Looks plausible, but would want to see this actually work (either pTML or a test)!

dansanduleac:
Tested on palantir-java-format:idea-plugin with/without the dependency, I see the expected log messages and behaviour is correct.

PR #982, opened 2019-10-17
Improvement: palantir-java-format now reflows strings

dansanduleac:
## Before this PR

Was not reflowing strings when formatting, hence some strings would end up exceeding the width limit.

## After this PR
==COMMIT_MSG==
For users who opted into palantir-java-format, we now reflow strings and reorder imports.
==COMMIT_MSG==

Also removed all of the googlejavaformat integration tests, as
1) palantir-java-format already encapsulates them and more, and
2) we don't really want to test the eclipse formatter anymore since we're moving away from it.

Instead, renamed some of the directories so that now the integration tests make sense:
```
< formatter-in
> eclipse-formatter-expected
> palantirjavaformat-expected
```


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineFormat.java
```diff
@@ -128,14 +122,10 @@ static Path eclipseConfigFile(Project project) {
...
                 JavaFormatterOptions.builder().style(Style.PALANTIR).build();
-        private final FormatterService formatterService =
-                Iterables.getOnlyElement(ServiceLoader.load(FormatterService.class));
 
         @Override
         public String apply(String input) throws Exception {
-            ImmutableList<Replacement> replacements = formatterService.getFormatReplacements(
-                    OPTIONS, input, ImmutableList.of(Range.closedOpen(0, input.length())));
-            return JavaOutput.applyReplacements(input, replacements);
+            return Formatter.createFormatter(OPTIONS).formatSourceAndFixImports(input);
```
iamdanfox, line 128: Ok so we're just bypassing the `palantir-java-format-api`, but I guess we were already because we used the JavaOutput class.

iamdanfox (approved):
Let's also drop one >120char string into the integration tests?

Issue #847, opened 2019-09-18
CheckImplicitDependenciesTask can list dependencies which are not required for compile


esword:
## What happened?
The maven ASMDependencyAnalyzer will return any string literal as a possible class file.  So if you have a literal for a class loaded by reflection with a newInstance call, the CheckImplicitDependenciesTask may report that you should add it even though you don't want it on compile classpath.  This can result in excessive dependencies or, in one case, a circular dependency between modules in a project that caused gradle to fail.

This is not easy to reproduce because if you just run the task once, then only the existing dependencies for a module/project are loaded into the Indexes, so the Indexes.classToDependency call cannot locate anything that is not already listed as a dependency or included transitively.  However, if the task is run on a multi-module project, the Indexes contain information for all dependencies for all projects since they are static, and an incorrect association can be made.

This is likely an uncommon case in small projects.  Mine is just so large and with so many strange connections that I hit all sorts of problems.  I bring this up just so you know about it and can possibly add to the docs so that users of the task are aware.

## What did you want to happen?
Need more confidence that dependencies are actually used.  There are two approaches:

1. Don't cache dependencies - make the Indexes non-static.  This may seem wasteful, but it doesn't slow things down a ton.  This also reduces memory load so you won't run out of memory as happens in pgdev.

2. Use a better analyzer.  I have started looking at [jdeps] (https://docs.oracle.com/javase/9/tools/jdeps.htm#JSWOR690) for a related project I have.  There is a gradle plugin for jdeps as well, but it is overly aggressive and scans configs recursively.  I need to shift to other things for a while, but will touch base if I incorporate it.  The other advantage of jdeps is that the verbose mode lists individual classes to dependencies, so you can tackle things like https://github.com/palantir/gradle-baseline/issues/614.

PR #983, opened 2019-10-17
Tasks that use jdeps analyzer

esword:
## Before this PR
Maven dependency analyzer has several shortcomings, including those that lead to:
https://github.com/palantir/gradle-baseline/issues/847
https://github.com/palantir/gradle-baseline/issues/614

Plus, the current analyzer tasks cache everything which can lead to OoMs with large projects.

## After this PR
==COMMIT_MSG==
New dependency analyzer yields cleaner results and tasks store output in files so better
caching and analysis can be reused for other purposes
==COMMIT_MSG==

## Possible downsides?
Possibly slightly slower since don't cache contents of all jars across projects.


esword:
I have not moved over the existing tasks to use the new analysis file yet.  The file is useful on it's own if we want to make that a separate PR, or can do as part of this one.

gradle-baseline-java/build.gradle
```diff
@@ -58,6 +61,10 @@ gradlePlugin {
             id = 'com.palantir.baseline-exact-dependencies'
             implementationClass = 'com.palantir.baseline.plugins.BaselineExactDependencies'
         }
+        baselineDependencyPluginv2 {
+            id = 'com.palantir.baseline-dependencies-v2'
+            implementationClass = 'com.palantir.baseline.plugins.BaselineDependencyPluginv2'
```
CRogers, line 66: We should just replace it entirely if we have confidence it's good. If we have lower confidence we should take a gradle property to excavate it out rather than an entirely new task (this allows us to track adoption and easily excavate the changes).

gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/dependencies/DependencyAnalysisTask.java
```diff
@@ -0,0 +1,167 @@
...
+    public final DirectoryProperty getDotFileDir() {
+        return dotFileDir;
+    }
+
+    @OutputFile
+    public final RegularFileProperty getReportFile() {
+        return report;
+    }
+
+    public static final class ReportContent {
```
CRogers, line 237: I would use immutables here as it will build this class for you and handle the yaml dumping by generating the correct jackson classes as well, meaning we don't need snakeyaml.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/dependencies/DependencyAnalyzer.java
```diff
@@ -0,0 +1,230 @@
...
+
+    /**
+     * Dependencies that are required but are not directly declared by the project.
+     */
+    public Set<String> getImplicitDependencies() {
+        return implicitDependencies;
+    }
+
+    /**
+     * Dependencies that are declared in one of the given configurations but not used in byte code.  This will only
```
CRogers: Does it detect things like `Class.forName(...)` or other dynamic class loading?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/dependencies/DependencyUtils.java
```diff
@@ -0,0 +1,164 @@
...
+import org.gradle.api.artifacts.ModuleVersionIdentifier;
+import org.gradle.api.artifacts.ProjectDependency;
+import org.gradle.api.artifacts.ResolvedArtifact;
+import org.gradle.api.artifacts.component.ProjectComponentIdentifier;
+import org.gradle.api.file.Directory;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.yaml.snakeyaml.Yaml;
+
+public final class DependencyUtils {
```
CRogers, line 42: I think most of these public classes should be package-private.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/dependencies/DependencyUtils.java
```diff
@@ -0,0 +1,164 @@
...
+    public static boolean isProjectDependency(String artifactName) {
+        return artifactName.startsWith("project :")
+                || artifactName.startsWith("project (')")
+                // a colon in the name (when the name doesn't start with "project" indicates it is a jar dependency
+                // e.g. group:id
+                || !artifactName.contains(":");
+    }
+
+    static void writeDepReport(File file, DependencyAnalysisTask.ReportContent content) {
+        Yaml yaml = new Yaml();
```
CRogers, line 130: We should try to use jackson rather than snakeyaml here to be consistent with everything else that uses jackson.

gradle-baseline-java/build.gradle
```diff
@@ -17,6 +17,9 @@ dependencies {
     compile 'org.github.ngbinh.scalastyle:gradle-scalastyle-plugin_2.11'
     implementation 'com.palantir.javaformat:palantir-java-format-spi'
     implementation 'com.palantir.javaformat:gradle-palantir-java-format'
+    implementation 'com.paypal.digraph:digraph-parser'
+    implementation 'org.zeroturnaround:zt-exec'
```
CRogers, line 21: We should avoid taking dependencies on extra stuff if it's not necessary. The java process API may be nasty but it's workable. Since we put everything into gradle `buildscript` internally there's great capacity to have breaking changes between gradle plugins wreak, especially from something as widely used as baseline.

gradle-baseline-java/build.gradle
```diff
@@ -17,6 +17,9 @@ dependencies {
...
+    implementation 'org.yaml:snakeyaml'
```
CRogers, line 22: Can we avoid snakeyaml and use jackson everywhere?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineDependencyPluginv2.java
```diff
@@ -0,0 +1,143 @@
...
+
+    @Override
+    public void apply(Project project) {
+        project.getPluginManager().withPlugin("java", plugin -> {
+            SourceSetContainer sourceSets = project.getConvention()
+                    .getPlugin(JavaPluginConvention.class)
+                    .getSourceSets();
+
+            createTasksForSourceSet(project, sourceSets.getByName(SourceSet.MAIN_SOURCE_SET_NAME));
+            createTasksForSourceSet(project, sourceSets.getByName(SourceSet.TEST_SOURCE_SET_NAME));
```
CRogers, line 47: What about people using custom `integrationTest` or `integTest` source sets?

CRogers: I guess we can perhaps just ignore them.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineDependencyPluginv2.java
```diff
@@ -0,0 +1,143 @@
...
+        return project.getTasks().register(taskName, CheckUnusedDependenciesTaskv2.class, t -> {
+            t.setDescription("Verifies that project does not declare any dependencies for the "
+                    + sourceSet.getName() + " source set that it does not use.");
+            t.setGroup(GROUP_NAME);
+            t.getReportFile().set(analyzerTask.get().getReportFile());
+            // this is liberally applied to ease the Java8 -> 11 transition
+            t.getIgnored().add("javax.annotation:javax.annotation-api");
+            //Projects may apply this in compileOnly to work around
+            //https://github.com/immutables/immutables/issues/291
+            t.getIgnored().add("org.immutables:value::annotations");
```
CRogers, line 122: It might make sense to ignore all dependencies which are in `compileOnly` and also `annotationProcessor` as this happens to more libraries than just immutables, iirc.

CRogers (commented):
I didn't manage to allocated as much time as I wanted to review this, still need to test it out locally.

iamdanfox:
I think we may want to resuscitate this because the maven-dependency-analyzer can't parse bytecode produced by java13.

PR #988, opened 2019-10-22
Format diff task

iamdanfox:
## Before this PR

We are not yet running the formatter inside excavator.

## After this PR
==COMMIT_MSG==
Run `./gradlew formatDiff -Pcom.palantir.baseline-format.palantir-java-format` to reformat the relevant sections of any uncommitted changed Java files (relies on `git diff -U0 HEAD` under the hood)
==COMMIT_MSG==

https://github.com/palantir/palantir-java-format/pull/39

## Possible downsides?
- in obscure cases, the formatter can still produce code that doesn't pass checkstyle (which is why we disable the [indentation thing](https://github.com/palantir/gradle-baseline/pull/987))


gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineFormatIntegrationTest.groovy
```diff
@@ -221,4 +221,41 @@ class BaselineFormatIntegrationTest extends AbstractPluginTest {
...
+
+        "git add .".execute(Collections.emptyList(), projectDir).waitFor()
+        "git commit -m Commit".execute(Collections.emptyList(), projectDir).waitFor()
+
+        file('src/main/java/Main.java').text = '''
+        class Main {
+            public static void crazyExistingFormatting  (  String... args) {
+                                        System.out.println("Reformat me please");
+            }
+        }
```
CRogers, line 259: I would make this example super minimal just so if the formatter changes any more it won't affect this test

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineFormat.java
```diff
@@ -45,6 +49,12 @@ public void apply(Project project) {
         this.project = project;
 
         project.getPluginManager().withPlugin("java", plugin -> {
+            project.getTasks().register("formatDiff", FormatDiffTask.class, task -> {
+                task.setDescription("Format only chunks of files that appear in git diff");
+                task.setGroup("Formatting");
+                task.onlyIf(t -> palantirJavaFormatterEnabled(project));
```
CRogers: Are we able to put these in the `FormatDiffTask` itself?

CRogers: Hmm probably makes sense not to actually

iamdanfox:
Ok so also trying this out locally:

-  I've discovered that it doesn't necessarily fix up imports. 
- might want some kind of slf4j logging to help diagnose this stuff??
- if a repo already uses `com.palantir.baseline-format.eclipse=true` then you get an error

CRogers (approved):

iamdanfox:
So we've actually pivoted and decided to create this task in palantir-java-format so it has a nice simple usability story independent of baseline. https://github.com/palantir/palantir-java-format/pull/48

CRogers:
This approach makes sense to me :+1:

iamdanfox:
I also approve this message :+1:

PR #989, opened 2019-10-22
Improvement: Load palantir-java-format dynamically

dansanduleac:
## Before this PR

The version of palantir-java-format used by spotless (through `com.palantir.baseline-format`), and the version configured in IntelliJ can differ, because they are sourced from two different configurations.

## After this PR
==COMMIT_MSG==
Load palantir-java-format dynamically from the same configuration set up by `com.palantir-java-format` which is also used to determine the version used by IntelliJ.
==COMMIT_MSG==

Depending on the implementation directly (`com.palantir.java-format:palantir-java-format`) is now prohibited, and attempting to use `com.palantir.baseline-format` when the implementation is on the gradle plugin class path will result in this error:

> Expected not be be able to load com.palantir.javaformat.java.Formatter via main class loader but was able to. Please ensure that `buildscript.configurations.classpath` doesn't depend on `com.palantir.javaformat:palantir-java-format`.


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/format/PalantirJavaFormatStep.java
```diff
@@ -0,0 +1,119 @@
...
+
+        State(Collection<String> deps, Provisioner provisioner) throws IOException {
+            this.jarState = JarState.from(deps, provisioner);
+        }
+
+        @SuppressWarnings({"unchecked", "rawtypes"})
+        FormatterFunc createFormat() throws Exception {
+            ClassLoader classLoader = jarState.getClassLoader();
+
+            Class<?> optionsClass = classLoader.loadClass(OPTIONS_CLASS);
```
dansanduleac: after https://github.com/palantir/palantir-java-format/pull/47, everything down from here becomes super simple and doesn't need reflection anymore: just service load the FormatterService from the classLoader.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineFormat.java
```diff
@@ -67,15 +63,24 @@ public void apply(Project project) {
                 if (eclipseFormattingEnabled(project) && palantirJavaFormatterEnabled(project)) {
                     throw new GradleException(
                             "Can't use both eclipse and palantir-java-format at the same time, please delete one of "
-                                    + ECLIPSE_FORMATTING + " or " + PJF_PROPERTY + " from your gradle.properties");
+                                    + ECLIPSE_FORMATTING
+                                    + " or "
+                                    + PJF_PROPERTY
+                                    + " from your gradle.properties");
```
dansanduleac, line 69: I reformatted the whole file using the formatter so a few other things have changed, sorry!

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineFormat.java
```diff
@@ -67,15 +63,24 @@ public void apply(Project project) {
...
                 }
 
                 if (eclipseFormattingEnabled(project)) {
                     java.eclipse().configFile(project.file(eclipseXml.toString()));
                 }
 
                 if (palantirJavaFormatterEnabled(project)) {
-                    java.customLazy("palantir-java-format", PalantirJavaFormatterFunc::new);
+                    Preconditions.checkState(
```
ferozco, line 77: who produces `java-format-provider`? This also forces users to apply java-format-provider first. Could we make it perform the logic in a `withPlugin()` block and then verify that it executed with an after evaluate?

ferozco (approved):

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineFormat.java
dansanduleac: It's produced in `com.palantir.javaformat:gradle-palantir-java-format`.
This is actually safe for now, as we ensure to apply `java-format-provider` in the main Baseline plugin.

PR #994, opened 2019-10-23
One more AssertjOptionalIsNotPresent case

iamdanfox:
I just saw another refactor produce this case, figured we might as well fix it!


carterkozak (approved):
nice, thanks Dan!

PR #997, opened 2019-10-24
DangerousThrowableMessageSafeArg disallows Throwables

carterkozak:
Update DangerousThrowableMessageSafeArg to disallow Throwable
objects as safe-arg values in addition to the result of
Throwable.getMessage.

## After this PR
==COMMIT_MSG==
DangerousThrowableMessageSafeArg disallows Throwables
==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DangerousThrowableMessageSafeArg.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DangerousThrowableMessageSafeArg.java
@@ -71,7 +71,7 @@ public final class DangerousThrowableMessageSafeArg extends BugChecker
                 state)) {
             return buildDescription(tree)
                     .setMessage("Do not use throwables as SafeArg values. "
-                            + "Throwables must be logged without an Arg wrapper as the last parameter, otherwise"
+                            + "Throwables must be logged without an Arg wrapper as the last parameter, otherwise "
                             + "unsafe data may be leaked from the unsafe message or the unsafe message of a cause.")
                     .build();
         }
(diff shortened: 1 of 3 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DangerousThrowableMessageSafeArg.java
```diff
@@ -56,17 +58,23 @@ public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState
...
-
         ExpressionTree safeValueArgument = args.get(1);
         if (THROWABLE_MESSAGE_METHOD.matches(safeValueArgument, state)) {
             return buildDescription(tree)
                     .setMessage("Do not use throwable messages as SafeArg values. "
                             + "SafeLoggable.getLogMessage is guaranteed to be safe.")
                     .build();
         }
+        if (ASTHelpers.isCastable(
+                ASTHelpers.getResultType(safeValueArgument),
```
gatesn: Can we match against unsafe args too? I can't see any reason to log them at all wrapped as an Arg

carterkozak: That stretches the `DangerousThrowableMessageSafeArg` a bit too far, I don't think it's unsafe to do that, just silly.

I think that belongs in a separate check (potentially capable of suggesting a fix to unwrap the throwable and apply it as the last param)

CRogers:
My only worry is that since this tightens checks, it could stop baseline upgrade excavators from automerging, which we're trying to be more on top of. Is it at all possible to autofix this for people? If they have a log statement with a safearg with a `Throwable` in it and no trailing throwable could we fix it for them? Or is that going to be an insane amount of work?

carterkozak:
I'd prefer for that to be a separate check for both safe and unsafe args, with an automated fix -- there's some overlap, but I think that check should be a warning, and this one should be an error.
Happy to put something together when I have time.
Fix will be very similar to the component of #998 which replaces the last arg `throwable.getMessage()` with `throwable`.

CRogers:
Sounds fair to me, :+1:

CRogers (approved):

carterkozak:
follow-up is pushed here: https://github.com/palantir/gradle-baseline/pull/1001
Holding off on merging this until we're happy with the follow-up to avoid error-prone validation breaks that we can automatically fix.

PR #998, opened 2019-10-24
Implement a suggested fix for CatchBlockLogException

carterkozak:
## Before this PR
No great path to fix exceedingly large projects with thousands of violations.

## After this PR
==COMMIT_MSG==
Implement a suggested fix for CatchBlockLogException
==COMMIT_MSG==

## Possible downsides?
In some places this may increase logging noise, if exceptions are expected or used for control flow. That risk is relatively low, and those codepaths should provide the exception at a less severe logging level. I believe its worth the trade-off against failures which don't provide enough data to debug.

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineErrorProneExtension.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineErrorProneExtension.java
@@ -23,6 +23,7 @@ import org.gradle.api.provider.ListProperty;
 public class BaselineErrorProneExtension {
     private static final ImmutableList<String> DEFAULT_PATCH_CHECKS = ImmutableList.of(
             // Baseline checks
+            "CatchBlockLogException",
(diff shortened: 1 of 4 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineErrorProneExtension.java
```diff
@@ -23,6 +23,7 @@
 public class BaselineErrorProneExtension {
     private static final ImmutableList<String> DEFAULT_PATCH_CHECKS = ImmutableList.of(
             // Baseline checks
+            "CatchBlockLogException",
```
carterkozak, line 26: Low risk to fix this by default because existing baseline consumers pass this check. We don't attempt to fix checks that have been opted out of.

iamdanfox (approved):
Thanks @carterkozak - I think we might actually benefit from this in Foundry-land too because some repos could still be on a mega-old baseline version, so an auto-fixed PR saves a bit of manual work :)

PR #1008, opened 2019-10-31
Implement `FinalClass` error prone check, replacing the checkstyle implementation

carterkozak:
The error-prone check provides a suggested fix which is applied automatically.

## Before this PR
`FinalClass` was provided by checkstyle, which we would like to replace with a formatter, and standardize validation on error-prone.

## After this PR
This implementation is an improvement over checkstyle because it provides the ability to fix itself.
==COMMIT_MSG==
Implement `FinalClass` error prone check, replacing the checkstyle implementation
==COMMIT_MSG==

## Possible downsides?
It's possible the checkstyle validation catches a case which this implementation does not.

Note: This is not based on the checkstyle implementation, nor have I read the checkstyle implementation, which uses lgpl2.1.

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/FinalClass.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/FinalClass.java
@@ -82,6 +82,7 @@ public final class FinalClass extends BugChecker implements BugChecker.ClassTree
     private static Optional<SuggestedFix> buildFix(ClassTree tree, VisitorState state) {
         return SuggestedFixes.addModifiers(tree, state, Modifier.FINAL)
                 .map(fix -> {
+                    // Remove redundant 'final' methods modifers that are no longer necessary.
                     SuggestedFix.Builder builder = SuggestedFix.builder().merge(fix);
                     tree.getMembers().stream()
                             .filter(member -> member instanceof MethodTree)
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/FinalClass.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/FinalClass.java
@@ -22,8 +22,6 @@ import com.google.errorprone.VisitorState;
 import com.google.errorprone.bugpatterns.BugChecker;
 import com.google.errorprone.fixes.SuggestedFixes;
 import com.google.errorprone.matchers.Description;
-import com.google.errorprone.matchers.Matcher;
-import com.google.errorprone.matchers.Matchers;
 import com.google.errorprone.util.ASTHelpers;
 import com.sun.source.tree.ClassTree;
 import com.sun.source.tree.MethodTree;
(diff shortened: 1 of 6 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/FinalClass.java
```diff
@@ -0,0 +1,72 @@
...
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "FinalClass",
+        // Support legacy suppressions from checkstyle
+        altNames = {"checkstyle:finalclass", "checkstyle:FinalClass"},
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
+        providesFix = BugPattern.ProvidesFix.REQUIRES_HUMAN_ATTENTION,
+        severity = BugPattern.SeverityLevel.ERROR,
```
carterkozak, line 45: I used ERROR here to match the severity of the checkstyle rule. We can reduce the severity if we decide to keep the checkstyle rule around.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/FinalClass.java
```diff
@@ -0,0 +1,118 @@
...
+        summary = "A class should be declared final if all of its constructors are private. Utility classes -- "
+                + "i.e., classes all of whose methods and fields are static -- have a private, empty, "
+                + "zero-argument constructor.\n"
+                + "https://github.com/palantir/gradle-baseline/tree/develop/docs/best-practices/"
+                + "java-coding-guidelines#private-constructors")
+public final class FinalClass extends BugChecker implements BugChecker.ClassTreeMatcher {
```
ferozco, line 51: we may want to consider checking/cleaning up all of the methods in the class to avoid inadvertently causing method modifiers to become redundant

carterkozak: I like that, remove final from methods that inherit finality from the class. Error prone doesn't have an unnecessary modifier check afaik, thoughts on implementing that separately? Counterargument is that it must be applied in another round after this check succeeds, but I like the separation of validation, and the ability to fix existing instances that aren't created by this check.

ferozco: It would be nice to have both. A generic redundant modifier check and then a special cleanup on this check to specifically remove `final` from methods of classes that have been converted. That way we avoid having to run the fix twice to get to a good state and can catch other issues like `public` on interface methods

carterkozak: Implemented the fix, four hits in large internal ;-)

ferozco (approved):

carterkozak:
policy bot, doing alright there?

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/FinalClass.java
carterkozak: Bug: We can't remove the `final` modifier from methods annotated with `java.lang.SafeVarargs`

PR #1010, opened 2019-11-01
Error prone validation to avoid redundant modifiers

carterkozak:
==COMMIT_MSG==
Error prone validation to avoid redundant modifiers
==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/RedundantModifier.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/RedundantModifier.java
@@ -36,7 +36,7 @@ import javax.lang.model.element.Modifier;
         link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
         linkType = BugPattern.LinkType.CUSTOM,
         providesFix = BugPattern.ProvidesFix.REQUIRES_HUMAN_ATTENTION,
-        severity = BugPattern.SeverityLevel.WARNING,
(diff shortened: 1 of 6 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/RedundantModifier.java
```diff
@@ -0,0 +1,113 @@
...
+import java.util.Locale;
+import javax.lang.model.element.Modifier;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "RedundantModifier",
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
+        providesFix = BugPattern.ProvidesFix.REQUIRES_HUMAN_ATTENTION,
+        severity = BugPattern.SeverityLevel.WARNING,
```
carterkozak: Should we upgrade this to `ERROR` and remove the checkstyle regex?
```xml
        <module name="RegexpSinglelineJava">
            <property name="format" value="static enum"/>
            <property name="message" value="Redundant ''static'' modifier."/>
        </module>
```

carterkozak: I didn't realize we also have a checkstyle rule with the same name in addition to  this regex check. I'll add support for the same suppressions, and bump to error.

ferozco (approved):

PR #1014, opened 2019-11-01
Fix `RedundantModifier` interpretation of implicit modifiers

carterkozak:
Updated RedundantModifier to check explicit modifiers only, added
test coverage for all output.

## After this PR
==COMMIT_MSG==
Fix `RedundantModifier` interpretation of implicit modifiers
==COMMIT_MSG==


carterkozak:
one more bug

carterkozak:
Verified using ptml on this repo.

dansanduleac (approved):

carterkozak:
bulldozer

PR #1017, opened 2019-11-02
Fix RedundantModifier failures types nested in interfaces

carterkozak:
## Before this PR
erroneous failure:
```java
interface Enclosing {
    enum Reproducer {
    // ^ erroneous failure that Reproducer should not be static.
        INSTANCE
    }
}
```

## After this PR
==COMMIT_MSG==
Fix RedundantModifier failures types nested in interfaces
==COMMIT_MSG==


carterkozak:
Validated there are no erroneous failures on a large internal project.

iamdanfox (approved):

PR #1018, opened 2019-11-02
Validate suggested fixes pass validation

carterkozak:
==COMMIT_MSG==
Validate suggested fixes pass validation
==COMMIT_MSG==

Goal:
Explicitly test that our suggested fixes result in code that passes validation.
When we expect the result not to pass validation, we should enforce that continues to be the case.


carterkozak:
WIP, some tests just validate that despite failure, there is no change.

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/RefactoringValidator.java
```diff
@@ -0,0 +1,112 @@
...
+        private TestStage(RefactoringValidator helper, BugCheckerRefactoringTestHelper delegate) {
+            this.helper = helper;
+            this.delegate = delegate;
+        }
+
+        void doTest() {
+            delegate.doTest();
+            helper.compilationHelper
+                    .addSourceLines(helper.outputPath, helper.outputLines)
+                    .doTest();
```
iamdanfox, line 93: nice

iamdanfox (approved):

PR #1020, opened 2019-11-04
Fix error-prone matching literal null as a subtype.

carterkozak:
The most common issue this fixes is failures on `SafeArg.of("name", null)`
assuming that the null literal value parameter may be a throwable.

## After this PR
==COMMIT_MSG==
Fix error-prone mathcing literal null as a subtype.
The most common issue this fixes is failures on `SafeArg.of("name", null)`
assuming that the null literal value parameter may be a throwable.
==COMMIT_MSG==

Reported upstream here: https://github.com/google/error-prone/issues/1397

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/Slf4jLogsafeArgs.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/Slf4jLogsafeArgs.java
@@ -29,11 +29,9 @@ import com.google.errorprone.matchers.Description;
 import com.google.errorprone.matchers.Matcher;
 import com.google.errorprone.matchers.Matchers;
 import com.google.errorprone.matchers.method.MethodMatchers;
-import com.google.errorprone.util.ASTHelpers;
 import com.sun.source.tree.ExpressionTree;
 import com.sun.source.tree.MethodInvocationTree;
 import com.sun.source.util.SimpleTreeVisitor;
-import com.sun.tools.javac.code.Type;
 import java.util.List;
 import java.util.Optional;
 import java.util.regex.Pattern;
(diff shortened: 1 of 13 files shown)


iamdanfox:
:+1:

Always good to be more precise about our subtypes, but I'm struggling to imagine a genuine use case for `SafeArg.of("key", null)` - the failure we saw in sls-logging was just test setup code, and I don't think we've seen this elsewhere right?

carterkozak:
I have not seen it elsewhere, and I agree that it's not terribly useful to log key=null, but I don't want to imply that it is dangerous.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/Slf4jLogsafeArgs.java
```diff
@@ -69,18 +69,14 @@ public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState
...
-                state) ? 2 : 1;
+        int startArg = MARKER.matches(allArgs.get(0), state) ? 2 : 1;
         ExpressionTree lastArg = allArgs.get(lastIndex);
         boolean lastArgIsThrowable = THROWABLE.matches(lastArg, state);
         int endArg = lastArgIsThrowable ? lastIndex - 1 : lastIndex;
 
         ImmutableList.Builder<Integer> badArgsBuilder = ImmutableList.builder();
-        Type argType = state.getTypeFromString("com.palantir.logsafe.Arg");
         for (int i = startArg; i <= endArg; i++) {
-            if (!ASTHelpers.isCastable(ASTHelpers.getType(allArgs.get(i)), argType, state)) {
```
carterkozak, line 83: This appears to have missed a lot of cases that are correctly flagged using the new `ARG.matches` implementation.

PR #1021, opened 2019-11-04
Error prone RedundantModifier check supports more interface components

carterkozak:
```diff
public interface Iface {
-  public static void a() {}
+  static void a() {}
}
```

```diff
public interface Iface {
-  public static final int VALUE = 1;
+  int VALUE = 1;
}
```

## After this PR
==COMMIT_MSG==
Error prone RedundantModifier check supports interface static methods and fields.
==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/RedundantModifier.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/RedundantModifier.java
@@ -28,6 +28,7 @@ import com.sun.source.tree.ClassTree;
 import com.sun.source.tree.MethodTree;
 import com.sun.source.tree.ModifiersTree;
 import com.sun.source.tree.Tree;
+import com.sun.source.tree.VariableTree;
 import java.util.Locale;
 import javax.lang.model.element.Modifier;
 
@@ -45,7 +46,7 @@ import javax.lang.model.element.Modifier;
         severity = BugPattern.SeverityLevel.ERROR,
         summary = "Avoid using redundant modifiers")
 public final class RedundantModifier extends BugChecker
-        implements BugChecker.ClassTreeMatcher, BugChecker.MethodTreeMatcher {
+        implements BugChecker.ClassTreeMatcher, BugChecker.MethodTreeMatcher, BugChecker.VariableTreeMatcher {
 
     private static final Matcher<ClassTree> STATIC_ENUM_OR_INTERFACE = Matchers.allOf(
             Matchers.anyOf(Matchers.kindIs(Tree.Kind.ENUM), Matchers.kindIs(Tree.Kind.INTERFACE)),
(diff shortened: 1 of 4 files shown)


carterkozak:
I haven't had a chance to test this on a large internal project yet, will post here once I've verified it.

carterkozak:
Verified on a large internal project.

carterkozak:
Verified the new change on a large internal project as well.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/RedundantModifier.java
```diff
@@ -45,38 +45,65 @@
         severity = BugPattern.SeverityLevel.ERROR,
```
iamdanfox, line 45: i'm a tiny bit concerned that having this at `ERROR` level might introduce unnecessary friction when people just want to run things like `./gradlew test` locally.  We're just detecting redundant code, not actually 'wrong' in anywa way... given that excavator can come around and auto-fix, what do you think about putting this as suggestion?

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/RedundantModifier.java
```diff
@@ -108,6 +135,12 @@ public Description matchMethod(MethodTree tree, VisitorState state) {
                     .addFix(SuggestedFixes.removeModifiers(tree, state, Modifier.PUBLIC, Modifier.ABSTRACT))
                     .build();
         }
+        if (INTERFACE_STATIC_METHOD_MODIFIERS.matches(tree, state)) {
+            return buildDescription(tree)
+                    .setMessage("Interface components are public by default. The 'public' modifier is unnecessary.")
```
iamdanfox, line 140: interesting that java9 lets people define `private` methods in interfaces! https://bugs.openjdk.java.net/browse/JDK-8071453

This should be all fine though, because we're explicitly only detecting and fixing the `public` keyword.

iamdanfox (approved):

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/RedundantModifier.java
carterkozak: Yep, I designed this with [JEP 213](https://openjdk.java.net/jeps/213) in mind.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/RedundantModifier.java
carterkozak: Agreed, I initially used error because this in some part replaces a checkstyle check which we verify premerge.
Given that robots can fix this for us, and it's not a correctness issue, I don't mind moving it to a suggestion, and allowing minor regressions until the robots sweep through.

PR #1027, opened 2019-11-06
fix #1025 Error Prone `StrictCollectionIncompatibleType` to guard collection usage

### LINKED ISSUE #1025: Rule to detect map.get(wrongType)
## What happened?

It's currently easy to write code like:
```
Map<Foo, Object> map;
Bar key;
Object value = map.get(key)
```
and not realize you have a bug. This is particularly insidious with things like wrapped strings, where you `get` with the right "thing", you just have to wrap or unwrap it.

For example, see commit 331070 in the major internal repo I work on.

Errorprone to flag that this can't work. IntelliJ does, so I expect this is doable.

CC @carterkozak

carterkozak:
This check is an improvement over `CollectionIncompatibleType` because it
validates that values exist in the same type hierarchy, where it could
theoretically be possible the input implements the collection type,
but the type system doesn't have enough information to be confident.
This check allows both subtypes and supertypes, but does not check
for shared supertypes, which are not common.

## After this PR
==COMMIT_MSG==
Error Prone `StrictCollectionIncompatibleType` to guard collection usage
==COMMIT_MSG==


TheBeruriahIncident:
Just tested locally. This does not catch the bug that inspired me to file the ticket, because this does not catch method references. However, it seems to catch the simpler case.
I.e. the case I had was roughly:
```
.map(map::get)
```
which is not caught, but this is:
```
.map(key -> map.get(key))
```
@carterkozak is aware and prefers to stick with this at first, then support method references if this doesn't cause problems.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StrictCollectionIncompatibleType.java
```diff
@@ -0,0 +1,209 @@
...
+import java.util.Stack;
+import java.util.Vector;
+import javax.annotation.Nullable;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "StrictCollectionIncompatibleType",
+        // Idea provides a similar check, avoid noise when that warning is already suppressed.
+        // https://github.com/JetBrains/intellij-community/blob/master/java/java-analysis-impl/src/com/intellij/codeInspection/miscGenerics/SuspiciousCollectionsMethodCallsInspection.java
+        altNames = "SuspiciousMethodCalls",
```
carterkozak: We should respect suppressions for `CollectionIncompatibleType` as well + test.

carterkozak:
Verified on a large internal project. Three failures, all of which appear to be legitimate.
Working on a follow-up to flag method references.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StrictCollectionIncompatibleType.java
```diff
@@ -0,0 +1,321 @@
...
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "StrictCollectionIncompatibleType",
+        // Idea provides a similar check, avoid noise when that warning is already suppressed.
+        // https://github.com/JetBrains/intellij-community/blob/master/java/java-analysis-impl/src/com/intellij/codeInspection/miscGenerics/SuspiciousCollectionsMethodCallsInspection.java
+        altNames = {"SuspiciousMethodCalls", "CollectionIncompatibleType"},
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
+        severity = BugPattern.SeverityLevel.WARNING,
```
carterkozak, line 56: These are pretty serious, and our large internal project had no false positives. We could bump this to ERROR.

carterkozak: Let's leave it WARNING for now, and bump it in a week or so.

dansanduleac (approved):
This is pretty cool! Can we add a couple of test cases involving primitive arguments too?

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/StrictCollectionIncompatibleTypeTest.java
```diff
@@ -0,0 +1,169 @@
...
+                "       // BUG: Diagnostic contains: incompatible types",
+                "       return in.keySet().contains(key);",
+                "   }",
+                "   boolean f4(List<Integer> in, String key) {",
+                "       // BUG: Diagnostic contains: incompatible types",
+                "       return in.remove(key);",
+                "   }",
+                "   boolean f4(List<Integer> in) {",
+                "       // BUG: Diagnostic contains: incompatible types",
+                "       return in.remove(5L);",
```
carterkozak, line 106: This tests boxing the wrong primitive results in a failure

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/StrictCollectionIncompatibleTypeTest.java
```diff
@@ -0,0 +1,169 @@
...
+    void testCollectionContains_edges() {
+        helper().addSourceLines(
+                "Test.java",
+                "import java.util.Collection;",
+                "class Test {",
+                "   boolean f0(Collection<Integer> in) {",
+                "     return in.contains(null);",
+                "   }",
+                "   boolean f1(Collection<Integer> in) {",
+                "     return in.contains(3);",
```
carterkozak, line 123: Tests boxing the expected primitive works as expected.

carterkozak (commented):
What sort of additional testing for primitives are you looking for, @dansanduleac?

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/StrictCollectionIncompatibleTypeTest.java
dansanduleac: oh somehow glossed over these, I was scanning for a function with a type of `int` or something

carterkozak:
dozer?

PR #1028, opened 2019-11-06
Drop RedundantModifier to SUGGESTION and FinalClass to WARNING

carterkozak:
## After this PR
==COMMIT_MSG==
Drop RedundantModifier to SUGGESTION and FinalClass to WARNING

These checks don't imply bugs, and automation will fix failing cases automatically, so it's not necessary to block compilation.
FinalClass moved to warning rather than suggestion because there are a few edge cases (e.g. dependent projects using mockito without inline mock maker) where releases can run into issues if the bots don't fix findings in time.
==COMMIT_MSG==

## Possible downsides?
Code on develop may have issues detected, but not fixed.

Based on comments from @iamdanfox.


iamdanfox (approved):

carterkozak:
bulldozer?

PR #1032, opened 2019-11-07
Error Prone InvocationHandlerDelegation to ensure proxies handle delegate exceptions.

carterkozak:
==COMMIT_MSG==
InvocationHandlers which delegate to another object must catch and unwrap
`InvocationTargetException`, otherwise an `UndeclaredThrowableException` will be thrown
each time the delegate throws an exception.
==COMMIT_MSG==


carterkozak:
I'll report back here once I've verified this on a large internal repo.

carterkozak:
This doesn't handle the cases:

```
try {
  return method.invoke(delegate, args);
} catch (Throwable t) {
   //either
  handle(t);
  //or
  if (t instanceof InvocationTargetException) {
    throw t.getCause();
  }
  throw t;
}
```

Though they're unnecessarily difficult to read, I don't want to cause friction for implementations that aren't actually broken.

carterkozak:
Verified in a large internal project (where it discovered four instances of the bug)

ferozco (approved):

dansanduleac (approved):

PR #1033, opened 2019-11-08
Assertj simplify assertions involving booleans

iamdanfox:
## Before this PR

I stumbled across some tests like the following (which would give terrible error messages if they fail) and realised we currently don't have anything that autofixes this.

```java
assertThat(foo.equals(bar) && baz.equals(barf)).isEqualTo(Boolean.TRUE)
```

## After this PR
==COMMIT_MSG==
Some assertions involving booleans are simplified using refaster to produce better error messages.
==COMMIT_MSG==

## Possible downsides?
- splitting lines relies on the code-formatter to pass checkstyle


baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/AssertjIsTrue.java
```diff
@@ -0,0 +1,43 @@
...
+
+    @BeforeTemplate
+    void before2(boolean bool) {
+        assertThat(bool).isEqualTo(Boolean.TRUE);
+    }
+
+    @AfterTemplate
+    @UseImportPolicy(ImportPolicy.STATIC_IMPORT_ALWAYS)
+    void after(boolean bool) {
+        assertThat(bool).isTrue();
```
iamdanfox, line 51: this one is actually particularly important because loads of our other refasters assume `.isTrue()` as a starting point, but if a line of code never reaches this, then we'd never enable all the other smarter refactorings.

baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/AssertjIsFalse.java
```diff
@@ -0,0 +1,43 @@
...
+public final class AssertjIsFalse {
+
+    @BeforeTemplate
+    void before(boolean bool) {
+        assertThat(bool).isEqualTo(false);
+    }
+
+    @BeforeTemplate
+    void before2(boolean bool) {
+        assertThat(bool).isEqualTo(Boolean.FALSE);
```
carterkozak, line 40: I wonder if we should verify the inverted case as well? assertThat(false).isEqualTo(expected)

carterkozak (approved):

baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/AssertjIsFalse.java
iamdanfox: Good shout - could easily get the inverted case from plain junit: `assertEquals(foo, true)`

carterkozak:
Thanks Dan! :+1:

PR #1054, opened 2019-11-18
fix #1053 Implement `Slf4jThrowable` check to ensure throwable args are logged last

### LINKED ISSUE #1053: ErrorProne Check for Log Arg Ordering
## What happened?

I had a log line in my codebase that put arguments *after* the throwable and thus got undesired/broken behavior.
```
Throwable t = ...
log.error("Exception!", t, SafeArg.of("safeArg", arg));
```

Would have been nice if my code failed to build because it in fact should have been:
```
Throwable t = ...
log.error("Exception!", SafeArg.of("safeArg", arg), t);
```

carterkozak:
fix #1053
==COMMIT_MSG==
Implement `Slf4jThrowable` check to ensure throwable args are logged last
==COMMIT_MSG==

## Possible downsides?
If folks are logging exceptions expecting the string value, they will get warnings. We don't recommend logging exceptions without a stack trace because we lose the most helpful information.


ferozco:
👍

PR #1055, opened 2019-11-19
Add JooqResultStreamLeak ErrorProne rule

ilyanep:
## After this PR
==COMMIT_MSG==
Adds an ErrorProne rule, `JooqResultStreamLeak`, which ensures that result streams and cursors returned from jOOQ results are closed in a try-with-resources block.
==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/JooqResultStreamLeak.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/JooqResultStreamLeak.java
@@ -27,6 +27,7 @@ import com.google.errorprone.matchers.method.MethodMatchers;
 import com.google.errorprone.util.ASTHelpers;
 import com.sun.source.tree.ExpressionTree;
 import com.sun.source.tree.MethodInvocationTree;
+import com.sun.tools.javac.code.Type;
 
 @AutoService(BugChecker.class)
 @BugPattern(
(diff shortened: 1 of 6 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/JooqResultStreamLeak.java
```diff
@@ -0,0 +1,73 @@
...
+
+        if (!isStreamOrCursor(tree, state)) {
+            return Description.NO_MATCH;
+        }
+
+        return matchNewClassOrMethodInvocation(tree, state);
+    }
+
+    private static boolean isStreamOrCursor(MethodInvocationTree tree, VisitorState state) {
+        boolean isStream = ASTHelpers.isSubtype(ASTHelpers.getReturnType(tree),
```
carterkozak: nit: could factor out `ASTHelpers.getReturnType(tree)` and call it once.

Actually, could we replace the checks for either `Stream` or `Cursor` with a single check for `AutoCloseable` which covers both cases (and any potential future types)?

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/JooqResultStreamLeak.java
```diff
@@ -0,0 +1,73 @@
...
+import com.sun.source.tree.MethodInvocationTree;
+import java.util.stream.Stream;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "JooqResultStreamLeak",
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
+        providesFix = BugPattern.ProvidesFix.REQUIRES_HUMAN_ATTENTION,
+        severity = BugPattern.SeverityLevel.ERROR,
```
carterkozak: Mind if we bump this down to a warning for the initial rollout? We can bump it up after we're confident it does exactly what we want.

carterkozak (commented):
Generally looks great, thanks @ilyanep.

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/JooqResultStreamLeakTest.java
```diff
@@ -0,0 +1,130 @@
...
+import org.junit.jupiter.api.parallel.ExecutionMode;
+
+@Execution(ExecutionMode.CONCURRENT)
+public final class JooqResultStreamLeakTest {
+
+    private final CompilationTestHelper testHelper =
+            CompilationTestHelper.newInstance(JooqResultStreamLeak.class, getClass());
+
+    private final BugCheckerRefactoringTestHelper refactoringTestHelper =
+            BugCheckerRefactoringTestHelper.newInstance(new JooqResultStreamLeak(), getClass());
```
carterkozak: one more nit: mind using our `RefactoringValidator` instead of `BugCheckerRefactoringTestHelper`? Ours validates that the refactored result also passes validation

carterkozak (approved):
lgtm, thanks @ilyanep

PR #1060, opened 2019-11-20
Refaster rule to eliminate unnecessary casting.

carterkozak:
```diff
- CharSequence cs = (CharSequence) "value";
+ CharSequence cs = "value";
```

==COMMIT_MSG==
Refaster rule to eliminate unnecessary casting.
==COMMIT_MSG==


carterkozak:
This breaks ambiguous method invocations.

PR #1061, opened 2019-11-20
Refaster rule to convert Stream.of() -> Stream.empty()

ferozco:
## After this PR
==COMMIT_MSG==
Add Refaster rule to convert Stream.of() -> Stream.empty()
==COMMIT_MSG==

### DIFF
--- a/baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/StreamEmpty.java
+++ b/baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/StreamEmpty.java
@@ -1,0 +1,21 @@
+package com.palantir.baseline.refaster;
+
+import com.google.errorprone.refaster.annotation.AfterTemplate;
+import com.google.errorprone.refaster.annotation.AlsoNegation;
+import com.google.errorprone.refaster.annotation.BeforeTemplate;
+import java.util.Collection;
+import java.util.stream.Stream;
+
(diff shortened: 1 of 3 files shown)


baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/StreamEmpty.java
```diff
@@ -0,0 +1,21 @@
...
+import com.google.errorprone.refaster.annotation.BeforeTemplate;
+import java.util.Collection;
+import java.util.stream.Stream;
+
+public class StreamEmpty<T> {
+
+    @BeforeTemplate
+    Stream<T> streamNoParams() {
+        return Stream.of();
+    }
```
carterkozak, line 28: I wonder if we have anything matching:
```java
Collections.emptySet().stream()
Collections.emptyList().stream()
ImmutableList.of().stream()
ImmutableSet.of().stream()
```

They might be _too_ silly

ferozco: That's pretty crazy

carterkozak: Ya, I doubt we have them for real, let's ship this as is :-]

carterkozak:
👍

PR #1064, opened 2019-11-21
Fix: Disable conflicting checkstyle rules if started migrating to PJF

dansanduleac:
## Before this PR

Excavator PRs that start migrating repos to palantir-java-format (but don't convert all files in one go) run into checkstyle errors because there is a check that conflicts with PJF style.

## After this PR
==COMMIT_MSG==
Allow declaring that conversion to palantir-java-format has been started, which ensures that conflicting checkstyle rules are removed.
==COMMIT_MSG==


iamdanfox (approved):
Sounds great - will you edit the corresponding excavator?

PR #1070, opened 2019-11-26
NonComparableStreamSort validates direct types

carterkozak:
## Before this PR
Previously NonComparableStreamSort only failed on types which
would cause a compilation error casting to Comparable. Now we
validate that the stream type is a subtype of Comparable.

## After this PR
==COMMIT_MSG==
Improve `NonComparableStreamSort` to validate that stream types implement comparable, as opposed to validating that casting to comparable does not cause a compiler error.

This commit reduces the severity to WARNING because it's
possible that the check will flag code that happens to work
today, but we strongly recommend against sorting streams of
a type that is not directly comparable without a custom
comparator because it is likely to break later due to lack
of enforcement by the type system.
==COMMIT_MSG==

## Possible downsides?
Possible this will cause friction if it flags new things. That friction is substantially lower than failures at runtime.


carterkozak:
not stale, just awaiting review

ferozco (approved):

PR #1073, opened 2019-11-27
dsinghvi/write rule for rid logging

dsinghvi:
WIP. A v0 of what an errorprone rule to flag rid usages without a helper method may look like.

### DIFF
--- a/baseline-error-prone/build.gradle
+++ b/baseline-error-prone/build.gradle
@@ -7,6 +7,7 @@ apply from: "${rootDir}/gradle/publish-jar.gradle"
 
 dependencies {
     compile 'com.google.errorprone:error_prone_core'
+    compile 'com.palantir.ri:resource-identifier'
 
     testCompile gradleApi()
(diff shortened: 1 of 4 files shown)


dansanduleac:
Hi @dsinghvi, what's the state of this? Are you planning to add more commits to this PR?
There is no changelog and the PR description says WIP and doesn't explain the motivation. The [PR template](https://github.com/palantir/gradle-baseline/blob/develop/.github/PULL_REQUEST_TEMPLATE.md) is a good guideline for how to write PR descriptions that help other people understand the motivation.

Also, FWIW, not everyone uses this RidArg, so if this were to go in, I'd make it disabled by default, and then enable it on the repos where you need it.

dsinghvi:
Hey @dansanduleac Apologize for not updating the PR and pushing it through, I got pulled onto some other things that delayed my progress here. I was initially told that I wouldn't be able to make rules that only apply to specific repos, if that's not true I will keep going with this and try to merge the rule. cc @ericanderson

PR #1076, opened 2019-11-27
Fix #1043: Interruption handling check

### LINKED ISSUE #1043: FR: ensure users who catch InterruptedException also call `Thread.currentThread().interrupt()`
## What happened?

It's easy for people to write code like this with no failures:

```java
} catch (InterruptedException e) {
  throw new RuntimeException(e);
}
```

_Just_ rethrowing an exception like this is usually wrong because if the user is trying to abort, the calling thread will be unable to detect that an interruption happened and short-circuit.

For the cases where people are re-throwing _some_ unchecked exception, error-prone should suggest:
```diff
 } catch (InterruptedException e) {
+    Thread.currentThread().interrupt();
     throw new RuntimeException(e);
 }
```

carterkozak:
Fix #1043

Currently using the ExceptionSpecificity base branch to avoid duplicating utility functionality.

==COMMIT_MSG==
error prone `HandleInterruption` to ensure users who catch InterruptedException also call `Thread.currentThread().interrupt()`.
==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/HandleInterruption.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/HandleInterruption.java
@@ -1,0 +1,200 @@
+/*
+ * (c) Copyright 2018 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.baseline.errorprone;
+
+import com.google.auto.service.AutoService;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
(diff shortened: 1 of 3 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/HandleInterruption.java
```diff
@@ -0,0 +1,200 @@
...
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "HandleInterruption",
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
+        severity = BugPattern.SeverityLevel.WARNING,
+        summary = "InterruptedException must be handled by either rethrowing the InterruptedException or setting "
+                + "Thread.currentThread().interrupt(). Failure to do so can result in retrial of long-running "
+                + "operations that are expected to be terminated.")
```
carterkozak, line 57: Need to reference the `// interruption reset` suppression in the summary

carterkozak: Should link https://www.ibm.com/developerworks/java/library/j-jtp05236/index.html (Brian Goetz)

carterkozak: @iamdanfox I'm not sure we want to limit ourselves to cases that rethrow -- those are the most likely to do the _right_ thing. The bad case is when InterruptedException is caught, the interrupted flag is not reset, and execution continues.

carterkozak:
bad robot.

carterkozak:
long-lived -- on my list to come back to this. It will be both incredibly valuable and tricky to get right.

PR #1078, opened 2019-11-28
refaster for assertThat(a == b)

iamdanfox:
## Before this PR

I just came across this line of code:
```
assertThat(someAtomicLong.get() == MAX_QUEUE_SIZE).isTrue()
```

Which is bad because if the assertion ever fails, the dev will get an error message like "false did not equal true".

## After this PR
==COMMIT_MSG==
refaster will automatically fix up assertions between primitives (a == b) to provide better error messages
==COMMIT_MSG==

## Possible downsides?
- we might get the expected / actual the wrong way round, but I think this is still net positive.


carterkozak:
On my todo list to write an error prone rule to make these easier. I think errror-prone would allow us to avoid the duplication, making greaterThan, lessThan, etc feasible for all types (and mixed types) without hundreds of lines of code.

:+1:

PR #1079, opened 2019-12-01
Error prone `AssertjPrimitiveComparison` for primitive comparisons

carterkozak:
This check replaces several refaster rules in order to provide a
more concise, more accurate implementation.

This check supports boxed types and mixed primitive types, in addition to full support for comparisons across all primitives.

==COMMIT_MSG==
Error prone `AssertjPrimitiveComparison` for primitive comparisons

Prefer using AssertJ fluent comparisons over logic in an assertThat
statement for better failure output. assertThat(a == b).isTrue()
failures report 'expected true' where assertThat(a).isEqualTo(b)
provides the expected and actual values.
==COMMIT_MSG==

## Possible downsides?
This check assumes the left side of a binary expression is the 'actual' value and the right side is 'expected'. This may not always be the case, but either way the resulting code will provide more actionable output on failure.


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/AssertjSingleAssertMatcher.java
```diff
@@ -0,0 +1,133 @@
...
+import com.sun.source.tree.MethodInvocationTree;
+import com.sun.source.tree.Tree;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import java.util.function.BiFunction;
+import javax.lang.model.type.TypeKind;
+
+
+final class AssertjSingleAssertMatcher {
```
iamdanfox, line 37: :+1: I like having this separated out, I imagine there will be a lot more checks that can benefit from this

iamdanfox (approved):
Thanks Carter!

iamdanfox:
Also nice to be able to delete so many refasters in one go!

PR #1082, opened 2019-12-03
errorprone 2.3.3 -> 2.3.4

iamdanfox:
## Before this PR

error-prone's last release was in March, with a _ton_ of cool stuff piled up on develop.

## After this PR
==COMMIT_MSG==
We now run error-prone 2.3.4, to benefit from smarter static analysis, and hopefully pick up the claimed performance Improvements: 

> 40% speedup when run against Google's codebase with errors enabled.
==COMMIT_MSG==

https://github.com/google/error-prone/releases

## Possible downsides?
- any release has a risk of unexpected false positives, slowing down error-prone adoption

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -54,6 +54,7 @@ allprojects {
                 check("PreferSafeLoggableExceptions", CheckSeverity.OFF)
                 check("PreferSafeLoggingPreconditions", CheckSeverity.OFF)
                 check("PreconditionsConstantMessage", CheckSeverity.OFF)
+                check("GradleCacheableTaskAction", CheckSeverity.OFF)
             }
         }
     }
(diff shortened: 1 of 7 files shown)


carterkozak:
👍

carterkozak:
@iamdanfox this will need an approval from you as well since it includes my PR.

build.gradle
```diff
@@ -54,6 +54,7 @@ allprojects {
                 check("PreferSafeLoggableExceptions", CheckSeverity.OFF)
                 check("PreferSafeLoggingPreconditions", CheckSeverity.OFF)
                 check("PreconditionsConstantMessage", CheckSeverity.OFF)
+                check("GradleCacheableTaskAction", CheckSeverity.OFF)
```
iamdanfox, line 57: We can re-enable this as soon as this repo upgrades to a version of baseline that includes this PR.

iamdanfox:
:+1: bulldozer?

build.gradle
dansanduleac: It's really scary that a bump in `error_prone_annotations` alone (which looks like it should be a sort of API, not implementation) changes the implementation logic.

PR #1085, opened 2019-12-04
Add dependency on mockito-errorprone

pkoenig10:
The `MockitoInternalUsage` error-prone check was moved out of `error_prone_core` in https://github.com/google/error-prone/commit/8e821c8e4f0d587c6d64de24b8dfd562e4a8aea7. This PR adds `org.mockito:mockito-errorprone` as a dependency of `baseline-error-prone` to pull in Mockito error-prone checks.

### DIFF
--- a/baseline-error-prone/build.gradle
+++ b/baseline-error-prone/build.gradle
@@ -7,6 +7,7 @@ apply from: "${rootDir}/gradle/publish-jar.gradle"
 
 dependencies {
     compile 'com.google.errorprone:error_prone_core'
+    compile 'org.mockito:mockito-errorprone'
 
     testCompile gradleApi()
     testCompile 'com.palantir.tokens:auth-tokens'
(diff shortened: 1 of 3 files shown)


baseline-error-prone/build.gradle
```diff
@@ -7,6 +7,7 @@ apply from: "${rootDir}/gradle/publish-jar.gradle"
 
 dependencies {
     compile 'com.google.errorprone:error_prone_core'
+    compile 'org.mockito:mockito-errorprone'
```
carterkozak, line 10: Might be cleaner to provide it as a runtime dependency since nothing compiles against it. baseline-error-prone isn't used as a library dependency, so it doesn't matter much either way
```suggestion
    runtimeOnly 'org.mockito:mockito-errorprone'
```

carterkozak (approved):
great catch, thanks!

PR #1088, opened 2019-12-04
auto-fix UnnecessaryBoxedVariable

iamdanfox:
This _seems_ like it will be low regret:
https://github.com/google/error-prone/blob/master/core/src/main/java/com/google/errorprone/bugpatterns/UnnecessaryBoxedVariable.java


carterkozak:
@iamdanfox Want to test this on a large internal project?

carterkozak:
This changes public method signatures and breaks compilation where null is passed as a boxed primitive.

PR #1091, opened 2019-12-04
Set `-XepCompilingTestOnlyCode` for tasks named `compileTestJava`

carterkozak:
This gives us a much faster way to check if we're in test code,
though doesn't allow us to tell whether or not we're not in test
code. Projects with lots of tests (obviously all of ours) will
build faster.

## After this PR
==COMMIT_MSG==
Set `-XepCompilingTestOnlyCode` for tasks named `compileTestJava`
==COMMIT_MSG==

TODO: needs test

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
@@ -216,6 +216,11 @@ public final class BaselineErrorProne implements Plugin<Project> {
             // Don't apply refaster to itself...
             return;
         }
+        // Hint to error-prone that we're compiling test code
+        if (javaCompile.getName().endsWith("compileTestJava")) {
+            errorProneOptions.getErrorproneArgumentProviders()
+                    .add(() -> ImmutableList.of("-XepCompilingTestOnlyCode"));
+        }
 
         if (isRefactoring(project)) {
             // Don't attempt to cache since it won't capture the source files that might be modified
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/CatchSpecificity.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/CatchSpecificity.java
@@ -77,6 +77,9 @@ public final class CatchSpecificity extends BugChecker implements BugChecker.Try
     @Override
     @SuppressWarnings("CyclomaticComplexity")
     public Description matchTry(TryTree tree, VisitorState state) {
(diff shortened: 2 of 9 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
```diff
@@ -216,6 +216,11 @@ private static void configureErrorProneOptions(
             // Don't apply refaster to itself...
             return;
         }
+        // Hint to error-prone that we're compiling test code
+        if (javaCompile.getName().endsWith("compileTestJava")) {
```
carterkozak, line 220: Any idea how we can do this better?

iamdanfox:
rather than matching on task names including `compileTestJava`, I've previously iterated through sourcesets, from which you can get a testtask and probably add the system property there?

https://github.com/palantir/gradle-baseline/blob/e95156c5106ad814d1ac855df83b4d923116ff6d/gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckJUnitDependencies.java#L43-L72

carterkozak:
Ah fantastic, thanks @iamdanfox!

carterkozak:
on my list to come back to this

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
ferozco: It seems like this is already being done for us: https://github.com/tbroyer/gradle-errorprone-plugin/blob/master/src/main/kotlin/net/ltgt/gradle/errorprone/ErrorPronePlugin.kt#L112

I think we should follow Dan's suggestion but upstream the improvement to the core gradle plugin since everyone would benefit from the improved accuracy

iamdanfox: mmm you'd think that but I don't think we've updated that plugin yet https://github.com/palantir/gradle-baseline/blob/2a560566527c34f707f5b78bc54345d213a2b08d/versions.props#L34

ferozco: Cool, lets upgrade then! I think we can decouple that change from updating the errorprone rules to respect the flag

carterkozak: 👍

PR #1092, opened 2019-12-05
autofix new errorprone rules

ferozco:
## After this PR
==COMMIT_MSG==

==COMMIT_MSG==


ferozco:
Need to go through and find all of the rules that have fixes

PR #1094, opened 2019-12-05
PreferJavaTimeOverload:OFF

iamdanfox:
## Before this PR

A bunch of our repos internally use custom assertj libraries (e.g. the c-j-r-api test-utils, audit logging test utils, safe-logging test-utils), which triggers this wonky error-prone behaviour: https://github.com/google/error-prone/issues/1435

I even saw this pretty non-sensical message internally in sls-logging:

```
/Volumes/git/sls-logging/sls-logging-log4j/src/test/java/com/palantir/sls/logging/log4j/Log4jSlsLoggingSubscribersMetricsTest.java:143: warning: [PreferJavaTimeOverload] If the numeric primitive (eventLengthHistogram.getCount()) represents a Instant, please call assertThat(Instant) instead.
                    .untilAsserted(() -> assertThat(eventLengthHistogram.getCount()).isOne());
```

## After this PR
==COMMIT_MSG==
The `PreferJavaTimeOverload` error-prone check is turned off to avoid false positives
==COMMIT_MSG==

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineErrorProneExtension.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineErrorProneExtension.java
@@ -50,7 +50,8 @@ public class BaselineErrorProneExtension {
             // Built-in checks
             "ArrayEquals",
             "MissingOverride",
-            "UnnecessaryParentheses");
+            "UnnecessaryParentheses",
+            "PreferJavaTimeOverload");
 
     private final ListProperty<String> patchChecks;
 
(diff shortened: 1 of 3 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineErrorProneExtension.java
```diff
@@ -50,7 +50,8 @@
             // Built-in checks
             "ArrayEquals",
             "MissingOverride",
-            "UnnecessaryParentheses");
+            "UnnecessaryParentheses",
+            "PreferJavaTimeOverload");
```
iamdanfox, line 54: So I think the `checkExplicitlyDisabled` function will actually prevent these auto-fixes being applied until we get rid of the explicit `CheckSeverity.OFF` setting... I'm kinda OK with just leaving this here so that when we re-enable the check it will start patching.

carterkozak:
👍

PR #1098, opened 2019-12-06
Disable ThrowSpecificty and CatchSpecificity fixes by default

carterkozak:
We can reenable these once we have scala detection. Scala code
throws checked exceptions without declaring them.

## Before this PR
Code which interacts with scala stops catching checked exceptions because they aren't declared.

## After this PR
==COMMIT_MSG==
Disable ThrowSpecificty and CatchSpecificity fixes by default
==COMMIT_MSG==

## Possible downsides?
`Note:` calling these out in the compiler logs until we implement scala detection.

### DIFF
--- a/changelog/@unreleased/pr-1098.v2.yml
+++ b/changelog/@unreleased/pr-1098.v2.yml
@@ -1,0 +1,5 @@
+type: fix
+fix:
+  description: Disable ThrowSpecificty and CatchSpecificity fixes by default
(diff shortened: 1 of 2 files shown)


iamdanfox (approved):
Looks good as these are both SUGGESTIONs at the moment, so shouldn't block anyone

carterkozak:
dozer?

PR #1102, opened 2019-12-09
Don't apply errorprone rules to refaster rules project

dansanduleac:
## Before this PR

Baseline upgrades rewrite our own refaster rules because they violate some of the errorproine rules.

See example: https://github.com/palantir/gradle-baseline/pull/1096

## After this PR
==COMMIT_MSG==
Don't apply errorprone rules to the refaster rules project
==COMMIT_MSG==


carterkozak (approved):
thanks, Dan

PR #1104, opened 2019-12-09
Allow _ prefix for local var

dansanduleac:
## Before this PR

Can't write code that satisfies both `CheckReturnValue` and `StrictUnusedVariable` - when a variable is created only to acklowledge that we don't care about the result of a `CheckReturnValue` function.
The only way to satisfy them both is to prepend `_` to the variable name, but checkstyle bans that. 😠 

## After this PR
==COMMIT_MSG==
Allow local variables to start with `_` if they are unused but still necessary in order to satisfy `StrictUnusedVariable`'s exclusion list.
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-1104.v2.yml
+++ b/changelog/@unreleased/pr-1104.v2.yml
@@ -1,0 +1,6 @@
+type: fix
+fix:
+  description: Allow local variables to start with `_` if they are unused but still
+    necessary in order to satisfy `StrictUnusedVariable`'s exclusion list.
+  links:
+  - https://github.com/palantir/gradle-baseline/pull/1104
--- a/.baseline/checkstyle/checkstyle.xml
+++ b/.baseline/checkstyle/checkstyle.xml
@@ -454,7 +454,7 @@
         <module name="LocalFinalVariableName"/> <!-- Java Style Guide: Local variable names -->
         <module name="LocalVariableName"> <!-- Java Style Guide: Local variable names -->
             <property name="tokens" value="VARIABLE_DEF"/>
-            <property name="format" value="^[a-z][a-zA-Z0-9]+$"/>
+            <property name="format" value="^_?[a-z][a-zA-Z0-9]+$"/>
             <property name="allowOneCharVarInForLoop" value="true"/>
             <message key="name.invalidPattern" value="Local variable name ''{0}'' must match pattern ''{1}''."/>
         </module>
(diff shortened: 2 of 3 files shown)


changelog/@unreleased/pr-1104.v2.yml
```diff
@@ -0,0 +1,6 @@
+type: fix
+fix:
+  description: Allow local variables to start with `_` if they are unused but still
+    necessary in order to satisfy `StrictUnusedVariable`'s exclusion list.
```
carterkozak, line 4: I'm not sure I understand why this is important. I get keeping around unused parameters to avoid breaking API, and to implement interfaces which provide more than we need, but I'm not sure when we want unused local variables.

carterkozak: I think these are dangerous because they can allow unused-return-value checks to pass when they shouldn't.
```
int _skipped = inputStream.skip(100);
```
won't get rewritten to
```
ByteStreams.skipFully(inputStream, 100);
```

ferozco: I agree with carter here, I don't think we want to make this change. In the case you describe, You would be better off directly ignoring the checkReturnValue rather than creating a variable which is then ignored

dansanduleac: We already allow this in the errorprone check, just not in checkstyle.

The specific case I hit was that I couldn't make errorprone happy with either StrictUnusedVariable or CheckReturnValue failing, *unless* I made the variable ignored (or explicitly suppressed one of the checks, but that seems worse).

Thoughts on tightening StrictUnusedVariable to allow unused locals *only* when assigning to a `CheckReturnValue` method?

carterkozak: Can we fix the CheckReturnValue check if there's a case where it's okay to ignore the returned value?

dansanduleac: Well it's not our check, it comes from errorprone.
In any case, the refactoring you mention @carterkozak is still happening at the moment, but I'm proposing
1. allowing users to explicitly ignore CheckReturnValue on just one such method call (this PR)
2. in another PR, to be smarter in refactoring such that it doesn't remove the variable entirely *iff* the initializer is a `CheckReturnValue` method, but instead makes it `_ignored`.

In other words, the main problem isn't so much that we can't do _something_ about such use cases (like suppress), but that they currently get rewritten in a way that causes other checks to fail, and frustration to ensue.

Allowing underscored variables for exactly that use case would make it clear to the `StrictUnusedVariable` fix logic that it can ignore these, but only when the method being called is a CheckReturnValue method.

This being said though, I've only found one case internally where this occurred, and it could potentially be rewritten to use assertj and get rid of the error, so maybe we don't need to change this just yet...

PR #1109, opened 2019-12-10
Allow non-default errorprone fixes to be applied from the command line

carterkozak:
## Before this PR

In an internal project we have used a similar flag to test refactors
prior to rollout, and I have wished that it was available to help
with Assertj rollout to avoid temporarily modifying gradle
configurations.

## After this PR
==COMMIT_MSG==
You may apply specific error-prone refactors including those which are
not enabled by default by providing a comma delimited list of check
names to the `-PerrorProneApply` option.
==COMMIT_MSG==

```bash
./gradlew compileJava compileTestJava -PerrorProneApply=PreferAssertj
```


iamdanfox (approved):
Oh hell yes

ferozco:
👍

PR #1110, opened 2019-12-10
Migrate baseline error-prone checks to use jdk13 compatible qualifiers

carterkozak:
## Before this PR
The compiler failed on jdk13 when checks with suggested fixes were found.

## After this PR
==COMMIT_MSG==
Migrate baseline error-prone checks to use jdk13 compatible qualifiers
==COMMIT_MSG==

## Possible downsides?
We don't run any tests on jdk13


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
```diff
@@ -210,7 +210,6 @@ private static void configureErrorProneOptions(
             // Errorprone 2.3.4 isn't officially compatible with Java13 either
             // https://github.com/google/error-prone/issues/1106
             errorProneOptions.check("TypeParameterUnusedInFormals", CheckSeverity.OFF);
-            errorProneOptions.check("PreferCollectionConstructors", CheckSeverity.OFF);
```
carterkozak, line 213: calling out that I re-enabled this because it uses the safe qualify method now.

ferozco (approved):

carterkozak:
dozer?

PR #1115, opened 2019-12-11
Pin format implementation

dansanduleac:
## Before this PR

Every time baseline updates the formatter gradle plugin, all users get a new version of the formatter implementation.
This can potentially introduce a lot of changes and slow down the rollout of baseline.

## After this PR
==COMMIT_MSG==
Decouple baseline from the palantir-java-format implementation, so they can be upgraded independently. Baseline by default recommends version 0.3.8 of the formatter.
==COMMIT_MSG==


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineFormat.java
```diff
@@ -41,6 +41,9 @@
     private static final String ECLIPSE_FORMATTING = "com.palantir.baseline-format.eclipse";
     private static final String PJF_PROPERTY = "com.palantir.baseline-format.palantir-java-format";
     private static final String GENERATED_MARKER = File.separator + "generated";
+    /** The implementation version of the palantir-java-format that we are recommending. */
+    private static final String PJF_IMPLEMENTATION_VERSION = "0.3.8";
```
carterkozak, line 45: Can we either generate this value based on versions.lock, or add a test that it matches the locked version?

iamdanfox:
I know we're trying to decouple pjf and the formatter, but I think I want to nail down what the goal state is before making all these little baby step pRs

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineFormat.java
```diff
@@ -49,6 +52,10 @@ public void apply(Project project) {
         if (project == project.getRootProject()) {
             if (BaselineFormat.palantirJavaFormatterState(project) == FormatterState.ON) {
                 project.getPluginManager().apply(PalantirJavaFormatIdeaPlugin.class);
+                // Also, configure a specific version of this dependency.
+                // To pick up updates, users should bump this using a constraint in their versions.props.
+                project.getDependencies()
```
ferozco, line 57: I thought that this was handled by the PJF plugin? does it make sense to move it over so that baseline just does super basic plumbing?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineFormat.java
dansanduleac: This coordinate will not appear in `versions.lock`, neither in baseline nor in baseline users' projects.
It's not used at compile time, rather added as a dependency into a non-locked configuration.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineFormat.java
dansanduleac: A default is currently provided by the PJF plugin, the idea here was to override that and provide a pinned version.
Discussed with Dan and we're gonna go with a different approach (will post a comment here).

dansanduleac:
Ok so what I was trying to do here has the downside that baseline and the gradle plugin are still coupled (there is a code dependency), and if the gradle plugin at some point changes in a way that isn't compatible with old versions of the implementation (which people could now bump individually) then they might get errors.

Instead, the plan is to move the entire spotless step logic from baseline into gradle-p-j-f, and have that plugin apply and configure spotless (independently of baseline).
Then, people would just depend on gradle-p-j-f in their classpath (that can be auto-updated by excavator), and apply a `palantir-java-format-spotless` plugin.

PR #1117, opened 2019-12-13
Decouple gradle-palantir-java-format code from baseline

dansanduleac:
## Before this PR

Baseline brings with itself an update to palantir-java-format, potentially causing big PRs that slow down the rollout of baseline.

Additionally, there is a complicated interplay where the `gradle-palantir-java-format` plugin code is used to set up a spotless rule in baseline.

## After this PR
==COMMIT_MSG==
Baseline now just applies `com.palantir.java-format` when enabled, and no longer needs access to `gradle-palantir-java-format` code.

Also, no longer bump the version of PJF that baseline brings in, as users are expected to depend on that individually themselves.
As a result, PJF can be easily updated independently of baseline.
==COMMIT_MSG==


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/format/PalantirJavaFormatStep.java
```diff
@@ -1,90 +0,0 @@
...
-import com.google.common.base.Suppliers;
-import com.palantir.javaformat.gradle.JavaFormatExtension;
-import com.palantir.javaformat.java.FormatterService;
-import java.io.File;
-import java.io.IOException;
-import java.io.Serializable;
-import java.util.function.Supplier;
-import org.gradle.api.artifacts.Configuration;
-
-public final class PalantirJavaFormatStep {
```
iamdanfox, line 31: Moved https://github.com/palantir/palantir-java-format/blob/develop/gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/spotless/PalantirJavaFormatStep.java

iamdanfox (approved):

PR #1123, opened 2019-12-13
Gauges should be registered with `registerWithReplacement`

carterkozak:
## After this PR
==COMMIT_MSG==
Using TaggedMetricRegistry.gauge is equivalent to map.putIfAbsent, and
can result in subtle resource leaks. Prefer replacing existing gauge
values using `registerWithReplacement`.

This check doesn't apply unless a new enough version of Tritium
is available on the compilation classpath.
==COMMIT_MSG==

## Possible downsides?
It's possible some consumers expect subsequent registrations not to replace the existing gauge, however in the scenarios we're aware of, the return value is used in these cases.


baseline-refaster-rules/src/main/java/com/palantir/baseline/refaster/TritiumReplaceGauge.java
```diff
@@ -0,0 +1,47 @@
...
+ * This refaster rule intentionally doesn't check the 'gauge' method in order to
+ * avoid creating changes that don't compile when older versions of Tritium are
+ * present.
+ */
+public final class TritiumReplaceGauge<T> {
+
+    @BeforeTemplate
+    void before(TaggedMetricRegistry registry, MetricName name, Gauge<T> gauge) {
+        registry.remove(name);
+        registry.registerWithReplacement(name, gauge);
```
schlosna, line 39: Should we have a template to replace following pattern with `registry.registerWithReplacement(name, gauge);`?

```
    @BeforeTemplate
    void before(TaggedMetricRegistry registry, MetricName name, Gauge<T> gauge) {
        registry.remove(name);
        registry.gauge(name, gauge);
    }
```

carterkozak: I don't want to create a change that doesn't compile on old versions of Tritium (pre 0.16.1). The error-prone check only runs on new tritium, and transforms `registry.gauge(name, gauge)` into `registry.registerWithReplacement(name, gauge)`, allowing this rule to do the rest.
Documented in the class level javadoc :-)

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/MoreSuggestedFixes.java
```diff
@@ -67,6 +72,41 @@ static SuggestedFix renameInvocationRetainingTypeArguments(
         return fix.replace(startPos, endPos, extra + newMethodName).build();
     }
 
+    /**
+     * Replaces the name of the method being invoked in {@code tree} with {@code replacement}.
+     * This implementation is forked from upstream error-prone to work around a bug that results in
+     * parameters being renamed in some scenarios: https://github.com/google/error-prone/issues/1451.
+     */
+    public static SuggestedFix renameMethodInvocation(
```
carterkozak, line 80: Upstream PR: https://github.com/google/error-prone/pull/1452

carterkozak:
Validated in a large internal project.

iamdanfox (approved):

PR #1125, opened 2019-12-17
Move assertj code -> palantir/assertj-automation

iamdanfox:
## Before this PR

Internally, many excavators to update people's baseline would not auto-merge because they make some trivial assertj changes (which require a human +1).

## After this PR
==COMMIT_MSG==
All assertj-related code-rewriting now lives in the https://github.com/palantir/assertj-automation repo.
==COMMIT_MSG==

## Possible downsides?
- if people want to 'lock-in' the PreferAssertj check, they'll need to take an explicit dependency on `com.palantir.assertj-automation:assertj-error-prone`


carterkozak:
:+1:

iamdanfox:
🚂

PR #1126, opened 2019-12-17
Use explicit CommandLineArgumentProviders instead of lambdas

iamdanfox:
## Before this PR

We get warnings from gradle about using lambdas here. Eventually I think we'll probably want to make these cacheable (even if they aren't completely right now).

## After this PR
==COMMIT_MSG==
Eliminate a gradle warning
==COMMIT_MSG==

## Possible downsides?
- more verbose, could regress.


carterkozak:
> Eventually I think we'll probably want to make these cacheable (even if they aren't completely right now).

We disable caching when refactoring is requested, so that may not be necessary.

:+1:

PR #1129, opened 2019-12-18
Enable checkNoUnusedPin on all projects

ferozco:
## Before this PR
Only users who were using nebula dependency recommender would have build time validation that their versions.props file was minimal

## After this PR
==COMMIT_MSG==
Projects with `com.palantir.baseline-versions.disable=true` will now have `checkNoUnusedPin` task run as part of check, ensuring that the `versions.props` file is minimal
==COMMIT_MSG==

## Possible downsides?
I tried this out on a few repos and it seemed to do the right thing, there may be cases where it does not and so it will prevent people from merging.

We'll probably also want to update the automation so that the "fix" is automatically applied as part of the upgrade

cc @jeremyjliu

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/versions/BaselineVersions.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/versions/BaselineVersions.java
@@ -75,10 +75,31 @@ public final class BaselineVersions implements Plugin<Project> {
 
     @Override
     public void apply(Project project) {
-        if (project.hasProperty(DISABLE_PROPERTY)) {
-            log.info("Not configuring com.palantir.baseline-versions because " + DISABLE_PROPERTY + " was set");
-            return;
+        project.getPluginManager().apply(BasePlugin.class);
+
+        File rootVersionsPropsFile = rootVersionsPropsFile(project);
+        TaskProvider<CheckVersionsPropsTask> checkVersionsProps = project.getTasks().register(
+                "checkVersionsProps", CheckVersionsPropsTask.class);
+        project.getTasks().named("check").configure(task -> task.dependsOn("checkVersionsProps"));
+
+        TaskProvider<CheckNoUnusedPinTask> checkNoUnusedPin = project.getTasks().register(
+                "checkNoUnusedPin", CheckNoUnusedPinTask.class, task -> task.setPropsFile(rootVersionsPropsFile));
+        checkVersionsProps.configure(task -> {
+            task.dependsOn(checkNoUnusedPin);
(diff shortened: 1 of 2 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/versions/BaselineVersions.java
```diff
@@ -75,10 +75,31 @@
...
+        project.getTasks().named("check").configure(task -> task.dependsOn("checkVersionsProps"));
+
+        TaskProvider<CheckNoUnusedPinTask> checkNoUnusedPin = project.getTasks().register(
+                "checkNoUnusedPin", CheckNoUnusedPinTask.class, task -> task.setPropsFile(rootVersionsPropsFile));
+        checkVersionsProps.configure(task -> {
+            task.dependsOn(checkNoUnusedPin);
+            checkNoUnusedPin.get().setShouldFix(task.getShouldFix());
+        });
+
+        // TODO(someone): Remove this one day
```
carterkozak:
```suggestion
        // TODO(#1131): Remove this one day
```
Let's track this

carterkozak:
It's unfortunate that the tests only run against the nebula implementation, which isn't what we use in most projects. Any chance we can parameterize those?

Not a huge deal since we're applying the same code in both cases, but it would be nice to prevent regressions. 👍

iamdanfox:
I think the outcome makes sense - having extraneous lines in your versions.props is pretty pointless, but I am a teeny bit uncomfortable about the implementation here because now there are two entirely independent codepaths that parse a versions.props file - one in GCV and one here in baseline.

Maybe it's fine to just run with this as it's a quick fix and something is better than nothing, but it might not be completely accurate because there is a list of configurations that GCV doesn't touch https://github.com/palantir/gradle-consistent-versions/blob/develop/src/main/java/com/palantir/gradle/versions/VersionRecommendationsExtension.java#L28-L35, so we might get a few false negatives (i.e. leftover lines which can be safely deleted by this task doesn't notice).

ferozco:
Ya agree it is a little weird to have the logic in gradle-baseline, especially as we move to remove nebula dependency recommender from it. I was mostly trying to make a quick fix to get this working for users, but I happy to move the feature over to GCV so that all versions.props stuff lives in one place

iamdanfox:
Defer to you on this one - if you feel like you have free time and wanna make it tidy, I think GCV would be a nice home, otherwise we can just go the slightly hacky way

ferozco:
Closing, I'll be moving this over to GCV

ferozco:
New PR here: https://github.com/palantir/gradle-consistent-versions/pull/365

PR #1130, opened 2019-12-18
Implement BracesRequired error-prone check with suggested fixes

carterkozak:
## Before this PR
No automatic fix for missing braces.

## After this PR
==COMMIT_MSG==
Implement BracesRequired error-prone check with suggested fixes
```diff
- if (condition) statement;
+ if (condition) {
+   statement;
+ }
```
==COMMIT_MSG==

## Possible downsides?
None, we already validate this using the default checkstyle configuration. The new check allows our automation to fix projects that don't yet comply.

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/BracesRequired.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/BracesRequired.java
@@ -49,7 +49,12 @@ public final class BracesRequired extends BugChecker implements
     @Override
     public Description matchIf(IfTree tree, VisitorState state) {
         check(tree.getThenStatement(), state);
-        check(tree.getElseStatement(), state);
+        StatementTree elseStatement = tree.getElseStatement();
+        if (elseStatement != null
+                // Additional check for else if
+                && elseStatement.getKind() != Tree.Kind.IF) {
+            check(elseStatement, state);
+        }
         return Description.NO_MATCH;
     }
 
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/BracesRequired.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/BracesRequired.java
@@ -1,0 +1,87 @@
+/*
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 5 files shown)


chrismale:
Can we support the 'statement' being on the immediate line beneath the if?

carterkozak:
> Can we support the 'statement' being on the immediate line beneath the if?

Yep, the auto-formatter takes care of that, we only need to build the right AST here :-)

chrismale:
Fantastic.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/BracesRequired.java
```diff
@@ -0,0 +1,87 @@
...
+        BugChecker.ForLoopTreeMatcher,
+        BugChecker.EnhancedForLoopTreeMatcher,
+        BugChecker.IfTreeMatcher,
+        BugChecker.WhileLoopTreeMatcher {
+
+    @Override
+    public Description matchIf(IfTree tree, VisitorState state) {
+        check(tree.getThenStatement(), state);
+        check(tree.getElseStatement(), state);
+        return Description.NO_MATCH;
```
ferozco, line 58: seems a little odd to always return NO_MATCH and rely on a side effect of check to actually do the fix

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/BracesRequiredTest.java
```diff
@@ -0,0 +1,200 @@
...
+                        "    else",
+                        "      System.out.println(\"else\");",
+                        "  }",
+                        "}")
+                .addOutputLines("Test.java",
+                        "class Test {",
+                        "  void f(boolean param) {",
+                        "    if (param) {",
+                        "      System.out.println(\"if\");",
+                        "    } else {",
```
ferozco, line 86: can you add a test for else if?

carterkozak: good call, incoming

carterkozak: Yup, I'd messed that up. Fixed, thanks!

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/BracesRequired.java
carterkozak: Agreed, it's a bit odd. It allows us to keep the code cleaner and reusable without creating potentially unnecessary SuggestedFix.Builder objects for every conditional and loop. Several of the upstream checks are structured similarly.

ferozco:
👍 programming at scale

PR #1134, opened 2019-12-19
Mark test resources as test resources in IntelliJ

pkoenig10:
## Before this PR
Test resources are marked as source resources in IntelliJ.

This logic from https://github.com/palantir/gradle-baseline/pull/156 is not correct because resource source folders do not have a `isTestSource` attribute.

The IDEA plugin added support for resources in https://github.com/gradle/gradle/pull/3724, which was first released in Gradle 4.7.0. 

## After this PR
Test resources are marked as test resources in IntelliJ.

I've verified the behavior using `publishToMavenLocal`.

Before:
```
<sourceFolder url="file://$MODULE_DIR$/src/test/resources" type="java-resource"/>
```
After:
```
<sourceFolder url="file://$MODULE_DIR$/src/test/resources" type="java-test-resource"/>
```

### DIFF
--- a/changelog/@unreleased/pr-1134.v2.yml
+++ b/changelog/@unreleased/pr-1134.v2.yml
@@ -1,0 +1,6 @@
+type: fix
+fix:
+  description: The `baseline-idea` plugin now correctly marks test resources as test
+    resources in IntelliJ.
+  links:
+  - https://github.com/palantir/gradle-baseline/pull/1134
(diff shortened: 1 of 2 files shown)


ferozco:
👍

dansanduleac:
I think this might have been the cause of my woes with importing from intellij where it thinks `test` source set modules have both a production and a test output, causing an error.

PR #1137, opened 2020-01-01
Error-prone OptionalOfNullableNonNull: Prefer Optional.of over ofNullable

carterkozak:
## After this PR
==COMMIT_MSG==
Error-prone OptionalOfNullableNonNull: Prefer Optional.of over ofNullable
==COMMIT_MSG==

## Possible downsides?
If code is refactored later to allow nulls, Optional.of will throw instead of returning an empty optional.
This is already true of any parameter that becomes nullable when it is accessed directly, so there's no real additional risk in this case.


carterkozak:
marked DNM until I've tested on a large internal codeabse

carterkozak:
I don't think we can trust the nullness check, it things `map.get(val)` is non-nullable in some cases depending on the surrounding code. I can work around this instance, but it doesn't give me confidence that there aren't similar bugs elsewhere.

PR #1139, opened 2020-01-02
Migrate the CollectionStreamForEach refactor to error-prone

carterkozak:
==COMMIT_MSG==
Migrate the CollectionStreamForEach refactor to error-prone
==COMMIT_MSG==

## Possible downsides?
Previously this was not merge blocking, but the updated check will fail compilation for projects which fail on warnings.
There's a potential failure scenario where a collection implementation can define forEach using `this.stream()`, however that code should be rewritten to avoid stream overhead.


carterkozak:
Verified on a large internal project

ferozco (approved):

PR #1145, opened 2020-01-07
New baseline-class-uniqueness.lock file allows incremental adoption

iamdanfox:
## Before this PR

It's pretty hard to roll out the `baseline-class-uniqueness` plugin because it currently is quite an "all or nothing" thing - either you're 100% compliant or else it fails. We haven't really invested in this up until now, because not _that_ much seems to go wrong.

HOWEVER - a recent c-j-r dependency upgrade brought in new 'jakarta' jars, which are renamed versions of javax ones (e.g. `javax.ws.rs:javax.ws.rs-api:2.1` and `jakarta.ws.rs:jakarta.ws.rs-api:2.1.6` are pretty much the same thing).  People are now starting to get confusing runtime errors from these.  Internally, I've got a plugin to automatically fix the problem (gradle-jakarta-renames), but in order to roll _that_ out, I want to ensure people can code-review the changes that it causes.

## After this PR
==COMMIT_MSG==
Users can now `apply plugin: 'com.palantir.baseline-class-uniqueness'` even if they are not yet fully compliant, as the plugin will now just record conflicts in a lock-file. This makes it easier to incrementally improve projects, rather than asking people to fix things in a big-bang way.
==COMMIT_MSG==

@raiju for SA who cared about this stuff before

before merging
- [x] tests
- [x] decide if we want to delete the old stuff too?

## Possible downsides?
- There is lots of room to make the class uniqueness analyzing more efficient (it's currently pretty naiive)... If 100 projects all run baseline-class-uniqueness and depend on one jar, I think that jar will get unpacked and analyzed 100 times. 
- more lock files might confuse people


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineClassUniquenessPlugin.java
```diff
@@ -17,27 +17,37 @@
...
         project.getTasks().addRule(rule);
 
         project.getPlugins().withId("java", plugin -> {
+            lockTask.configure(t -> t.configurations.add(JavaPlugin.RUNTIME_CLASSPATH_CONFIGURATION_NAME));
+
             String checkRuntimeClasspathTask = "checkRuntimeClasspathClassUniqueness";
             rule.apply(checkRuntimeClasspathTask);
-            project.getTasks().getByName("check")
-                    .dependsOn(project.getTasks().getByName(checkRuntimeClasspathTask));
+            project.getTasks().getByName("check").dependsOn(project.getTasks().getByName(checkRuntimeClasspathTask));
```
dansanduleac: just string is fine

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineClassUniquenessPlugin.java
```diff
@@ -17,27 +17,37 @@
...
     @Override
     public final void apply(Project project) {
-        BaselineClassUniquenessRule rule = new BaselineClassUniquenessRule(project);
+        TaskProvider<ClassUniquenessLockTask> lockTask =
+                project.getTasks().register("checkClassUniquenessLock", ClassUniquenessLockTask.class);
+        project.getPlugins().apply(LifecycleBasePlugin.class);
+        project.getTasks().getByName(LifecycleBasePlugin.CHECK_TASK_NAME).dependsOn(lockTask);
 
+        // TODO(dfox): is this 'rule' redundant now? maybe replace it with a kinda strict mode?
+        BaselineClassUniquenessRule rule = new BaselineClassUniquenessRule(project);
```
dansanduleac: let's get rid of the rule

gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/ClassUniquenessLockTask.java
```diff
@@ -0,0 +1,155 @@
...
+    private void ensureLockfileContains(String expected) {
+        if (getProject().getGradle().getStartParameter().isWriteDependencyLocks()) {
+            GFileUtils.writeFile(expected, lockFile);
+            getLogger().lifecycle("Updated {}", getProject().getRootDir().toPath().relativize(lockFile.toPath()));
+        } else {
+            String onDisk = GFileUtils.readFile(lockFile);
+            if (!onDisk.equals(expected)) {
+                throw new GradleException(lockFile
+                        + " is out of date, please run `./gradlew "
+                        + "checkClassUniquenessLock --write-locks` to update this file");
```
carterkozak: I think we may want to provide a different message if new duplicates appear vs when problems are fixed, or mention here that it's dangerous to have duplicate classes on the classpath and explain why it's worth spending time to fix.
Otherwise I think folks will see "run X to update locks" and they'll `./X`.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineClassUniquenessPlugin.java
```diff
@@ -16,28 +16,30 @@
...
+                project.getTasks().register("checkClassUniqueness", CheckClassUniquenessLockTask.class);
+        project.getPlugins().apply(LifecycleBasePlugin.class);
+        project.getTasks().getByName(LifecycleBasePlugin.CHECK_TASK_NAME).dependsOn(lockTask);
 
         project.getPlugins().withId("java", plugin -> {
-            String checkRuntimeClasspathTask = "checkRuntimeClasspathClassUniqueness";
-            rule.apply(checkRuntimeClasspathTask);
-            project.getTasks().getByName("check")
-                    .dependsOn(project.getTasks().getByName(checkRuntimeClasspathTask));
+            lockTask.configure(t -> t.configurations.add(
```
ferozco, line 41: We're losing the ability to easily verify the class uniqueness on different configurations. This could be valuable when running tests since you could inadvertently test with different classes than what exists at runtime.

ferozco, line 41: you might want to copy https://github.com/palantir/metric-schema/blob/develop/gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/MetricSchemaMarkdownPlugin.java#L46 so that `./gradlew --write-locks` also updates the lockfile

gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckClassUniquenessLockTask.java
```diff
@@ -0,0 +1,165 @@
...
+import org.gradle.util.GFileUtils;
+
+@CacheableTask
+public class CheckClassUniquenessLockTask extends DefaultTask {
+
+    private static final String HEADER = "# Danger! Multiple jars contain identically named classes. This may "
+            + "cause different behaviour depending on classpath ordering.\n"
+            + "# Run ./gradlew checkClassUniqueness --write-locks to update this file\n\n";
+
+    // not marking this as an Input, because we want to re-run if the *contents* of a configuration changes
```
ferozco, line 47: shouldn't the contents only change if a dependency on the configuration changes?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineClassUniquenessPlugin.java
iamdanfox: I actually preserved this functionality because I remember @raiju specifically added it, you just have to add the configurations you want as per:

```gradle
checkClassUniqueness {
  configurations.add project.configurations.myConf
}
```

Then they'll be analyzed and will appear in the lockfile

iamdanfox: I was a bit torn about this, because all our other `--write-locks` operations are pretty fast, but this one involves actually unpacking every single jar in your classpath. On the one hand a single entrypoint is quite nice, but also I think we really want people to acknowledge these and make a conscious decision to fix or not fix.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckClassUniquenessLockTask.java
iamdanfox: Configurations are not Serializable, so I can't just mark a list of them as an `@Input`.

iamdanfox:
FWIW I've been iterating on this with share-link-service and it nicely highlights all the jakarta problems, then when I apply `gradle-jakarta-renames` they all go away.

carterkozak:
Fantastic!

gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckClassUniquenessLockTask.java
ferozco: you can take the configuration as an input by annotating it with @classpath

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineClassUniquenessPlugin.java
ferozco: ya fair enough, my concern is that not having it linked to `--write-locks` will block excavators since the lockfile will not automatically be updated

gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckClassUniquenessLockTask.java
iamdanfox: I think that `@classpath` annotation would work if I had just a single configuration, but in order to preserve the ability to analyze [multiple configurations](https://github.com/palantir/gradle-baseline/pull/1145#discussion_r363867958), I don't have just one single configuration here but a bunch of them?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineClassUniquenessPlugin.java
iamdanfox: Ok I've just copied and pasted that in so that `./gradlew --write-locks` will run it.  Will make our excavators simpler, but it might make local dev a bit slower

gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckClassUniquenessLockTask.java
ferozco: Instead of storing a set of configurations could you create a single configuration which extends from all of the provided configurations

iamdanfox: I dunno, it seems that has a slight risk of false positives... e.g. if we're analyzing two configurations foo and bar and we move a jar from foo -> bar.  In the big clobbered configuration there would be no change, but _actually_ I would expect the lockfile to change.

ferozco: ok no big deal

ferozco:
👍

PR #1146, opened 2020-01-07
Exempt slf4j logger from unused field removal

dansanduleac:
## Before this PR
Unused logger fields are removed but the logger initialization remains in a static{} block.

## After this PR
==COMMIT_MSG==
Don't mark unused slf4j logger fields as StrictUnusedVariable.
==COMMIT_MSG==

## Possible downsides?
Unused loggers stick around bloating classes.


ferozco:
We might want to be more aggressive and remove the entire logger instead of just the variable definition

CRogers:
I think it's probably ok to leave to static loggers around if it's going to be hard to do the fully correct thing. It's not going to cause a problem like leaving say auth headers around :+1:

PR #1151, opened 2020-01-08
IntelliJ: Format on save when using PJF

dansanduleac:
## Before this PR

A common complaint we've seen with Palantir Java Format is that it's quite opinionated in what your code should look like, so people often forget to format, then get red checks on CI.

## After this PR
==COMMIT_MSG==
When palantir-java-format is applied to a repo, configure the intellij project to auto-format on save.
==COMMIT_MSG==


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
```diff
@@ -311,4 +329,31 @@ class BaselineIdea extends AbstractBaselinePlugin {
...
+            // Configure defaults if this plugin is configured for the first time only
+            appendNode('option', [name: 'actions']).appendNode('set').with {
+                appendNode('option', [value: 'activate'])
+                appendNode('option', [value: 'noActionIfCompileErrors'])
+                appendNode('option', [value: 'organizeImports'])
+                appendNode('option', [value: 'reformat'])
+            }
+            appendNode('option', [name: 'configurationPath', value: ''])
+            appendNode('inclusions').appendNode('set').with {
+                appendNode('option', [value: 'src/.*\\.java'])
```
ferozco, line 354: nice

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/GroovyXmlUtils.groovy
```diff
@@ -0,0 +1,33 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.baseline.plugins
+
+final class GroovyXmlUtils {
+    static Node matchOrCreateChild(Node base, String name, Map attributes = [:], Map defaults = [:],
```
ferozco: spacing on this is weird

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
```diff
@@ -68,6 +68,21 @@ class BaselineIdea extends AbstractBaselinePlugin {
             moveProjectReferencesToEnd(ideaModuleModel)
         }
 
+        // Suggest and configure the "save actions" plugin if Palantir Java Format is turned on.
+        if (project == project.rootProject) {
+            IdeaModel ideaRootModel = project.extensions.findByType(IdeaModel)
```
ferozco: we already have the ideaRootModel above

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
```diff
@@ -68,6 +68,21 @@ class BaselineIdea extends AbstractBaselinePlugin {
             moveProjectReferencesToEnd(ideaModuleModel)
         }
 
+        // Suggest and configure the "save actions" plugin if Palantir Java Format is turned on.
+        if (project == project.rootProject) {
```
ferozco, line 46: it seems like nearly the whole plugin should be a no-op if not applied to the root. perhaps we can restructure this so its a little more clear

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
```diff
@@ -311,4 +329,31 @@ class BaselineIdea extends AbstractBaselinePlugin {
             module.dependencies.addAll(projectRefs)
         }
     }
+
+    /**
+     * Configures some defaults on the save-actions plugin, but only if it hasn't been configured before.
+     */
+    private static void configureSaveActions(Node rootNode) {
+        GroovyXmlUtils.matchOrCreateChild(rootNode, 'component', [name: 'SaveActionSettings'], [:]) {
+            // Configure defaults if this plugin is configured for the first time only
```
ferozco, line 345: is this to allow users to override the configuration?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
```diff
@@ -311,4 +329,31 @@ class BaselineIdea extends AbstractBaselinePlugin {
...
+            }
+        }
+    }
+
+    private static void configureExternalDependencies(Node rootNode) {
+        def externalDependencies = GroovyXmlUtils.matchOrCreateChild(rootNode, 'component', [name: 'ExternalDependencies'])
+        // I kid you not, this is the id for the save actions plugin:
+        // https://github.com/dubreuia/intellij-plugin-save-actions/blob/v1.9.0/src/main/resources/META-INF/plugin.xml#L5
+        // https://plugins.jetbrains.com/plugin/7642-save-actions/
+        GroovyXmlUtils.matchOrCreateChild(externalDependencies, 'plugin', [id: 'com.dubreuia'], ['min-version': '1.9.0'])
```
ferozco: might want to pull out the version so its more visible/easier to upgrade

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
dansanduleac: Yep, so it's a bit tricky as if people already use this plugin (and have configured it for a project) then we do nothing.. but in the majority of cases, this is probably what people want.

An unsolved problem I will add to the description is how to deal with changing flags / how do we know what's safe to update because we set it up that way vs the user set it up that way.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
dansanduleac: yeah but it's needlessly trapped inside `afterEvaluate`. Basically I think the code above is beyond gross (it reapplies the same steps to the root model from every single subproject) and so will not touch it, with an eye to refactor it in another PR.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
dansanduleac: yes, or they might have configured this plugin already for this project and then we don't want to override their configuration

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
dansanduleac: restructured

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
dansanduleac: done

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/GroovyXmlUtils.groovy
dansanduleac: fixed

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
ferozco: I'd have to look into what other options the plugin exposes, but we may want to narrow what users can configure to a very narrow set of flags

ferozco (approved):

PR #1158, opened 2020-01-10
Fix: baseline-idea correctly filters save actions to java files

dansanduleac:
## Before this PR
Made a mistake in configuring the include filters for the Save Actions plugin.
Thus, the plugin would format any file on save, not just java files.

## After this PR
==COMMIT_MSG==
Fix a bug in the Save Actions plugin configuration where it would format any file on save, instead of just java files.
==COMMIT_MSG==

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
@@ -350,7 +350,7 @@ class BaselineIdea extends AbstractBaselinePlugin {
                 appendNode('option', [value: 'reformat'])
             }
             appendNode('option', [name: 'configurationPath', value: ''])
-            appendNode('inclusions').appendNode('set').with {
+            appendNode('option', [name: 'inclusions']).appendNode('set').with {
                 appendNode('option', [value: 'src/.*\\.java'])
             }
         }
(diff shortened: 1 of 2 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
```diff
@@ -350,7 +350,7 @@ class BaselineIdea extends AbstractBaselinePlugin {
                 appendNode('option', [value: 'reformat'])
             }
             appendNode('option', [name: 'configurationPath', value: ''])
-            appendNode('inclusions').appendNode('set').with {
+            appendNode('option', [name: 'inclusions']).appendNode('set').with {
                 appendNode('option', [value: 'src/.*\\.java'])
```
iamdanfox: will the `src/` fail on windows? Should we just use File.separator instead?

iamdanfox (approved):

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
```diff
@@ -350,8 +350,8 @@ class BaselineIdea extends AbstractBaselinePlugin {
                 appendNode('option', [value: 'reformat'])
             }
             appendNode('option', [name: 'configurationPath', value: ''])
-            appendNode('inclusions').appendNode('set').with {
-                appendNode('option', [value: 'src/.*\\.java'])
+            appendNode('option', [name: 'inclusions']).appendNode('set').with {
+                appendNode('option', [value: "src${File.pathSeparator}.*\\.java"])
```
iamdanfox: I think you got the wrong one mate -

```
    @Test
    public void foo() {
        System.out.println(File.pathSeparator); // prints ':'
        System.out.println(File.separator); // prints '/'
    }
```

dansanduleac:
👍

PR #1160, opened 2020-01-10
Fix #1159: StrictUnusedVariable respects UnusedVariable suppression

### LINKED ISSUE #1159: errorProneApply removes suppressed UnusedVariable field
## What happened?

running `classes -PerrorProneApply` in conjure-python removed this "unused" field which is actually necessary for argument parsing to work as expected.

```java
        @CommandLine.Unmatched
        @SuppressWarnings("UnusedVariable")
        private List<String> unmatchedOptions;
```

Don't remove fields that are explicitly marked with `@SuppressWarnings("UnusedVariable")`

carterkozak:
## After this PR
==COMMIT_MSG==
StrictUnusedVariable respects UnusedVariable suppression
==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StrictUnusedVariable.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StrictUnusedVariable.java
@@ -106,7 +106,7 @@ import javax.lang.model.element.Name;
 @AutoService(BugChecker.class)
 @BugPattern(
         name = "StrictUnusedVariable",
-        altNames = {"unused", "StrictUnusedVariable"},
+        altNames = {"unused", "UnusedVariable"},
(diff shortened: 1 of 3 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StrictUnusedVariable.java
```diff
@@ -106,7 +106,7 @@
 @AutoService(BugChecker.class)
 @BugPattern(
         name = "StrictUnusedVariable",
-        altNames = {"unused", "StrictUnusedVariable"},
```
carterkozak, line 109: It's unnecessary to duplicate the `name`.

dansanduleac (approved):

PR #1163, opened 2020-01-10
fix #1161: Add error prone LoggerEnclosingClass

### LINKED ISSUE #1161: Consider a check to validate slf4j LoggerFactory.getLogger(class) uses the enclosing class
## What happened?

Couldn't find expected logging because the logger was copied from another class without updating the target class name.

This should fail compilation and suggest the enclosing class name.

carterkozak:
Loggers created using getLogger(Class<?>) must reference their enclosing
class.

## After this PR
==COMMIT_MSG==
Add error prone LoggerEnclosingClass check. Loggers created using getLogger(Class<?>) must reference their enclosing class.
==COMMIT_MSG==

## Possible downsides?
It's possible there are unexpected usage patterns which may need to be suppressed.

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/LoggerEnclosingClass.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/LoggerEnclosingClass.java
@@ -1,0 +1,96 @@
+/*
+ * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.baseline.errorprone;
+
+import com.google.auto.service.AutoService;
+import com.google.errorprone.BugPattern;
+import com.google.errorprone.VisitorState;
+import com.google.errorprone.bugpatterns.BugChecker;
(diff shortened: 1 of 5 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/LoggerEnclosingClass.java
```diff
@@ -0,0 +1,96 @@
...
+@BugPattern(
+        name = "LoggerEnclosingClass",
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
+        providesFix = BugPattern.ProvidesFix.REQUIRES_HUMAN_ATTENTION,
+        severity = BugPattern.SeverityLevel.WARNING,
+        summary = "Loggers created using getLogger(Class<?>) must reference their enclosing class.")
+public final class LoggerEnclosingClass extends BugChecker implements BugChecker.VariableTreeMatcher {
+
+    private static final Matcher<ExpressionTree> getLoggerMatcher = MethodMatchers.staticMethod()
```
ferozco:
```suggestion
    private static final Matcher<VariableTree> LOGGER = Matchers.allOf(
            Matchers.isField(),
            Matchers.isSameType(Logger.class),
            Matchers.variableInitializer(MethodMatchers.staticMethod()
                    .onClass(LoggerFactory.class.getName())
                    .named("getLogger")
                    .withParameters(Class.class.getName())));
```

ferozco:
CI failure is unrelated, looks like `http://www.vogella.com/tutorials/JUnit/article.html` is dead :(

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/LoggerEnclosingClass.java
carterkozak: Snazzy, I hadn't seen `Matchers.variableInitializer!

carterkozak:
Removing the dead link here: https://github.com/palantir/gradle-baseline/pull/1164

ferozco (approved):
Looks great!

carterkozak:
Thanks for the quick review :-)

Issue #1167, opened 2020-01-13
2.39.0 release broke java compilation on my project


tomkozlowski:
## What happened?

The change in #1085 appears to have broken my usage in a kotlin project during the `compileJava` step:

```
Picked up _JAVA_OPTIONS: -Dorg.gradle.internal.launcher.welcomeMessageEnabled=false -Xmx8192m
An exception has occurred in the compiler ((version info not available)). Please file a bug against the Java compiler via the Java bug reporting page (http://bugreport.java.com) after checking the Bug Database (http://bugs.java.com) for duplicates. Include your program and the following diagnostic in your report. Thank you.
java.lang.IllegalArgumentException: Cannot combine scanner suppliers with different implementations of 'MockitoInternalUsage': org.mockito.errorprone.bugpatterns.MockitoInternalUsage, com.google.errorprone.bugpatterns.MockitoInternalUsage
	at com.google.errorprone.scanner.ScannerSupplier.lambda$plus$6(ScannerSupplier.java:245)
	at com.google.common.collect.RegularImmutableBiMap.forEach(RegularImmutableBiMap.java:157)
	at com.google.errorprone.scanner.ScannerSupplier.plus(ScannerSupplier.java:239)
	at com.google.errorprone.ErrorPronePlugins.loadPlugins(ErrorPronePlugins.java:46)
	at com.google.errorprone.ErrorProneAnalyzer.lambda$scansPlugins$0(ErrorProneAnalyzer.java:77)
	at com.google.common.base.Suppliers$NonSerializableMemoizingSupplier.get(Suppliers.java:167)
	at com.google.errorprone.ErrorProneAnalyzer.finished(ErrorProneAnalyzer.java:151)
	at com.sun.tools.javac.api.MultiTaskListener.finished(MultiTaskListener.java:120)
	at com.sun.tools.javac.main.JavaCompiler.flow(JavaCompiler.java:1404)
	at com.sun.tools.javac.main.JavaCompiler.flow(JavaCompiler.java:1353)
	at com.sun.tools.javac.main.JavaCompiler.compile(JavaCompiler.java:946)
	at com.sun.tools.javac.api.JavacTaskImpl.lambda$doCall$0(JavacTaskImpl.java:100)
	at com.sun.tools.javac.api.JavacTaskImpl.handleExceptions(JavacTaskImpl.java:142)
	at com.sun.tools.javac.api.JavacTaskImpl.doCall(JavacTaskImpl.java:96)
[...]

PR #1169, opened 2020-01-14
Delete baseline-versions (and nebula.dependency-recommender)

### LINKED ISSUE #1131: Remove support for nebula dependency recommender
## What happened?
We moved to recommending all new projects use Gradle consistent versions instead of Nebula dependency recommender and nearly all internal projects have migrated. We should remove the unused code path and cut a major rev

iamdanfox:
## Before this PR

To avoid breaking consumers, baseline has bundled nebula.dependency-recommender for a _long_ time, even though we discourage people from using it (as we now recommend GCV).

Here's a recent example of nebula causing people confusion: https://github.com/palantir/gradle-baseline/issues/1167.

## After this PR
==COMMIT_MSG==
The deprecated `com.palantir.baseline-versions` plugin has been deleted, as we recommended using gradle-consistent-versions instead. (Alternatively, you can still manually set up nebula.dependency-recommender).
==COMMIT_MSG==

As a follow-up, we might want to excavate a cleanup of the `com.palantir.baseline-versions.disable=true` line from people's gradle.properties.

fixes https://github.com/palantir/gradle-baseline/issues/1131

## Possible downsides?
- taking away functionality might upset people who haven't migrated to GCV yet.


gradle-baseline-java/build.gradle
```diff
@@ -11,7 +11,6 @@ dependencies {
     compile 'com.diffplug.spotless:spotless-plugin-gradle'
     compile 'com.google.errorprone:error_prone_refaster'
     compile 'com.google.guava:guava'
-    compile 'com.netflix.nebula:nebula-dependency-recommender'
     compile 'com.palantir.configurationresolver:gradle-configuration-resolver-plugin'
```
iamdanfox, line 14: if we're queuing up breaks, perhaps we should cut this one too?

ferozco:
Big fan, this has been a long time coming 👍

PR #1171, opened 2020-01-16
Fix `LoggerEnclosingClass` edge cases for type parameters and anonymous classes

carterkozak:
## Before this PR
Possible to suggest `LoggerFactory.getLogger(Bar<T>.class)` instead of `LoggerFactory.getLogger(Bar.class)`.
Also possible to fail when loggers are defined as fields of anonymous classes.

## After this PR
==COMMIT_MSG==
Fix `LoggerEnclosingClass` edge cases for type parameters and anonymous classes
==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/LoggerEnclosingClass.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/LoggerEnclosingClass.java
@@ -21,7 +21,6 @@ import com.google.errorprone.BugPattern;
 import com.google.errorprone.VisitorState;
 import com.google.errorprone.bugpatterns.BugChecker;
 import com.google.errorprone.fixes.SuggestedFix;
-import com.google.errorprone.fixes.SuggestedFixes;
 import com.google.errorprone.matchers.Description;
 import com.google.errorprone.matchers.Matcher;
 import com.google.errorprone.matchers.Matchers;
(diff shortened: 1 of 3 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/LoggerEnclosingClass.java
```diff
@@ -83,8 +83,16 @@ public Description matchVariable(VariableTree tree, VisitorState state) {
...
-                        SuggestedFixes.prettyType(state, fix, enclosingClassSymbol.type) + ".class")
+                        enclosingClassSymbol.name + ".class")
                         .build())
                 .build();
     }
+
+    private static Symbol.ClassSymbol enclosingConcreteClass(Symbol.ClassSymbol input) {
+        Symbol.ClassSymbol current = input;
+        while (current != null && current.isAnonymous()) {
+            current = ASTHelpers.enclosingClass(current);
```
carterkozak, line 94: I haven't encountered this in the wild (it's an odd thing to do) but fixed it while I'm here.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/LoggerEnclosingClass.java
```diff
@@ -83,8 +83,16 @@ public Description matchVariable(VariableTree tree, VisitorState state) {
         return buildDescription(classArgument)
                 .addFix(fix.replace(
                         classArgument,
-                        SuggestedFixes.prettyType(state, fix, enclosingClassSymbol.type) + ".class")
+                        enclosingClassSymbol.name + ".class")
```
carterkozak, line 86: Simple name is sufficient here because the code we're updating is nested within the type -- the previous approach using prettyType wasn't necessary and included type parameters.

iamdanfox (approved):

carterkozak:
dozer?

PR #1172, opened 2020-01-16
Remove checkstyle LineLength when formatted

dansanduleac:
## Before this PR

Repos being formatted with [palantir-java-format](https://github.com/palantir/palantir-java-format) can still run into line length issues, for instance due to a long `{@link ...}` in a javadoc.

However, we consider these acceptable.

## After this PR
==COMMIT_MSG==
Stop enforcing 120 character limit for repos which enrolled in palantir-java-format.
==COMMIT_MSG==

## Possible downsides?

A formatter regression could cause bad PRs with very long lines to still get merged automatically.

cc @robert3005

### DIFF
--- a/changelog/@unreleased/pr-1172.v2.yml
+++ b/changelog/@unreleased/pr-1172.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Stop enforcing 120 character limit for repos which enrolled in palantir-java-format.
+  links:
+  - https://github.com/palantir/gradle-baseline/pull/1172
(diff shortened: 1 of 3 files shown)


uschi2000:
Is the assumption that everyone has now moved to palantir-java-format?

dansanduleac:
No, this only opts you out of the LineLength check *iff* you have palantir-java-format applied (which ensures that `check` will fail if your code is not properly formatted).

uschi2000:
ok

iamdanfox:
Honestly I kinda liked having this rule as a backstop, because it proved that pjf complied with our historic opinions about what well formatted code looks.

dansanduleac:
Ok, I'll try to get the formatter to work around this for now..

PR #1185, opened 2020-01-23
Fix #1182: Don't run baseline-error-prone tests concurrently

### LINKED ISSUE #1182: Running :baseline-error-prone:test OOMs
When I clone the repo on my MacBook Pro and try to `./gradlew build`, I get consistent errors in the `:baseline-error-prone:test` task with `Aborting due to java.lang.OutOfMemoryError: Java heap space`. This typically happens about 20 seconds into the task.

This looks related to the fact that it's trying to run 16 tests in parallel in a single process, according to the Gradle output. The process appears to max out around 800-900 MB RAM in Activity Monitor and the hprof is around 600 MB.

I was able to work around this by adding the following to the baseline-error-prone/build.gradle:

```
tasks.withType(Test) {
    // Use doFirst so this doesn't get overwritten by the baseline-testing plugin
    doFirst {
        systemProperty("junit.jupiter.execution.parallel.enabled", "false");
    }
}
```

carterkozak:
Error prone validation is memory-intensive and parallel execution
can cause OOMs and segfaults.

## After this PR
==COMMIT_MSG==
baseline-error-prone tests no longer run concurrently in order to avoid OOMs.
==COMMIT_MSG==

## Possible downsides?
Tests take longer to run on systems with sufficient resources in configurations where gradle provides enough heap.

### DIFF
--- a/baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/CatchBlockLogExceptionTest.java
+++ b/baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/CatchBlockLogExceptionTest.java
@@ -21,10 +21,7 @@ import com.google.errorprone.CompilationTestHelper;
 import java.util.Optional;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
-import org.junit.jupiter.api.parallel.Execution;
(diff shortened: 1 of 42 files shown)


ferozco:
👍 Seems like this slows down unit-tests by ~30s but Its probably worth its probably alright for now

PR #1186, opened 2020-01-23
fix #894: Error Prone UnnecessaryLambdaArgumentParentheses

### LINKED ISSUE #894: Drop redundant parentheses around single lambda parameters
Drop / disallow parentheses in lambdas such as `(x) -> System.out.println(x)`

carterkozak:
Lambdas with a single parameter do not require argument parentheses.

```diff
- (value) -> value != null
+ value -> value != null
```

## After this PR
==COMMIT_MSG==
Lambdas with a single parameter do not require argument parentheses.
==COMMIT_MSG==

## Possible downsides?
This is already validated by checkstyle, however checkstyle doesn't provide functionality to automatically fix occurrences.

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/UnnecessaryLambdaArgumentParentheses.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/UnnecessaryLambdaArgumentParentheses.java
@@ -1,0 +1,100 @@
+/*
+ * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.baseline.errorprone;
+
+import com.google.auto.service.AutoService;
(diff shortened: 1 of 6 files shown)


carterkozak:
pending testing on a large internal project

ferozco:
We should also remove the existing checkstyle rule

carterkozak:
> We should also remove the existing checkstyle rule

Good call, updated and increased the level to WARN.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/UnnecessaryLambdaArgumentParentheses.java
```diff
@@ -0,0 +1,97 @@
...
+ * contributed upstream. There's an argument against combining the two because parentheses around lambda arguments
+ * cannot be parsed directly from the AST where other parenthesis checked by UnnecessaryParentheses can.
+ */
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "UnnecessaryLambdaArgumentParentheses",
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
+        providesFix = BugPattern.ProvidesFix.REQUIRES_HUMAN_ATTENTION,
+        severity = BugPattern.SeverityLevel.WARNING,
```
ferozco, line 43: I think we should go with error since checkstyle previous enforced this

carterkozak: Most projects fail on WARNING -- iirc we tried using ERROR for the `BracesRequired` check but ran into issues (possibly with generated code).

Since excavator will auto-fix anything in projects that don't fail on warnings, I don't think there's much risk.

ferozco: ok sure!

ferozco:
👍

PR #1192, opened 2020-01-27
Fix `UnnecessaryLambdaArgumentParentheses` handling primitive lambda args

carterkozak:
## Before this PR
```diff
- forEach((long value) -> action.accept(value));
+ forEach(long value -> action.accept(value)); // does not compile
```

This makes the check safer when unknown inputs are encountered.

## After this PR
==COMMIT_MSG==
Fix `UnnecessaryLambdaArgumentParentheses` handling explicitly typed primitive lambda args
==COMMIT_MSG==


ferozco:
👍 thanks carter!

PR #1197, opened 2020-01-29
Add an ErrorProne check for raw types

AlexLandau:
This adds an ErrorProne check that fails when it detects raw types. Currently, javac has a built-in warning type for this, but it's not as flexible as ErrorProne; for example, you can't fail on raw-types errors without also failing on deprecation warnings, and you can't add exemptions for generated code.

Upstream ErrorProne doesn't seem to have any check for this or requests for such a check, beyond this issue: https://github.com/google/error-prone/issues/253

For @SuppressWarnings, the check respects "rawtypes", which is already used by javac and Eclipse.

The current level is "warning"; feel free to change that.

I added the check to the BaselineErrorProneExtension due to seeing it in another PR adding a check.

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/RawTypes.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/RawTypes.java
@@ -61,10 +61,11 @@ public final class RawTypes extends BugChecker
 
     @Override
     public Description matchClass(ClassTree tree, VisitorState state) {
-        List<Tree> types = new ArrayList<>();
-        types.add(tree.getExtendsClause());
-        types.addAll(tree.getImplementsClause());
-        return testTypes(types);
+        Description extendsResult = testType(tree.getExtendsClause());
+        if (extendsResult != Description.NO_MATCH) {
+            return extendsResult;
+        }
+        return testTypes(tree.getImplementsClause());
     }
 
     @Override
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/RawTypes.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/RawTypes.java
@@ -1,0 +1,104 @@
+/*
+ * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.baseline.errorprone;
+
+import com.google.auto.service.AutoService;
+import com.google.errorprone.BugPattern;
+import com.google.errorprone.VisitorState;
+import com.google.errorprone.bugpatterns.BugChecker;
+import com.google.errorprone.matchers.Description;
+import com.google.errorprone.util.ASTHelpers;
+import com.sun.source.tree.ClassTree;
+import com.sun.source.tree.MethodTree;
+import com.sun.source.tree.NewClassTree;
+import com.sun.source.tree.ParameterizedTypeTree;
+import com.sun.source.tree.Tree;
+import com.sun.source.tree.VariableTree;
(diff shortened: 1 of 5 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/RawTypes.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    @Override
+    public Description matchNewClass(NewClassTree tree, VisitorState state) {
+        return testType(tree.getIdentifier());
+    }
+
+    @Override
+    public Description matchClass(ClassTree tree, VisitorState state) {
+        List<Tree> types = new ArrayList<>();
```
carterkozak: nit: we should try to avoid array allocations since this will be invoked for every class/interface/enum/etc. A large internal project will benefit from avoiding this.

Perhaps something along these lines?
```java
checkType(tree.getExtendsClause());
checkTypes(tree.getImplementsClause());
```

gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineErrorProneExtension.java
```diff
@@ -38,6 +38,7 @@
             "PreferListsPartition",
             "PreferSafeLoggableExceptions",
             "PreferSafeLoggingPreconditions",
+            "RawTypes",
```
carterkozak: This isn't necessary because the RawTypes check does not provide a suggested fix, this list enumerates the checks that we automatically fix by default.

I wonder how many projects this will impact, we could implement a automated fix to apply `?` as a replacement. I'm sure there are edge cases that it won't fix, but it should make most cases slightly better. The counterargument is that most of these should be replaced with more precise type parameters that humans are better equipped to write.

dansanduleac:
I'm supportive of this. Have we run it on a large repo to make sure that it doesn't cause any unexpected breaks?

Also the changelog message should be a one-line sentence explaining what you're adding, just "Add an ErrorProne check for raw types" should be enough

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/RawTypes.java
```diff
@@ -0,0 +1,104 @@
...
+        altNames = {"rawtypes"},
+        summary =
+                "Avoid raw types; add appropriate type parameters if possible.\n"
+                        + "This can be suppressed with @SuppressWarnings(\"rawtypes\") where necessary, such as when "
+                        + "interacting with older library code.")
+public final class RawTypes extends BugChecker
+        implements BugChecker.VariableTreeMatcher,
+                BugChecker.NewClassTreeMatcher,
+                BugChecker.ClassTreeMatcher,
+                BugChecker.MethodTreeMatcher {
```
carterkozak: Should this cover `TypeCastTreeMatcher` as well, to protect against:
```java
Map<String, String> value = new HashMap<>();
Object object = (Map) value;
```

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/RawTypes.java
```diff
@@ -0,0 +1,104 @@
...
+    private Description testType(Tree type) {
+        if (type == null) {
+            return Description.NO_MATCH;
+        }
+        Type realType = ASTHelpers.getType(type);
+        if (realType != null && realType.isRaw()) {
+            return buildDescription(type)
+                    .setMessage("Avoid raw types; add appropriate type parameters if possible. "
+                            + "The type was: "
+                            + realType
```
carterkozak: types can render in odd ways, especially when they include other generic type information. prettyType should make it a bit cleaner:
```suggestion
                            + MoreSuggestedFixes.prettyType(null, null, realType)
```

carterkozak:
Supportive of this check as well, thanks for putting it together @AlexLandau :-)

gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineErrorProneExtension.java
AlexLandau: Removed from this list.

I'd rather leave existing cases marked as warnings for human review than have them replaced by `?` or `Object`.

AlexLandau:
@dansanduleac Yes, I've been testing this against the big monorepo. Thanks for taking care of the changelog.

@carterkozak Thanks for the suggestions!

gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineErrorProneExtension.java
carterkozak: Sounds good

carterkozak:
lgtm, thanks! 👍

PR #1200, opened 2020-01-30
Allow leading underscores in variable names

JacekLach:
## Before this PR

We use leading underscores as a marking of intentionally unread variables, as required by StrictUnusedVariable. However checkstyle currently doesn't allow it in local variable names.
## After this PR
==COMMIT_MSG==
Allow leading underscores in local variable names
==COMMIT_MSG==


JacekLach:
I guess this is only relevant when trying to ignore `@CheckReturnValue`, so maybe should just use a suppression

PR #1210, opened 2020-02-04
Turn off UnnecessaryLambda check in tests

dansanduleac:
## Before this PR

UnnecessaryLambda errorprone check blocking baseline in projects with `-Werror`.

## After this PR
==COMMIT_MSG==
Turn off UnnecessaryLambda in tests where it doesn't provide value because there is a ubiquitous use case that is painful to fix.
==COMMIT_MSG==

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineErrorProneExtension.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineErrorProneExtension.java
@@ -57,8 +57,7 @@ public class BaselineErrorProneExtension {
             "ArrayEquals",
             "MissingOverride",
             "UnnecessaryParentheses",
-            "PreferJavaTimeOverload",
-            "UnnecessaryLambda");
+            "PreferJavaTimeOverload");
 
     private final ListProperty<String> patchChecks;
 
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineErrorProneExtension.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineErrorProneExtension.java
@@ -57,7 +57,8 @@ public class BaselineErrorProneExtension {
             "ArrayEquals",
             "MissingOverride",
             "UnnecessaryParentheses",
-            "PreferJavaTimeOverload");
+            "PreferJavaTimeOverload",
+            "UnnecessaryLambda");
 
     private final ListProperty<String> patchChecks;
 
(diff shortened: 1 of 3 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
```diff
@@ -154,6 +154,17 @@ private static void applyToJavaProject(Project project) {
                                 errorProneOptions.check("PreferSafeLoggingPreconditions", CheckSeverity.OFF);
                                 errorProneOptions.check("PreconditionsConstantMessage", CheckSeverity.OFF);
                             }));
+
+            project.getTasks().withType(JavaCompile.class).configureEach(javaCompile -> {
```
iamdanfox: is this intentionally only disabled for people that use `java-gradle-plugin`? It feels a little odd to disable it in such a specific case.

If we think this UnnecessaryLambda is genuinely not great signal, why not just turn it off for all test code?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineErrorProneExtension.java
```diff
@@ -57,7 +57,8 @@
             "ArrayEquals",
             "MissingOverride",
             "UnnecessaryParentheses",
-            "PreferJavaTimeOverload");
+            "PreferJavaTimeOverload",
+            "UnnecessaryLambda");
```
carterkozak: I recall this creating code that didn’t compile, might want to test it on something large and internal (or atlas) first

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
dansanduleac: that's what i meant to do, sorry

gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineErrorProneExtension.java
iamdanfox: @dansanduleac you can use the hot new `./gradlew compileJava compileTestJava -PerrorProneApply=UnnecessaryLambda`

dansanduleac: I've tested it on a project where this was causing trouble and it produced no fixes... if you recall this way, then maybe i can just remove it from here and let people deal with it

iamdanfox (approved):

PR #1107, opened 2019-12-10
Fix: Disable CatchSpecificity on java 13

robert3005:
## Before this PR
```
An unhandled exception was thrown by the Error Prone static analysis plugin.
ERROR: An unhandled exception was thrown by the Error Prone static analysis plugin.
                    try {
                    ^
     Please report this at https://github.com/google/error-prone/issues/new and include the following:
  
     error-prone version: 2.3.4
     BugPattern: CatchSpecificity
```

## After this PR
==COMMIT_MSG==
Disable CatchSpecificity on java 13
==COMMIT_MSG==

PR #1095, opened 2019-12-05
Turn off UnnecessaryLambda, ThrowSpecificity on Java 13

iamdanfox:
## Before this PR

A few early adopters internally are building/running with Java13, but the recent error-prone 2.3.4 upgrade seems to have introduced a couple of blockers, seemingly with a similar root cause:
```
Caused by: java.lang.NoSuchMethodException: com.sun.tools.javac.comp.Resolve.findIdent(com.sun.tools.javac.comp.Env,com.sun.tools.javac.util.Name,com.sun.tools.javac.code.Kinds$KindSelector)
```

```
An unhandled exception was thrown by the Error Prone static analysis plugin.
    private static final Supplier<RuntimeException> NOT_AUTHORIZED = () -> new ServiceException(
                                                    ^
     Please report this at https://github.com/google/error-prone/issues/new and include the following:

     error-prone version: 2.3.4
     BugPattern: UnnecessaryLambda
     Stack Trace:
     java.lang.LinkageError: com.sun.tools.javac.comp.Resolve.findIdent(com.sun.tools.javac.comp.Env,com.sun.tools.javac.util.Name,com.sun.tools.javac.code.Kinds$KindSelector)
        at com.google.errorprone.util.FindIdentifiers.findIdent(FindIdentifiers.java:104)
        at com.google.errorprone.fixes.SuggestedFixes.qualifyType(SuggestedFixes.java:289)
        at com.google.errorprone.fixes.SuggestedFixes$5.visitClassType(SuggestedFixes.java:1082)
        at com.google.errorprone.fixes.SuggestedFixes$5.visitClassType(SuggestedFixes.java:1065)
        at jdk.compiler/com.sun.tools.javac.code.Type$ClassType.accept(Type.java:1010)
        at com.google.errorprone.fixes.SuggestedFixes.prettyType(SuggestedFixes.java:1064)
        at com.google.errorprone.bugpatterns.UnnecessaryLambda.lambdaToMethod(UnnecessaryLambda.java:215)
        at com.google.errorprone.bugpatterns.UnnecessaryLambda.matchVariable(UnnecessaryLambda.java:162)
        at com.google.errorprone.scanner.ErrorProneScanner.processMatchers(ErrorProneScanner.java:451)
        ...
[...]

PR #1213, opened 2020-02-05
-Pcom.palantir.baseline-error-prone.disable

iamdanfox:
## Before this PR

I want to start running some unit tests with Java13, but currently error-prone barfs, see:

- https://github.com/palantir/gradle-baseline/pull/1107
- https://github.com/palantir/gradle-baseline/pull/1106
- https://github.com/palantir/gradle-baseline/pull/1095

## After this PR
==COMMIT_MSG==
`./gradlew compileJava -Pcom.palantir.baseline-error-prone.disable` turns off error-prone, to allow compilation on Java 13 which is not yet supported by error-prone.
==COMMIT_MSG==

## Possible downsides?
- not sure I want people to start using this as part of their regular dev workflow :/

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
@@ -64,6 +64,7 @@ public final class BaselineErrorProne implements Plugin<Project> {
     private static final String ERROR_PRONE_JAVAC_VERSION = "9+181-r4173-1";
     private static final String PROP_ERROR_PRONE_APPLY = "errorProneApply";
     private static final String PROP_REFASTER_APPLY = "refasterApply";
+    private static final String DISABLE_PROPERY = "com.palantir.baseline-error-prone.disable";
 
     @Override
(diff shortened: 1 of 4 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
```diff
@@ -155,42 +181,6 @@ private static void applyToJavaProject(Project project) {
...
-        project.getTasks().withType(JavaCompile.class).configureEach(javaCompile -> {
-            ((ExtensionAware) javaCompile.getOptions())
-                    .getExtensions()
-                    .configure(ErrorProneOptions.class, errorProneOptions -> {
-                        // Relax some checks for test code
-                        if (errorProneOptions.isCompilingTestOnlyCode()) {
-                            errorProneOptions.check("UnnecessaryLambda", CheckSeverity.OFF);
-                        }
-                    });
-        });
```
iamdanfox, line 168: this block was pretty much an exact duplicate of a block above, so I moved this logic further down.

CRogers (approved):

PR #1217, opened 2020-02-06
License derived from .baseline/copyright

dansanduleac:
## Before this PR

Enforcing copyright file using checkstyle which is hard to change and not in sync with whatever files are in `.baseline/copyright`.

## After this PR
==COMMIT_MSG==
Enforcing copyright based on the first file (lexicographically) founds in `.baseline/copyright`.

This now uses spotless, which means that we can now autofix all copyrights using gradle.

It also allows us to enforce this copyright on groovy files, which we didn't have before.
==COMMIT_MSG==

## Possible downsides?
* the copyright at the top of all (java, groovy) files must be identical (except the year) or else it will get rewritten
* it might make it more difficult to deal with copied files that had another license already on them
  * can deal with it by having two comments, the standard repo one, and then another one with the original attribution


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineFormat.java
```diff
@@ -41,6 +55,104 @@
...
+        // Compute the copyright lazily
+        Supplier<FormatterStep> realStep = Suppliers.memoize(() -> {
+            String header = computeCopyrightHeader(project);
+            return LicenseHeaderStep.createFromHeader(header, delimiter);
+        });
+
+        return (FormatterStep) Proxy.newProxyInstance(
+                Thread.currentThread().getContextClassLoader(),
+                new Class[] {FormatterStep.class},
+                (_proxy, method, args) -> method.invoke(realStep.get(), args));
```
iamdanfox: can we use a LazyInitializedFormatterStep containing a memozied suppleire instead pls?

gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineFormatCopyrightIntegrationTest.groovy
```diff
@@ -0,0 +1,136 @@
...
+        /*
+         * (c) Copyright ${LocalDate.now().year} Palantir
+         *
+         * EXTRA
+         */
+    """.stripIndent()
+
+    static goodCopyright = """\
+        /*
+         * (c) Copyright 2019 Palantir
```
iamdanfox: can we use MyCompany instead of Palantir to emphasise this is not prod usage

iamdanfox:
Looks cool, might be worth a mention in the README that the copyright directory is important, and that running `./gradlew spotlessApply` should fix things :)

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineFormat.java
dansanduleac: yep, and turns out doing so now showcased that this thing was trivially serializable which is not quite what you want.

README.md
```diff
@@ -224,9 +224,11 @@ as that file will be overridden on updates.
 
 ### Copyright Checks
 
-By default Baseline enforces Palantir copyright at the beginning of files. To change this, edit the template copyright
-in `.baseline/copyright/*.txt` and the RegexpHeader checkstyle configuration in `.baseline/checkstyle/checkstyle.xml`
+Baseline enforces Palantir copyright at the beginning of files when applying `com.palantir.baseline-format`. To change this, edit the template copyright
+in `.baseline/copyright/*.txt`.
+Note that only the first file (sorted lexicographically) is chosen as the copyright for the entire project.
 
+To automatically update all files with mismatching/missing copyrights, run `./gradlew format`.
```
iamdanfox, line 230: `./gradlew format` feels a bit odd here. I think I'd almost recommend `spotlessApply` here

dansanduleac: Isn't it the same thing though?

dansanduleac: Recommending another command that `format` intentionally aliases feels like it might confuse people.

iamdanfox:
So running this internally on sls-logging, it seems it currently wrecks the world (and replaces our internal shorthand license with the apache one, which isn't correct). Think we need to:

- ensure this task tolerates _any_ of the templatets in the .baseline/copyright dir
- prefers the _highest_ prefix (not lowest)
- probably update excavator to delete the apache license template from internal repos (as internal code is not apache license by default!)

iamdanfox:
@mattsills this should mean you don't need to keep fixing up copyrights each time your baseline gets updated!

iamdanfox (approved):

PR #1220, opened 2020-02-07
./gradlew test -Drecreate=true

iamdanfox:
## Before this PR

A number of our repos use the `./gradlew test -Drecreate=true` pattern, but I recently found (on our circle-templates repo) that we've been doing it wrong the entire time.  Specifically, the task short-circuits and doesn't run when there are no code changes, even though the behaviour _should_ be different when `-Drecreate=true` is passed in.

## After this PR
==COMMIT_MSG==
For repos that use snapshot-style testing, `./gradlew test -Drecreate=true` will ensure the `"recreate"` system property is passed through to Java correctly.
==COMMIT_MSG==

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineTesting.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineTesting.java
@@ -46,6 +46,12 @@ public final class BaselineTesting implements Plugin<Project> {
                 // Never cache test tasks, until we work out the correct inputs for ETE / integration tests
                 task.getOutputs().cacheIf(t -> false);
             }
+
+            // repos that use 'snapshot' style testing should all use one convenient task to refresh the snapshots,
+            // ./gradlew test -Drecreate=true
+            String shouldRecreate = System.getProperty("recreate", "false");
+            task.systemProperty("recreate", shouldRecreate);
+            task.getInputs().property("recreate", shouldRecreate);
         });
 
         project.getPlugins().withType(JavaPlugin.class, unusedPlugin -> {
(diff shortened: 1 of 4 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineTesting.java
```diff
@@ -46,6 +46,12 @@ public void apply(Project project) {
                 // Never cache test tasks, until we work out the correct inputs for ETE / integration tests
                 task.getOutputs().cacheIf(t -> false);
             }
+
+            // repos that use 'snapshot' style testing should all use one convenient task to refresh the snapshots,
+            // ./gradlew test -Drecreate=true
+            String shouldRecreate = System.getProperty("recreate", "false");
+            task.systemProperty("recreate", shouldRecreate);
+            task.getInputs().property("recreate", shouldRecreate);
```
carterkozak: Should be be defensive and disable caching when shouldRecreate=true?

carterkozak: We may want to require/force boolean values here, possibly `Boolean.getBoolean("recreate")` and use the string value for the sysprop+input?

ferozco:
Its seems slightly out of scope for baseline to handle this convention and the upside for adding this is pretty low since in the worst case things will fail at CI time

dansanduleac (approved):

dansanduleac:
The reason to put this in a common place is that we keep copy pasting a chunk of gradle to help us do this, and we'd rather do away with the duplication.

iamdanfox:
@ferozco I mean we have quite a diverse chunk of functionality in baseline already, stuff to give us reproducibility, compilation fixes, circleci.

While it's a bit weird, I think it's in line with the overall goal of trying to minimise floating blobs of identical gradle across projects (especially when you find a bug in one of them!).

ferozco:
👍

PR #1227, opened 2020-02-10
License replacement attempts to preserve existing year info

iamdanfox:
## Before this PR

@tusharnarayan got a PR to upgrade baseline on his project and it nuked all his existing license (which were pretty bonkers anyway) and replaced them with 2020 ones. This upsets legal, who prefer to preserve the original creation date of files where possible.

## After this PR
==COMMIT_MSG==
Auto-fixing license headers will attempt to preserve the existing year of creation.
==COMMIT_MSG==

I predict this will be ***incredibly useful*** when we want to open source an internal project, because we should be able to just swap out the license template it .baseline/copyright and then run `./gradlew format`.

Confirming I ran this on tushar's project and it did the right thing.

## Possible downsides?
- we diverge further from spotless' existing LicenseHeaderStep

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/MultiLicenseHeaderStep.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/MultiLicenseHeaderStep.java
@@ -80,6 +80,7 @@ public final class MultiLicenseHeaderStep implements Serializable {
     private static class LicenseHeader implements Serializable {
         private static final long serialVersionUID = 1L;
         private static final Pattern YEAR_RANGE = Pattern.compile("[0-9]{4}(-[0-9]{4})?");
+        private static final String YEAR_TOKEN = "${today.year}";
 
         private final String licenseHeader;
         private final boolean hasYearToken;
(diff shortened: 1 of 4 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/MultiLicenseHeaderStep.java
```diff
@@ -55,113 +60,95 @@
...
-            if (hasYearToken) {
-                return licenseHeaderWithYearTokenReplaced;
-            }
-            return licenseHeader;
+        private String render(String existingHeader) {
+            Matcher yearInfo = YEAR_RANGE.matcher(existingHeader);
+            boolean existingHeaderContainsYear = yearInfo.find();
+            String year = existingHeaderContainsYear
+                    ? yearInfo.group(0)
+                    : String.valueOf(YearMonth.now().getYear());
```
iamdanfox, line 129: this line is really the money. 

the rest of the PR is mainly just me deleting indirection that doesn't seem necessary now that we're not working around spotless' API

ferozco (approved):

ferozco:
Thanks @iamdanfox!

PR #1252, opened 2020-02-21
Add an id for @VisibleForTesting annotation checks.

jkozlowski:
## Before this PR

Cannot selectively turn this check off.

## After this PR
==COMMIT_MSG==
@VisibleForTesting annotation has an id.
==COMMIT_MSG==

### DIFF
--- a/.baseline/checkstyle/checkstyle.xml
+++ b/.baseline/checkstyle/checkstyle.xml
@@ -19,6 +19,7 @@
         <property name="lineSeparator" value="lf"/>
     </module>
     <module name="RegexpMultiline"> <!-- Development Practices: Writing good unit tests -->
+        <property name="id" value="VisibleForTestingPackagePrivate"/>
         <property name="fileExtensions" value="java"/>
         <property name="format" value="@VisibleForTesting\s+(protected|public)"/>
         <property name="message" value="@VisibleForTesting members should be package-private."/>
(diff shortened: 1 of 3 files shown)


.baseline/checkstyle/checkstyle.xml
```diff
@@ -19,6 +19,7 @@
         <property name="lineSeparator" value="lf"/>
     </module>
     <module name="RegexpMultiline"> <!-- Development Practices: Writing good unit tests -->
+        <property name="id" value="VisibleForTestingPackagePrivate"/>
```
iamdanfox: you actually need to make this change to the other checkstyle.xml - this one will get overwritten next time you run `./gradlew baselineUpdateConfig` in this repo.

carterkozak (dismissed):

.baseline/checkstyle/checkstyle.xml
jkozlowski: 🤦‍♂

ferozco:
Should we just migrate this to an errorprone rule?

carterkozak:
Generally in favor of moving everything we can to errorprone :-)

.baseline/checkstyle/checkstyle.xml
jkozlowski: Done

carterkozak:
Alternative proposal: https://github.com/palantir/gradle-baseline/pull/1253

jkozlowski:
Ah, great @carterkozak! Gonna close since, that should allow me to suppress correctly.

PR #1253, opened 2020-02-21
Migrate VisibleForTesting validation from checkstyle to errorprone

carterkozak:
==COMMIT_MSG==
Migrate VisibleForTesting validation from checkstyle to errorprone
==COMMIT_MSG==

The fix exists primarily for documentation, it's not applied by default because it could break API and cause compilation to fail.


jkozlowski:
@iamdanfox ping

ferozco:
👍 Thanks @carterkozak!

PR #1254, opened 2020-02-24
New plugin BaselineFixGradleJava

dansanduleac:
## Before this PR

Users can resolve configurations like `compile`, `runtime` and `compileOnly`, which are meant to be buckets of dependencies (you can only add dependencies to them), never actually resolved or depended on.

Also, resolving these can trigger a Gradle bug - https://github.com/gradle/gradle/issues/11844#issuecomment-585219427 - when running code that attempts to resolve all resolvable configurations. This bug seems to be encountered less often when these configurations are not being resolved.

## After this PR
==COMMIT_MSG==
A new `com.palantir.baseline-fix-gradle-java` plugin fixes up the deprecated configurations of java source sets so they cannot be resolved or depended on.
==COMMIT_MSG==

## Possible downsides?

This can break external plugins. I think this is a good opportunity to discover these problems though, as later the same break will be introduced in Gradle 7. And as mentioned in Before this PR, the fact that these are resolvable is already causing some problems.


fawind (approved):
Looks good to me!

I'm missing a bit of context to evaluate how this will affect existing repos. If you are worried about this, feel free to tag another reviewer. Otherwise, merge away!

PR #1262, opened 2020-02-26
BaselineExactDependencies applies to all source sets

ferozco:
## Before this PR
BaselineExactDependencies only applied to the main source set

## After this PR
==COMMIT_MSG==
BaselineExactDependencies applies to all source sets
==COMMIT_MSG==

## Possible downsides?
This will flag even more unused dependencies (including in tests), and could block upgrades

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineExactDependencies.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineExactDependencies.java
@@ -45,6 +45,7 @@ import org.gradle.api.attributes.Usage;
 import org.gradle.api.plugins.JavaPlugin;
 import org.gradle.api.plugins.JavaPluginConvention;
 import org.gradle.api.tasks.SourceSet;
+import org.gradle.api.tasks.TaskProvider;
 
 /** Validates that java projects declare exactly the dependencies they rely on, no more and no less. */
 public final class BaselineExactDependencies implements Plugin<Project> {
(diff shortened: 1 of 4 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineExactDependencies.java
```diff
@@ -60,47 +63,65 @@
...
+                    conf.getAttributes()
+                            .attribute(
+                                    Usage.USAGE_ATTRIBUTE, project.getObjects().named(Usage.class, Usage.JAVA_API));
+                });
+        TaskProvider<CheckUnusedDependenciesTask> sourceSetUnusedDependencies = project.getTasks()
+                .register(
+                        GUtil.toLowerCamelCase("checkUnusedDependencies " + sourceSet.getName()),
+                        CheckUnusedDependenciesTask.class,
+                        task -> {
+                            task.dependsOn(JavaPlugin.CLASSES_TASK_NAME);
```
dansanduleac: this should be per source set

gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckUnusedDependenciesTask.java
```diff
@@ -190,6 +191,11 @@ public final void dependenciesConfiguration(Configuration dependenciesConfigurat
     }
 
     public final void sourceOnlyConfiguration(Configuration configuration) {
+        Preconditions.checkNotNull(configuration);
+        Preconditions.checkArgument(
+                configuration.isCanBeResolved(),
+                "May only add sourceOnlyConfiguration if it is resolvable: %s",
+                configuration);
```
dansanduleac, line 198: To be fair this method shouldn't exist, because I don't know in what situation users might actually use it.
The name is confusing now too since it implies source-only configurations need to be added, but we already add (ignore) all compileOnly configurations from all source sets.

But for now, let's at least throw eagerly if the user gave us a bad configuration.

dansanduleac:
👍

iamdanfox:
I don't really want to apply this to test source sets, as I think it will cause more noisy and frustration.  The intention of this task is to reduce the cruft that people ship in prod, and requiring test sourcesets to be minimal I think is overly strict.

dansanduleac:
So, as mentioned in person I think it's fine in principle to require people's tests to also have exact dependencies, as this is an opt-in thing and can help reduce the amount of stuff downloaded when running tests.

However, having tried this in practice, `checkUnusedDependenciesTest` won't work in its current form because
* `testImplementation` extends from `implementation`
* `testCompile` extends from `compile`
 and so all of the dependencies coming from the `main` source set will show up as unused when running this on the `test` source set

ferozco:
Ok, seems like we're not in sync on the desired outcome and it doesn't help solve the initial problem that got me looking at this. Happy to close this out

dansanduleac:
I think we still want this outcome, it's just that we need to be clear on what dependencies we consider 'explicit' and thus markable as unused.
I wrote up a change that only resolves dependencies that come _directly_ from the `compile` and `implementation` of each source set, and this seems to solve the problem (I've tested it on a project using gradle-atlas internally).

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineExactDependencies.java
```diff
@@ -60,45 +65,100 @@
...
+                            .attribute(
+                                    Usage.USAGE_ATTRIBUTE, project.getObjects().named(Usage.class, Usage.JAVA_API));
+                });
+
+        // Figure out what our compile dependencies are while ignoring dependencies we've inherited from other source
+        // sets. For example, if we are `test`, some of our configurations extend from the `main` source set:
+        // testImplementation     extendsFrom(implementation)
+        //  \-- testCompile       extendsFrom(compile)
+        // We therefore want to look at only the dependencies _directly_ declared in the implementation and compile
+        // configurations (belonging to our source set)
```
CRogers, line 109: Can we write a test for this?

dansanduleac: Yep, wrote one

gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineExactDependenciesTest.groovy
```diff
@@ -109,6 +109,32 @@ class BaselineExactDependenciesTest extends AbstractPluginTest {
...
+        package pkg;
+        public class Foo {
+            void foo() {
+                com.google.common.collect.ImmutableList.of();
+            }
+        }
+        '''.stripIndent()
+
+        then:
+        def result = with('checkUnusedDependencies', '--stacktrace').build()
```
dansanduleac, line 136: Without my latest changes (with the previous implementation at https://github.com/palantir/gradle-baseline/pull/1262/files/bc8e9e2cca856833a02d3d8775c9641378a8953f) this test failed with:

```
* What went wrong:
Execution failed for task ':checkUnusedDependenciesTest'.
> Found 1 dependencies unused during compilation, please delete them from 'build.gradle' or choose one of the suggested fixes:
  	com.google.guava:guava
```

CRogers (approved):

PR #1267, opened 2020-02-28
Fix check dependencies

dansanduleac:
## Before this PR

Sadly merged #1262 too quickly and it's missing some inter-project task dependencies if you depend on a project which applies `java-library`.

## After this PR
==COMMIT_MSG==
Fix `checkUnusedDependencies` regression introduced in 3.7.0 where it was failing when finding a dependency on another project that applies `java-library`.
==COMMIT_MSG==


ferozco:
👍

PR #1273, opened 2020-03-02
Add back checkUnusedDependencies.ignore and checkImplicitDependencies.ignore

dansanduleac:
## Before this PR

We broke some "API" that was being used by gradle-conjure in #1262, namely the `ignore` method on a task named exactly `checkUnusedDependencies`.

## After this PR
==COMMIT_MSG==
Bring back the `ignore(String, String)` method on the `checkUnusedDependencies`  and `checkImplicitDependencies` tasks. This now ignores the coordinate for all source sets.
==COMMIT_MSG==


dansanduleac:
@CRogers I think we should bring back this functionality on the `checkImplicitDependencies` task too.. No plugin is configuring that one, but the break might still be annoying to people.

CRogers (approved):

PR #1274, opened 2020-03-02
Don't apply BaselineFixGradleJava by default

dansanduleac:
## Before this PR

3.5.0 (#1254) was breaking a lot of code that was resolving the bad configurations which we banned from resolving.

## After this PR
==COMMIT_MSG==
Fix regression in 3.5.0 where people's idea{} configuration code would now crash, by not applying `com.palantir.baseline-fix-gradle-java` by default.
==COMMIT_MSG==


carterkozak:
:+1:

dansanduleac:
Got another fix for the CMEs in #1262 coming in, will release together.

PR #1278, opened 2020-03-03
Disable left curl when using palantir-java-format

ferozco:
## Before this PR
@mglazer flagged an issue where PJF and checkstyle were fighting where a scope was defined within a switch statement

## After this PR
==COMMIT_MSG==
Disable `LeftCurly` CheckStyle rule when Palantir-Java-Format is applied
==COMMIT_MSG==


carterkozak:
Mind linking/describing the scenario where then analyzers disagree?
It might make sense to collocate PJF and our checkstyle configurations at some point so we can test outputs pass both.

carterkozak:
:+1:

mglazer (approved):

PR #1283, opened 2020-03-03
Fix class uniqueness non-deterministic ordering

CRogers:
## Before this PR
There isn't a deterministic ordering on various parts of the `baseline-class-uniqueness.lock` file that's produced. This can cause failures as the generated file is different from the one on disk, despite having the same information.

## After this PR
==COMMIT_MSG==
The `checkClassUniqueness` will no longer spuriously fail due to inconsistent ordering in the `baseline-class-uniqueness.lock` file.
==COMMIT_MSG==

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckClassUniquenessLockTask.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckClassUniquenessLockTask.java
@@ -18,12 +18,9 @@ package com.palantir.baseline.tasks;
 
 import com.google.common.collect.ImmutableList;
 import java.io.File;
-import java.util.Collection;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Optional;
-import java.util.Set;
+import java.util.*;
 import java.util.stream.Collectors;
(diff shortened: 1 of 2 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckClassUniquenessLockTask.java
```diff
@@ -92,25 +94,18 @@ public final void doIt() {
...
+                                    this::clashingJarHeader, clashingJars -> clashingClasses(analyzer, clashingJars)));
+
+                    return Optional.of(clashingHeadersToClasses.entrySet().stream()
+                            .sorted(Comparator.comparing(Map.Entry::getKey))
+                            .flatMap(entry -> {
+                                String clashingJarHeader = entry.getKey();
+                                String clashingClasses = entry.getValue();
+                                return Stream.of(clashingJarHeader, clashingClasses);
+                            })
+                            .collect(Collectors.joining("\n")));
```
dansanduleac: why not just put them in a sorted map?

dansanduleac (approved):

fawind (approved):
lgtm

PR #1292, opened 2020-03-06
Disable catch specificity by default

ferozco:
## Before this PR
`CatchSpecificity` did not fail builds by default, but did result in a lot of output at compile time making it hard to quickly see the errors.

## After this PR
==COMMIT_MSG==
Disable `CatchSpecificity` errorprone check by default
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-1292.v2.yml
+++ b/changelog/@unreleased/pr-1292.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Disable `CatchSpecificity` errorprone check by default
+  links:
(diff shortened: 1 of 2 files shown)


fawind (approved):

carterkozak (approved):

pkoenig10:
Should we also disable `ThrowSpecificity`?

carterkozak:
If it's causing noise, feel free to push a PR to disable it. It has been helpful in a number of places but I don't think it buys us a lot until we're ready to start working on checks for proper InterruptedException handling, which won't be for a while.

PR #1299, opened 2020-03-19
Tidy duplicate method types

Jolyon-S:
## Before this PR
Methods with multiple arguments with the same types (or some types that are subtypes of others) are prone to mistakes.

## After this PR
==COMMIT_MSG==
Errorprone warning when methods have some arguments that have the same type or are subtypes of other arguments.
==COMMIT_MSG==

## Possible downsides?
Pretty aggressive check that could cause some code bloat from argument interfaces.

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DuplicateArgumentTypes.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DuplicateArgumentTypes.java
@@ -22,8 +22,10 @@ import com.google.errorprone.bugpatterns.BugChecker;
 import com.google.errorprone.matchers.Description;
 import com.google.errorprone.matchers.Matcher;
 import com.google.errorprone.matchers.Matchers;
+import com.google.errorprone.util.ASTHelpers;
 import com.sun.source.tree.MethodTree;
 import com.sun.source.tree.Tree;
+import com.sun.tools.javac.tree.JCTree;
 import java.util.HashSet;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicBoolean;
@@ -62,16 +64,23 @@ public final class DuplicateArgumentTypes extends BugChecker implements BugCheck
             Tree type = param.getType();
             tree.getParameters().forEach(param2 -> {
                 if (!param.equals(param2)) {
-                    Tree type2 = param.getType();
-                    bad.set((isSubtypeOf(type.getClass(), type2, state) || isSubtypeOf(type2.getClass(), type, state)));
+                    Tree type2 = param2.getType();
+                    // Matcher<Tree> m = Matchers.isSubtypeOf($ -> ASTHelpers.getType(type));
+                    bad.set(bad.get() || (isSubtypeOf(type, type2, state) || isSubtypeOf(type2,
+                            type, state)));
                 }
             });
         });
 
+        System.out.println(bad.get());
         return Description.NO_MATCH;
     }
 
-    private boolean isSubtypeOf(Class clazz, Tree tree, VisitorState state) {
-        return Matchers.isSubtypeOf(clazz).matches(tree, state);
+    private boolean isSubtypeOf(Tree clazz, Tree tree, VisitorState state) {
+        boolean x = Matchers.isSubtypeOf($ -> ASTHelpers.getType(clazz)).matches(tree, state);
+        if(x) {
+            System.out.println("madness abides");
+        }
+        return x;
     }
 }
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DuplicateArgumentTypes.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DuplicateArgumentTypes.java
@@ -1,0 +1,53 @@
+/*
+ * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.palantir.baseline.errorprone;
+
+import com.google.auto.service.AutoService;
+import com.google.errorprone.BugPattern;
+import com.google.errorprone.VisitorState;
+import com.google.errorprone.bugpatterns.BugChecker;
+import com.google.errorprone.matchers.Description;
+import com.sun.source.tree.MethodTree;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * Development Practices: Writing good unit tests.
+ */
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "DuplicateArgumentTypes",
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
+        severity = BugPattern.SeverityLevel.WARNING,
+        summary = "Duplicate argument types")
+public final class DuplicateArgumentTypes extends BugChecker implements BugChecker.MethodTreeMatcher {
+
+    @Override
+    public Description matchMethod(MethodTree tree, VisitorState state) {
+        Set<String> types = new HashSet<>();
+        AtomicInteger counter = new AtomicInteger(0);
+        tree.getParameters().forEach(la -> {
+            types.add(la.getType().toString());
+            counter.incrementAndGet();
+        });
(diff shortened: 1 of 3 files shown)


iamdanfox:
Is this intended to discourage people from writing methods like `foo(String, String, String, String)` (i'd be v. keen for this) or is it aimed at strange overloads?

Jolyon-S:
The former: don't want methods like `foo(String, String, String, String)` as you said, nor really methods like `bar(int, Integer, Number, Long)` or something like that (because there is equal opportunity to make mistakes there). Would be open to any improved approaches over what I've done - this is very much a WIP/draft, and I haven't worked with this code before.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DuplicateArgumentTypes.java
```diff
@@ -0,0 +1,115 @@
...
+                    Tree innerTree = param2.getType();
+                    badMethod.set(badMethod.get() || isSubTypeOf(outerTree, innerTree, state));
+                }
+            });
+        });
+
+        if (badMethod.get()) {
+            return buildDescription(tree)
+                    .setMessage("Some argument types are equal or are subtypes of each other. Consider using a "
+                            + "builder instead.")
```
dansanduleac: You can put this as the summary on the `@BugPattern` instead

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/DuplicateArgumentTypesTest.java
```diff
@@ -0,0 +1,156 @@
...
+public final class DuplicateArgumentTypesTest {
+
+    private static final BugCheckerRefactoringTestHelper.TestMode TEST_MODE =
+            BugCheckerRefactoringTestHelper.TestMode.AST_MATCH;
+
+    @Test
+    void testSameType() {
+        validator()
+                .addInputLines(
+                        "Test.java", "public class Test {", "  public void badMethod(Integer a, Integer b) {}", "}")
```
dansanduleac: can put a `//` after `"Test.java",` to make the formatter align these one per line

dansanduleac (commented):
Isn't it a bit too strict that this would flag methods like these?

```java
interface Foo {
  Expression add(Expression a, Expression b);

  // I've actually seen this in a bunch of places in code since this doesn't exist natively until java 9+
  <T> Optional<T> orElse(Optional<T> a, Optional<T> b);
}
```

Perhaps we should at least require 3 arguments in order to flag a method.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DuplicateArgumentTypes.java
```diff
@@ -0,0 +1,115 @@
...
+                    .build();
+        }
+
+        return Description.NO_MATCH;
+    }
+
+    private boolean isSubTypeOf(Tree tree1, Tree tree2, VisitorState state) {
```
dansanduleac: this method itself would get flagged by your check :)

Jolyon-S: I know - I should use a builder, but the way this code is written, it doesn't matter about the ordering. Still, I should practice what I preach!

Jolyon-S:
> Perhaps we should at least require 3 arguments in order to flag a method.

Do you mean three arguments in general, or three arguments that satisfy some sort of sub-typing relationship? I would be in favour of still having it trigger when there are only two arguments (at least, two arguments of the same / sub type), but as you point out that becomes very strict. @jkozlowski for thoughts

carterkozak:
Using a builder instead of a standard method call can reduce compile-time safety in other ways -- if we're not using staged/telescopic builders, library consumers aren't notified that they need to add new fields. Failures occur at runtime rather than compile time.

Jolyon-S:
> Using a builder instead of a standard method call can reduce compile-time safety in other ways -- if we're not using staged/telescopic builders, library consumers aren't notified that they need to add new fields. Failures occur at runtime rather than compile time.

True - but passing in two arguments the wrong way around can cause problems that aren't necessarily obvious at runtime, but even further down. If there is a way to do both, that would be ideal - i.e. surface problems still at compile time but also help to avoid the case of passing arguments the wrong way around, which can be pretty difficult to debug and also cause very non-obvious problems. @jkozlowski for context.

Jolyon-S:
@carterkozak Realised that `Immutables` library supports staged builders out of the box, so that makes it easy for people to replace their multi-argument methods with `Immutables` objects, then they need only add `@Value.Style(stagedBuilder = true)` and they get the guarantees that you are after.

Also, we only want this as a warning, not an error (not sure if I have set that correctly atm).

jkozlowski:
it would be great to get this out in a sort of strict mode (so the checks are as strict as they are right now), but where it doesn't fail the build. And we can test drive it and see in AtlasDB (and internally) how it ends up working. If it's a total failure then so be it, but I think being forced to add a suppression will at least make the person changing this code think through the implications.

It might end up being a total ergonomic hit, but we won't see until we try.

carterkozak:
Can we code-gen an immutables-based args object for these methods?

jkozlowski:
@carterkozak I don't know how to do that, but that would be cool! That could be a nice flup for this

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DuplicateArgumentTypes.java
jkozlowski: I was kinda thinking that we could initially have this for publicly visible methods.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DuplicateArgumentTypes.java
```diff
@@ -0,0 +1,130 @@
...
+
+/**
+ * Development Practices: Writing good unit tests.
+ */
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "DuplicateArgumentTypes",
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
+        severity = BugPattern.SeverityLevel.SUGGESTION,
```
dansanduleac: BLUF: not useful if people aren't forced to ack it somehow, whether by explicitly suppressing or yelling at us about it.

I don't think this is very useful, because nobody will see it when using `./gradlew idea`, and people who import via IntelliJ will see this as an error thanks to a long-standing intellij bug that I failed to fix correctly... I can tell you this is super annoying as your cursor is always swept away to the ~error~ note even if your code compiles correctly.
https://github.com/JetBrains/intellij-community/pull/1285

Either make it a warning and deal with the wrath of the -Werror'ers, or an error.

Jolyon-S: Just to confirm (since I don't have much experience with this) - you can enable this independently for repos, right? Will it default to being off, and be able to be enabled at will? I think we'd be happy to deal with the wrath but I don't want to impose it on everyone unless they opt-in.

jkozlowski: @dansanduleac I don't think I agree, we should be able to merge this to see how it feels, with opt-in, then followup with an excavator/something that rewrites your code, and then make this mandatory.

Same as with all the other improvements like enum casing etc.

ferozco:
I think we may want to consider a more targeted approach rather than disallowing any function with two parameters of the same type. I'm concerned that as is the check will flag a tons of completely benign patterns making it hard to adopt/roll out. I also worry about the impact the check will have on legibility by forcing users bloat their code with additional interfaces that don't necessarily solve the problem (case in point the check itself).

jkozlowski:
@ferozco that's why I suggested filtering out private methods to start with. But also I suspect we might be overestimating the prevalence of this. Every time I think about this in the past, it was a disaster waiting to happen, and every time I refactored the code to be explicit.

So basically are you guys against this completely or want to make some changes, but ultimately get this in? I would like us to get to some sort of agreement and next steps, so we can get a resolution.

dansanduleac:
I think we're not keen on it as it is, because it's too broad and a lot of currently benign cases would end up looking worse if they decide to address this check.
While not completely against the idea, I'd like you to come up with more targeted and fully-fledged examples of the badness this is trying to fix.

Then, based on that we could discuss heuristics that would limit this to only those cases as much as possible.

jkozlowski:
Sure, the badness was multiple production incidents because an order of arguments to a function (or constructor) of the same type were reversed. This broke some assumptions in other code and ultimately caused issues.

Here's the fix: https://github.com/palantir/atlasdb/pull/4638/files, and the bug was because of code like this: https://github.com/palantir/atlasdb/pull/4638/files#diff-a10f39471b5fc19a572d6f28de9cfb1eL41

We followed up with some extra hardening: https://github.com/palantir/atlasdb/pull/4645.

Having the code be flagged would probably have made the code use a builder in the first place, or just made the dev stop to think.

Is there a way for us to implement a custom error-prone check in our repo to test how well this works and how much code it would flag?

ferozco:
@jkozlowski you can publish this locally and trial it against whatever repo you would like (Its even possible to define the errorprone rule in the repo you would like to apply the rule on) but I'm don't think that we should move forward with this check in baseline

PR #1309, opened 2020-03-27
Remove SwitchStatementDefaultCase error-prone check

### LINKED ISSUE #499: SwitchStatementDefaultCase clashes with MissingCasesInEnumSwitch
## What happened?

Removed default from switch statement per SwitchStatementDefaultCase, now receive MissingCasesInEnumSwitch instead.

Revert/disable SwitchStatementDefaultCase or otherwise make it smarter.

### LINKED ISSUE #541: SwitchStatementDefaultCase should not apply to integers
## What happened?

SwitchStatementDefaultCase applied to a switch statements on an integer, which is a totally reasonable use case for `default` (special case 1, 2, 3, 4, 5, then do something generic for all other values).

SwitchStatementDefaultCase should only apply to enums.

### LINKED ISSUE #580: SwitchStatementDefaultCase doesn't make sense when switching on strings
## What happened?

I wrote some code which seemed vaguely sensible, but errorprone complained:

```
/Users/dfox/eggtimer-bot/eggtimer-bot/src/main/java/com/palantir/eggtimer/ReceiveWebhookResource.java:26: Note: [SwitchStatementDefaultCase] Avoid using default case in switch statement.
        switch (webhook) {
        ^
```

```java
switch (webhook) {
    // InstallationRepositoriesEvent: Triggered when a repository is added or removed from an installation.
    case "installation_repositories.added":
        Events.InstallationRepositoriesEvent added = deserialize(payload);
        log.info("Added", added.repositoriesAdded());
        return;

    case "installation_repositories.removed":
        Events.InstallationRepositoriesEvent removed = deserialize(payload);
        log.info("Removed", removed.repositoriesRemoved());
        return;
        
    default:
        log.debug("Received webhook", SafeArg.of("webhook", webhook), SafeArg.of("payload", payload));
        return;
}
```

Similar to https://github.com/palantir/gradle-baseline/issues/541

The whole point of our custom `SwitchStatementDefaultCase` is to make error-prone's exhaustiveness checks on _enums_ more effective. If someone is switching on strings (or other non-enum things), I think it should be disabled.

pkoenig10:
Fixes #499
Fixes #541 
Fixes #580

## Before this PR
Baseline provides a `SwitchStatementDefaultCase` error-prone check.

## After this PR
Baseline does not provide a `SwitchStatementDefaultCase` error-prone check.

I added this back in https://github.com/palantir/gradle-baseline/pull/481 with a very narrow focus on enums used in APIs. However there a many valid uses of default case in switch statements outside the narrow scope of interacting with an API.

We've ended disabling this check on many of our internal repos. And for repos that haven't disabled this check, having it at `SUGGESTION` results in a noisy compiler output.

A number of other people have experience friction with this check, see the issues linked above.

The built-in error-prone switch statement checks should be sufficient for our uses.


pkoenig10:
@iamdanfox @ferozco any thoughts on this?

pkoenig10:
@iamdanfox @ferozco another gentle ping

iamdanfox:
Ok so we're now just relying on product teams to know to avoid using `default` when they want to have exhaustiveness validated?

pkoenig10:
Yes.

This check was only ever a `SUGGESTION`, so it was never a hard failure and relied on noticing the message in the compiler output.

pkoenig10:
@iamdanfox any other concerns here?

iamdanfox (approved):

PR #1316, opened 2020-04-06
Enable check/refactorings for UnnecessaryBoxedVariable

gatesn:
## Before this PR
I've found this in a few PRs recently, would be good to turn it on.

## After this PR
errorProneApply will rewrite code to use primitive types where possible.

## Possible downsides?
It's not clear to me if we should set this to WARN or ERROR, or not at all?

The refactoring is still applied whatever level, but it might not be a great UX for your compilation to succeed and then find out it's wrong later? Perhaps that's fine and we should leave it at the default SUGGESTION level?

### DIFF
--- a/changelog/@unreleased/pr-1316.v2.yml
+++ b/changelog/@unreleased/pr-1316.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Enable ErrorProne check UnnecessaryBoxedVariable.
+  links:
+  - https://github.com/palantir/gradle-baseline/pull/1316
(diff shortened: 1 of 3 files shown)


CRogers:
I would prefer not to add yet another thing to block people's PRs - if we can get it so that it goes through but we then simply rewrite their code to be better in an excavator later, that seems preferable to me. Which means WARN should be fine (except the repos that have added `-Werror` - so maybe it should be SUGGESTION).

gatesn:
👍 Kept it as a suggestion, but the refactoring will apply

CRogers:
:+1:

carterkozak:
I recall this broke compilation changing public apis from nullable boxed values

CRogers:
Hmm, that's a good point, sounds like a blocker.

gatesn:
Wouldn't we catch that with rev-api? Particularly since code changes require human approval.

carterkozak:
Yes, but it would block baseline upgrades until a human is able to resolve the issue.

PR #1322, opened 2020-04-16
Provider implicit toString

gatesn:
## Before this PR
I was refactoring a Gradle plugin to start using Providers and got hit by this too many times.

## After this PR
Implicit use of a Provider as a String is forbidden.

## Possible downsides?
No auto-fix

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/GradleProviderToString.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/GradleProviderToString.java
@@ -1,0 +1,55 @@
+/*
+ * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.baseline.errorprone;
+
+import com.google.auto.service.AutoService;
+import com.google.errorprone.BugPattern;
+import com.google.errorprone.VisitorState;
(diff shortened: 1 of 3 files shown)


baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/GradleProviderToStringTest.java
```diff
@@ -0,0 +1,49 @@
...
+                        "import org.gradle.api.provider.Provider;",
+                        "class Foo implements Plugin<Project> {",
+                        "  public final void apply(Project project) {",
+                        "    Provider<String> provider = project.provider(() -> \"hello\");",
+                        "    // BUG: Diagnostic contains: Calling toString on a Provider",
+                        "    String value = \"My bad provider value: \" + provider;",
+                        "  }",
+                        "}")
+                .doTest();
+    }
```
CRogers, line 79: I would just add a second test to check that toString on non-providers is not broken.

CRogers (approved):
Nice, I like this

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/GradleProviderToString.java
```diff
@@ -0,0 +1,55 @@
...
+import com.sun.source.tree.ExpressionTree;
+import com.sun.source.tree.Tree;
+import java.util.Optional;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "GradleProviderToString",
+        summary = "Calling toString on a Provider does not render the contained value",
+        severity = BugPattern.SeverityLevel.ERROR)
+public final class GradleProviderToString extends AbstractToString {
```
iamdanfox, line 37: oh this is just too easy :D

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/GradleProviderToString.java
```diff
@@ -0,0 +1,55 @@
...
+
+    @Override
+    protected TypePredicate typePredicate() {
+        return IS_PROVIDER;
+    }
+
+    @Override
+    protected Optional<Fix> implicitToStringFix(ExpressionTree tree, VisitorState state) {
+        // We could automatically call Provider#get, but is that always right?
+        return Optional.empty();
```
carterkozak: if it's right enough of the time, we could do it for convenience. We don't need to automatically apply it.

gatesn:
Think I've addressed comments now

PR #1323, opened 2020-04-24
Bump on demand import threshold

a-k-g:
## Before this PR
The IDE tries to use on-demand imports if there are more than 99 things from the same place but our Checkstyle config completely disallows it. The 99 is actually just an arbitrarily high number to "disable" this auto-formatting rule in IDEs. We hit that in an internal product unfortunately, so increasing it further.

## After this PR
==COMMIT_MSG==
Bumping the on demand import threshold to 999
==COMMIT_MSG==

## Possible downsides?
don't think so.


carterkozak (approved):
Sounds great! I entered 999 manually to get around this, it's great to have the robots take care of it :-)

PR #1338, opened 2020-05-13
Fix unused dependencies breaking with gradle 6.4

dansanduleac:
## Before this PR

`checkUnusedDependencies` would break on gradle 6.4, because it tried to find the jars of other subprojects rather than the `classes` directories, but the plugin didn't wire up task dependencies onto the jars.

## After this PR
==COMMIT_MSG==
Fix task dependencies onto other subprojects for `com.palantir.baseline-exact-dependencies` tasks (`checkUnusedDependencies`, `checkImplicitDependencies`), so they work with gradle 6.4.
==COMMIT_MSG==

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineExactDependencies.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineExactDependencies.java
@@ -48,7 +48,8 @@ import org.gradle.api.artifacts.ResolvedDependency;
 import org.gradle.api.artifacts.component.ComponentIdentifier;
 import org.gradle.api.artifacts.component.ProjectComponentIdentifier;
 import org.gradle.api.attributes.Attribute;
-import org.gradle.api.attributes.Usage;
+import org.gradle.api.attributes.AttributeContainer;
+import org.gradle.api.plugins.JavaPlugin;
 import org.gradle.api.plugins.JavaPluginConvention;
 import org.gradle.api.tasks.SourceSet;
 import org.gradle.api.tasks.TaskProvider;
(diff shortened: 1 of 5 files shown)


dansanduleac:
It's just markdown failing on linked URLs it can't resolve... this is really frustrating, I wish we could override it with a check or something.
unit-test-11 was flaking before due to

> Caused by: org.gradle.internal.resource.transport.http.HttpErrorStatusCodeException: Could not GET 'https://jcenter.bintray.com/com/google/googlejavaformat/google-java-format/1.7/google-java-format-1.7.pom'. Received status code 502 from server: Bad Gateway

gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckUnusedDependenciesTask.java
```diff
@@ -176,7 +176,7 @@ private boolean shouldIgnore(ResolvedArtifact artifact) {
         return ignore.get().contains(BaselineExactDependencies.asString(artifact));
     }
 
-    @Input
+    @InputFiles
```
dansanduleac: this is really the key, if you don't set this then gradle doesn't wire up the tasks that produce various files inside these configurations as inputs...
I think we were just getting lucky before.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineExactDependencies.java
```diff
@@ -99,11 +101,20 @@ private static void configureSourceSet(
...
+                        attributes.attribute(
+                                Usage.USAGE_ATTRIBUTE, project.getObjects().named(Usage.class, Usage.JAVA_API));
+                        // Ensure we resolve the classes directory for local projects where possible, rather than the
+                        // 'jar' file. We can only do this on Gradle 5.6+, otherwise do nothing.
+                        if (GradleVersion.current().compareTo(GradleVersion.version("5.6")) >= 0) {
+                            attributes.attribute(
+                                    LibraryElements.LIBRARY_ELEMENTS_ATTRIBUTE,
+                                    project.getObjects().named(LibraryElements.class, LibraryElements.CLASSES));
+                        }
+                    });
```
dansanduleac, line 117: This is mostly an optimisation so we don't force the production of jars since we can use the classes dirs.
Gradle 6.4 has [made it so that](https://github.com/gradle/gradle/blob/42f7c3d0c3066b7b38bd0726760d4881e86fd19f/subprojects/plugins/src/main/java/org/gradle/api/plugins/internal/JvmPluginsHelper.java#L250-L255) it sometimes forces jars when resolving `compileClasspath` in order to be able to confidently extract the `Automatic-Module-Name` if there is one set in the manifest.
However I think we can always just use the classes dir for this case, as it won't change the behaviour.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckImplicitDependenciesTask.java
```diff
@@ -142,7 +142,7 @@ private boolean shouldIgnore(ResolvedArtifact artifact) {
         return ignore.get().contains(BaselineExactDependencies.asString(artifact));
     }
 
-    @Input
+    @InputFiles
```
ferozco: should this be `@Classpath`?

dansanduleac: good catch, yep that will fix any ordering issues!

ferozco (commented):
Could we put together some sort of test for this?

dansanduleac:
so there was a test specifically for this, but it was accidentally running `checkUnusedDependencies` instead of `:checkUnusedDependencies`, and thus hiding the fact that the task dependency wasn't there as expected. Fixed now.

ferozco:
👍 Awesome, thanks @dansanduleac

PR #1341, opened 2020-05-15
Insert inner class imports in Intellij

pkoenig10:
I find myself manually enabling this frequently. It's particularly useful for things like AtlasDB generated tables in which the row and column classes are nested classes like:
- `MyAtlasTable`
- `MyAtlasTable.MyAtlasRow`
- `MyAtlasTable.MyAtlasRowResult`
- etc.

PR #1347, opened 2020-05-19
Actually insert inner class imports

pkoenig10:
#1341 wasn't quite right. I've verified locally that this works as expected now.


ferozco:
👍

PR #1349, opened 2020-05-20
 Disable HTML test reports by default

pkoenig10:
## Before this PR
The `baseline-circle` enabled HTML test reports. Uploading these HTML test reports can take quite a while (nearly 2 minutes) on large projects because these files are not compressed and are uploaded individually.

## After this PR
The `baseline-circle` plugin does not enable HTML test reports.

The plugin still enables the XML test reports, to ensure tests results are displayed nicely in CircleCI. Test artifacts are compressed before uploading and don't suffer the same problem.

### DIFF
--- a/changelog/@unreleased/pr-1349.v2.yml
+++ b/changelog/@unreleased/pr-1349.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: The `baseline-circle` plugin no longer enables HTML test reports.
+  links:
+  - https://github.com/palantir/gradle-baseline/pull/1349
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineCircleCi.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineCircleCi.java
@@ -64,7 +64,7 @@ public final class BaselineCircleCi implements Plugin<Project> {
         }
 
         project.getRootProject().allprojects(proj -> proj.getTasks().withType(Test.class, test -> {
-            test.getReports().getHtml().setEnabled(true);
+            test.getReports().getHtml().setEnabled(false);
             test.getReports().getHtml().setDestination(junitPath(circleArtifactsDir, test.getPath()));
         }));
     }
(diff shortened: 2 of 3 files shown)


fawind:
You are talking about these reports, right? https://10386-164943450-gh.circle-artifacts.com/0/~/artifacts/junit/dialogue-core/test/index.html

I remember them being useful when the exception got swallowed by the Gradle output and you could still get the stacktrace from the report. So not sure if we simply want to remove them for all people.

Maybe we should make it easier to disable uploading reports on a per-project basis?

cc @iamdanfox who has more awareness on this.

carterkozak:
What if we conditionally save html reports on failure?

ferozco:
I find the test reports to be very valuable on failure, especially since they are the best/only way to see the full output of a test. Similarly, I'd be hesitant to upload a compressed version of the reports since it makes it much more annoying to use. 

@carterkozak suggestion of omitting the html reports on success seems reasonable though

pkoenig10:
Discussed with @iamdanfox and @robert3005 and the consensus was that disabling by default is acceptable because it is easy for projects to enable this if they need it. They just need to add:
```
test {
    report.html.enabled = true
}
```

ferozco:
Hold on a second, I don't want to be encouraging most projects to add custom gradle so that they can see the output of a test. Why not switch it the other way and ask repos that are sensitive the the cost of uploading files to opt out?

pkoenig10:
Do many repos have tests that generate output? My experience is that the vast majority of projects don't generate output from their unit tests.

ferozco:
I'd prefer to give as much info as possible by default rather than relying on dev's to look up that there is a magic gradle flag that lets them see what they want.

What is wrong with only emitting them on failure?

pkoenig10:
Is that possible? There's doesn't appear to be a way to configure the Gradle test reporting to do that.

pkoenig10:
Looking at internal CircleCI build data for the time it takes to upload artifacts in unit test builds, we have only ~15 repos that are consistently >30 seconds and only ~4 repos that that are consistently >90 seconds.

Given that I'm happy to close this PR in favor of custom Gradle config in these services. I'll admit it's a bit strange for the default to be `false` here.

ferozco:
cool, thanks for doing the extra investigation @pkoenig10

PR #1350, opened 2020-05-20
Stop IntelliJ suggesting commonly shaded packages

CRogers:
## Before this PR
Jars commonly contain shaded dependencies - you should never use them, but they end up polluting your autocomplete suggestions:

Screen Shot 2020-05-20 at 5 13 36 PM

Internally, someone was confused why their code wasn't compiling because they were using a shaded version of an interface that was generated by `gradle-shadow-jar`.

## After this PR
==COMMIT_MSG==
Classes in common shaded packages are no longer for autocompletion/auto-import recommended by IntelliJ.
==COMMIT_MSG==

Screen Shot 2020-05-20 at 5 27 48 PM

Screen Shot 2020-05-20 at 5 28 39 PM

## Possible downsides?
I can't really think of any. It's only "bad" if people want to use this shaded classes (they shouldn't) and would perhaps be annoying they're not suggested? Even then, this just stops suggestions, you can still find them/use them:

Screen Shot 2020-05-20 at 5 30 30 PM

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
@@ -19,9 +19,6 @@ package com.palantir.baseline.plugins
 import com.palantir.baseline.util.GitUtils
 import groovy.transform.CompileStatic
 import groovy.xml.XmlUtil
-import java.nio.file.Files
-import java.nio.file.Path
-import java.nio.file.Paths
 import org.gradle.api.Action
 import org.gradle.api.Project
 import org.gradle.api.Task
@@ -33,6 +30,10 @@ import org.gradle.plugins.ide.idea.GenerateIdeaWorkspace
 import org.gradle.plugins.ide.idea.IdeaPlugin
 import org.gradle.plugins.ide.idea.model.IdeaModel
 import org.gradle.plugins.ide.idea.model.ModuleDependency
+
+import java.nio.file.Files
+import java.nio.file.Path
+import java.nio.file.Paths
 
 class BaselineIdea extends AbstractBaselinePlugin {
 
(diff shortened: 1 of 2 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
```diff
@@ -286,6 +288,20 @@ class BaselineIdea extends AbstractBaselinePlugin {
...
+    private static void ignoreCommonShadedPackages(Node node) {
+        // language=xml
+        node.append(new XmlParser().parseText('''
+            <component name="JavaProjectCodeInsightSettings">
+              <excluded-names>
+                <name>shadow</name><!-- from gradle-shadow-jar -->
+                <name>org.gradle.internal.impldep</name>
+                <name>autovalue.shaded</name>
+                <name>org.inferred.freebuilder.shaded</name>
+              </excluded-names>
```
carterkozak, line 301: @ilyanep If memory serves you had a list of packages that are helpful to exclude, is there anything missing here?

ferozco:
Nice this is very slick! I really appreciate it for gradle plugin development

ferozco:
👍 @CRogers did you test this out locally to make sure the plumbing is all set?

ferozco:
wrecked, I approved at the same time you added the labels :(

CRogers:
Race conditions!

Yep, I tested locally on a couple of projects ([this "after" screenshot](https://user-images.githubusercontent.com/397795/82471990-64469600-9abf-11ea-84d4-cf44420de67e.png) was from one of them)

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
ilyanep: @carterkozak My full list is:

1. com.palantir.shaded
2. com.sun.javafx.collections
3. com.sun.tools.javac.util
4. com.sun.xml.internal
5. de.intarsys
6. io.vavr
7. jersey.repackaged
8. org.apache.poi.ss.formula
9. org.aspectj.lang.annotation
10. org.codehaus.jackson
11. org.jooq.Exception.IOException
12. shaded.com.palantir
13. shaded.es244
14. shaded.es6x

Copied from what I have configured right now, though not sure how many of those are generally applicable for baseline. Hope that's helpful!

PR #1355, opened 2020-05-27
Strict unused variable enforces lambda parameters

ferozco:
## Before this PR
`StrictUnusedVariable` would not ensure that all lambda parameters were unused

## After this PR
==COMMIT_MSG==
`StrictUnusedVariable` checks for unused lambda parameters
==COMMIT_MSG==

## Possible downsides?
Some code bases may have had existing conventions for denoting unused lambda parameters. However the fix that comes along with the rule should prevent upgrades from being blocked

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StrictUnusedVariable.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StrictUnusedVariable.java
@@ -508,7 +508,7 @@ public final class StrictUnusedVariable extends BugChecker implements BugChecker
             Symbol varSymbol, Symbol.MethodSymbol methodSymbol, List<TreePath> usagePaths, VisitorState state) {
         boolean isPrivateMethod = methodSymbol.getModifiers().contains(Modifier.PRIVATE);
         int index = methodSymbol.params.indexOf(varSymbol);
-        Preconditions.checkState(index != -1, "symbol {} must be a parameter to the owning method", varSymbol);
+        Preconditions.checkState(index != -1, "symbol %s must be a parameter to the owning method", varSymbol);
         SuggestedFix.Builder fix = SuggestedFix.builder();
         for (TreePath path : usagePaths) {
             fix.delete(path.getLeaf());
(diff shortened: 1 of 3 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StrictUnusedVariable.java
```diff
@@ -472,11 +479,34 @@ public Boolean visitEnhancedForLoop(EnhancedForLoopTree tree, Void unused) {
...
+        return ImmutableList.of(fix.build());
+    }
+
     private static ImmutableList<SuggestedFix> buildUnusedParameterFixes(
-            Symbol varSymbol, List<TreePath> usagePaths, VisitorState state) {
-        Symbol.MethodSymbol methodSymbol = (Symbol.MethodSymbol) varSymbol.owner;
+            Symbol varSymbol, Symbol.MethodSymbol methodSymbol, List<TreePath> usagePaths, VisitorState state) {
         boolean isPrivateMethod = methodSymbol.getModifiers().contains(Modifier.PRIVATE);
         int index = methodSymbol.params.indexOf(varSymbol);
+        Preconditions.checkState(index != -1, "symbol must be a parameter to the owning method");
```
dansanduleac: might be nice to log the symbol and the method if we're gonna throw? :)

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/StrictUnusedVariableTest.java
```diff
@@ -133,6 +151,30 @@ public void renames_unused_param() {
...
 
+    @Test
+    void renames_unused_lambda_params() {
+        refactoringTestHelper
+                .addInputLines(
+                        "Test.java",
+                        "import java.util.function.BiFunction;",
+                        "class Test {",
+                        "  private static BiFunction<String, String, Integer> doStuff() {",
+                        "    BiFunction<String, String, Integer> first = (String value1, String value2) -> 1;",
```
dansanduleac, line 162: I'd maybe use one of them to check that it only renames the other one

dansanduleac:
👍

pkoenig10:
👍

dansanduleac:
This just broke some code by dropping the contents before a lambda with a single (used) parameter `i -> i + 1` upon auto-fixing.

PR #1357, opened 2020-05-28
Revert "Strict unused variable enforces lambda parameters (#1355)"

dansanduleac:
This reverts commit 620c65b288d234b721d311ccac3f301635d8b35e.

## Before this PR

#1355 broke some consumers.

## After this PR
==COMMIT_MSG==
Revert recent regression #1355 from 5.18.0 which broke some consumers.
==COMMIT_MSG==


carterkozak (approved):
We can unrevert with a fix separately, let’s ship this to stop the bleeding.

ferozco:
👎 why rollback. lets just fix it

PR #1358, opened 2020-05-28
Fix strict unused lambda

dansanduleac:
## Before this PR

The suggested fix from #1355 failed for some lambdas where a seemingly unused parameter (starting with `_`) was actually used, and would end up deleting a bunch of code.

## After this PR
==COMMIT_MSG==
Correctly handle converting lambda parameters from unused (`_param`) to used (param`).
==COMMIT_MSG==


ferozco:
👍

ferozco:
Thanks @dansanduleac

PR #1359, opened 2020-05-28
Rewrite lambdas to method references for instance methods

ferozco:
## Before this PR
We would only re-write lambdas with zero parameters that invoked static methods to method references

## After this PR
==COMMIT_MSG==
Rewrite single parameter lambdas that only invoke an zero param instance method to a method reference
==COMMIT_MSG==

## Possible downsides?
N/A

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/LambdaMethodReference.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/LambdaMethodReference.java
@@ -17,6 +17,7 @@
 package com.palantir.baseline.errorprone;
 
 import com.google.auto.service.AutoService;
+import com.google.common.collect.Iterables;
 import com.google.errorprone.BugPattern;
 import com.google.errorprone.VisitorState;
 import com.google.errorprone.bugpatterns.BugChecker;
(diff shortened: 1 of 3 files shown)


carterkozak:
Nice, we should test this on a large internal project before merging. I found a lot of bugs in the original implementation that way!

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/LambdaMethodReference.java
```diff
@@ -93,23 +95,29 @@ private Description checkMethodInvocation(
...
+            Symbol.MethodSymbol methodSymbol, LambdaExpressionTree root, MethodInvocationTree methodInvocation) {
+        if (!methodSymbol.isStatic()) {
+            if (root.getParameters().size() == 1) {
+                Symbol paramSymbol = ASTHelpers.getSymbol(Iterables.getOnlyElement(root.getParameters()));
+                Symbol receiverSymbol = ASTHelpers.getSymbol(ASTHelpers.getReceiver(methodInvocation));
+                return !paramSymbol.equals(receiverSymbol);
+            }
+            return true;
+        }
+        return false;
```
carterkozak:
```suggestion
        return !root.getParameters().isEmpty();
```

carterkozak (approved):
Approving pending tests in a large internal project

carterkozak:
Everything is green.

carterkozak:
Needs an approval from you as well since my commits merged into this branch :-)

ferozco:
woohooo 👍

ferozco:
🤖

PR #1363, opened 2020-05-28
Fix for method references with type parameters

carterkozak:
`Optional<String>::toString` does not compile


ferozco:
👍 wrecked, we had checks for the method having type parameters but not the type

PR #1365, opened 2020-05-29
Convert multi param lambdas and local method invocations to method references

ferozco:
## Before this PR
We did not catch or convert any lambda that took multiple parameters to method references and we did not convert any local instance method either

## After this PR
==COMMIT_MSG==
Convert multi param lambdas and local method invocations to method references
==COMMIT_MSG==

## Possible downsides?
We want to make sure this works in all cases so we should test against our large internal repo before merging

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/LambdaMethodReference.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/LambdaMethodReference.java
@@ -17,6 +17,7 @@
 package com.palantir.baseline.errorprone;
 
 import com.google.auto.service.AutoService;
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Iterables;
 import com.google.errorprone.BugPattern;
 import com.google.errorprone.VisitorState;
@@ -34,6 +35,8 @@ import com.sun.source.tree.StatementTree;
 import com.sun.source.tree.Tree;
 import com.sun.tools.javac.code.Symbol;
 import com.sun.tools.javac.code.Type;
+import java.util.List;
+import java.util.Objects;
 import java.util.Optional;
 
 @AutoService(BugChecker.class)
@@ -91,10 +94,13 @@ public final class LambdaMethodReference extends BugChecker implements BugChecke
 
     private Description checkMethodInvocation(
             MethodInvocationTree methodInvocation, LambdaExpressionTree root, VisitorState state) {
-        if (!methodInvocation.getArguments().isEmpty()
-                || !methodInvocation.getTypeArguments().isEmpty()) {
+        if (!methodInvocation.getTypeArguments().isEmpty()) {
             return Description.NO_MATCH;
         }
+        if (!methodInvocation.getArguments().isEmpty()) {
+            return convertParameterizedMethodInvocation(methodInvocation, root, state);
+        }
+
         Symbol.MethodSymbol methodSymbol = ASTHelpers.getSymbol(methodInvocation);
         if (methodSymbol == null || shouldIgnore(methodSymbol, root, methodInvocation)) {
             return Description.NO_MATCH;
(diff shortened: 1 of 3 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/LambdaMethodReference.java
```diff
@@ -91,49 +89,97 @@ public Description matchLambdaExpression(LambdaExpressionTree tree, VisitorState
...
+        List<Symbol> lambdaParam = getSymbols(root.getParameters());
+
+        // We are guaranteed that all of root params are symbols so equality should handle cases where methodInvocation
+        // arguments are not symbols or are out of order
+        if (!methodParams.equals(lambdaParam)) {
+            return Description.NO_MATCH;
         }
-        return !root.getParameters().isEmpty();
+
+        boolean isLocal = ASTHelpers.getReceiver(methodInvocation) == null;
```
carterkozak: I think we should handle a receiver IdentifierTree with name “this” as well. Worth a test case if we don’t already have one (review on thumbs)

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/LambdaMethodReferenceTest.java
```diff
@@ -67,7 +68,83 @@ void testFunction() {
...
+                        "class Test {",
+                        "  public Optional<Integer> foo(Optional<String> optional) {",
+                        "    // BUG: Diagnostic contains: Lambda should be a method reference",
+                        "    return optional.map(v -> bar(v));",
+                        "  }",
+                        "  private Integer bar(String value) {",
+                        "    return value.length();",
+                        "  }",
+                        "}")
+                .doTest();
```
carterkozak, line 84: Not sure these negative tests are necessary when we have refactoring tests as well.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/LambdaMethodReference.java
```diff
@@ -91,49 +89,97 @@ public Description matchLambdaExpression(LambdaExpressionTree tree, VisitorState
...
+        }
+
+        if (methodInvocation.getArguments().size() == root.getParameters().size()) {
+            return convertMethodInvocations(methodSymbol, methodInvocation, root, state);
+        }
+
+        return Description.NO_MATCH;
+    }
+
+    private Description convertSuppliersAndVariableInstanceMethods(
```
carterkozak: Can we remove the supplier case here in favor of the zero-length method invocation code path?

carterkozak (commented):
Looks slick!

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/LambdaMethodReference.java
```diff
@@ -91,49 +89,97 @@ public Description matchLambdaExpression(LambdaExpressionTree tree, VisitorState
...
 
     private static String qualifyTarget(
             Symbol.MethodSymbol symbol,
             MethodInvocationTree invocation,
             SuggestedFix.Builder builder,
-            VisitorState state) {
+            VisitorState state,
+            boolean isLocal) {
+        if (!symbol.isStatic() && isLocal) {
+            return "this." + symbol.name.toString();
```
carterkozak: I think there might be some oddness if the method is. It provided by this, but an enclosing class, requiring prefix “Enclosing.this.“. Worth a test to verify, I’d probably exclude that from the check rather than attempt to refactor since it’s fairly uncommon.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/LambdaMethodReference.java
```diff
@@ -91,49 +89,93 @@ public Description matchLambdaExpression(LambdaExpressionTree tree, VisitorState
...
+        Symbol.MethodSymbol methodSymbol = ASTHelpers.getSymbol(methodInvocation);
+        if (methodSymbol == null || !methodInvocation.getTypeArguments().isEmpty()) {
             return Description.NO_MATCH;
         }
-        Symbol.MethodSymbol methodSymbol = ASTHelpers.getSymbol(methodInvocation);
-        if (methodSymbol == null || shouldIgnore(methodSymbol, root, methodInvocation)) {
+
+        ExpressionTree receiver = ASTHelpers.getReceiver(methodInvocation);
+        boolean isLocal = isLocal(methodInvocation);
+        if (!isLocal && !(receiver instanceof IdentifierTree)) {
```
carterkozak, line 109: Does this prevent us from fixing static method calls? I don't think static methods have receivers (but I may be incorrect)

```diff
- value -> Preconditions.checkNotNull(value)
+ Preconditions::checkNotNull
```

carterkozak:
Seeing a lot of places where we're not qualifying the method reference correctly. This results in:

```diff
-        open().runVoid(conn -> consumer.accept(conn));
+        open().runVoid(Consumer::accept);
```

where we want:

```diff
-        open().runVoid(conn -> consumer.accept(conn));
+        open().runVoid(consumer::accept);
```

however I think we only want to do this when `consumer` is final or effectively final. In the original the value of consumer was read when the funciton is called, where in the new version we capture the reference when the lambda is created.

carterkozak (approved):
Validated on a large internal project

ferozco:
👍

PR #1369, opened 2020-05-29
Ckozak/fo/lambda methods/fix

carterkozak:


carterkozak:
Still need to validate against lambdas with parameter types specified, this won't compile:
```diff
-                .consumer((String filePathString) -> files.add(filePathString))
+                .consumer(files::add)
```

ferozco:
nice! 👍

PR #1370, opened 2020-05-29
Improvement: Upgrade error-prone to 2.4.0

robert3005:
## Before this PR
We are using error-prone 2.3.4

## After this PR
https://github.com/google/error-prone/releases/tag/v2.4.0
==COMMIT_MSG==
Upgarde error-prone to 2.4.0 to support latest jdks
==COMMIT_MSG==

## Possible downsides?
There's 300 commits in the diff, surely there are bunch of checks that need autofixes applied


ferozco:
Lets try this out on a few repos before letting it rip more widely

robert3005:
We also should remove our jdk 12, 13 overrides as they shouldn't be necessary

carterkozak:
Many MoreASTHelpers/MoreSuggestedFixes methods are no longer necessary since our fixes were merged upstream.

robert3005:
Cool, will fix those as well. Also found that your JavaTime check bug was fixed so we can reenable that

robert3005:
~@carterkozak couldn't find things in MoreAstHelpers that could be removed - have specific methods/bugs in mind?~ Found some

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/CatchSpecificityTest.java
```diff
@@ -560,7 +560,7 @@ void typeParameterExceptionToRuntimeException() {
                         "  <T extends RuntimeException> void f(ThrowingRunnable<T> in) {",
                         "    try {",
                         "        in.run();",
-                        "    } catch (Exception e) {",
+                        "    } catch (RuntimeException e) {",
```
robert3005, line 577: this looked like a bug to me - in this test we know we throw T which in this scope is guaranteed to be RuntimeException (while unbounded T is Exceptio)

carterkozak: :+1:

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
```diff
@@ -224,18 +221,6 @@ private static void configureErrorProneOptions(
...
-            // works apart from this one check: https://github.com/google/error-prone/issues/1106
-            errorProneOptions.check("Finally", CheckSeverity.OFF);
-        }
-
-        if (jdkVersion.compareTo(JavaVersion.toVersion("13.0.0")) >= 0) {
-            // Errorprone 2.3.4 isn't officially compatible with Java13 either
-            // https://github.com/google/error-prone/issues/1106
-            errorProneOptions.check("TypeParameterUnusedInFormals", CheckSeverity.OFF);
-        }
-
```
carterkozak, line 238: :heart:

robert3005:
I found some cases of `JdkObsolete` check complaining about types from libraries, i.e. right now it doesn't like SortedMap that dropwizard-metrics uses.

ferozco:
We should also clean up some of the work arounds in the gradle plugin itself. ex: https://github.com/palantir/gradle-baseline/blob/a927d383ffd95756deac9d3fcf3d20162c0571c2/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java#L227

robert3005:
@ferozco I have removed it already

ferozco:
👍

pkoenig10:
FYI these commits in error-prone 2.4.0 are going to cause a lot of failures for common nested imports.
https://github.com/google/error-prone/blame/7037696f4b00c8137394b10d143e37072b6048f6/core/src/main/java/com/google/errorprone/bugpatterns/BadImport.java#L63-L71

carterkozak:
Would you mind trying the automated fixes and reporting back? If they seem safe and sane we can apply them automatically.

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/CatchSpecificityTest.java
carterkozak: So it turns out this is a bit more subtle -- using a java9+ compiler to target java8 this works as expected, however the java 8 compiler cannot compile this output!
Not recommending that we revert since everything will be 11+ soon, and it's rare enough that I'd rather suppress.

pkoenig10:
The suggested fixes seemed to work pretty well, let's apply them automatically.
https://github.com/palantir/gradle-baseline/pull/1387

pkoenig10:
I also ran into a bunch of issues with `JdkObsolete` because it matches any use of methods on obsolete classes, even if you're just using an obsolete type returned from a library or passing an obsolete type to a library.

This was due to this change:
https://github.com/google/error-prone/commit/6671c25d6e7fb60d0f5e34c6f75fd1a0cd983a04#diff-065165cb9efc7cfdb26debe40d8ea8ccR230-R245

ferozco:
I would be in favour of disabling the rule. I think its pretty lame to get warnings for code you don't control

pkoenig10:
Looks like this is being tracked in https://github.com/google/error-prone/issues/1646

PR #1371, opened 2020-06-01
Upgrade gradle-errorprone

ferozco:
## Before this PR
We had blocked upgrades of gradle-errorprone because we were worried about forcing a "high" dependency on gradle (5.2). Since then gradle has released a bunch and the fleet has been keeping up to date so we no longer have to be worried about the dependency.

## After this PR
==COMMIT_MSG==
Upgrade gradle-errorprone
==COMMIT_MSG==

## Possible downsides?
N/A


robert3005:
errorProneArgs is now a property so you can't do https://github.com/palantir/gradle-baseline/blob/develop/gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineErrorProneIntegrationTest.groovy#L173

ferozco:
Ya I noticed :( its quite unfortunate since its pretty common for people to configure checks that way. I think we may want to hold off on merging this until we have a replacement. My current thoughts are to propose the following to the upstream plugin:
```
tasks.withType(JavaCompile) {
  options.errorprone.check 'OptionalOrElseMethodInvocation', 'OFF'
}
```

Eh, actually looking at it they already expose pretty nice methods configuring checks. ex: https://github.com/palantir/gradle-baseline/blob/d039b4b123daac27e8e3083eaeda9b68936eec99/gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineErrorProneIntegrationTest.groovy#L173

robert3005:
yeah, I think current version of the plugin has nice apis. Just the original version only really had the array based method and we stuck to it.

README.md
```diff
@@ -124,11 +124,7 @@ Rules can be suppressed at the project level, or have their severity modified, b
 
 ```gradle
 tasks.withType(JavaCompile).configureEach {
-    options.errorprone.errorproneArgs += ['-Xep:Slf4jLogsafeArgs:OFF']
-    // alternatively, using the DSL:
-    options.errorprone {
-        check('Slf4jLogsafeArgs', CheckSeverity.OFF)
-    }
+    options.errorprone.disable '-Slf4jLogsafeArgs'
```
robert3005: I think you don't want the dash at the beginning?

robert3005 (approved):

PR #1372, opened 2020-06-01
validate against ambiguous references

carterkozak:
## After this PR
==COMMIT_MSG==

==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/LambdaMethodReference.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/LambdaMethodReference.java
@@ -144,11 +144,11 @@ public final class LambdaMethodReference extends BugChecker implements BugChecke
         if (!paramSymbol.equals(receiverSymbol)) {
             return Description.NO_MATCH;
         }
-
-        return buildDescription(root)
-                .setMessage(MESSAGE)
-                .addFix(buildFix(methodSymbol, methodInvocation, root, state, isLocal(methodInvocation)))
-                .build();
+        return buildFix(methodSymbol, methodInvocation, root, state, isLocal(methodInvocation))
+                .filter(fix -> SuggestedFixes.compilesWithFix(fix, state, ImmutableList.of(), true))
+                .map(fix ->
+                        buildDescription(root).setMessage(MESSAGE).addFix(fix).build())
+                .orElse(Description.NO_MATCH);
     }
 
     private Description convertMethodInvocations(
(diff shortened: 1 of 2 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/LambdaMethodReference.java
```diff
@@ -184,12 +185,56 @@ private Description convertMethodInvocations(
...
+            Symbol.MethodSymbol symbol, @Nullable ExpressionTree receiver, VisitorState state) {
+        if (symbol.isStatic()) {
+            if (symbol.params().size() != 1) {
+                return false;
+            }
+            Symbol.ClassSymbol classSymbol = ASTHelpers.enclosingClass(symbol);
+            if (classSymbol == null) {
+                return false;
+            }
+            Set<Symbol.MethodSymbol> matching = ASTHelpers.findMatchingMethods(
```
ferozco, line 210: this is a super cool helper method! didn't know it existed

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/LambdaMethodReferenceTest.java
```diff
@@ -23,23 +23,21 @@
...
 import java.util.Optional;
 import java.util.function.Supplier;
-import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
 public class LambdaMethodReferenceTest {
 
-    private CompilationTestHelper compilationHelper;
-    private RefactoringValidator refactoringValidator;
+    private CompilationTestHelper compile() {
```
ferozco, line 30: gotta keep those tests fast 😛

ferozco (approved):

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/LambdaMethodReferenceTest.java
carterkozak: heh, most of our tests are structured this way so I moved it over. I think I migrated most of them after hitting a few cases where tests flaked after helper/validator instances were reused (due to either parallelism or multiple asserts in a single test). This way they're always clean regardless of other state!

carterkozak:
There's one more bug that I'm aware of where we trasform `foo()` to `this.foo()` regardless of whether the enclosing class (`this`) provides `foo()`. This breaks when `foo()` exists on an enclosing class, should be simple to check though.

PR #1373, opened 2020-06-01
Handle nested instance methods

ferozco:
## After this PR
==COMMIT_MSG==

==COMMIT_MSG==


carterkozak:
I pushed a commit to handle this case :-s

PR #1367, opened 2020-05-29
Upgrade asm to support JDK14

ferozco:
## Before this PR
Projects using JDK14 sourceCompat would have `checkUnusedDependencies` fail since the parser did not support the file major version.

## After this PR
==COMMIT_MSG==
`checkUnusedDependencies` works with JDK14
==COMMIT_MSG==
See example usage in our API related bot

## Possible downsides?
N/A

PR #1374, opened 2020-06-01
Add explicit asm dep

ferozco:
## Before this PR
We had tried to bump the version of asm we pulled in transitively in #1367, however this did not fix the issue for consumers

## After this PR
==COMMIT_MSG==
Add an explicit dependency on `org.ow2.asm:asm` to ensure consumers can use JDK14 source compat
==COMMIT_MSG==

## Possible downsides?
N/A

### DIFF
--- a/gradle-baseline-java/build.gradle
+++ b/gradle-baseline-java/build.gradle
@@ -15,6 +15,7 @@ dependencies {
     compile 'org.apache.maven.shared:maven-dependency-analyzer'
     compile 'org.github.ngbinh.scalastyle:gradle-scalastyle-plugin_2.11'
     implementation 'com.palantir.javaformat:palantir-java-format-spi'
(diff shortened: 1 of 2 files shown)


carterkozak (approved):

gradle-baseline-java/build.gradle
```diff
@@ -15,6 +15,7 @@ dependencies {
     compile 'org.apache.maven.shared:maven-dependency-analyzer'
     compile 'org.github.ngbinh.scalastyle:gradle-scalastyle-plugin_2.11'
     implementation 'com.palantir.javaformat:palantir-java-format-spi'
+    implementation 'org.ow2.asm:asm'
```
carterkozak, line 19: Might be worth a quick comment on this so we don't remove it.

PR #1375, opened 2020-06-02
auto-fix PublicConstructorForAbstractClass

iamdanfox:
Seems very non-scary https://github.com/google/error-prone/blob/master/core/src/main/java/com/google/errorprone/bugpatterns/PublicConstructorForAbstractClass.java

```
> Task :build2-ete:compileJava FAILED
/home/circleci/project/build2-ete/src/main/java/com/palantir/foundry/build2/ete/worker/EteWorkerBase.java:52: warning: [PublicConstructorForAbstractClass] Constructors of on an abstract class can be declared protected as there is never a need for them to be public
    public EteWorkerBase(WorkerReportingService workerReportingService) {
           ^
    (see https://errorprone.info/bugpattern/PublicConstructorForAbstractClass)
  Did you mean 'protected EteWorkerBase(WorkerReportingService workerReportingService) {'?
1 error
1 warning
```


dansanduleac:
👍

PR #1379, opened 2020-06-02
Improvement: Add error-prone check that forbids extending java.lang.Error

robert3005:
## Before this PR
We would allow extending Error. While there are niche cases where this is useful in vast majority of cases it can lead to more confusing and bugs than value.

## After this PR
==COMMIT_MSG==
Add error-prone check that forbids extending java.lang.Error (reports ERROR). Provides a fix to replace `extends Error` with `extends RuntimeException` for clearer contract to the consumers.
==COMMIT_MSG==

## Possible downsides?
The refactoring will get applied automatically but it might require human attention to merge. Cases where people would `} catch (Error e) {` are what we want to fix so those who do that might require to adjust their catch clauses after libraries pick up this check.


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ExtendsError.java
```diff
@@ -0,0 +1,64 @@
...
+        if (tree.getKind() != Tree.Kind.CLASS) {
+            // Don't apply to out interfaces and enums
+            return Description.NO_MATCH;
+        }
+
+        if (!ERROR_MATCHER.matches(tree, state)) {
+            return Description.NO_MATCH;
+        }
+
+        return buildDescription(tree).addFix(buildFix(tree, state)).build();
```
carterkozak: I'm not sure we want to suggest a replacement for classes that extend specific error implementations, just those which extend Error. We could fail without a suggested fix in the case of specific error types.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ExtendsError.java
```diff
@@ -0,0 +1,64 @@
...
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "ExtendsError",
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
+        providesFix = BugPattern.ProvidesFix.REQUIRES_HUMAN_ATTENTION,
+        severity = BugPattern.SeverityLevel.ERROR,
+        summary = "Class should not extend java.lang.Error. While allowed by java it can lead to surprising behaviour"
+                + " if users end up catching Error.")
+public final class ExtendsError extends BugChecker implements BugChecker.ClassTreeMatcher {
```
carterkozak: What if I extend Throwable?

robert3005: I think we should also forbid that, I would be surprised if java allowed that

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ExtendsErrorOrThrowable.java
```diff
@@ -0,0 +1,73 @@
...
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "ExtendsErrorOrThrowable",
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
+        providesFix = BugPattern.ProvidesFix.REQUIRES_HUMAN_ATTENTION,
+        severity = BugPattern.SeverityLevel.ERROR,
+        summary = "Class should not extend Error or Throwable directly. While allowed by java "
+                + "it can lead to surprising behaviour if users end up catching Error or Throwable.")
+public final class ExtendsErrorOrThrowable extends BugChecker implements BugChecker.ClassTreeMatcher {
```
ferozco, line 43: I think you have to add this to the list that gets applied by default in the errorprone plugin

ferozco:
👍

dansanduleac (approved):

PR #1380, opened 2020-06-02
Don't include SUGGESTION checks in default checks

pkoenig10:
When new `SUGGESTION` checks are added, it adds noise to the compiler output.

Even if these checks have suggested fixes, there is nothing that enforces `SUGGESTION` checks. So subsequent changes can introduce regressions that clutter the compiler output.

As a result I've found myself adding the config like this to all our repos to ensure that we enforce the default patch checks when new `SUGGESTION` checks are added.
```
options.errorprone.errorproneArgs += [
    '-Xep:AssertjRefactoring:WARN',
    '-Xep:CatchSpecificity:WARN',
    '-Xep:LambdaMethodReference:WARN',
    '-Xep:PreferAssertj:WARN',
    '-Xep:PreferCollectionConstructors:WARN',
    '-Xep:PreferBuiltInConcurrentKeySet:WARN',
    '-Xep:RedundantMethodReference:WARN',
    '-Xep:RedundantModifier:WARN',
    '-Xep:StreamOfEmpty:WARN',
    '-Xep:ThrowSpecificity:WARN']
```

Relying on excavator upgrades to fix these results in code churn and doesn't seem like the right approach. I'd would strongly prefer to either:
1. Enforce these checks at the warning level so projects can opt-in to failing at compile time for all new checks using `-Werror`.
2. Don't include `SUGGESTION` checks in default checks.

This PR takes approach (1), but I'm indifferent between these two options (or some mix of them depending on the check).

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/RedundantModifier.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/RedundantModifier.java
@@ -18,6 +18,7 @@ package com.palantir.baseline.errorprone;
 
 import com.google.auto.service.AutoService;
 import com.google.errorprone.BugPattern;
+import com.google.errorprone.BugPattern.SeverityLevel;
 import com.google.errorprone.VisitorState;
 import com.google.errorprone.bugpatterns.BugChecker;
 import com.google.errorprone.fixes.SuggestedFixes;
@@ -41,7 +42,7 @@ import javax.lang.model.element.Modifier;
         link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
         linkType = BugPattern.LinkType.CUSTOM,
         providesFix = BugPattern.ProvidesFix.REQUIRES_HUMAN_ATTENTION,
-        severity = BugPattern.SeverityLevel.SUGGESTION,
+        severity = SeverityLevel.WARNING,
         summary = "Avoid using redundant modifiers")
 public final class RedundantModifier extends BugChecker
         implements BugChecker.ClassTreeMatcher, BugChecker.MethodTreeMatcher, BugChecker.VariableTreeMatcher {
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StreamOfEmpty.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StreamOfEmpty.java
@@ -18,6 +18,7 @@ package com.palantir.baseline.errorprone;
 
 import com.google.auto.service.AutoService;
 import com.google.errorprone.BugPattern;
+import com.google.errorprone.BugPattern.SeverityLevel;
 import com.google.errorprone.VisitorState;
 import com.google.errorprone.bugpatterns.BugChecker;
 import com.google.errorprone.fixes.SuggestedFixes;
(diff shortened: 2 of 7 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/RedundantModifier.java
```diff
@@ -41,7 +42,7 @@
         link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
         linkType = BugPattern.LinkType.CUSTOM,
         providesFix = BugPattern.ProvidesFix.REQUIRES_HUMAN_ATTENTION,
-        severity = BugPattern.SeverityLevel.SUGGESTION,
+        severity = SeverityLevel.WARNING,
```
ferozco, line 45: seems like this should be an error

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StreamOfEmpty.java
```diff
@@ -33,7 +34,7 @@
         name = "StreamOfEmpty",
         link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
         linkType = BugPattern.LinkType.CUSTOM,
-        severity = BugPattern.SeverityLevel.SUGGESTION,
+        severity = SeverityLevel.WARNING,
```
ferozco: same here, definitely a bug

ferozco (commented):
Overall supportive of making these checks (especially since most autofix) warning/errors

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/RedundantModifier.java
carterkozak: I think we want WARN here, iirc we use a flag to ignore error-prone warnings in generated code but it doesn't apply to errors.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StreamOfEmpty.java
pkoenig10: I think the risk with making some of these errors is that repos which have a bunch of existing violation will be blocked. `WARNING` seems a bit safer because it's won't fail compilation unless you have `-Werror`.

Happy to change it if you would still prefer `ERROR`, just wanted to point this out.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/RedundantModifier.java
ferozco: I believe we explicitly opt out of errorprone on generated source sets: https://github.com/palantir/gradle-baseline/blob/a63a0adcf47044bdaae2f8a2c453159ad347ed02/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java#L210

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StreamOfEmpty.java
ferozco: in practice these rules will be auto applied on upgrade so I'm not particularly worried about blocking upgrades.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/RedundantModifier.java
pkoenig10: Can we change this in a FLUP? I'd prefer not to accidentally break things in this PR.

ferozco:
👍

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/RedundantModifier.java
carterkozak: +1 for handling warn->error separately. I don't think our regex catches 100% of generated code and I'm not sure it's worth the risk breaking some folks in order to fail builds in repos that don't fail on errors

PR #1386, opened 2020-06-03
Enable PreferCollectionConstructors by default

pkoenig10:
This excavator has been out for a while so making this a `WARNING` should not result in too much additional code churn.

Also update severity on some checks I missed in #1380.

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/CatchSpecificity.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/CatchSpecificity.java
@@ -57,7 +57,7 @@ import javax.lang.model.element.Name;
         name = "CatchSpecificity",
         link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
         linkType = BugPattern.LinkType.CUSTOM,
-        severity = BugPattern.SeverityLevel.SUGGESTION,
+        severity = BugPattern.SeverityLevel.WARNING,
         summary = "Prefer more specific error types than Exception and Throwable. When methods are updated to throw new"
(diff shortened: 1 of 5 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/CatchSpecificity.java
```diff
@@ -57,7 +57,7 @@
         name = "CatchSpecificity",
         link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
         linkType = BugPattern.LinkType.CUSTOM,
-        severity = BugPattern.SeverityLevel.SUGGESTION,
+        severity = BugPattern.SeverityLevel.WARNING,
```
ferozco, line 60: I recall there being a reason we had this one set only to suggeation. @carterkozak  do you remember why? or is my memory getting corrupted?

ferozco:
Adding do not merge, while waiting for a response to my question. Otherwise LGTM 👍 . Thanks @pkoenig10

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/CatchSpecificity.java
pkoenig10: Looks like we just wanted to validate that it did the right thing and then planned to increase the severity.
https://github.com/palantir/gradle-baseline/pull/941#discussion_r331304242

carterkozak: I think we wanted to roll out the refactors for a while before adding the requirement to avoid friction.

PR #1387, opened 2020-06-03
Auto-fix BadImport

pkoenig10:
I ran into this a bunch when upgrading, specifically because of these commits which flag a lot of commonly used inner imports:
https://github.com/google/error-prone/blame/7037696f4b00c8137394b10d143e37072b6048f6/core/src/main/java/com/google/errorprone/bugpatterns/BadImport.java#L63-L71


carterkozak:
I think they have a bug which qualifies enum values in some cases when the nested enum is named `Type`, but that's rare and the bad result can't compile so we're not any worse off.

👍

PR #1394, opened 2020-06-08
Added support for logsafe Arg arrays in Slf4jLogsafeArgs.

aioobe:
## Before this PR
`Slf4jLogsafeArgs` complains that an `Arg<?>[]` isn't a logsafe argument.

## After this PR
==COMMIT_MSG==
Slf4jLogsafeArgs now allows Arg<?>[] to be passed as vararg argument to logging methods.
==COMMIT_MSG==

## Possible downsides?
None.

## Context
Only allowing `Arg<?>[]` gets us half way. Compiler would still give the the following diagnostic warning:

```
... non-varargs call of varargs method with inexact argument type for last parameter;
        cast to java.lang.Object for a varargs call
        cast to java.lang.Object[] for a non-varargs call and to suppress this warning 
```

By following the above suggestion and casting to `Object[]` the expression is no longer of `Arg<?>[]` type, hence we add support also for `Arg<?>[]` cast into an `Object[]`.

For `// BUG: Diagnostic contains: varargs method with inexact argument` to work, we can't only pick up `Slf4jLogsafeArgs` diagnostics, so we add `.matchAllDiagnostics()`.

With `.matchAllDiagnostics()` we need to add `@SuppressWarnings("deprecation")` in order to avoid diagnostics about `hasChildren` being deprecated.


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/Slf4jLogsafeArgs.java
```diff
@@ -68,24 +73,27 @@ public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState
...
-        List<Integer> badArgs = badArgsBuilder.build();
-        if (badArgs.isEmpty() || TestCheckUtils.isTestCode(state)) {
+
+        boolean lastArgIsThrowable = THROWABLE.matches(lastArg, state);
+        int lastNonThrowableArgIndex = lastArgIsThrowable ? lastArgIndex - 1 : lastArgIndex;
+
+        List<Integer> badArgIndices = IntStream.rangeClosed(startArgIndex, lastNonThrowableArgIndex)
+                .filter(i -> !ARG.matches(allArgs.get(i), state))
+                .boxed()
+                .collect(Collectors.toList());
```
carterkozak: Any chance we can avoid the stream here? This code is _super_ hot because we log a lot of things, one large internal project in particular benefits from any compile speedup we can provide!

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/Slf4jLogsafeArgsTest.java
```diff
@@ -47,6 +47,7 @@ private void test(String logArgs, String failingArgs) throws Exception {
                         "    public String getName() { return null; }",
                         "    public void add(Marker reference) {}",
                         "    public boolean remove(Marker reference) { return true; }",
+                        "    @SuppressWarnings(\"deprecation\")",
```
carterkozak, line 50: nice

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/Slf4jLogsafeArgs.java
aioobe: Oh, I thought I spotted a nice simplification. Not my intention to slow it down. I'll revert to the previous for-each loop.

carterkozak:
Description is a little out of date now, and it needs a changelog. I'm happy to merge this once those are in :-)

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/Slf4jLogsafeArgs.java
aioobe: Reverted but had to factor it out to a method to pass the cyclomatic check.

carterkozak (approved):

carterkozak:
I think this needs to have develop merged into it to pick up the unit-test-14 check

ferozco:
👍

Issue #1395, opened 2020-06-08
Add errorprone check to avoid immutables Style inline + meta-annnotation


a10y:
## What happened?

We had a production issue where an immutables data object for configuration had a common style meta-annotation applied. A code-change was introduced that added a `@Value.Style()` annotation to expose public access to a Builder.

Because immutables [does no merging of style annotations](https://immutables.github.io/style.html), this overrode the meta-annotation, breaking config deserialization for dozens of production deployments.

## What did you want to happen?

The user should be told that this is going to produce unexpected behavior and that they should either
* Incorporate their style changes into the meta-annotation
* Inline the meta-annotations fields into the inline style declaration

a10y:
Fixed in #1396

PR #1396, opened 2020-06-08
Add errorprone check for immutables style collisions

a10y:
Fix for #1395

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ImmutablesStyleCollision.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ImmutablesStyleCollision.java
@@ -46,9 +46,7 @@ public final class ImmutablesStyleCollision extends BugChecker implements BugChe
     @Override
     public Description matchClass(ClassTree tree, VisitorState state) {
         if (MATCHER.matches(tree, state)) {
-            return buildDescription(tree)
-                    .setMessage("Immutable type cannot have both inline @Value.Style and meta-annotation applied")
-                    .build();
+            return describeMatch(tree);
         }
         return Description.NO_MATCH;
     }
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ImmutablesStyleCollision.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ImmutablesStyleCollision.java
@@ -1,0 +1,40 @@
+package com.palantir.baseline.errorprone;
+
+import com.google.auto.service.AutoService;
+import com.google.errorprone.BugPattern;
+import com.google.errorprone.VisitorState;
+import com.google.errorprone.bugpatterns.BugChecker;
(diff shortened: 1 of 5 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ImmutablesStyleCollision.java
```diff
@@ -0,0 +1,55 @@
...
+    private static final Matcher<ClassTree> STYLE_META_ANNOTATION =
+            Matchers.annotations(ChildMultiMatcher.MatchType.AT_LEAST_ONE, Matchers.hasAnnotation(Value.Style.class));
+    private static final Matcher<ClassTree> MATCHER = Matchers.allOf(INLINE_STYLE_ANNOTATION, STYLE_META_ANNOTATION);
+
+    @Override
+    public Description matchClass(ClassTree tree, VisitorState state) {
+        if (MATCHER.matches(tree, state)) {
+            return buildDescription(tree)
+                    .setMessage("Immutable type cannot have both inline @Value.Style and meta-annotation applied")
+                    .build();
```
carterkozak: I think we can use describeMatch instead which inherits the summary message from the class-level BugPattern annotation.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ImmutablesStyleCollision.java
```diff
@@ -0,0 +1,55 @@
...
+import com.google.errorprone.matchers.ChildMultiMatcher;
+import com.google.errorprone.matchers.Description;
+import com.google.errorprone.matchers.Matcher;
+import com.google.errorprone.matchers.Matchers;
+import com.sun.source.tree.ClassTree;
+import org.immutables.value.Value;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "ImmutablesStyleCollision",
```
carterkozak, line 32: Please document this alongside the other baseline error-prone checks in the readme!

ferozco: You'll also need to add it to the list of errorprone rules we always apply

carterkozak:
lgtm pending readme update and changelog, thanks!

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ImmutablesStyleCollision.java
a10y: I added it to the readme but where is the list @ferozco ?

carterkozak: I think that's just for automatic refactors, but this can't be auto-fixed. `severity = BugPattern.SeverityLevel.ERROR,` should be sufficient.

carterkozak (approved):

ferozco:
Thanks @a10y !

PR #1401, opened 2020-06-11
Added OptionalOrElseGetConstant error prone check.

aioobe:
## Before this PR
Lots of `Optional.orElseGet(() -> someConstant)` floating around in the world.

## After this PR
==COMMIT_MSG==
Recommends using optional.orElse(constant) instead of optional.orElseGet(() -> constant)
==COMMIT_MSG==

## Possible downsides?
None


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/OptionalOrElseGetConstant.java
```diff
@@ -0,0 +1,88 @@
...
+import com.sun.source.tree.Tree;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "OptionalOrElseGetConstant",
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
+        providesFix = BugPattern.ProvidesFix.REQUIRES_HUMAN_ATTENTION,
+        severity = SeverityLevel.ERROR,
+        summary = "If lambda passed to Optional#orElseGet returns a compile time constant, use Optional#orElse instead")
```
ferozco: I think we should expand this check to include any variable

carterkozak: +1 no need for it to be constant, field loads are less expensive than lambda allocations.

We would want to accept `foo`, `this.foo`, `foo.bar` but not `foo().bar`, `foo[index]`, `foo[index].bar`.

aioobe: Was a bit worried about race conditions in case it's a member variable. Thinking of it now again though, I'm realizing that it's probably not an issue. I'll look into it.

aioobe: Out of curiosity, why not `foo[index]` and `foo[index].bar`?

carterkozak: I'm not entirely sure they'll always be better, but they probably are. Even if there's llc miss it's probably better than the lambda.
I tend to prefer flagging and fixing the most unambiguous cases, but I don't have a strong opinion here.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/OptionalOrElseGetConstant.java
```diff
@@ -0,0 +1,88 @@
...
+import com.google.errorprone.matchers.Matcher;
+import com.google.errorprone.matchers.method.MethodMatchers;
+import com.sun.source.tree.ExpressionTree;
+import com.sun.source.tree.LambdaExpressionTree;
+import com.sun.source.tree.MethodInvocationTree;
+import com.sun.source.tree.Tree;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "OptionalOrElseGetConstant",
```
carterkozak: If the suggested changes make sense, perhaps we should rename this to `OptionalOrElseGetValue`

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/OptionalOrElseGetConstant.java
```diff
@@ -0,0 +1,88 @@
...
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
+        providesFix = BugPattern.ProvidesFix.REQUIRES_HUMAN_ATTENTION,
+        severity = SeverityLevel.ERROR,
```
carterkozak: Should probably be `WARN` rather than `ERROR`.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/OptionalOrElseGetConstant.java
carterkozak: This check should be documented with the others in the readme

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/OptionalOrElseGetConstant.java
```diff
@@ -0,0 +1,88 @@
...
+import com.sun.source.tree.LambdaExpressionTree;
+import com.sun.source.tree.MethodInvocationTree;
+import com.sun.source.tree.Tree;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "OptionalOrElseGetConstant",
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
+        providesFix = BugPattern.ProvidesFix.REQUIRES_HUMAN_ATTENTION,
```
carterkozak: After we test the suggested fixes on a large internal project (per-merge) we should add this check to the list that baseline automatically fixes by referencing the name from the list in `BaselineErrorProneExtension.java`

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/OptionalOrElseGetConstant.java
aioobe: Check now covers `foo`, `this.foo`, `foo.bar` but not array accesses or anything that could have side-effects.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/OptionalOrElseGetConstant.java
aioobe: Changed.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/OptionalOrElseGetConstant.java
aioobe: Added a line in the readme.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/OptionalOrElseGetConstant.java
aioobe: Sounds good. :+1:

README.md
```diff
@@ -159,6 +159,7 @@ Safe Logging can be found at [github.com/palantir/safe-logging](https://github.c
 - `NonComparableStreamSort`: Stream.sorted() should only be called on streams of Comparable types.
 - `DangerousStringInternUsage`: Disallow String.intern() invocations in favor of more predictable, scalable alternatives.
 - `OptionalOrElseThrowThrows`: Optional.orElseThrow argument must return an exception, not throw one.
+- `OptionalOrElseGetValue`: Prefer `Optional.orElse(expression)` over `Optional.orElseGet(() -> expression)` for trivial expressions.
```
carterkozak: pedantic nit:
```suggestion
- `OptionalOrElseGetValue`: Prefer `Optional.orElse(value)` over `Optional.orElseGet(() -> value)` for trivial expressions.
```

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/OptionalOrElseGetValue.java
```diff
@@ -0,0 +1,99 @@
...
+
+            LambdaExpressionTree lambdaExpressionTree = (LambdaExpressionTree) orElseGetArg;
+            LambdaExpressionTree.BodyKind bodyKind = lambdaExpressionTree.getBodyKind();
+
+            if (bodyKind != LambdaExpressionTree.BodyKind.EXPRESSION) {
+                break match;
+            }
+
+            ExpressionTree expressionBody = (ExpressionTree) lambdaExpressionTree.getBody();
+            if (COMPILE_TIME_CONSTANT.matches(expressionBody, state)
```
carterkozak: Is the compile time constant check ever necessary? What does it catch that the following checks cannot?

aioobe: It covers expressions that may look more complicated than the other cases, but can be evaluated at compile time. For example `PREFIX + "_someSuffix"` or `MIN + SIZE * 1.5`.

Added a test for such case.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/OptionalOrElseGetValue.java
```diff
@@ -0,0 +1,98 @@
...
+
+            if (orElseGetArg.getKind() != Tree.Kind.LAMBDA_EXPRESSION) {
+                break match;
+            }
+
+            LambdaExpressionTree lambdaExpressionTree = (LambdaExpressionTree) orElseGetArg;
+            LambdaExpressionTree.BodyKind bodyKind = lambdaExpressionTree.getBodyKind();
+
+            if (bodyKind != LambdaExpressionTree.BodyKind.EXPRESSION) {
+                break match;
```
carterkozak: Not sure the named block + break buys us anything over `return Description.NO_MATCH;`, by returning no-match we can shed a layer of indents, and style more uniformly with other checks.

aioobe: It was part of my battle against cyclomatic complexity. I'll check if I can drop it without tipping over.

carterkozak: I’d suppress the cyclomatic complexity check here

carterkozak:
do we need to merge develop into this branch to pick up the unit-test-14 checks?

carterkozak (approved):

PR #1407, opened 2020-06-11
fix #1406 Upgrade google-java-format to support java 14 lang features

### LINKED ISSUE #1406: Support instanceof pattern matching
## What happened?

When using code of the shape `foo instance String myString`, both checkstyle & spotless fail.

Checkstyle:
```
[ant:checkstyle] Foo.java:70:50: unexpected token: foo
```

Spotless:
```
...
Caused by: com.google.googlejavaformat.java.FormatterException: 71:45: error: ')' expected
	at com.google.googlejavaformat.java.FormatterException.fromJavacDiagnostics(FormatterException.java:50)
	at com.google.googlejavaformat.java.RemoveUnusedImports.parse(RemoveUnusedImports.java:253)
	at com.google.googlejavaformat.java.RemoveUnusedImports.removeUnusedImports(RemoveUnusedImports.java:211)
	at com.google.googlejavaformat.java.RemoveUnusedImports.removeUnusedImports(RemoveUnusedImports.java:204)
	... 206 more
```

Checkstyle and spotless accept valid Java 14 code.

carterkozak:
==COMMIT_MSG==
Upgrade google-java-format to support java 14 lang features
==COMMIT_MSG==


carterkozak:
I have not tested this locally.

robert3005:
why is there google java format dependency here? I thought we should have it in https://github.com/palantir/palantir-java-format/ ?

carterkozak:
Pretty sure it comes from the spotless plugin, not pjf.

robert3005:
This is just in the tests from my understanding
```
com.google.googlejavaformat:google-java-format:1.4
\--- com.google.errorprone:error_prone_test_helpers:2.4.0
````

robert3005:
Ah, so this is what we use to remove unused imports

robert3005:
Just heads up that per the release docs google-java-format 1.8 drops java < 11 `The minimum supported runtime version is now JDK 11.`

carterkozak:
Whelp, looks like we're blocked supporting 14 here until we get rid of 8. Womp 😞

ferozco:
We shouldn't be using googl-java-format for anything so it should be fine to keep it at the current version. We'll need to fix PJF then adopt the fix here

PR #1404, opened 2020-06-11
Improvement: Upgrade Checkstyle to 8.33

robert3005:
## Before this PR
We are on checkstyle 8.13

## After this PR
==COMMIT_MSG==
Upgrade checkstyle to 8.33
==COMMIT_MSG==

## Possible downsides?
This is second time we bump checkstyle, there are bound to be some issues but since we have it we should follow latest

PR #1409, opened 2020-06-12
Fix: Adjust checkstyle config for new version

robert3005:
## Before this PR
in #1404 we upgraded checkstyle but didn't take into account necessary config changes. Turns out we don't have tests that validate our checkstyle parses with default checkstyle version

## After this PR
==COMMIT_MSG==
Fix checkstyle config to conform to breaks made between 8.13 and 8.33
==COMMIT_MSG==

## Possible downsides?
N/A this should be purely a migration

### DIFF
--- a/gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
+++ b/gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
@@ -47,6 +47,10 @@
     <module name="BeforeExecutionExclusionFileFilter">
         <property name="fileNamePattern" value="module\-info\.java$"/>
     </module>
+    <module name="LineLength"> <!-- Java Style Guide: No line-wrapping -->
+        <property name="max" value="120"/>
+        <property name="ignorePattern" value="^package.*|^import.*|a href|href|http://|https://|ftp://"/>
+    </module>
     <module name="TreeWalker">
         <module name="SuppressionCommentFilter"/> <!-- baseline-gradle: README.md -->
         <module name="SuppressionCommentFilter">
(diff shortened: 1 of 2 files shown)


gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
```diff
@@ -374,11 +374,11 @@
         </module>
         <module name="RightCurly"> <!-- Java Style Guide: Nonempty blocks: K & R style -->
             <property name="option" value="same"/>
-            <property name="tokens" value="LITERAL_TRY, LITERAL_CATCH, LITERAL_FINALLY, LITERAL_IF, LITERAL_ELSE, LITERAL_DO"/>
+            <property name="tokens" value="CLASS_DEF, METHOD_DEF, CTOR_DEF, LITERAL_TRY, LITERAL_CATCH, LITERAL_FINALLY, LITERAL_IF, LITERAL_ELSE, LITERAL_DO"/>
```
robert3005, line 377: this is very weird - this didn't use to complain with checkstyle 8.13 on codebases I tried but with 8.33 this fails. I would be tempted to just ditch this since palantir-java-format is the autoritative source. I have made this setting conform to what palantir-java-format generates.

gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
```diff
@@ -431,11 +431,8 @@
         <module name="JavadocMethod"> <!-- Java Style Guide: Where Javadoc is used -->
             <property name="scope" value="public"/>
             <property name="allowMissingParamTags" value="true"/>
-            <property name="allowMissingThrowsTags" value="true"/>
```
robert3005, line 434: changes here are because javadoc check got split into two, this and `MissingJavadocMethod`. We actually only want this one that validates javadoc if it exists but doesn't enforce existence of javadoc

gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
```diff
@@ -47,6 +47,10 @@
     <module name="BeforeExecutionExclusionFileFilter">
         <property name="fileNamePattern" value="module\-info\.java$"/>
     </module>
+    <module name="LineLength"> <!-- Java Style Guide: No line-wrapping -->
```
robert3005, line 50: This was mentioned in the changes it's just hard to decipher things from their long list of changes

ferozco:
Tried it out locally seem to be working 👍

carterkozak (approved):

robert3005:
Will follow up with a test so this doesn't happen in the future

PR #1411, opened 2020-06-12
Improve gradle-baseline-java integration with IntelliJ import

aldexis:
## Before this PR
When directly importing the gradle project in IntelliJ, neither checkstyle nor copyright would be set up properly

## After this PR
==COMMIT_MSG==
Adds the proper configuration files upon IntelliJ import of a gradle project for checkstyle and copyright.

This generates the following additional files:
- .idea/copyright/profiles_settings.xml
- an xml file under .idea/copyright/ per copyright file under .baseline/copyright
- .idea/checkstyle-idea.xml (and adds Checkstyle-IDEA to the external dependencies) if baseline-checkstyle is applied
- Either .idea/codeStyleSettings.xml or a .idea/codeStyles/ folder with the contents being copied from .baseline/idea
  - If .baseline/idea/codeStyles is present, it will copy its contents, otherwise, it will fall back to .baseline/idea/intellij-java-palantir-style.xml as currently
  - The fallback is using a legacy IntelliJ format and requires closing and reopening the project to be taken into account

==COMMIT_MSG==

## Possible downsides?
- Small changes to the existing code structure for non-import code paths, which shouldn't change the output
- Not impossible this would override or conflict with other setups people have manually done


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
```diff
@@ -203,6 +294,29 @@ class BaselineIdea extends AbstractBaselinePlugin {
...
+        addCheckstyleNode(node)
+        addCheckstyleExternalDependencies(node)
+    }
+
+    private static void addCheckstyleIntellijImport(project) {
+        def checkstyle = project.plugins.findPlugin(BaselineCheckstyle)
+        if (checkstyle == null) {
+            project.logger.debug "Baseline: Skipping IDEA checkstyle configuration since baseline-checkstyle not applied"
+            return
+        }
```
dansanduleac: I'd flip this so it performs the action _when_ (and if) the BaselineCheckstyle plugin is applied.
In the current state, you're dependent on the order of application, so if BaselineCheckstyle is applied later, intellij won't be configured correctly.

```groovy
project.plugins.withType(BaselineCheckstyle) {
  // rest of the function
}
```

aldexis: I copied this from the existing checkstyle application. Would you recommend I flip it for both then?

dansanduleac: only for code paths that happen at configuration time

aldexis: I ended up flipping it for both `addCheckstyle` (called from within `ideaRootModel.project.ipr.withXml` so maybe that wasn't needed?) and `addCheckstyleIntellijImport` (mostly because I didn't see your edit before pushing, but I also think consistency is better - I'd even want to not have duplicate code but didn't find a nice way to do it because I'm not familiar enough with the code and how gradle configuration works yet)

dansanduleac (approved):

dansanduleac:
Played around with this locally while importing with intellij and fixed some issues where we were adding extra nodes (that intellij didn't know how to deal with) instead of modifying existing nodes when they exist.

Specifically we were always adding 
* a new <state> node to `.idea/codeStyles/codeStyleConfig.xml`'s `<component>`, and 
* a new `<code_scheme>` to `.idea/codeStyles/Project.xml`'s `<component>`

even if such node already existed. This is now fixed

aldexis:
Thanks for fixing @dansanduleac! Seems like we need another person's approval as well since you contributed?

fawind (approved):

PR #1413, opened 2020-06-12
Disable broken rule on switch expressions

ferozco:
## Before this PR
There was a false negative with `UnnecessaryParentheses` rule

## After this PR
==COMMIT_MSG==
Disable `UnnecessaryParentheses` by default
==COMMIT_MSG==

We still don't fully support Java 14 switch expressions because of checkstyle: https://github.com/checkstyle/checkstyle/issues/6615

## Possible downsides?
Short term we will lose some validation without any clear upside. We could just move forward to demonstrate the failure and hope it and checkstyle get fixed up stream

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
@@ -213,7 +213,7 @@ public final class BaselineErrorProne implements Plugin<Project> {
         // UnnecessaryParentheses does not currently work with switch expressions
         errorProneOptions.check("UnnecessaryParentheses", project.provider(() -> {
             JavaPluginExtension ext = project.getExtensions().getByType(JavaPluginExtension.class);
-            return ext.getSourceCompatibility().compareTo(JavaVersion.valueOf("14")) < 0
+            return ext.getSourceCompatibility().compareTo(JavaVersion.toVersion(14)) < 0
                     ? CheckSeverity.DEFAULT
                     : CheckSeverity.OFF;
         }));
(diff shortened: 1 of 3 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
```diff
@@ -210,18 +210,18 @@ private static void configureErrorProneOptions(
...
-        errorProneOptions.check("CatchSpecificity", CheckSeverity.OFF);
-        errorProneOptions.check("UnusedVariable", CheckSeverity.OFF);
-        errorProneOptions.check("EqualsHashCode", CheckSeverity.ERROR);
-        errorProneOptions.check("EqualsIncompatibleType", CheckSeverity.ERROR);
-        errorProneOptions.check("StreamResourceLeak", CheckSeverity.ERROR);
-        errorProneOptions.check("InputStreamSlowMultibyteRead", CheckSeverity.ERROR);
-        errorProneOptions.check("JavaDurationGetSecondsGetNano", CheckSeverity.ERROR);
-        errorProneOptions.check("URLEqualsHashCode", CheckSeverity.ERROR);
 
+        errorProneOptions.disable("UnnecessaryParentheses", "CatchSpecificity", "UnusedVariable");
```
carterkozak: I've found UnnecessaryParentheses useful in a few places, thoughts on disabling it only on jdk14 with experimental features enabled?

pkoenig10: Switch expressions aren't an experimental feature in Java 14.

ferozco:
@carterkozak i've updated the PR so we only disable the check when sourceCompat is 14

carterkozak (approved):

carterkozak:
Thanks!

PR #1412, opened 2020-06-12
StrictUnusedVariable handles Java 14 records

ferozco:
Partially addresses https://github.com/palantir/gradle-baseline/issues/1405
## Before this PR
We would incorrectly identify any record definition as a violation of `StrictUnusedVariable` due to some peculiarities with how the record definition was interpreted by errorprone.

## After this PR
==COMMIT_MSG==
StrictUnusedVariable handles Java 14 records
==COMMIT_MSG==

## Possible downsides?
N/A

PR #1414, opened 2020-06-12
Actually support records

ferozco:
## Before this PR
I accidentally pushed and merged a test case with a suppression that broke the test: https://github.com/palantir/gradle-baseline/pull/1412. 

## After this PR
==COMMIT_MSG==
Actually support records with `StrictUnusedVariable`
==COMMIT_MSG==

## Possible downsides?
N/A

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StrictUnusedVariable.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StrictUnusedVariable.java
@@ -67,7 +67,6 @@ import com.google.errorprone.bugpatterns.UnusedVariable;
 import com.google.errorprone.fixes.SuggestedFix;
 import com.google.errorprone.fixes.SuggestedFixes;
 import com.google.errorprone.matchers.Description;
-import com.google.errorprone.suppliers.Suppliers;
 import com.google.errorprone.util.ASTHelpers;
 import com.sun.source.tree.AnnotationTree;
 import com.sun.source.tree.ArrayAccessTree;
@@ -112,6 +111,7 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.regex.Pattern;
 import javax.annotation.Nullable;
 import javax.lang.model.element.ElementKind;
 import javax.lang.model.element.Modifier;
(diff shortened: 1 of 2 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StrictUnusedVariable.java
```diff
@@ -751,11 +748,12 @@ public Void visitClass(ClassTree tree, Void unused) {
             if (isSuppressed(tree)) {
                 return null;
             }
-            if (EXEMPTING_SUPER_TYPES.stream()
-                    .anyMatch(t ->
-                            isSubtype(getType(tree), Suppliers.typeFromString(t).get(state), state))) {
+
+            Pattern isRecord = Pattern.compile("record\\s+" + tree.getSimpleName() + "\\(");
+            if (isRecord.matcher(state.getSourceForNode(tree)).find()) {
```
carterkozak, line 753: This could become expensive, I wonder if we can get away with `state.getSourceForNode(tree).startsWith("record ")`?

ferozco: Unfortunately records can have a visibility modifier so that check doesn't quite work

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/StrictUnusedVariableTest.java
```diff
@@ -305,12 +305,7 @@ public void testRecord() {
                 .setArgs(ImmutableList.of("--enable-preview", "--release", "14"));
 
         compilationHelper
-                .addSourceLines(
-                        "Test.java",
-                        "class Test {",
-                        "@SuppressWarnings(\"StrictUnusedVariable\")",
```
iamdanfox, line 311: 😂

iamdanfox:
So are we saying that new error-prone doesn't actually use a modern java compiler that would allow us to use first-class methods to determine if an AST is actually a record?? It does seem very sad to have to regex things :(

ferozco:
Yes its pretty unfortunate to have to do a regex. We could leave things as is and say that we don't have 100% support for `record` but that you at least now have an escape hatch with `@SuppressWarnings`

ferozco:
I'm going to close this out and say that we only partially support records at the given moment, which seems reasonable given that they are still a preview feature

PR #1416, opened 2020-06-16
Don't wrap method declaration

hesselink:
## Before this PR
The default wrapping for long method declarations wrapped on the method modifiers, return type and name instead of the parameters, as most people on CRs seem to prefer.

## After this PR
==COMMIT_MSG==
Wrap method declarations on parameters, not on modifiers and return type.
==COMMIT_MSG==

## Possible downsides?
My eclipse saved in a higher format so that caused a bunch of changes. I read through these and tried to match to what was done before, but I could have missed something.

### DIFF
--- a/changelog/@unreleased/pr-1416.v2.yml
+++ b/changelog/@unreleased/pr-1416.v2.yml
@@ -1,6 +1,6 @@
 type: improvement
 improvement:
-  description: Wrap method declarations on parameters, not on modifiers and return
-    type.
+  description: When using Eclipse's built-in formatter, method declarations are now
+    wrapped on parameters, not on modifiers and return type.
   links:
   - https://github.com/palantir/gradle-baseline/pull/1416
--- a/changelog/@unreleased/pr-1416.v2.yml
+++ b/changelog/@unreleased/pr-1416.v2.yml
@@ -1,0 +1,6 @@
+type: improvement
+improvement:
+  description: Wrap method declarations on parameters, not on modifiers and return
+    type.
+  links:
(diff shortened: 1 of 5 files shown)


iamdanfox:
The thing that sucks here is that the eclipse and intellij formatters don't end up agreeing with each other, which is one of the big reasons we invested in [palantir-java-format](https://github.com/palantir/palantir-java-format).

I assume you work on the 'large internal project' which doesn't yet have pjf applied? In some ways maybe it's fine to just tinker with the eclipse.xml, just wanted to flag that it's always going to be suboptimal until that project gets onto pjf.

hesselink:
Yeah, this is for 'large internal project', where we use the Eclipse formatter plugin for IntelliJ. I tested this change on that project and it doesn't reformat any existing code (it does when the alignment is set to `0`).

iamdanfox (approved):
I haven't installed Eclipse to try this out locally, but I don't want to block it given that you are one of the very few eclipse users so if it's broken I trust you'll fix it :)

carterkozak:
Let's try running this on our large internal project before we merge and release, that way we don't get surprised by unexpected changes.
There are a lot of options that result in broken NLS markers.

hesselink:
@carterkozak As I mentioned above, I did run it on that project (`./gradlew spotlessApply), there were no changes (I actually tweaked the alignment from 0 to 12 to make that happen. It's quite finicky, at 16 it no longer works (and created the old formatting).

carterkozak:
:+1:

PR #1421, opened 2020-06-19
Fix configuration used for check junit dependencies

j-baker:
At present, the 'implementation' configuration is used in order to
figure out if junit 5 is on the classpath. This is a bit bogus because
the implementation configuration is not resolved to indicate whether
junit5 is present at runtime - it's more a statement around what is
available at compile time. This PR changes the configuration used to
be the testRuntimeClasspath configuration instead.

I hit this when I wanted to set up a project that exports some generic
test utilities to downstream projects (all internal, obviously); depends
on JUnit, depends on AssertJ, depends on Mockito, sets up
mock-maker-inline globally, has a FakeClock in, sets up Mockito's
default answers to return smart nulls instead of nulls for better error
messages - and found that this code was rejecting transitive
dependencies.


CRogers:
Why did you close this @j-baker? I was looking at this code yesterday afternoon and came to a similar conclusion.

j-baker:
I closed this because while my bug report seems valid, my solution just changes the error message!

j-baker:
I think the problem is somewhere actually in the bit that sets up the 'use junit platform' code - but that bit looks right at first glance.

PR #1425, opened 2020-06-21
do not safe log branch information

bavardage:
## Before this PR
It is possible to safe log branch info.

## After this PR
(per our discussion) add an errorprone check to try to prevent this.

==COMMIT_MSG==
Forbid safe logging of branch information.
==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreventBranchSafeLogging.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreventBranchSafeLogging.java
@@ -1,0 +1,63 @@
+/*
+ * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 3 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreventBranchSafeLogging.java
```diff
@@ -0,0 +1,63 @@
...
+        if (matcher.matches(tree, state)) {
+            if (tree.getArguments().isEmpty()) {
+                return Description.NO_MATCH;
+            }
+
+            ExpressionTree arg = tree.getArguments().get(0);
+
+            String parameterName = ASTHelpers.constValue(arg, String.class);
+
+            if (parameterName.startsWith("branch")) {
```
carterkozak, line 54: @ferozco Has another check that allows a configurable set of SafeArg names to be banned, though I don't recall where it's implemented.

carterkozak (changes requested):
I think this check is too specific and would prefer to use the general solution instead if we can roll that out instead. Thoughts?

bavardage:
I was unaware of the general check - happy to make the change there if someone can point me at it

bavardage:
closing in light of the other solution

carterkozak:
I think it lives in another repo but I can't find it with a quick search, I'll ask the team tomorrow.

bavardage:
yeah @ferozco shared the link - I sent an email

PR #1427, opened 2020-06-23
check junit dependencies task can be skipped

j-baker:

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckJUnitDependencies.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckJUnitDependencies.java
@@ -35,7 +35,6 @@ import org.gradle.api.artifacts.result.ResolvedComponentResult;
 import org.gradle.api.plugins.JavaPluginConvention;
 import org.gradle.api.provider.Provider;
 import org.gradle.api.tasks.Classpath;
-import org.gradle.api.tasks.Input;
 import org.gradle.api.tasks.SourceSet;
 import org.gradle.api.tasks.TaskAction;
 import org.gradle.api.tasks.testing.Test;
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckJUnitDependencies.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckJUnitDependencies.java
@@ -41,7 +41,7 @@ import org.gradle.api.tasks.TaskAction;
 import org.gradle.api.tasks.testing.Test;
 
 public class CheckJUnitDependencies extends DefaultTask {
-    private static final String IGNORE_MAIN_CONFIGURATION = "main"
+    private static final String IGNORE_MAIN_CONFIGURATION = "main";
 
     public CheckJUnitDependencies() {
         setGroup("Verification");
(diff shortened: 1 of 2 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckJUnitDependencies.java
```diff
@@ -63,6 +71,17 @@ public final void validateDependencies() {
                 });
     }
 
+    @Classpath
+    public final Provider<List<Configuration>> getConfigurations() {
```
robert3005, line 66: I think gradle will get angry at you for using Provider<List<>> instead of ListProperty. If you want to have dynamic set of inputs you have to do it via `getInputs()` similarly to how you handle outputs

carterkozak (dismissed):

gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckJUnitDependencies.java
```diff
@@ -47,7 +56,7 @@ public final void validateDependencies() {
                 .getPlugin(JavaPluginConvention.class)
                 .getSourceSets()
                 .forEach(ss -> {
-                    if (ss.getName().equals("main")) {
+                    if (ss.getName().equals(IGNORE_MAIN_CONFIGURATION)) {
```
CRogers: Not that it's every going to change, but there's https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/SourceSet.html#MAIN_SOURCE_SET_NAME

gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckJUnitDependencies.java
```diff
@@ -63,6 +72,17 @@ public final void validateDependencies() {
                 });
     }
 
+    @Classpath
+    public final Provider<List<Configuration>> getConfigurations() {
+        return getProject()
+                .provider(() ->
+                        getProject().getConvention().getPlugin(JavaPluginConvention.class).getSourceSets().stream()
+                                .filter(ss -> !ss.getName().equals(IGNORE_MAIN_CONFIGURATION))
```
CRogers: Can we extract this snippet up to `getSourceSets().stream().filter(...)` out to a method and reuse it in `validateDependencies` too?

j-baker: done

gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckJUnitDependencies.java
j-baker: done

gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckJUnitDependencies.java
j-baker: this is using the same pattern as check unused dependencies. Propeties are differnet to providers, right?

CRogers: I think this should be fine, right? There was a bug in Gradle 6.something that tried to cast `Provider`s to `CollectionPropertyInternal` at some point, forcing you to use `Property`s, but this was fixed?

j-baker:
I validated that this works locally

CRogers:
:+1:

PR #1432, opened 2020-06-25
fix #1431: Avoid rewriting lambdas into ambiguous references

### LINKED ISSUE #1431: LambdaMethodReference creates ambiguous
## What happened?

After upgrading gradle-baseline, the automated LambdaMethodReference refactoring triggered a compilation error on a project due to ambiguous method reference for a [`KeyedStream#flatMap` which has two overloads -- one taking a `Function<? super V, ? extends Stream<? extends R>> mapper` and the other taking a `BiFunction<? super K, ? super V, ? extends Stream<? extends R>> entryMapper`](https://github.com/palantir/streams/blob/master/src/main/java/com/palantir/common/streams/KeyedStream.java#L116-L124) 

Before refactoring:
```
KeyedStream.stream(elements)
    .flatMap(o -> getValues(o))
```

After refactoring, fails compilation:
```
KeyedStream.stream(elements)
    .flatMap(this::getValues)

// where there are two overloaded
private Stream<Value> getValues(Key key);
private Stream<Value> getValues(Key key, Object o2);
```

```
ERROR: reference to flatMap is ambiguous
                .flatMap(this::getValues)
                ^
  both method <R#1>flatMap(Function<? super V,? extends Stream<? extends R#1>>) in KeyedStream and method <R#2>flatMap(BiFunction<? super K,? super V,? extends Stream<? extends R#2>>) in KeyedStream match
  where R#1,V,K,R#2 are type-variables:
    R#1 extends Object declared in method <R#1>flatMap(Function<? super V,? extends Stream<? extends R#1>>)
    V extends Object declared in interface KeyedStream
    K extends Object declared in interface KeyedStream
    R#2 extends Object declared in method <R#2>flatMap(BiFunction<? super K,? super V,? extends Stream<? extends R#2>>)
```

Example `LambdaMethodReferenceTest` test case reproducing case.
```
    @Test
    void testNegative_ambiguousThisOverload() {
        refactor()
                .addInputLines(
                        "Test.java",
                        "import " + BiFunction.class.getName() + ';',
                        "import " + Function.class.getName() + ';',
                        "import " + Stream.class.getName() + ';',
                        "class Test {",
                        "  private Stream<String> test() {",
                        "      return Stream.concat(",
                        "              flatMap(s -> bar(s)), flatMap((s1, s2) -> bar(s1, s2)));",
                        "  }",
                        "  private static Stream<String> flatMap(Function<String, Stream<String>> fn) {",
                        "      return fn.apply(\"a\");",
                        "  }",
                        "  private static Stream<String> flatMap(BiFunction<String, String, Stream<String>> fn) {",
                        "      return fn.apply(\"a\", \"b\");",
                        "  }",
                        "  private Stream<String> bar(String in) {",
                        "      return Stream.of(in);",
                        "  }",
                        "  private Stream<String> bar(String s1, String s2) {",
                        "      return Stream.of(s1, s2);",
                        "  }",
                        "}")
                .expectUnchanged()
                .doTest();
    }
```

Skip the automated refactor as the lambda is simpler, or refactor to valid code that is not an ambiguous method reference such as the following with explicit cast (though stylistically in this case the lambda is shorter and clearer).

```
  .flatMap((Function<Key, Stream<? extends Value>>) this::getValues)
```

carterkozak:
## Before this PR
Produced code that did not work in some scenarios.

## After this PR
==COMMIT_MSG==
Avoid rewriting lambdas into ambiguous references
==COMMIT_MSG==

## Possible downsides?
Note that this uses the expensive `SuggestedFixes.compilesWithFix`
check which can result in poor performance, particularly on large
projects.
Fortunately this case isn't very common (no occurrences in a
very large internal project).

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/LambdaMethodReference.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/LambdaMethodReference.java
@@ -42,12 +42,11 @@ import com.sun.tools.javac.code.Flags;
 import com.sun.tools.javac.code.Symbol;
 import com.sun.tools.javac.code.Type;
 import com.sun.tools.javac.parser.Tokens;
-
-import javax.annotation.Nullable;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
+import javax.annotation.Nullable;
 
 @AutoService(BugChecker.class)
 @BugPattern(
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/LambdaMethodReference.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/LambdaMethodReference.java
@@ -42,11 +42,12 @@ import com.sun.tools.javac.code.Flags;
 import com.sun.tools.javac.code.Symbol;
 import com.sun.tools.javac.code.Type;
 import com.sun.tools.javac.parser.Tokens;
+
+import javax.annotation.Nullable;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
-import javax.annotation.Nullable;
 
 @AutoService(BugChecker.class)
 @BugPattern(
@@ -58,8 +59,6 @@ import javax.annotation.Nullable;
         summary = "Lambda should be a method reference")
 @SuppressWarnings("checkstyle:CyclomaticComplexity")
 public final class LambdaMethodReference extends BugChecker implements BugChecker.LambdaExpressionTreeMatcher {
-
-    private static final String MESSAGE = "Lambda should be a method reference";
 
     @Override
     public Description matchLambdaExpression(LambdaExpressionTree tree, VisitorState state) {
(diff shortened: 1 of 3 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/LambdaMethodReference.java
```diff
@@ -170,8 +168,8 @@ private Description convertMethodInvocations(
         }
 
         return buildFix(methodSymbol, methodInvocation, root, state, isLocal(methodInvocation))
-                .map(fix ->
-                        buildDescription(root).setMessage(MESSAGE).addFix(fix).build())
+                .filter(fix -> SuggestedFixes.compilesWithFix(fix, state))
```
carterkozak, line 170: This is a quick and dirty fix, we can write something more precise in the future if we find performance suffers.

schlosna: Ahh, I didn't see a good way to do the ambiguous check on the lambda/method reference receiver, but this would catch it. Should we put this filter in the `buildFix` method itself to catch any other potential invalid transforms?

carterkozak: I'm not aware of other paths that result in ambiguous behavior, I'd prefer to catch specific causes so we can add test cases for them, and hopefully write more precise fixes that don't involve recompilation.

schlosna (approved):

iamdanfox (approved):

iamdanfox:
// policy?

carterkozak:
policy bot has been sad today.

carterkozak:
:robot: :popcorn:

PR #1436, opened 2020-06-29
Checkstyle: PreferInclusiveLanguage regex

iamdanfox:
## Before this PR

As per @samrogerson's "Improving internal language" email from this morning, the industry is reflecting on the terminology we all use, how it affects members of our community and, in particular, how it can be made more inclusive.
- https://www.drupal.org/node/2275877
- https://twitter.com/natfriedman/status/1271253144442253312
- https://go-review.googlesource.com/c/go/+/236857/

Useful reference: https://developers.google.com/style/word-list

## After this PR
==COMMIT_MSG==
A new checkstyle regex 'PreferInclusiveLanguage' spots usages of the words 'whitelist', 'blacklist', 'slave' and suggests words that don't have any racial undertones. This check is intended to guide users when names are initially chosen (as this is often easier than renaming things later) - users can add `@SuppressWarnings("PreferInclusiveLanguage")` if it's not practical to fix these straight away.
==COMMIT_MSG==

This is obviously not a complete list - people will still need to be thoughtful about how their naming choices affect other people (e.g. ableist language).

## Possible downsides?
- this will likely block baseline upgrades from automerging in a few repos, as humans will need to make changes/suppress

### DIFF
--- a/gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
+++ b/gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
@@ -353,6 +353,11 @@
             <property name="ignoreComments" value="true"/>
         </module>
         <module name="RegexpSinglelineJava">
+            <property name="id" value="PreferInclusiveLanguage"/>
+            <property name="format" value="(whitelist|blacklist|slave)"/>
+            <property name="message" value="Prefer choosing names that don't have racial undertones, e.g. 'blocklist/allowlist' or 'server/client'"/>
+        </module>
+        <module name="RegexpSinglelineJava">
             <property name="format" value="\bCharsets\."/>
             <property name="message" value="Use JDK StandardCharsets instead of alternatives."/>
         </module>
--- a/gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
+++ b/gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
@@ -355,7 +355,7 @@
         <module name="RegexpSinglelineJava">
             <property name="id" value="PreferInclusiveLanguage"/>
             <property name="format" value="(whitelist|blacklist|slave)"/>
-            <property name="message" value="Prefer choosing names that don't have racial undertones, e.g. 'blocklist/allowlist' or 'server/client'"/>
+            <property name="message" value="Prefer choosing names that don't have racial undertones, e.g. 'allowlist/blocklist', 'server/client', 'leader/follower', 'primary/secondary', 'replica', 'allow/deny'"/>
         </module>
         <module name="RegexpSinglelineJava">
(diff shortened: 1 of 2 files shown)


gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
```diff
@@ -352,6 +352,11 @@
             <property name="message" value="Logging with System.err is not allowed because it has no metadata and can't be configured at runtime. Please use an SLF4J logger instead, e.g. log.info(&quot;Message&quot;)."/>
             <property name="ignoreComments" value="true"/>
         </module>
+        <module name="RegexpSinglelineJava">
+            <property name="id" value="PreferInclusiveLanguage"/>
+            <property name="format" value="(whitelist|blacklist|slave)"/>
+            <property name="message" value="Prefer choosing names that don't have racial undertones, e.g. 'blocklist/allowlist' or 'server/client'"/>
```
iamdanfox: we could provide more suggested alternatives here, e.g. leader/follower, replica, primary/secondary?

carterkozak: +1 for additional suggestions.

Some (large) projects exclusively use error-prone instead of checkstyle, I wonder if we could implement this using error-prone and automatically suggest changes inline?

iamdanfox:
Reading though a bunch of comments where other people have made similar changes, someone flagged an interesting case where `blocklist` has a possible ambiguity with 'list of blocks' as in filesystem blocks.

Also the term `whitelist` is often used to imply some kind of mandatory or exclusionary condition (i.e. programs will not run unless they're whitelisted). I have a slight concern that `allowlist` doesn't have this same kind of mandatory implication (i.e. these programs are on the allowlist, but could other things run too?). This might not be a big deal because if the whole industry switches terms then the meaning will probably just evolve and people will recognise the term `allowlist`.

gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
3yakuya: Should we also add "master" to the format?

iamdanfox: I think there are 7658 usages of master in java files internally. I think I'd like to rely on humans to prioritise this themselves, as blocking rollout of baseline is quite painful.

(For comparison, seems like there are 2833 of 'blacklist', 2375 of 'whitelist')

carterkozak:
Instead of `blocklist` and `allowlist` how about `disallowed` and `allowed`? I think this avoids the ambiguity with blocks, and the meaning of `allowed` is clear. If items aren't in `allowed` it shouldn't be terribly ambiguous.

PR #1442, opened 2020-06-30
Disable FallThrough on Java 14 for switch expressions

pkoenig10:
Related to https://github.com/palantir/gradle-baseline/pull/1413

## Before this PR
There is a false negative with `FallThrough` on Java 14 when using switch expressions where the case arms are statements (as opposed to expressions).

See https://github.com/google/error-prone/issues/1649.

## After this PR
Disable `FallThrough` on Java 14.

### DIFF
--- a/baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/FallThroughTest.java
+++ b/baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/FallThroughTest.java
@@ -1,0 +1,48 @@
+/*
+ * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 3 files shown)


baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/FallThroughTest.java
```diff
@@ -0,0 +1,48 @@
...
+                        "class Test {",
+                        "  static void foo(int value) {",
+                        "    switch (value) {",
+                        "      case 42 -> {}",
+                        "      // BUG: Diagnostic matches: X",
+                        "      default -> {}",
+                        "    };",
+                        "  }",
+                        "}")
+                .expectErrorMessage("X", input -> input.contains("Execution may fall through from the previous case"))
```
pkoenig10, line 45: I had to use `Diagnostic matches` and `expectErrorMessage` because otherwise the diagnostic message matches the magic comment that will suppress the check.
https://github.com/google/error-prone/blob/v2.4.0/core/src/main/java/com/google/errorprone/bugpatterns/FallThrough.java#L45-L46

iamdanfox (approved):

ferozco:
👍

PR #1445, opened 2020-07-03
Disallow usage of dangerous Set.copyOf, Map.copyOf methods

j-baker:
I tried to write a refaster rule, but bumped up against this being a Java1 10+ method, only.


j-baker:
It seems that i'm a complete idiot and somehow managed to misread the part where they are both actually hash based. Oops.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DangerousCollectionCopyOfUsage.java
```diff
@@ -0,0 +1,54 @@
...
+@BugPattern(
+        name = "DangerousCollectionCopyOfUsage",
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
+        severity = BugPattern.SeverityLevel.ERROR,
+        summary = "Disallow Java 10+ java.util.{Map,Set}.copyOf methods.")
+public final class DangerousCollectionCopyOfUsage extends BugChecker implements BugChecker.MethodInvocationTreeMatcher {
+    private static final String ERROR_MESSAGE = "Should not use java.util.{Map,Set}.copyOf methods, and should instead "
+            + "prefer Guava Immutable{Map,Set}.copyOf. Java's internal immutable collections are not hash based, they "
+            + "are linear lookup based. This means that instead of O(1) lookup time, they have O(n) lookup time. This "
```
iamdanfox, line 41: I'm reading through the `ImmutableCollections.MapN` class to try to try and understand this, and while the design does seem unusual, I'd be surprised if you get O(n) for lookups??

```
/**
     * An array-based Map implementation. There is a single array "table" that
     * contains keys and values interleaved: table[0] is kA, table[1] is vA,
     * table[2] is kB, table[3] is vB, etc. The table size must be even. It must
     * also be strictly larger than the size (the number of key-value pairs contained
     * in the map) so that at least one null key is always present.
```

They use linear probing and javadoc describes how it uses a flat array as internal storage, but I think the key thing is that they use a load factor of 2, so you shouldn't have to iterate over the whole thing before hitting a null

```
        // returns index at which the probe key is present; or if absent,
        // (-i - 1) where i is location where element should be inserted.
        // Callers are relying on this method to perform an implicit nullcheck
        // of pk.
        private int probe(Object pk) {
            int idx = Math.floorMod(pk.hashCode(), table.length >> 1) << 1;
            while (true) {
                @SuppressWarnings("unchecked")
                K ek = (K)table[idx];
                if (ek == null) {
                    return -idx - 1;
                } else if (pk.equals(ek)) {
                    return idx;
                } else if ((idx += 2) == table.length) {
                    idx = 0;
                }
            }
        }
```

The only method that did look gross to me was `containsValue`, cos that is _definitely_ O(n) whereas in guava that's definitely O(1) amortised

PR #1446, opened 2020-07-05
Implement PreferStaticLoggers error-prone check

carterkozak:
## Before this PR
No easy migration path for existing projects with instance loggers.

## After this PR
==COMMIT_MSG==
Implement PreferStaticLoggers error-prone check
==COMMIT_MSG==


iamdanfox (approved):
Approving since the sandman is OOTO

ferozco:
🤖

PR #1447, opened 2020-07-07
add TooManyArguments errorprone rule

ferozco:
## Before this PR
Users could define unwieldy interfaces that took a large number of parameters.

## After this PR
==COMMIT_MSG==
Add `TooManyArguments` errorprone rule which prevents interfaces from having more than 10 arguments
==COMMIT_MSG==

## Possible downsides?
There are definitely cases of people with larger interfaces so upgrades will be blocked and manual action will be required

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/TooManyArguments.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/TooManyArguments.java
@@ -1,0 +1,64 @@
+/*
+ * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.baseline.errorprone;
+
+import com.google.auto.service.AutoService;
+import com.google.errorprone.BugPattern;
+import com.google.errorprone.BugPattern.LinkType;
+import com.google.errorprone.BugPattern.ProvidesFix;
+import com.google.errorprone.BugPattern.SeverityLevel;
+import com.google.errorprone.VisitorState;
+import com.google.errorprone.bugpatterns.BugChecker;
+import com.google.errorprone.matchers.Description;
+import com.google.errorprone.matchers.Matcher;
+import com.google.errorprone.matchers.Matchers;
+import com.google.errorprone.util.ASTHelpers;
+import com.sun.source.tree.MethodTree;
+import com.sun.source.tree.Tree;
+import com.sun.tools.javac.code.Symbol;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "TooManyArguments",
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = LinkType.CUSTOM,
+        providesFix = ProvidesFix.REQUIRES_HUMAN_ATTENTION,
+        severity = SeverityLevel.WARNING,
+        summary = "Prefer Interface that take few arguments rather than many.")
+public final class TooManyArguments extends BugChecker implements BugChecker.MethodTreeMatcher {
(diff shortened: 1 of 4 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/TooManyArguments.java
```diff
@@ -0,0 +1,64 @@
...
+@BugPattern(
+        name = "TooManyArguments",
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = LinkType.CUSTOM,
+        providesFix = ProvidesFix.REQUIRES_HUMAN_ATTENTION,
+        severity = SeverityLevel.WARNING,
+        summary = "Prefer Interface that take few arguments rather than many.")
+public final class TooManyArguments extends BugChecker implements BugChecker.MethodTreeMatcher {
+    private static final int MAX_NUM_ARGS = 10;
+    public static final Matcher<Tree> IS_OVERRIDE = Matchers.hasAnnotation(Override.class);
```
carterkozak:
```suggestion
    private static final Matcher<Tree> IS_OVERRIDE = Matchers.hasAnnotation(Override.class);
```

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/TooManyArguments.java
```diff
@@ -0,0 +1,64 @@
...
+        }
+        Symbol.MethodSymbol symbol = ASTHelpers.getSymbol(tree);
+        if (symbol == null || symbol.isConstructor()) {
+            return Description.NO_MATCH;
+        }
+
+        if (tree.getParameters().size() > MAX_NUM_ARGS) {
+            return buildDescription(tree)
+                    .setMessage("Interfaces can take at most " + MAX_NUM_ARGS + " arguments.")
+                    .build();
```
carterkozak, line 62: No suggested fix to lop off a few unimportant params? ;-)

We might want to add something to the message suggesting an args object with a builder, potentially recommending immutables.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/TooManyArguments.java
```diff
@@ -0,0 +1,64 @@
...
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "TooManyArguments",
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = LinkType.CUSTOM,
+        providesFix = ProvidesFix.REQUIRES_HUMAN_ATTENTION,
+        severity = SeverityLevel.WARNING,
+        summary = "Prefer Interface that take few arguments rather than many.")
+public final class TooManyArguments extends BugChecker implements BugChecker.MethodTreeMatcher {
+    private static final int MAX_NUM_ARGS = 10;
```
robert3005, line 43: how about 5?

ferozco: we can ratchet it down over time

carterkozak: I'd prefer to start high and ratchet down

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/TooManyArguments.java
ferozco: I'd prefer not to have an autofix since it would leave behind code that didn't compile.

Happy to improve the message with ideas on how to solve the issue

carterkozak: > I'd prefer not to have an autofix since it would leave behind code that didn't compile.

Sorry, I should have included a `/s` tag! Agree we shouldn't break code.

ferozco:
@carterkozak I think this is ready to go

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/TooManyArguments.java
```diff
@@ -0,0 +1,67 @@
...
+import com.google.errorprone.matchers.Matcher;
+import com.google.errorprone.matchers.Matchers;
+import com.google.errorprone.util.ASTHelpers;
+import com.sun.source.tree.MethodTree;
+import com.sun.source.tree.Tree;
+import com.sun.tools.javac.code.Symbol;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "TooManyArguments",
```
carterkozak, line 36: Mind adding this to the readme?

carterkozak (approved):
lgtm aside from the nit

PR #1459, opened 2020-07-28
feature: add LogsafeArgName errorprone rule

ferozco:
## Before this PR
Occasionally developers would incorrectly tag a known unsafe parameter (ex. branch, username) as safe. 

## After this PR

==COMMIT_MSG==
add LogsafeArgName errorprone rule which allows users to specify a list of argument names that must always be tagged as unsafe.
==COMMIT_MSG==

This isn't a perfect solution since developers may use a different name for an unsafe value, but this should help catch some cases.

The expectation is that we will configure the set of unsafe arg names internally.

## Possible downsides?
N/A

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/LogsafeArgName.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/LogsafeArgName.java
@@ -25,6 +25,7 @@ import com.google.errorprone.VisitorState;
 import com.google.errorprone.bugpatterns.BugChecker;
 import com.google.errorprone.bugpatterns.BugChecker.MethodInvocationTreeMatcher;
 import com.google.errorprone.fixes.SuggestedFix;
+import com.google.errorprone.fixes.SuggestedFixes;
 import com.google.errorprone.matchers.CompileTimeConstantExpressionMatcher;
 import com.google.errorprone.matchers.Description;
 import com.google.errorprone.matchers.Matcher;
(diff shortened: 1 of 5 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/LogsafeArgName.java
```diff
@@ -0,0 +1,89 @@
...
+
+        List<? extends ExpressionTree> args = tree.getArguments();
+        ExpressionTree argNameExpression = args.get(0);
+        if (compileTimeConstExpressionMatcher.matches(argNameExpression, state)) {
+            String argName = (String) ((JCTree.JCLiteral) argNameExpression).getValue();
+            if (unsafeParamNames.stream().anyMatch(unsafeArgName -> unsafeArgName.equalsIgnoreCase(argName))) {
+                return buildDescription(tree)
+                        .setMessage("Arguments with name '" + argName + "' must be marked as unsafe.")
+                        .addFix(SuggestedFix.builder()
+                                .replace(tree.getMethodSelect(), "UnsafeArg.of")
```
carterkozak: SuggestedFixes.qualifyType on "com.palantir.logsafe.UnsafeArg" will add the import automagically, or fully qualify the type if there happens to be a `foo.bar.UnsafeArg` already imported. We can append `".of"` to the result.

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/LogsafeArgNameTest.java
```diff
@@ -0,0 +1,111 @@
...
+
+import com.google.common.collect.ImmutableList;
+import com.google.errorprone.CompilationTestHelper;
+import com.google.errorprone.ErrorProneFlags;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public final class LogsafeArgNameTest {
+    private CompilationTestHelper compilationHelper;
+    private RefactoringValidator refactoringHelper;
```
carterkozak: Slight preference to use factory methods for these, that way we can fluently build the test from the factory.
I recall some of the preconditions tests attempting multiple asserts on a shared helper and getting unexpected results because these guys are stateful!

ferozco:
I'd like to merge the associated internal PR first so that excavator will correctly apply the fix on baseline upgrades

carterkozak (approved):
lgtm, merge when you're ready.

PR #1465, opened 2020-08-03
Ignore non-literal arg names in LogsafeArgName

pkoenig10:
## Before this PR
The `LogsafeArgName` check would fail if using a compile time constant identifier (static final variables) for args names.

This fails with the following exception:
```
java.lang.AssertionError: An unhandled exception was thrown by the Error Prone static analysis plugin.
  
     error-prone version: 2.4.0
     BugPattern: LogsafeArgName
     Stack Trace:
     java.lang.ClassCastException: class com.sun.tools.javac.tree.JCTree$JCIdent cannot be cast to class com.sun.tools.javac.tree.JCTree$JCLiteral (com.sun.tools.javac.tree.JCTree$JCIdent and com.sun.tools.javac.tree.JCTree$JCLiteral are in module jdk.compiler of loader 'app')
  	at com.palantir.baseline.errorprone.LogsafeArgName.matchMethodInvocation(LogsafeArgName.java:77)
  	at com.google.errorprone.scanner.ErrorProneScanner.processMatchers(ErrorProneScanner.java:451)
```

## After this PR
The `LogsafeArgName` ignores arg names that are not literals.

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/LogsafeArgName.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/LogsafeArgName.java
@@ -73,15 +73,17 @@ public final class LogsafeArgName extends BugChecker implements MethodInvocation
         List<? extends ExpressionTree> args = tree.getArguments();
         ExpressionTree argNameExpression = args.get(0);
         if (compileTimeConstExpressionMatcher.matches(argNameExpression, state)) {
-            String argName = (String) ((JCTree.JCLiteral) argNameExpression).getValue();
-            if (unsafeParamNames.stream().anyMatch(unsafeArgName -> unsafeArgName.equalsIgnoreCase(argName))) {
-                SuggestedFix.Builder builder = SuggestedFix.builder();
-                String unsafeArg = SuggestedFixes.qualifyType(state, builder, "com.palantir.logsafe.UnsafeArg");
-                return buildDescription(tree)
-                        .setMessage("Arguments with name '" + argName + "' must be marked as unsafe.")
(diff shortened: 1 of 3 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/LogsafeArgName.java
```diff
@@ -73,15 +73,17 @@ public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState
...
-            String argName = (String) ((JCTree.JCLiteral) argNameExpression).getValue();
-            if (unsafeParamNames.stream().anyMatch(unsafeArgName -> unsafeArgName.equalsIgnoreCase(argName))) {
-                SuggestedFix.Builder builder = SuggestedFix.builder();
-                String unsafeArg = SuggestedFixes.qualifyType(state, builder, "com.palantir.logsafe.UnsafeArg");
-                return buildDescription(tree)
-                        .setMessage("Arguments with name '" + argName + "' must be marked as unsafe.")
-                        .addFix(builder.replace(tree.getMethodSelect(), unsafeArg + ".of")
-                                .build())
-                        .build();
+            if (argNameExpression instanceof JCTree.JCLiteral) {
```
carterkozak: nit: we might as well combine these two conditional expressions.

pkoenig10: We can just get rid of the compile time constant matcher, right? All literals are compile time constants.

carterkozak: Yep, you're right

carterkozak (approved):

PR #1487, opened 2020-09-02
Fix baseline-exact-dependencies with new GCV

dansanduleac:
## Before this PR

`com.palantir.baseline-exact-dependencies` tries to copy over constraints from `<source-set>CompileClasspath`, in order to pick up versions.lock constraints generated by GCV.

However, after https://github.com/palantir/gradle-consistent-versions/pull/557, these are no longer implemented as explicit constraints inherited by that configuration (which would have shown up in `compileClasspath.getAllDependencyConstraints()`), but instead GCV uses a _dependency_ to a configuration on the root project, which _itself_ holds these constraints in a single centralized place.

This means that the configuration resolved by tasks like `checkUnusedDependenciesMain` etc now lacks `versions.lock` constraints, leading to weird outcomes including flakes.

## After this PR
==COMMIT_MSG==
Fix `com.palantir.baseline-exact-dependencies` to work with GCV 1.26.0+.
==COMMIT_MSG==

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineExactDependencies.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineExactDependencies.java
@@ -157,7 +157,9 @@ public final class BaselineExactDependencies implements Plugin<Project> {
             // but we can rely on this since we control GCV.
             // Alternatively, we could tell GCV to lock this configuration, at the cost of a slightly more
             // expensive 'unifiedClasspath' resolution during lock computation.
-            explicitCompile.extendsFrom(project.getConfigurations().getByName("lockConstraints"));
+            if (project.getRootProject().getPluginManager().hasPlugin("com.palantir.versions-lock")) {
+                explicitCompile.extendsFrom(project.getConfigurations().getByName("lockConstraints"));
+            }
             // Inherit the excludes from compileClasspath too (that get aggregated from all its super-configurations).
             compileClasspath.getExcludeRules().forEach(rule -> explicitCompile.exclude(excludeRuleAsMap(rule)));
         });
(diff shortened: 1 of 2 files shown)


CRogers:
Is it possible to write a test for this? Does it work with GCV <1.26.0 too?

dansanduleac:
We're currently [testing it with old GCV](https://github.com/palantir/gradle-baseline/blob/db7ac4c5a55ac94078045530d33d5faa7825e9a3/gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineExactDependenciesTest.groovy#L31) - I suppose I could add a test for new GCV but the problem is the issue we've seen is a heisenbug... I couldn't reproduce it locally, for instance.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineExactDependencies.java
```diff
@@ -153,9 +153,13 @@ private static void configureSourceSet(
...
         explicitCompile.withDependencies(deps -> {
-            // Mirror the transitive constraints form compileClasspath in order to pick up GCV locks.
-            // We should really do this with an addAllLater but that would require Gradle 6, or a hacky workaround.
-            explicitCompile.getDependencyConstraints().addAll(compileClasspath.getAllDependencyConstraints());
+            // Pick up GCV locks. We're making an internal assumption that this configuration exists,
+            // but we can rely on this since we control GCV.
+            // Alternatively, we could tell GCV to lock this configuration, at the cost of a slightly more
+            // expensive 'unifiedClasspath' resolution during lock computation.
+            if (project.getRootProject().getPluginManager().hasPlugin("com.palantir.versions-lock")) {
+                explicitCompile.extendsFrom(project.getConfigurations().getByName("lockConstraints"));
```
CRogers, line 161: Feels quite weird to have this direct knowledge of GCV here? Will this still work with old GCV?

CRogers:
I am also struggling to reproduce this bug locally anywhere :(

CRogers:
:+1: Let's see if this fixes people's problems

PR #1492, opened 2020-09-14
BuilderPrivateConstructor error prone rule

carterkozak:
Easy to accidentally create builders with default public constructors.

==COMMIT_MSG==
BuilderPrivateConstructor error prone rule
==COMMIT_MSG==

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,6 +192,7 @@ Safe Logging can be found at [github.com/palantir/safe-logging](https://github.c
 - `TooManyArguments`: Prefer Interface that take few arguments rather than many.
 - `PreferStaticLoggers`: Prefer static loggers over instance loggers.
 - `LogsafeArgName`: Prevent certain named arguments as being logged as safe. Specify unsafe argument names using `LogsafeArgName:UnsafeArgNames` errorProne flag.
+- `BuilderPrivateConstructor`: Prevent builders with static factory methods from leaking public constructors.
 
 ### Programmatic Application
 
--- a/README.md
+++ b/README.md
@@ -192,7 +192,7 @@ Safe Logging can be found at [github.com/palantir/safe-logging](https://github.c
 - `TooManyArguments`: Prefer Interface that take few arguments rather than many.
 - `PreferStaticLoggers`: Prefer static loggers over instance loggers.
 - `LogsafeArgName`: Prevent certain named arguments as being logged as safe. Specify unsafe argument names using `LogsafeArgName:UnsafeArgNames` errorProne flag.
-- `BuilderPrivateConstructor`: Prevent builders with static factory methods from leaking public constructors.
+- `ImplicitPublicBuilderConstructor`: Prevent builders from unintentionally leaking public constructors.
 
 ### Programmatic Application
 
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/BuilderPrivateConstructor.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/BuilderPrivateConstructor.java
@@ -26,6 +26,8 @@ import com.google.errorprone.matchers.Description;
 import com.google.errorprone.util.ASTHelpers;
 import com.sun.source.tree.ClassTree;
 import com.sun.source.tree.MethodTree;
+import com.sun.source.tree.Tree;
+import com.sun.source.tree.Tree.Kind;
 import com.sun.tools.javac.code.Symbol.ClassSymbol;
 import com.sun.tools.javac.code.Symbol.MethodSymbol;
 import com.sun.tools.javac.tree.JCTree;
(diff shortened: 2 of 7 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/BuilderPrivateConstructor.java
```diff
@@ -0,0 +1,108 @@
...
+                        .replace(
+                                constructorPosition,
+                                constructorPosition,
+                                String.format("\nprivate %s() {}", tree.getSimpleName()))
+                        .build())
+                .build();
+    }
+
+    private static boolean isValidBuilderClass(ClassTree tree) {
+        return tree.getSimpleName().contentEquals("Builder")
```
iamdanfox: possible endsWith? I'm thinking we might want to catch some classes like FooBuilder or SimpleBuilder?

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/BuilderPrivateConstructor.java
```diff
@@ -0,0 +1,108 @@
...
+                && tree.getImplementsClause().isEmpty()
```
iamdanfox: requiring it to not implement any methods is probably going to mean we miss any staged builders right? (these would have one class implementing `FooStage`, `BarStage`, `BazStage` which each declare one method?)

carterkozak: That's correct, however staged builders generally use several interfaces with an internal hidden implementation.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/BuilderPrivateConstructor.java
carterkozak: :+1: We might as well catch those :-)

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/BuilderPrivateConstructor.java
```diff
@@ -0,0 +1,108 @@
...
+                && tree.getExtendsClause() == null
+                && tree.getModifiers().getFlags().contains(Modifier.STATIC);
+    }
+
+    private static boolean hasStaticBuilderFactory(ClassSymbol classSymbol, VisitorState state) {
+        Set<MethodSymbol> matching = ASTHelpers.findMatchingMethods(
+                state.getName("builder"),
+                methodSymbol -> methodSymbol != null
+                        && methodSymbol.getReturnType().tsym.getSimpleName().contentEquals("Builder")
```
iamdanfox: might be fine to skip this type return validation? people can always tell us if there are false positives (and suppress as aa workaround :)

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/BuilderPrivateConstructor.java
carterkozak:
```suggestion
        return tree.getSimpleName().toString().endsWith("Builder")
```

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/BuilderPrivateConstructor.java
```diff
@@ -0,0 +1,108 @@
...
+        // If no fields exist, the constructor is placed after the curly brace
+        int constructorPosition = ((JCTree) tree).getStartPosition()
+                + state.getSourceForNode(tree).indexOf('{')
+                + 1;
+
+        for (Tree member : tree.getMembers()) {
+            if (member.getKind() == Kind.VARIABLE) {
+                int endPosition = state.getEndPosition(member);
+                if (endPosition > constructorPosition) {
+                    constructorPosition = endPosition;
```
iamdanfox: nice ;)

README.md
```diff
@@ -192,6 +192,7 @@ Safe Logging can be found at [github.com/palantir/safe-logging](https://github.c
 - `TooManyArguments`: Prefer Interface that take few arguments rather than many.
 - `PreferStaticLoggers`: Prefer static loggers over instance loggers.
 - `LogsafeArgName`: Prevent certain named arguments as being logged as safe. Specify unsafe argument names using `LogsafeArgName:UnsafeArgNames` errorProne flag.
+- `BuilderPrivateConstructor`: Prevent builders with static factory methods from leaking public constructors.
```
iamdanfox: not 100% sure about the naming conventions we've stuck to, but somehow I feel like flipping this round and naming this after the 'bad' case feels more appropriate to me, so that people could write `@SuppressWarnings("badthing")` to acknowledge there's something dodgy and they're ignoring it?

Perhaps something like `ImplicitPublicBuilderConstructor` or `UnintentionalPublicBuilder` or something?

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/BuilderPrivateConstructor.java
carterkozak: Idea is that some (largely legacy) builders are meant to be used via their constructors, I don't want to create a bunch of false positives. This way we only flag builders that we expect aren't supposed to have exposed constructors.

README.md
carterkozak: :+1:

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/BuilderPrivateConstructor.java
carterkozak: odd that `Name` has a `startsWith`, but not `endsWith` ¯\_(ツ)_/¯

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/BuilderPrivateConstructor.java
carterkozak: I've updated this to take the builder ClassTree into account given the endsWith matcher.

iamdanfox (approved):

PR #1494, opened 2020-09-15
run assertj error-prone checks on every build

pnepywoda:


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
```diff
@@ -63,6 +63,7 @@
     private static final Logger log = Logging.getLogger(BaselineErrorProne.class);
     public static final String EXTENSION_NAME = "baselineErrorProne";
     private static final String ERROR_PRONE_JAVAC_VERSION = "9+181-r4173-1";
+    private static final String ASSERTJ_ERROR_PRONE_VERSION = "0.2.1";
```
pnepywoda, line 66: we should discuss how to not hard-code this version

iamdanfox:
I think the thinking here was that we already have a lot of static analysis provided by baseline, which can make it frictionful for people to just get a green build.  The assertj-automation things provide _mild_ style improvements, but aren't detecting 'dangerous' bug patterns in the same way that the baseline ones usually do.  As such, having excavator just come round and fix them up seemed like a decent middle ground.

Can you flesh out the motivation of why you want this to apply to everyone?

pnepywoda:
@pkoenig10 has applied it to several of our repos and would know better about the friction (or not) that it causes

pkoenig10:
I'm pretty indifferent about whether we apply this to everyone, I certainly understand the arguments why we wouldn't want to.

Manually adding this dependency to the place where we want the build-time validation is pretty frictionless, so it seems fine to not include it here.

iamdanfox:
You can see some of the example fixes it lands here: https://github.com/palantir/assertj-automation/blob/develop/assertj-error-prone/src/test/java/com/palantir/assertj/errorprone/AssertjSameSizeAsTest.java#L24-L56.

In the happy case, they're exactly the same.  Most of the improvement is when the test _fails_,  because you get something like "expected [a, b, c, potato] to have 3 elements" instead of "expected 4 to equal 3". This often jogs a devs memory if they've recently added the string 'potato' somewhere.

They're currently set a the level of [WARNING](https://github.com/palantir/assertj-automation/blob/9a427951e00de8be59e7229f5be264aaa69b84f6/assertj-error-prone/src/main/java/com/palantir/assertj/errorprone/AssertjRefactoring.java#L87) which makes them compile-blocking if you use -Werror.

If it's OK with you folks I think I'd like to keep this low-friction for BD-folks, and say if you want to opt-in to strictness you absolutely can.

pnepywoda:
+1 to low friction. we'll manually apply in our few repos

PR #1507, opened 2020-10-13
Fix ImmutablesBuilderMissingInitialization on Java 15 when immutables isn't on the classpath

jackwickham:
## Before this PR
The ImmutablesBuilderMissingInitialization check consistently fails with java 15 (but not <=14) if the immutables annotation jar isn't available on the classpath, such as in tests.

## After this PR
==COMMIT_MSG==
Fix ImmutablesBuilderMissingInitialization in java15
==COMMIT_MSG==

This fixes the issue because it now accesses the attributes directly from the AST rather than trying to reflectively instantiate the annotation to access the attributes. I'm not sure why this is java 15 specific...

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ImmutablesBuilderMissingInitialization.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ImmutablesBuilderMissingInitialization.java
@@ -311,7 +311,7 @@ public final class ImmutablesBuilderMissingInitialization extends BugChecker imp
                                     .tsym
                                     .getQualifiedName()
                                     .contentEquals("org.immutables.value.Generated"))
-                            .flatMap(annotation -> Stream.of(annotation.member(state.getName("generator"))))
+                            .map(annotation -> annotation.member(state.getName("generator")))
                             .filter(Objects::nonNull)
                             .anyMatch(attribute -> Objects.equals(attribute.getValue(), "Immutables"))
                     || extendsImmutablesGeneratedClass(((ClassSymbol) type.tsym).getSuperclass(), state);
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ImmutablesBuilderMissingInitialization.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ImmutablesBuilderMissingInitialization.java
@@ -46,7 +46,6 @@ import java.util.Set;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
-import org.immutables.value.Generated;
 
 /**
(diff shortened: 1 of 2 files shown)


fawind:
Curious why this only fails on java 15. Do you have a stacktrace you can share here?

jackwickham:
The stack trace that shows up in circle is
```
An exception has occurred in the compiler (15). Please file a bug against the Java compiler via the Java bug reporting page (http://bugreport.java.com) after checking the Bug Database (http://bugs.java.com) for duplicates. Include your program, the following diagnostic, and the parameters passed to the Java compiler in your report. Thank you.
java.lang.NoSuchMethodError: 'java.lang.Object com.sun.tools.javac.code.Symbol$CompletionFailure.getDetailValue()'
	at com.google.errorprone.ErrorProneAnalyzer.finished(ErrorProneAnalyzer.java:171)
	at jdk.compiler/com.sun.tools.javac.api.MultiTaskListener.finished(MultiTaskListener.java:132)
	at jdk.compiler/com.sun.tools.javac.main.JavaCompiler.flow(JavaCompiler.java:1421)
	at jdk.compiler/com.sun.tools.javac.main.JavaCompiler.flow(JavaCompiler.java:1368)
	at jdk.compiler/com.sun.tools.javac.main.JavaCompiler.compile(JavaCompiler.java:960)
	at jdk.compiler/com.sun.tools.javac.api.JavacTaskImpl.lambda$doCall$0(JavacTaskImpl.java:104)
	at jdk.compiler/com.sun.tools.javac.api.JavacTaskImpl.handleExceptions(JavacTaskImpl.java:147)
	at jdk.compiler/com.sun.tools.javac.api.JavacTaskImpl.doCall(JavacTaskImpl.java:100)
	at jdk.compiler/com.sun.tools.javac.api.JavacTaskImpl.call(JavacTaskImpl.java:94)
	at org.gradle.internal.compiler.java.IncrementalCompileTask.call(IncrementalCompileTask.java:74)
	at org.gradle.api.internal.tasks.compile.AnnotationProcessingCompileTask.call(AnnotationProcessingCompileTask.java:94)
	at org.gradle.api.internal.tasks.compile.ResourceCleaningCompilationTask.call(ResourceCleaningCompilationTask.java:57)
```
but that's actually an exception while handling the other exception (I've not figured that one out but it happens in all JDKs under the right conditions and is either a JDK or Errorprone issue).

The actual root cause exception is a `Symbol.CompletionException` with message `class file for org.immutables.value.Generated not found` - I don't have a stack trace for it, but it's coming from `com.sun.tools.javac.code.Symbol` when trying to instantiate the annotation, in theory because that annotation isn't available at runtime.

fawind (approved):
Thanks for providing the context. I'm also not sure why this is j15 specific but happy to get the fix out!

PR #1510, opened 2020-10-14
error-prone UnnecessarilyQualified

carterkozak:
## Before this PR
Mixed qualified and unqualified uses of the same type in the same java file.

## After this PR
Note that we don't disallow fully qualified classes, they can be useful when two types with the same simple name are used to avoid ambiguity.

==COMMIT_MSG==
error-prone UnnecessarilyQualified
==COMMIT_MSG==

## Possible downsides?
Might cause code churn, however it makes the code substantially easier to follow, and fixes itself.


carterkozak:
Validated on a large internal project.

ferozco (approved):
Nice!

ferozco:
🤖

PR #1513, opened 2020-10-16
disable PlaceholderCountMatchesArgumentCount

dansanduleac:
## Before this PR

Intellij keeps warning about missing placeholders `{}` in our slf4j logs.
However, in the overwhelming majority of places in our code, we use sls logs which don't require placeholders.

## After this PR

==COMMIT_MSG==
Disable intellij's missing slf4j placeholders inspection by default.
==COMMIT_MSG==

## Possible downsides?

cc @fsamuel-bs


fawind (approved):
Sounds reasonable, I certainly find these warnings quite irritating!

PR #1521, opened 2020-10-22
Implement DeprecatedGuavaObjects error-prone migration

carterkozak:
==COMMIT_MSG==
Implement DeprecatedGuavaObjects error-prone migration
==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DeprecatedGuavaObjects.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DeprecatedGuavaObjects.java
@@ -1,0 +1,70 @@
+/*
+ * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.baseline.errorprone;
+
+import com.google.auto.service.AutoService;
+import com.google.errorprone.BugPattern;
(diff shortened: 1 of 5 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DeprecatedGuavaObjects.java
```diff
@@ -0,0 +1,70 @@
...
+    private static final String GUAVA_OBJECTS_NAME = "com.google.common.base.Objects";
+    private static final String GUAVA_OBJECTS_EQUAL = GUAVA_OBJECTS_NAME + ".equal";
+    private static final String GUAVA_OBJECTS_HASH_CODE = GUAVA_OBJECTS_NAME + ".hashCode";
+    private static final String JAVA_OBJECTS_NAME = "java.util.Objects";
+    private static final Matcher<ExpressionTree> EQUAL_MATCHER = MethodMatchers.staticMethod()
+            .onClass("com.google.common.base.Objects")
+            .named("equal");
+    private static final Matcher<ExpressionTree> HASH_CODE_MATCHER = MethodMatchers.staticMethod()
+            .onClass("com.google.common.base.Objects")
+            .named("hashCode");
```
schlosna, line 51: should we treat single arg version differently to map to `java.util.Objects.hashCode(Object)` and varargs maps to `java.util.Objects.hash(Objects...)`?

carterkozak: I was planning to keep this simple as a direct migration and write a separate check for `java.util.Objects.hash(value)` -> `java.util.Objects.hashCode(value)` to eventually resolve to the ideal implementation.

That would result in the first refactor run failing in this case, so it might be fine to share the complexity with this check as well. What do you think?

schlosna: 👍 your proposal makes sense to me and is additive beyond just the Guava -> java.util conversion

schlosna (approved):

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DeprecatedGuavaObjects.java
carterkozak: PR is #1522

ferozco (approved):

PR #1522, opened 2020-10-22
Implement error-prone ObjectsHashCodeUnnecessaryVarargs migration

carterkozak:
==COMMIT_MSG==
Implement error-prone ObjectsHashCodeUnnecessaryVarargs migration
==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/ObjectsHashCodeUnnecessaryVarargsTest.java
+++ b/baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/ObjectsHashCodeUnnecessaryVarargsTest.java
@@ -37,7 +37,7 @@ class ObjectsHashCodeUnnecessaryVarargsTest {
                         "Test.java",
                         "import java.util.Objects;",
                         "public class Test {",
-                        "  void f(Object a, Object b) {",
+                        "  void f(Object a, Object[] b, String[] c) {",
                         "    Objects.hashCode(a);",
                         "    Objects.hash(a, a);",
                         "    Objects.hash(b);",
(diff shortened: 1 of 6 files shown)


baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/ObjectsHashCodeUnnecessaryVarargsTest.java
```diff
@@ -0,0 +1,53 @@
...
+    public void test() {
+        fix().addInputLines(
+                        "Test.java",
+                        "import java.util.Objects;",
+                        "public class Test {",
+                        "  void f(Object a, Object[] b, String[] c) {",
+                        "    Objects.hash(a);",
+                        "    Objects.hash(a, a);",
+                        "    Objects.hash(b);",
+                        "    Objects.hash(c);",
```
schlosna, line 34: these are both a bit sketchy cases where compiler should have issued lint message `warning: non-varargs call of varargs method with inexact argument type for last parameter;`

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/ObjectsHashCodeUnnecessaryVarargsTest.java
```diff
@@ -0,0 +1,53 @@
...
+                        "}")
+                .addOutputLines(
+                        "Test.java",
+                        "import java.util.Objects;",
+                        "public class Test {",
+                        "  void f(Object a, Object[] b, String[] c) {",
+                        "    Objects.hashCode(a);",
+                        "    Objects.hash(a, a);",
+                        "    Objects.hash(b);",
+                        "    Objects.hashCode(c);",
```
schlosna: Should `Object[]` be treated differently than other object type array such as `String[]` in this case? I'm a little worried this might be papering over some other potential issues where something should be using `Arrays.hashCode` directly on an array.

carterkozak: Good call, I've updated the check to avoid arrays entirely.

iamdanfox (approved):
V. neat :)

PR #1524, opened 2020-10-29
Always set JVM 8 target in Scala compiler

rshkv:
## Before this PR
Projects with gradle-baseline cannot compile Scala code if they set Java targetCompatibility to 11. The error is:
```
> Task :optimizer-rules:compileScala
'jvm-11' is not a valid choice for '-target'
bad option: '-target:jvm-11'
```

BaselineScalaStyle sets scalac `-target` to match that of javac. However scalac 2.12 only accepts target values up-to `jvm-8` [(see here)](https://github.com/scala/scala/blob/v2.12.12/project/ScalaOptionParser.scala#L109). Scala version 2.13 supports higher targets, but 2.12 was released July this year and we still need to support it.

## After this PR
**Edit:** ~No longer set target version for Scala compiler to match that of Java compiler.~

==COMMIT_MSG==
Always set target version for Scala compiler to JVM 8.
==COMMIT_MSG==

## Possible downsides?
If using scalac 2.13 (which permits `jvm-9+`), this change can result in a lower target version. I don't think that's dangerous but might be overlooking something.

`jvm-7` and lower are [disallowed by 2.12](https://docs.scala-lang.org/overviews/jdk-compatibility/overview.html#version-compatibility-table) - they wouldn't have been valid targets before. The last 2.11 version was released in 2016.

### DIFF
--- a/changelog/@unreleased/pr-1524.yml
+++ b/changelog/@unreleased/pr-1524.yml
@@ -1,0 +1,5 @@
+type: fix
+fix:
+  description: No longer set target version for Scala compiler to match that of Java compiler.
+  links:
+  - https://github.com/palantir/gradle-baseline/pull/1524
--- a/changelog/@unreleased/pr-1524.yml
+++ b/changelog/@unreleased/pr-1524.yml
@@ -1,5 +1,5 @@
 type: fix
 fix:
-  description: No longer set target version for Scala compiler to match that of Java compiler.
+  description: Always set target version for Scala compiler to JVM 8.
   links:
   - https://github.com/palantir/gradle-baseline/pull/1524
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineScalastyle.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineScalastyle.java
@@ -16,7 +16,6 @@
 
 package com.palantir.baseline.plugins;
 
-import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import groovy.util.Node;
 import groovy.xml.QName;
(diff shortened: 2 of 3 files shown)


robert3005:
Scala 2.11 still requires it so either you have to drop it or just set it to 1.8

rshkv:
True, it defaults to Java 6 [in 2.11.12](https://github.com/scala/scala/blob/b8155a5502d6608e491a241a94c82cdec1f4a161/src/compiler/scala/tools/nsc/settings/StandardScalaSettings.scala#L41-L42).

rshkv:
I went with _always_ setting the target to 8 as I don't find a straight-forward way to get the Scala version and offer a target based on that.

Personally, I'd still prefer to not have this here. I don't think we need to support 2.11. And if this happens to fail anyone, they can configure the right target themselves.

robert3005:
I think setting it to target 1.8 is correct for now and we can remove it once we are sure we don't need scala 2.11 👍

rshkv:
@CRogers, the `trial-publish` job is failing because `javadoc` is OOMing. I see the same on https://github.com/palantir/gradle-baseline/pull/1517.

rshkv:
Also would appreciate a release. This is blocking source compatibility bumps 🙏

iamdanfox:
I think the changelog is malformed. changelog/@unreleased/pr-1524.yml should be changelog/@unreleased/pr-1524.v2.yml

rshkv:
Contributors can't edit the changelog-app message.

rshkv:
Will file PR.

iamdanfox:
@rshkv you shouldn't have to fork if you're part of the palantir org. then you can edit the msg.

rshkv:
I'm part of the org but can't push to this repo.

robert3005:
https://github.com/palantir/gradle-baseline/pull/1525

PR #1528, opened 2020-10-29
Use old version-string scheme for Scala compiler target

rshkv:
## Before this PR
In #1524, we provide `-target:jvm-8` as parameter to scalac. It turns out the new version-string scheme (`8` instead of `1.8`) is [only supported from Scala 2.13](https://github.com/scala/scala/pull/8060). We need to support 2.12 and below. Apologies.

## After this PR
Set `-target:jvm-1.8` instead.

==COMMIT_MSG==
Use old version-string scheme for Scala compiler target (`jvm-1.8` instead of `jvm-8`).
==COMMIT_MSG==


robert3005:
👍

PR #1531, opened 2020-11-03
Add error-prone check for system time zone usage

pkoenig10:
Error Prone has a built-in [JavaTimeDefaultTimeZone](https://errorprone.info/bugpattern/JavaTimeDefaultTimeZone) check. However this check only matches _implicit_ uses of the system default time zone.

There's no good reason to use the system default time zone so we should also prevent _explicit_ uses. This PR add a `JavaTimeSystemDefaultTimeZone` error-prone check that matches these explicit uses and suggests a replacement.

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/JavaTimeSystemDefaultTimeZone.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/JavaTimeSystemDefaultTimeZone.java
@@ -1,0 +1,71 @@
+/*
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/JavaTimeSystemDefaultTimeZone.java
```diff
@@ -0,0 +1,71 @@
...
+import com.google.errorprone.matchers.Matchers;
+import com.sun.source.tree.ExpressionTree;
+import com.sun.source.tree.MethodInvocationTree;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "JavaTimeSystemDefaultTimeZone",
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
+        severity = BugPattern.SeverityLevel.WARNING,
```
pkoenig10: Should this be `WARNING` or `ERROR`?

robert3005: As a person who spent too much time explaining time and time zones to people I suggest making this an error

ferozco (approved):
thanks @pkoenig10

fawind (approved):
:+1:

PR #1532, opened 2020-11-04
Automatically fix JavaTimeSystemDefaultTimeZone

pkoenig10:
Follow up to https://github.com/palantir/gradle-baseline/pull/1531

Automatically fixes `JavaTimeDefaultTimeZone` and `JavaTimeSystemDefaultTimeZone`.

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineErrorProneExtension.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineErrorProneExtension.java
@@ -38,6 +38,8 @@ public class BaselineErrorProneExtension {
             "ExtendsErrorOrThrowable",
             "FinalClass",
             "ImplicitPublicBuilderConstructor",
+            "JavaTimeDefaultTimeZone",
+            "JavaTimeSystemDefaultTimeZone",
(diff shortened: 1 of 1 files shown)


ferozco:
👍

gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineErrorProneExtension.java
```diff
@@ -38,6 +38,8 @@
             "ExtendsErrorOrThrowable",
             "FinalClass",
             "ImplicitPublicBuilderConstructor",
+            "JavaTimeDefaultTimeZone",
+            "JavaTimeSystemDefaultTimeZone",
```
carterkozak, line 42: Nit: can we move this one from ERROR->WARNING since there are cases (non-server) where the users timezone is preferred? If we're auto-fixing it, there's not a great deal of difference between the two.

pkoenig10: I don't have any preference here. I originally had this as WARNING but changed it in response to https://github.com/palantir/gradle-baseline/pull/1531#discussion_r517019478.

PR #1533, opened 2020-11-04
Allow ErrorProne to qualify JavaTimeSystemDefaultTimeZone fix types

carterkozak:
SuggestedFixes.qualifyType will import if there's not already an
import with the same simple name, otherwise use a fully qualified
reference to avoid breaking compilation.

==COMMIT_MSG==
Allow ErrorProne to qualify JavaTimeSystemDefaultTimeZone fix types
==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/JavaTimeSystemDefaultTimeZone.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/JavaTimeSystemDefaultTimeZone.java
@@ -21,11 +21,13 @@ import com.google.errorprone.BugPattern;
 import com.google.errorprone.VisitorState;
 import com.google.errorprone.bugpatterns.BugChecker;
 import com.google.errorprone.fixes.SuggestedFix;
+import com.google.errorprone.fixes.SuggestedFixes;
 import com.google.errorprone.matchers.Description;
 import com.google.errorprone.matchers.Matcher;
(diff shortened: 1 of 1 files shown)


pkoenig10:
Ah thanks, I wasn't sure how to get the suggested fix I wanted here.

pkoenig10 (approved):

mglazer:
Isn't this change potentially extremely dangerous if you don't actually vet every single place where this is called?

For example, this can potentially cause all persisted dates in a system to start being persisted 8 hours into the future all of a sudden with just a simple excavator change.

carterkozak:
Qualifying the names helps us compile, the suggested change itself could change behavior and I don't agree with the choice to auto-apply it, but I don't have a strong opinion. It won't change persisted dates unless the values are used incorrectly though.

pkoenig10:
Providing some context, this change was motivated because we had an issue where one service was serializing an `OffsetDateTime` with a non-UTC system default time zone and another service attempted to deserialize it as an `Instant`.

This definitely should not cause any behavior change, especially since you can't really guarantee what the system default time zone is. The system default time zone could be UTC, in which case this logic should be exactly equivalent.

mglazer:
Sorry @carterkozak I should have commented on the actual PR where this was changed. I realize you just made a bug fix to that one.

I don't understand how this is not a behavioral change.

If my system timezone is UTC-8, and we change this to UTC, how is that not a functional change?

PR #1537, opened 2020-11-10
Update maven-dependency-analyzer

fawind:
## Before this PR

checkUnusedDependencies errors when the code contains new preview features. This is because the current version of maven-dependency-analyzer does not use the new ASM8 features.

> \> checkUnusedDependenciesMain
> java.lang.UnsupportedOperationException: This feature requires ASM8

## After this PR
Update maven-dependency-analyzer to 1.11.3 to pick up the update to ASM 8 (https://github.com/apache/maven-dependency-analyzer/pull/11).

==COMMIT_MSG==
Update maven-dependency-analyzer
==COMMIT_MSG==

## Possible downsides?
* Should be a non-breaking and backwards compatible upgrade.
* For java 16 language support we need maven-dependency-analyzer with ASM 9 which is not yet released. But this already gives us support for java 15 language features (e.g. records, text blocks, ...)


iamdanfox (approved):
Seems v uncontroversial! Could you add `org.apache.maven.shared:maven-dependency-analyzer` to the managed-artifacts.txt internally so we get this kind of upgrade for free next time?

fawind:
> Seems v uncontroversial! Could you add org.apache.maven.shared:maven-dependency-analyzer to the managed-artifacts.txt internally so we get this kind of upgrade for free next time?

Done.

PR #1542, opened 2020-11-12
Disable --release flag when --add-exports is used

fawind:
## Before this PR
To bring the palantir-java-formatter to java14+ we need to export the now internal `com.sun.tools.javac` api using [`--add-exports`](https://docs.oracle.com/javase/9/migrate/toc.htm#JSMIG-GUID-2F61F3A9-0979-46A4-8B49-325BA0EE8B66). However, on newer jdks it is not allowed to use `--add-exports` in combination with `--release` ([ref](https://stackoverflow.com/a/45403134/3388748)).

## After this PR
~~Add a gradle property to disable the `com.palantir.baseline-release-compatibility` plugin when set. This is not intended to be used generally but only in the very few places where we need full control over the provided compiler args.~~

**Updated PR:**

Don't set the `--release` flag if `--add-exports` flags are used.

==COMMIT_MSG==
Disable --release flag when --add-exports is used
==COMMIT_MSG==

## Possible downsides
It's a bit hidden so users might get confused if their --release flag disappears.


iamdanfox (approved):
Looks neat, and not too much of a blast radius

PR #1546, opened 2020-11-16
checkstyle 8.36 -> 8.37 (and keep it up to date using excavator)

iamdanfox:
## Before this PR

@fawind is putting in a bunch of work to allow people to use some more 'cutting edge' java syntax (e.g. records).  Checkstyle seems to be landing a _lot_ of bugfixes related to these new features at the moment, e.g.

https://checkstyle.sourceforge.io/releasenotes.html

Currently, as the checkstyle version is read out of a private static field, it has to be manually updated. This means we're probably going to miss bugfixes until someone complains about them.

## After this PR
==COMMIT_MSG==
checkstyle 8.36 -> 8.37, to benefit from some bugfixes to do with new java language features (e.g. records)
==COMMIT_MSG==

## Possible downsides?
- we don't read the changelog every time a new rule is introduced, so we might accidentally roll out some new strictness before validating it.  I think we can just eat this risk and disable/customize the rule if we get feedback.


gradle-baseline-java/build.gradle
```diff
@@ -131,3 +131,21 @@ project.ext.'gradle.publish.secret' = System.env["GRADLE_SECRET"]
...
+
+task writeCheckstyleVersion() {
+    doLast {
+        file('src/main/resources/checkstyle.version').text = getVersion('com.puppycrawl.tools', 'checkstyle', configurations.checkstyleVersion)
+    }
+}
+
+if (gradle.startParameter.isWriteDependencyLocks()) {
+    gradle.startParameter.taskNames += 'writeCheckstyleVersion'
+}
```
iamdanfox, line 151: an alternative approach here is to generate the resource just-in-time as a dependency of the ./gradlew idea task... no super strong opinions, i just kinda like that this one is reasonably bulletproof

fawind: Also no strong opinion but I like doing this as part of "--write-locks" as people already expect their version numbers to change through this run.

iamdanfox: yeah, in general I hate it when i open intellij and something crucial is missing because it failed to generate somehow.  this seemed more ~idiot~ dan-proof

fawind (approved):
nice!

PR #1549, opened 2020-11-17
New plugin: 'com.palantir.baseline-enable-preview-flag'

iamdanfox:
## Before this PR

I have a PR up internally on template-witchcraft#1414 which demonstrates how verbose it is to opt-in to feature previews, e.g. records. As described in [this stackoverflow answer](https://stackoverflow.com/a/61536968/1941552), I had to add:

```gradle
tasks.withType(JavaCompile) {
    options.compilerArgs += "--enable-preview"
}
tasks.withType(Test) {
    jvmArgs += "--enable-preview"
}
tasks.withType(JavaExec) {
    jvmArgs += "--enable-preview"
}
```

Which was pretty lame. Additionally, you'd have to add this to _every project_.

## After this PR
==COMMIT_MSG==
Add the new line `apply plugin: 'com.palantir.baseline-enable-preview-flag'` to your subprojects block to enable the usage of unreleased java features (e.g. records, switch expressions, var keyword etc).

Note, this plugin is a no-op on any project where you have a low sourceCompatibility.
==COMMIT_MSG==

_cc @robert3005 who usually likes shiny new things_

## Possible downsides?
- I think the decision to start using unstable java features should be taken quite cautiously, as I really don't want to find abandoned projects which are deeply coupled to some exotic unstable feature which has since been changed/removed. I think the rule-of-thumb for using this is if you really had to back it out, you could undo it all in a day.
- For a moment, I wondered if actually the verbosity of enabling this stuff is actually _desirable_, but it occurred to me that people will just copy and paste these snippets everywhere and then we'll be completely screwed. Better to have a one-liner which is easy to excavate

### DIFF
--- a/README.md
+++ b/README.md
@@ -385,7 +385,7 @@ The plugin also adds a `checkJUnitDependencies` to make the migration to JUnit5 
 3. For repos that use 'snapshot' style testing, it's convenient to have a single command to accept the updated snapshots after a code change.
 This plugin ensures that if you run tests with `./gradlew test -Drecreate=true`, the system property will be passed down to the running Java process (which can be detected with `Boolean.getBoolean("recreate")`).
 
-## com.palantir.baseline-fix-gradle-java
+## com.palantir.baseline-fix-gradle-java (off by default)
 
 Fixes up all Java [SourceSets](https://docs.gradle.org/current/userguide/building_java_projects.html#sec:java_source_sets)
 by marking their deprecated [configurations](https://docs.gradle.org/current/userguide/java_plugin.html#tab:configurations)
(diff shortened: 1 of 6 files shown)


fawind:
Looks good!

Need to double check but do we also need to set the preview flag for the `javadoc` task?

Edit: Tried it and we also need to set preview on javadoc. Something like:

```gradle
tasks.withType(Javadoc) {
    options.addBooleanOption('-enable-preview', true)
    options.addStringOption("source", SOURCE_COMPAT)
}
```

fawind (approved):
Javadoc is just nasty. But looks good!

PR #1553, opened 2020-11-17
Apply baseline-class-unqiueness by default

pkoenig10:
Follow up to https://github.com/palantir/gradle-baseline/pull/267.

## Before this PR
The `baseline-class-unqiueness` plugin is not applied by default.

## After this PR
The `baseline-class-unqiueness` plugin is applied by default.

### DIFF
--- a/changelog/@unreleased/pr-1553.v2.yml
+++ b/changelog/@unreleased/pr-1553.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: The `baseline-class-unqiueness` plugin is now applied by default.
+  links:
+  - https://github.com/palantir/gradle-baseline/pull/1553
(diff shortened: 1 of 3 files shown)


iamdanfox:
In an ideal world I would be 100% on board with this - just lock it in everywhere! My caution/hesitance comes from a couple of things:

1. in repos where there are many many projects, I think the duplication of the same stuff in many different subproject's baseline-class-uniqueness.lock might get really noisy if we add this to everything
1. the task is _super_ slow
1. anecdotally I think we've seen some flakes with this library in combination with jakarta-renames

I think if were gonna do this, I'd want a gradle property that would allow people to break glass and opt-out if necessary

pkoenig10:
Is this not sufficient as an opt-out mechanism?
```
tasks.checkClassUniqueness.enabled = false
```

iamdanfox:
Yeah that is actually a pretty reasonable opt-out, and also allows opting out on a per-project basis.

I guess if we could fix the crappy user experience of excavator PRs needing humans fix-ups then maybe this doesn't actually have that many downsides.  The fact you're using it for everything in MP is pretty encouraging :)

pkoenig10:
@dfox any other concerns here?

iamdanfox (approved):
I've landed the tweaked logic to excavator-gradle to run `baselineClassUniqueness --write-locks`, so let's give it a go.

iamdanfox:
@ferozco / @fawind shout if you have any last minute concerns, otherwise will merge when ready this in a few mins :)

fawind:
> @ferozco / @fawind shout if you have any last minute concerns, otherwise will merge when ready this in a few mins :)

Looks good with the added `baselineClassUniqueness --write-locks`! This should also be simple to roll-back if we hit an edge case for some reason that we did not consider here.

PR #1554, opened 2020-11-19
Add `PreventUsingIncubatingMethods` error-prone check (conjure)

blacksmithgu:
An explicit check which warns against using incubating APIs; you can disable it on a per-project basis, or on a per-usage basis using @SuppressWarnings.

## Before this PR
Incubating APIs are a stability liability, and should generally only be used explicitly (i.e., opt-in instead of opt-out). This check prevents their usage without an explicit annotation or opt-out.

## After this PR
Adds the `PreventUsingIncubatingMethods` error-prone check, which blocks methods annotated with Conjure's `@Incubating` annotation.

## Possible downsides?
Users upgrading `gradle-baseline` may get errors related to incubating methods until they either disable the check or tag their incubating API usages.


palantirtech:
Thanks for your interest in [palantir/gradle-baseline](https://github.com/palantir/gradle-baseline), @blacksmithgu! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

blacksmithgu:
cc @carterkozak

README.md
```diff
@@ -198,6 +198,7 @@ Safe Logging can be found at [github.com/palantir/safe-logging](https://github.c
 - `UnnecessarilyQualified`: Types should not be qualified if they are also imported.
 - `DeprecatedGuavaObjects`: `com.google.common.base.Objects` has been obviated by `java.util.Objects`.
 - `JavaTimeSystemDefaultTimeZone`: Avoid using the system default time zone.
+- `PreventUsingIncubatingMethods`: Prevents calling Conjure incubating APIs unless you explicitly opt-out of the check on a per-use or per-project basis.
```
carterkozak: Thoughts on renaming this to something like `IncubatingMethod`? We haven't been great about naming, but most error-prone checks are named to describe the bad thing they detect.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreventUsingIncubatingMethods.java
```diff
@@ -0,0 +1,66 @@
...
+import com.sun.tools.javac.code.Symbol;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "PreventUsingIncubatingMethods",
+        providesFix = BugPattern.ProvidesFix.REQUIRES_HUMAN_ATTENTION,
+        severity = BugPattern.SeverityLevel.ERROR,
+        summary = "You should avoid using incubating methods where possible, since they have very weak stability"
+                + " guarantees. You can explicitly disable this check on a case-by-case basis using"
+                + " @SuppressWarnings(\"PreventUsingIncubatingMethods\"), or disable it globally for new projects.")
```
carterkozak:
```suggestion
                + " @SuppressWarnings(\"PreventUsingIncubatingMethods\").")
```

Let's leave the global piece implied, we don't want to give folks too many ideas!

README.md
blacksmithgu: That's fine. I don't think we have code using the suppression yet, so no breakages.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreventUsingIncubatingMethods.java
```diff
@@ -0,0 +1,66 @@
...
+    public Description matchMemberReference(MemberReferenceTree tree, VisitorState state) {
+        return checkTree(tree, ASTHelpers.getSymbol(tree), state);
+    }
+
+    private Description checkTree(Tree tree, Symbol.MethodSymbol method, VisitorState state) {
+        if (!ASTHelpers.hasAnnotation(method, INCUBATING, state)) {
+            return Description.NO_MATCH;
+        }
+
+        return buildDescription(tree)
```
carterkozak: No need to set a message, the summary from the annotation above is sufficient
```suggestion
        return describeMatch(tree);
```

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreventUsingIncubatingMethods.java
blacksmithgu: Alright.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreventUsingIncubatingMethods.java
blacksmithgu: Ack, didn't know that method existed.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreventUsingIncubatingMethods.java
```diff
@@ -0,0 +1,66 @@
...
+        return checkTree(tree, ASTHelpers.getSymbol(tree), state);
+    }
+
+    @Override
+    public Description matchMemberReference(MemberReferenceTree tree, VisitorState state) {
+        return checkTree(tree, ASTHelpers.getSymbol(tree), state);
+    }
+
+    private Description checkTree(Tree tree, Symbol.MethodSymbol method, VisitorState state) {
+        if (!ASTHelpers.hasAnnotation(method, INCUBATING, state)) {
```
carterkozak: this check can be extracted to a static reusable matcher:
```java
Matchers.hasAnnotation("com.palantir.conjure.java.lib.internal.Incubating");
```

We should probably not fail builds when folks test their endpoints, so we can avoid reporting a match if `TestCheckUtils.isTestCode(state)` is true (note that this method is relatively expensive, so we should only check it after we've found an incubating match).

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/IncubatingMethodTest.java
```diff
@@ -0,0 +1,112 @@
...
+package com.palantir.baseline.errorprone;
+
+import com.google.errorprone.CompilationTestHelper;
+import org.junit.Test;
+
+public class IncubatingMethodTest {
+
+    // We define our own simple conjure incubating annotation at the right path:
+    private static final String[] ANNOTATION_DEFINITION =
+            new String[] {"package com.palantir.conjure.java.lib.internal;", "public @interface Incubating {}"};
```
carterkozak: Might be simpler to add a `testCompile` dependency on `conjure-lib` so we don't have to define this in tests, and we know we're using the right fqcn

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreventUsingIncubatingMethods.java
blacksmithgu: That is a good idea.

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/IncubatingMethodTest.java
blacksmithgu: Sure, that makes sense. Had to remember how to use `version.props` correctly...

carterkozak (approved):
Lgtm, thanks!

PR #1559, opened 2020-11-23
Add error-prone CompileTimeConstantViolatesLiskovSubstitution check

carterkozak:
==COMMIT_MSG==
Add error-prone CompileTimeConstantViolatesLiskovSubstitution check
==COMMIT_MSG==

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,6 +199,7 @@ Safe Logging can be found at [github.com/palantir/safe-logging](https://github.c
 - `DeprecatedGuavaObjects`: `com.google.common.base.Objects` has been obviated by `java.util.Objects`.
 - `JavaTimeSystemDefaultTimeZone`: Avoid using the system default time zone.
 - `IncubatingMethod`: Prevents calling Conjure incubating APIs unless you explicitly opt-out of the check on a per-use or per-project basis.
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -199,6 +199,7 @@ Safe Logging can be found at [github.com/palantir/safe-logging](https://github.c
 - `DeprecatedGuavaObjects`: `com.google.common.base.Objects` has been obviated by `java.util.Objects`.
 - `JavaTimeSystemDefaultTimeZone`: Avoid using the system default time zone.
 - `IncubatingMethod`: Prevents calling Conjure incubating APIs unless you explicitly opt-out of the check on a per-use or per-project basis.
+- `CompileTimeConstantViolatesLiskovSubstitution`: Requires consistent application of the `@CompileTimeConstant` annotation.
```
iamdanfox: I'd flesh this out to say something like "annotating a subclass method with `@CompileTimeConstant` but leaving a parent class/interface un-annotated may result in the validation being missed"

iamdanfox:
This is pretty fun! How much do you think it will tank compiler performance?

carterkozak:
I imagine it will be... measurable

README.md
carterkozak: :+1:

iamdanfox (approved):

PR #1561, opened 2020-11-24
baseline-format: can format build.gradle files

iamdanfox:
## Before this PR

Every now and again, we stumble across some elaborate build.gradle files full of custom stuff. These are sometimes hard to read due to each project's own idiosyncratic styling opinions.

Well it turns out spotless can just zap em all with a consistent format!

## After this PR
==COMMIT_MSG==
`com.palantir.baseline-format` now understands how to format build.gradle files. This is opt-in by default, so you need to run `./gradlew format -Pcom.palantir.baseline-format.gradle-files=true` to try it, or add this property to your gradle.properties if you want to lock it in.
==COMMIT_MSG==

I've intentionally made this opt-in because I'm intending to write an auto-merging excavator which will just come round and re-format your files, without incurring the annoyance of failing builds because someone has got their indentation wrong.

## Possible downsides?
- people can be pretty opinionated about styling, and I think there's a risk this might make certain build.gradles _longer_ (e.g. by expanding one-liners to multiple lines).  Overall though, I think most people have recognised the value of applying palantir-java-format everywhere, so maybe they'll be receptive???
- i've spent like 5 minutes on this PR, so there are no tests. #gottagofast

### DIFF
--- a/gradle-baseline-java/src/main/resources/greclipse.properties
+++ b/gradle-baseline-java/src/main/resources/greclipse.properties
@@ -1,0 +1,3 @@
+org.eclipse.jdt.core.formatter.tabulation.char=space
+org.eclipse.jdt.core.formatter.indent_empty_lines=false
+groovy.formatter.remove.unnecessary.semicolons=true
--- a/README.md
+++ b/README.md
@@ -303,6 +303,7 @@ buildscript {
     }
 }
 
+apply plugin: 'com.palantir.java-format'
 apply plugin: 'com.diffplug.spotless'
 
 spotless {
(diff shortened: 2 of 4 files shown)


gradle-baseline-java/src/main/resources/greclipse.properties
```diff
@@ -0,0 +1,3 @@
+org.eclipse.jdt.core.formatter.tabulation.char=space
+org.eclipse.jdt.core.formatter.indent_empty_lines=false
+groovy.formatter.remove.unnecessary.semicolons=true
```
iamdanfox, line 3: I did a trial run with this config on devx/template-witchcraft and it makes _zero_ changes!! I didn't quite believe it, so I indented stuff like crazy and it restored it :)

fawind:
The unit-test-15 errors are unrelated, put up a fix in https://github.com/palantir/gradle-baseline/pull/1565. But don't think we need to block on this.

fawind (approved):
I love it!

Will the excavator also be opt-in? I think its fine to just always let the excavator reformat the gradle files but maybe that's what you are already intending to do :)

PR #1565, opened 2020-11-25
Fix unit tests on java 15

fawind:
## Before this PR
We switched testing on java 14 to testing on java 15 in https://github.com/palantir/gradle-baseline/commit/02c9f15d91056bb8bdbcc7eae910fe06424fe562 which broke some tests on java 15 and automerged before we could fix the tests.

Failures: https://app.circleci.com/pipelines/github/palantir/gradle-baseline/1492/workflows/f8aaf40f-f7fb-4a28-8860-80f00df4532d/jobs/21807

## After this PR

Fix unit tests when running with java 15.

==COMMIT_MSG==
Fix unit tests on java 15
==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/UnnecessaryParenthesesTest.java
+++ b/baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/UnnecessaryParenthesesTest.java
@@ -29,7 +29,7 @@ public class UnnecessaryParenthesesTest {
     public void testSwitchExpression() {
         CompilationTestHelper compilationHelper = CompilationTestHelper.newInstance(
                         UnnecessaryParentheses.class, getClass())
-                .setArgs(ImmutableList.of("--enable-preview", "--release", "14"));
+                .setArgs(ImmutableList.of("--enable-preview", "--release", "15"));
 
         compilationHelper
                 .addSourceLines(
--- a/gradle-baseline-java/src/test/groovy/com/palantir/baseline/plugins/BaselineEnablePreviewFlagTest.groovy
+++ b/gradle-baseline-java/src/test/groovy/com/palantir/baseline/plugins/BaselineEnablePreviewFlagTest.groovy
@@ -156,19 +156,38 @@ class BaselineEnablePreviewFlagTest extends IntegrationSpec {
 
     def 'does not always apply'() {
         when:
-        File java8Dir = addSubproject('my-java-8-api', "sourceCompatibility = 8")
+        File java8Dir = addSubproject('my-java-8-api', """
+        sourceCompatibility = 8
+
+        tasks.withType(JavaCompile) {
+           doFirst {
+              println "Args for for \$name are \$options.allCompilerArgs"
+           }
+        } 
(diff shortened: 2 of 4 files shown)


gradle-baseline-java/src/test/groovy/com/palantir/baseline/plugins/BaselineEnablePreviewFlagTest.groovy
```diff
@@ -156,19 +156,38 @@ class BaselineEnablePreviewFlagTest extends IntegrationSpec {
...
+        } 
+        """.stripIndent())
+
         buildFile << '''
         subprojects {
             apply plugin: 'java-library'
             apply plugin: 'com.palantir.baseline-enable-preview-flag'
         }
         '''
-        writeSourceFileContainingRecord(java8Dir)
```
fawind, line 166: `:compileJava` was failing here with ([ref](https://app.circleci.com/pipelines/github/palantir/gradle-baseline/1492/workflows/f8aaf40f-f7fb-4a28-8860-80f00df4532d/jobs/21807)):
```
error: cannot find symbol
|            public record Coordinate(int x, int y) {}
|                   ^
|        symbol:   class record
```

which makes sense to me as we are running with `sourceCompatibility = 8`. I think the output has changed from java 14 to java 15 and no longer contains the `use --enable-preview to enable records` line.

Changed the test to explicitly print the compiler args and check that `--enable-preview` is not set when running with `sourceCompatibility=8`.

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/UnnecessaryParenthesesTest.java
```diff
@@ -29,7 +29,7 @@
     public void testSwitchExpression() {
         CompilationTestHelper compilationHelper = CompilationTestHelper.newInstance(
                         UnnecessaryParentheses.class, getClass())
-                .setArgs(ImmutableList.of("--enable-preview", "--release", "14"));
+                .setArgs(ImmutableList.of("--enable-preview", "--release", "15"));
```
fawind, line 32: This is to fix this error ([ref](https://app.circleci.com/pipelines/github/palantir/gradle-baseline/1492/workflows/f8aaf40f-f7fb-4a28-8860-80f00df4532d/jobs/21807)):
```
invalid source release 14 with --enable-preview
    (preview language features are only supported for release 15)
```

I think that's a restriction by the java-compiler that enable-preview has to be used with the latest release but could not find good documentation on this.

iamdanfox: Yep I stumbled across this restriction too when writing https://github.com/palantir/gradle-baseline/pull/1549

> javac in JDK $N accepts the --enable-preview flag only in combination with --release $N or -source $N. (For brevity, only --release is mentioned in the rest of this JEP.)

https://openjdk.java.net/jeps/12

iamdanfox (approved):

PR #1568, opened 2020-11-26
Fix path sensitivity of compile tasks

iamdanfox:
## Before this PR

Our build caching is currently kinda broken, because you can't get cache hits between local & CI. [build scan link](https://gradle.p.b/c/m2khsesqbx2jw/kbk7v6omaei5k/task-inputs?expanded=WyJuejY2Z3RvdTZyNmtjLXNvdXJjZSIsImNwNGxucXJqYXU3d2Etc291cmNlIl0#ovd53fafhaeic-options.compilerargumentproviders.errorprone$1.errorproneoptions.excludedpaths)

image

Specifically, on a local run the java compiler would be invoked with the following argument, which contains an absolute path 🌶🌶🌶:

```
-XepDisableWarningsInGeneratedCode -XepExcludedPaths:\Q/Volumes/git/container-vuln-scanner/container-vuln-scanner-api\E\Q/\E(build|src\Q/\Egenerated.*)\Q/\E.* -Xep:FallThrough -Xep:UnnecessaryParentheses -Xep:CatchSpecificity:OFF -Xep:UnusedVariable:OFF -Xep:EqualsHashCode:ERROR -Xep:EqualsIncompatibleType:ERROR -Xep:StreamResourceLeak:ERROR -Xep:InputStreamSlowMultibyteRead:ERROR -Xep:JavaDurationGetSecondsGetNano:ERROR -Xep:URLEqualsHashCode:ERROR -Xep:BoxedPrimitiveEquality:ERROR -Xep:ReferenceEquality:ERROR -XepOpt:LogsafeArgName:UnsafeArgNames=branch,branchid,branchname,branchrid,email,emailaddress,organization,organizationname,path,ref,secret,transactionref,token,username
```

When the repo is the checked out to a different location (e.g. `/home/circleci/container-vuln-scanner`), it computes a different set of args and causes a cache miss.

_cc @jmcampanini this is example number (3) of gradle enterprise saving the day_

## After this PR
==COMMIT_MSG==
JavaCompile tasks should now get more build cache hits irrespective of the location of your repo on disk, as `baseline-errorprone` no longer injects an absolute path into `errorproneOptions.excludedPaths`.
==COMMIT_MSG==

Interestingly I _think_ that windows compilation was also broken, and that this _might_ fix it.

PR #1571, opened 2020-11-26
Exclude generated sources from error-prone

### LINKED ISSUE #1570: 3.57.1 breaks error-prone generated excluded paths
## What happened?

As seen in excavator bump PR https://github.com/palantir/tritium/pull/925 , the changes in https://github.com/palantir/gradle-baseline/pull/1568 do not seem to handle cases that previously were excluded such as `build/metricSchema/generated_src` used by `metric-schema` which is causing the baseline bump excavator on tritium to fail compilation: https://app.circleci.com/pipelines/github/palantir/tritium/906/workflows/7c65b97b-b475-46fc-a1e1-b5a4bf7198a0/jobs/11565

Baseline bump to successfully compile so that the 🤖 work while we eat 🦃

schlosna:
## Before this PR

As seen in excavator bump PR https://github.com/palantir/tritium/pull/925 , the changes in https://github.com/palantir/gradle-baseline/pull/1568 do not seem to handle cases that previously were excluded such as `build/metricSchema/generated_src` used by `metric-schema` which is causing the baseline bump excavator on tritium to fail compilation: https://app.circleci.com/pipelines/github/palantir/tritium/906/workflows/7c65b97b-b475-46fc-a1e1-b5a4bf7198a0/jobs/11565

## After this PR

==COMMIT_MSG==
Exclude generated sources from error-prone
==COMMIT_MSG==

Closes https://github.com/palantir/gradle-baseline/issues/1570

## Possible downsides?

Probably want some additional test cases.

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
@@ -201,9 +201,7 @@ public final class BaselineErrorProne implements Plugin<Project> {
         }
 
         errorProneOptions.getDisableWarningsInGeneratedCode().set(true);
-        errorProneOptions
-                .getExcludedPaths()
-                .set(excludedPathsRegex());
+        errorProneOptions.getExcludedPaths().set(excludedPathsRegex());
 
         // FallThrough does not currently work with switch expressions
         // See https://github.com/google/error-prone/issues/1649
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
@@ -201,13 +201,9 @@ public final class BaselineErrorProne implements Plugin<Project> {
         }
 
         errorProneOptions.getDisableWarningsInGeneratedCode().set(true);
-        // don't want backslashes on windows to break our regex
-        String separator = File.separatorChar == '\\' ? Pattern.quote("\\") : File.separator;
         errorProneOptions
                 .getExcludedPaths()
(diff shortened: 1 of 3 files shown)


iamdanfox (approved):
Ah thanks for rescuing this! I wanted to make those regexes really specific, because I figured there was a danger of someone defining a class `com.palantir.foundry.build.Whatever` and getting it unintentionally excluded - tests definitely worth it

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
```diff
@@ -292,6 +286,12 @@ private static void configureErrorProneOptions(
         }
     }
 
+    static String excludedPathsRegex() {
+        // don't want backslashes on windows to break our regex
+        String separator = File.separator.contains("\\") ? Pattern.quote("\\") : File.separator;
+        return String.format(".*%s(build|generated_.*[sS]rc|src%sgenerated.*)%s.*", separator, separator, separator);
```
schlosna, line 292: @iamdanfox we might be able to remove `build` from regex here, or might want to constrain the pattern to avoid falsely matching say `com/foo/build/Bar.java`.

robert3005:
~Just an FYI, had a project with generated sources in `build/generated-src/antlr/`. Going to change that to `_`.~  Womp, looks I am not following the pattern

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
pkoenig10: Yeah this feels a bit dangerous. I'd suggest we do something like `build.*generated` since almost all code-gen plugins I've seen put source files on a path with a `generated` component.

PR #1574, opened 2020-11-27
baseline-reproducibility validates that sourceCompatibility is set explicitly

iamdanfox:
## Before this PR

To get a new JVM rolled out fleet wide, we blast out excavators which just dial up the JVM used at CI time.  This has been pretty much fine so far as most projects set `sourceCompatibility` explicitly, but nothing actually _validates_ this right now.

Occasionally, this manifests as a question in `#dev-foundry-infra` about the product-dependencies.lock file differing locally on CI (because on CI it's inferring java 15) and locally it's inferring whatever JVM the dev is using, often 11.

## After this PR
==COMMIT_MSG==
baseline-reproducibility validates that sourceCompatibility is set explicitly using a new `checkExplicitSourceCompatibility` task.
==COMMIT_MSG==

## Possible downsides?
- it's one more task that will get registered on every project, also can't see any reasonable way of caching this badboy

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckExplicitSourceCompatibilityTask.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckExplicitSourceCompatibilityTask.java
@@ -86,7 +86,7 @@ public class CheckExplicitSourceCompatibilityTask extends DefaultTask {
         if (shouldFix.get()) {
             Files.write(
                     getProject().getBuildFile().toPath(),
-                    Collections.singletonList(String.format("%nsourceCompatibility = %s", JavaVersion.current())),
+                    Collections.singletonList(String.format("%nsourceCompatibility = %s%n", JavaVersion.current())),
                     StandardCharsets.UTF_8,
                     StandardOpenOption.APPEND,
                     StandardOpenOption.CREATE);
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckExplicitSourceCompatibilityTask.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckExplicitSourceCompatibilityTask.java
@@ -25,9 +25,12 @@ import javax.inject.Inject;
 import org.gradle.api.DefaultTask;
 import org.gradle.api.GradleException;
 import org.gradle.api.JavaVersion;
+import org.gradle.api.Task;
 import org.gradle.api.model.ObjectFactory;
 import org.gradle.api.plugins.JavaPluginConvention;
 import org.gradle.api.provider.Property;
+import org.gradle.api.publish.PublishingExtension;
+import org.gradle.api.specs.Spec;
 import org.gradle.api.tasks.TaskAction;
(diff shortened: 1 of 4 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckExplicitSourceCompatibilityTask.java
```diff
@@ -0,0 +1,110 @@
...
+                        + "that Gradle is currently running with (%s). To auto-fix, run%n"
+                        + "%n"
+                        + "     ./gradlew %s --fix%n"
+                        + "%n"
+                        + "This will automatically add a suggested line "
+                        + "(you may need to adjust the number, e.g. to '1.8' for maximum compatibility).",
+                getProject(),
+                getProject().getRootProject().relativePath(getProject().getBuildFile()),
+                JavaVersion.current(),
+                getPath()));
```
iamdanfox, line 108: I broke tracing-java and tried it out, error message looks like this:
```
-----------
* What went wrong:
Execution failed for task ':tracing-jaxrs:checkExplicitSourceCompatibility'.
> project ':tracing-jaxrs' must set sourceCompatibility explicitly in 'tracing-jaxrs/build.gradle', otherwise compilation will not be reproducible but instead depends on the Java version that Gradle is currently running with (11). To auto-fix, run

       ./gradlew :tracing-jaxrs:checkExplicitSourceCompatibility --fix

  This will automatically add a suggested line (you may need to adjust the number, e.g. to '1.8' for maximum compatibility).

```

gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckExplicitSourceCompatibilityTask.java
```diff
@@ -0,0 +1,110 @@
...
+        if (convention.getRawSourceCompatibility() != null) {
+            // In theory, users could configure the fancy new 'java toolchain' as an alternative to explicit
+            // sourceCompatibility, but there's no method to access this yet (as of Gradle 6.8).
+            return;
+        }
+
+        if (shouldFix.get()) {
+            Files.write(
+                    getProject().getBuildFile().toPath(),
+                    Collections.singletonList(String.format("%nsourceCompatibility = %s", JavaVersion.current())),
```
schlosna: Should this also add a trailing newline at the end of the file?

```suggestion
                    Collections.singletonList(String.format("%nsourceCompatibility = %s%n", JavaVersion.current())),
```

fawind (approved):
+1 on having sourceCompat set explicitly.

Looks good to me! Feel free to merge after fixing the nit 🚀

robert3005:
How does this work with https://docs.gradle.org/current/userguide/toolchains.html ? Would I have to specify source compat even if I specify my jvm?

iamdanfox:
Currently yes - although as soon as they expose a way to extract the toolchain stuff we can update this plugin https://github.com/palantir/gradle-baseline/pull/1574/files#diff-14d5052a9b7ac4fa7663ebf9a728428314e468cc84e6a420f0de25aacaebbc57R81-R82

PR #1576, opened 2020-11-30
Saner test event logging default for CI

CRogers:
## Before this PR
We've seen numerous times that when projects have long running integration test tasks that take longer than 10 mins they get killed by circle ("context deadline exceeded"). If there are no failing tests we won't print anything in the gradle output, so it's possible to hit this deadline.

## After this PR

==COMMIT_MSG==
Print more test logging output to avoid builds with long running tests getting terminated by circle ("context deadline exceeded").
==COMMIT_MSG==

## Possible downsides?
Perhaps more noisy, but the failed tests already have junit xml produced in circle, so this should be strictly better.

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineTesting.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineTesting.java
@@ -16,6 +16,7 @@
 
 package com.palantir.baseline.plugins;
 
+import com.google.common.collect.ImmutableSet;
 import com.palantir.baseline.tasks.CheckJUnitDependencies;
 import com.palantir.baseline.tasks.CheckUnusedDependenciesTask;
 import java.util.Objects;
@@ -32,6 +33,7 @@ import org.gradle.api.tasks.TaskContainer;
 import org.gradle.api.tasks.TaskProvider;
 import org.gradle.api.tasks.testing.Test;
 import org.gradle.api.tasks.testing.junitplatform.JUnitPlatformOptions;
+import org.gradle.api.tasks.testing.logging.TestLogEvent;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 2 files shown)


ferozco (commented):
👎 Pretty sure we explicitly wanted to avoid printing out test details instead relying on test reports. Seems like there is an unrelated problem if user tests are taking that long to execute

iamdanfox:
I've found this kind of logging extremely helpful when a repo has integration tests.  Fine to leave it off for unit-tests, but I don't think it's that controversial for integ!

CRogers:
@ferozco The issue that if all your tests pass, we print out no events at all from the test tasks, so even if each test takes 10 secs but you have 60 of them, the test task will not log anything for 10 mins and the build will be killed by circle. I've seen this happen internally on at least 5 different projects, and since circle just `kill -15`s gradle you get really gnarly errors.

Additionally, it's useful for cases when tests do hang and then eventually get killed by gradle to be able to see which test is bad

CRogers:
Definitely more of a problem for integ tests than for unit tests, but I've seen this happen in our gradle nebula tests as well (which use the default `test` tasks).

ferozco:
I hear what you're saying perhaps we can find a middle ground? I think printing out every single test case results in a ton of noise and causes CI output to become large (which devtools doesn't like). Can we only print test suites (i.e. classes).

Alternatively, we could just recommend the folks with 10m+ tests to configure this themselves

iamdanfox:
@ferozco just for extra context, here are the repos in foundry _already_ using this: https://sourcegraph.p.b/search?q=testLogging+repo:foundry/&patternType=literal

ferozco:
Thanks for the link @iamdanfox I didn't realize how wide spread it was to configure some tests. I think I still stand by my opinion that this configuration is undesirable inside of unit tests though, so I'd be supportive of @CRogers idea of pushing this into dev-env

robert3005 (approved):

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineTesting.java
```diff
@@ -140,8 +142,17 @@ private static void enableJunit5ForTestTask(Test task) {
...
-        // provide some stdout feedback when tests fail
-        task.testLogging(testLogging -> testLogging.events("failed"));
+        // provide some stdout feedback when tests fail when running on CI and locally
+        task.getTestLogging().getEvents().add(TestLogEvent.FAILED);
+
+        // Only on CI, print out more detailed test information to avoid hitting the circleci 10 min deadline if
+        // there are lots of tests. Don't do this locally to avoid spamming massive amount of info for people running
+        // unit tests through the command line
+        if ("true".equals(System.getenv("CI"))) {
+            task.getTestLogging()
```
iamdanfox, line 153: @CRogers can we try to not apply this to the `./gradlew test` tasks which are normally just unit-tests??

CRogers: Yep, I have a commit to do this actually, I was just distracted by testing it on c-v-s that had the silently not running tests problem :p

CRogers: Pushed

iamdanfox (approved):

PR #1587, opened 2020-12-15
Escape gradle failure text properly with CDATA section in junit reports

CRogers:
## Before this PR
I think with an internal PR I have printed out some characters that cause the junit report produced by the `JunitReportsPlugin` to produce invalid xml, at least according to circle :

Screen Shot 2020-12-15 at 2 58 57 PM

One of these characters: https://en.wikipedia.org/wiki/List_of_XML_and_HTML_character_entity_references#Character_reference_overview

## After this PR

==COMMIT_MSG==
A CDATA section is used in junit reports to avoid string escaping problems.
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-1587.v2.yml
+++ b/changelog/@unreleased/pr-1587.v2.yml
@@ -1,0 +1,5 @@
+type: fix
+fix:
+  description: A CDATA section is used in junit reports to avoid string escaping problems.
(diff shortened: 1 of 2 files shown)


CRogers:
Eh this was actually a problem in an attribute rather than a tag content, I've worked around it.

PR #1591, opened 2020-12-21
Add configuration to checkImplicitDependencies suggestion

pkoenig10:
## Before this PR
The `checkImplicitDependencies` suggestion message does not include the configuration name. This make it cumbersome for users to translate the suggestion to their Gradle configuration.
```
Found 1 implicit dependencies - consider adding the following explicit dependencies or avoid using classes from these jars:
    dependencies {
        'com.palantir.foo:foo'
    }
```

## After this PR
The `checkImplicitDependencies` suggestion message includes the `implementation` configuration name. This allows users to just copy and paste the suggestion into their Gradle configuration.
```
Found 1 implicit dependencies - consider adding the following explicit dependencies or avoid using classes from these jars:
    dependencies {
        implementation 'com.palantir.foo:foo'
    }
```

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckImplicitDependenciesTask.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckImplicitDependenciesTask.java
@@ -102,7 +102,7 @@ public class CheckImplicitDependenciesTask extends DefaultTask {
                         "'%s:%s'",
                         artifact.getModuleVersion().getId().getGroup(),
                         artifact.getModuleVersion().getId().getName());
-        return String.format("        %s", artifactNameString);
+        return String.format("        implementation %s", artifactNameString);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


carterkozak (approved):

gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckImplicitDependenciesTask.java
```diff
@@ -56,6 +57,7 @@ public CheckImplicitDependenciesTask() {
         sourceClasses = getProject().getObjects().property(FileCollection.class);
         ignore = getProject().getObjects().setProperty(String.class);
         ignore.set(Collections.emptySet());
+        suggestionConfigurationName = getProject().getObjects().property(String.class);
```
pkoenig10, line 60: Not sure whether it's better to add another property or change the existing `sourceClasses` property to a `SourceSet`. I was worried that might be incorrect since that property is annotated with `@Classpath`.

PR #1593, opened 2020-12-22
Don't modify exact dependencies index

pkoenig10:
## Before this PR
The values in the `classToDependency` index can be modified if the same class is produced by multiple artifacts and subprojects are not all using the same artifact to provide this class.

This can happen quite easily when using our internal jakarta renames plugin.

## After this PR
The `classToDependency` index stores sets of artifacts that provide a given class. The `checkUnusedDependencies` and `checkImplicitDependencies` tasks have been updated to handle multiple artifacts as gracefully as possible.

This was pretty trivial for the `checkUnusedDependencies` task. It was a bit more involved for the `checkImplicitDependencies` task since we have to treat the potential artifacts as a conjunctive normal form.


fawind (approved):
Looks very reasonable, thanks for fixing!

PR #1598, opened 2021-01-04
Implement error-prone ClassInitializationDeadlock check

carterkozak:
## Before this PR
Deadlocks which require API/ABI changes to resolve. It's a far better use of resources to alert on these using static analysis than to deadlock every N service startups depending on architecture, JVM, and luck.

## After this PR

==COMMIT_MSG==
Implement error-prone ClassInitializationDeadlock check
==COMMIT_MSG==

## Possible downsides?
Might produce noise in some projects, however in those cases it's tremendously easy to inadvertently shoot ourselves in the foot.

### DIFF
--- a/baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/ClassInitializationDeadlockTest.java
+++ b/baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/ClassInitializationDeadlockTest.java
@@ -1,0 +1,306 @@
+/*
+ * (c) Copyright 2021 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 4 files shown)


baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/ClassInitializationDeadlockTest.java
```diff
@@ -0,0 +1,306 @@
...
+                        "  private final Sub SUB = new Sub();",
+                        "  static class Sub extends Base {",
+                        "  }",
+                        "}")
+                .doTest();
+    }
+
+    @Test
+    void testStaticFieldNewSubtypeInstancePrivateAllowed() {
+        // This cas is handled to avoid false positives, however we don't
```
fawind:
```suggestion
        // This case is handled to avoid false positives, however we don't
```

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/ClassInitializationDeadlockTest.java
```diff
@@ -0,0 +1,306 @@
...
+                        "    new Sub();",
+                        "  }",
+                        "  private static class Sub extends Base {",
+                        "  }",
+                        "}")
+                .doTest();
+    }
+
+    @Test
+    void deleteMe_exampleOnly() {
```
fawind: delete me :) Should test same codepath as `testStaticFieldNewSubtypeInstance`.

fawind (approved):
Looks very clean, feel free to merge!

fawind:
For future reference, an example of where we ran into the issue: https://github.com/palantir/conjure-java/pull/1211

PR #1600, opened 2021-01-05
Add baseline-encoding plugin to force UTF-8 during compilation

pkoenig10:
## Before this PR
`JavaCompile` tasks use the platform charset by default. This is often `UTF-8`, but we have seen internal CI builds where the platform charset is `US_ASCII`. This can lead to compilation failures if source files contains non-ASCII characters.

## After this PR
Baseline now provides a `baseline-encoding` plugin that set the encoding to `UTF-8` for all `JavaCompile` tasks.

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineEncoding.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineEncoding.java
@@ -24,8 +24,8 @@ public final class BaselineEncoding implements Plugin<Project> {
 
     @Override
     public void apply(Project project) {
-        project.getTasks().withType(JavaCompile.class).configureEach(t -> {
-            t.getOptions().setEncoding("UTF-8");
+        project.getTasks().withType(JavaCompile.class).configureEach(javaCompileTask -> {
+            javaCompileTask.getOptions().setEncoding("UTF-8");
         });
     }
 }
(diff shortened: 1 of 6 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineEncoding.java
```diff
@@ -0,0 +1,31 @@
...
+import org.gradle.api.Plugin;
+import org.gradle.api.Project;
+import org.gradle.api.tasks.compile.JavaCompile;
+
+public final class BaselineEncoding implements Plugin<Project> {
+
+    @Override
+    public void apply(Project project) {
+        project.getTasks().withType(JavaCompile.class).configureEach(t -> {
+            t.getOptions().setEncoding("UTF-8");
```
carterkozak: nit: would prefer to rename t -> compileJavaTask

carterkozak (approved):
lgtm, one minor naming nit. Merge when you're ready.

gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineEncodingIntegrationTest.groovy
```diff
@@ -0,0 +1,84 @@
...
+
+    def 'compileJava fails with other encoding'() {
+        when:
+        buildFile << otherEncodingBuildFile
+        file('src/main/java/test/Test.java').text = javaFile
+
+        then:
+        BuildResult result = with('compileJava').build()
+        result.task(":compileJava").outcome == TaskOutcome.SUCCESS
+        result.output.contains("unmappable character")
```
pkoenig10, line 82: Compilation doesn't actually fail when encountering unmappable characters - those characters are just replaced with the unicode replacement character.

PR #1603, opened 2021-01-06
Disable WhitespaceAround when using palantir-java-format

pkoenig10:
## Before this PR

The `WhitespaceAround` checkstyle check conflicts with how palantir-java-format (correctly) formats empty blocks in switch expressions.

The following code:
```
enum Case { ONE, TWO }
switch (foo) {
    case ONE -> {},
    case TWO -> {
        ...
    }
}
```
Results in the following errors:
```
WhitespaceAround: '}' is not preceded with whitespace.
WhitespaceAround: '{' is not followed by whitespace. Empty blocks may only be represented as {} when not part of a multi-block statement (4.1.3)
```

See internal build https://c.p.b/gh/foundry/mp/236221

## After this PR

The `WhitespaceAround` checkstyle check is removed when using palantir-java-format.

This is similar to what we've done previously for other checks in https://github.com/palantir/gradle-baseline/pull/987, https://github.com/palantir/gradle-baseline/pull/1205, and https://github.com/palantir/gradle-baseline/pull/1278.

I've also taken the liberty to refactor this to use Java's XML tooling to make it a bit more robust and easier to read/modify. This did result in some slight formatting changes to the output. See https://g.p.b/foundry/mp/commit/acc37f76 for an example of the changes.

### DIFF
--- a/changelog/@unreleased/pr-1603.v2.yml
+++ b/changelog/@unreleased/pr-1603.v2.yml
@@ -1,0 +1,6 @@
+type: fix
+fix:
+  description: Disable `WhitespaceAround` Checkstyle rule when palantir-java-format
+    is applied.
+  links:
+  - https://github.com/palantir/gradle-baseline/pull/1603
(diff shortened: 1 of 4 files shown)


fawind (approved):
Very nice, thanks for fixing!

PR #1606, opened 2021-01-07
Preserve DOCTYPE in checkstyle.xml

pkoenig10:
## Before this PR
The XML parsing/writing logic added in https://github.com/palantir/gradle-baseline/pull/1603 doesn't preserve the DOCTYPE. This causes the `checkstyleMainCircleFinalizer` task to fail.

https://app.circleci.com/pipelines/github/palantir/gradle-baseline/1699/workflows/2623eafb-018b-41d4-bd1f-9fb2e6511f37/jobs/23047

## After this PR
Preserve the DOCTYPE when writing the checkstyle XML.

### DIFF
--- a/gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
+++ b/gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
@@ -3,13 +3,6 @@
         "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
         "https://checkstyle.org/dtds/configuration_1_3.dtd">
 
-<!--
-    Palantir Baseline Checkstyle configuration.
-    Authors: Robert Fink, Brian Worth, Merrick Zoubeiri, and many other contributors. Based in part on http://checkstyle.sourceforge.net/google_style.html
-    Please keep checks alphabetized with one exception: "relaxed" checks are grouped together at the bottom for easier disabling.
-    Check-specific comments reference documents internal to Palantir and can be safely ignored or removed.
- -->
-
 <module name="Checker">
     <property name="charset" value="UTF-8"/>
     <property name="severity" value="error"/>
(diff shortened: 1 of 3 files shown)


gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
```diff
@@ -3,13 +3,6 @@
         "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
         "https://checkstyle.org/dtds/configuration_1_3.dtd">
 
-<!--
-    Palantir Baseline Checkstyle configuration.
-    Authors: Robert Fink, Brian Worth, Merrick Zoubeiri, and many other contributors. Based in part on http://checkstyle.sourceforge.net/google_style.html
-    Please keep checks alphabetized with one exception: "relaxed" checks are grouped together at the bottom for easier disabling.
-    Check-specific comments reference documents internal to Palantir and can be safely ignored or removed.
- -->
```
pkoenig10: Removed this comment because it didn't write well - it ends up looking like:
```
<?xml version="1.0" encoding="UTF-8" standalone="no"?><!--
    Palantir Baseline Checkstyle configuration.
    Authors: Robert Fink, Brian Worth, Merrick Zoubeiri, and many other contributors. Based in part on http://checkstyle.sourceforge.net/google_style.html
    Please keep checks alphabetized with one exception: "relaxed" checks are grouped together at the bottom for easier disabling.
    Check-specific comments reference documents internal to Palantir and can be safely ignored or removed.
 --><module name="Checker">
```

This comment isn't particularly useful. In particular, the checks are no longer alphabetized.

carterkozak: I'd prefer not to make this change in the same PR as a bugfix.

pkoenig10: Are you opposed to the change? I'm happy to make another PR, but if you're just going to approve that one it seems a bit silly.

I probably should have removed this in https://github.com/palantir/gradle-baseline/pull/1603.

carterkozak: I don't really have an opinion one way or the other on the comment.

carterkozak (approved):

PR #1617, opened 2021-01-18
improvement: Disable inner import on classes that cause errorprone to fail

ferozco:
## Before this PR
Errorprone produces `BadImport` which fails if you import nested classes with commonly used names. We however configured intellij to blindly auto-import nested classes where possible.

## After this PR

==COMMIT_MSG==
Disable inner import on classes that cause ErrorProne to fail
==COMMIT_MSG==

List of classes to disable sourced from [here](https://github.com/google/error-prone/blob/master/core/src/main/java/com/google/errorprone/bugpatterns/BadImport.java#L67)

### DIFF
--- a/changelog/@unreleased/pr-1617.v2.yml
+++ b/changelog/@unreleased/pr-1617.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Disable inner import on classes that cause ErrorProne to fail
+  links:
+  - https://github.com/palantir/gradle-baseline/pull/1617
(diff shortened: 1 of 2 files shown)


carterkozak:
Why don't we disable inner class imports entirely? The outer class provides important context, otherwise the imported class shouldn't be nested.

ferozco:
I agree with the  motivation behind the [original](https://github.com/palantir/gradle-baseline/pull/1341) change and have found  other cases (ex. Timber, Jackson, etc.) where importing the nested classes improves legibility and succinctness without a loss in clarity

robert3005 (approved):

carterkozak:
I don't agree with the original change, it's unfortunate that we penalize everyone for suboptimal generated code layout. I'd noticed inner classes getting auto-imported by my IDE lately and assumed it was an idea EA bug. This change shifts the needle the other direction which is an improvement, it's unfortunate that I missed the original discussion.

carterkozak (approved):

PR #1632, opened 2021-01-26
Migrate preconditions to the new package

carterkozak:
## Before this PR
See https://github.com/palantir/safe-logging/pull/515

## After this PR
This change does two things:
* The existing errorprone checks which reference preconditions
  have been updated to suggest the new package name
* A new `SafeLoggingPreconditionsMigration` check has been added
  to automatically migrate code to the new package when a
  sufficiently new `safe-logging` version is available.
==COMMIT_MSG==
Migrate preconditions to the new package
==COMMIT_MSG==


carterkozak:
Draft until palantir/safe-logging#515 is ready

PR #1644, opened 2021-02-11
Feature: Add consistent logging errorprone rule

### LINKED ISSUE #1642: Enforce casing and naming of slf4j static `Logger` field
## What happened?

Our tooling currently allows people to name their logger fields anything they want (and also tolerates different capitalizations):
```java
private static final Logger log = LoggerFactory.getLogger(MyClass.class);
private static final Logger LOG = LoggerFactory.getLogger(MyClass.class);
private static final Logger LOGGER = LoggerFactory.getLogger(MyClass.class);
private static final Logger logger = LoggerFactory.getLogger(MyClass.class);
```

I've observed a couple of code-review nit threads where the one of these has been picked and the maintainers would prefer another.

I'd like tooling to just enforce this (and ideally auto-fix everywhere), because I don't think it's a valuable thing for authors/reviewers to spend time on and reduces the signal to noise ratio of code reviews. My preferred variant would be:

```java
private static final Logger log = LoggerFactory.getLogger(MyClass.class);
```

ferozco:
## Before this PR
Closes #1642 

## After this PR
==COMMIT_MSG==
Add consistent logging errorprone rule
==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ConsistentLoggerName.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ConsistentLoggerName.java
@@ -26,7 +26,7 @@ import com.google.errorprone.matchers.Matcher;
 import com.google.errorprone.matchers.Matchers;
 import com.google.errorprone.matchers.method.MethodMatchers;
 import com.sun.source.tree.VariableTree;
-import java.lang.reflect.Modifier;
+import javax.lang.model.element.Modifier;
 
 @AutoService(BugChecker.class)
 @BugPattern(
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ConsistentLoggerName.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ConsistentLoggerName.java
@@ -39,6 +39,8 @@ public final class ConsistentLoggerName extends BugChecker implements BugChecker
 
     private static final Matcher<VariableTree> matcher = Matchers.allOf(
             Matchers.isField(),
+            Matchers.isStatic(),
+            MoreMatchers.isFinal(),
             Matchers.isSubtypeOf("org.slf4j.Logger"),
             Matchers.variableInitializer(MethodMatchers.staticMethod()
                     .onClass("org.slf4j.LoggerFactory")
@@ -48,16 +50,13 @@ public final class ConsistentLoggerName extends BugChecker implements BugChecker
 
     @Override
     public Description matchVariable(VariableTree tree, VisitorState state) {
-        if (!matcher.matches(tree, state)) {
-            return Description.NO_MATCH;
+        if (matcher.matches(tree, state)) {
+            if (!tree.getName().contentEquals("log")) {
+                return buildDescription(tree)
+                        .addFix(SuggestedFixes.renameVariable(tree, "log", state))
+                        .build();
+            }
         }
-
-        if (tree.getName().contentEquals("log")) {
-            return Description.NO_MATCH;
-        }
-
-        return buildDescription(tree)
-                .addFix(SuggestedFixes.renameVariable(tree, "log", state))
-                .build();
+        return Description.NO_MATCH;
     }
 }
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ConsistentLoggerName.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ConsistentLoggerName.java
@@ -50,12 +50,10 @@ public final class ConsistentLoggerName extends BugChecker implements BugChecker
 
     @Override
     public Description matchVariable(VariableTree tree, VisitorState state) {
-        if (matcher.matches(tree, state)) {
-            if (!tree.getName().contentEquals("log")) {
-                return buildDescription(tree)
-                        .addFix(SuggestedFixes.renameVariable(tree, "log", state))
-                        .build();
-            }
+        if (matcher.matches(tree, state) && !tree.getName().contentEquals("log")) {
+            return buildDescription(tree)
+                    .addFix(SuggestedFixes.renameVariable(tree, "log", state))
+                    .build();
         }
         return Description.NO_MATCH;
     }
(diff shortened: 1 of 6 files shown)


iamdanfox:
Depending on how controversial this is, we might consider auto-fixing this by adding here https://github.com/palantir/gradle-baseline/blob/928e115f5598644f613e2715a8783742759e95f4/gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineErrorProneExtension.java#L29

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ConsistentLoggerName.java
```diff
@@ -0,0 +1,63 @@
...
+@BugPattern(
+        name = "ConsistentLoggerName",
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
+        providesFix = BugPattern.ProvidesFix.REQUIRES_HUMAN_ATTENTION,
+        severity = BugPattern.SeverityLevel.WARNING,
+        summary = "Loggers created using getLogger(Class<?>) must be named 'log'.")
+public final class ConsistentLoggerName extends BugChecker implements BugChecker.VariableTreeMatcher {
+
+    private static final Matcher<VariableTree> matcher = Matchers.allOf(
```
carterkozak, line 41: Thoughts on filtering down to `static final` fields in this case? I think we should add other rules to require loggers to be private+static+final, but in the case of odd legacy code we should flag the most specific failure. I'm not confident we'd want instance loggers to be renamed automatically, I just don't have a good enough sense for all the places they might exist.

carterkozak:
+1 for auto-fixing this, please give the automatic fix a run on our large internal codebase to verify before we merge.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ConsistentLoggerName.java
```diff
@@ -48,16 +50,13 @@
 
     @Override
     public Description matchVariable(VariableTree tree, VisitorState state) {
-        if (!matcher.matches(tree, state)) {
-            return Description.NO_MATCH;
+        if (matcher.matches(tree, state)) {
+            if (!tree.getName().contentEquals("log")) {
```
carterkozak: pedantic nit: might as well combine conditions to reduce indent depth `if (matcher.matches(tree, state) && !tree.getName().contentEquals("log")) {`

I'll give this PR a try internally

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ConsistentLoggerName.java
```diff
@@ -0,0 +1,60 @@
...
+        linkType = BugPattern.LinkType.CUSTOM,
+        providesFix = BugPattern.ProvidesFix.REQUIRES_HUMAN_ATTENTION,
+        severity = BugPattern.SeverityLevel.WARNING,
+        summary = "Loggers created using getLogger(Class<?>) must be named 'log'.")
+public final class ConsistentLoggerName extends BugChecker implements BugChecker.VariableTreeMatcher {
+
+    private static final Matcher<VariableTree> matcher = Matchers.allOf(
+            Matchers.isField(),
+            Matchers.isStatic(),
+            MoreMatchers.isFinal(),
```
carterkozak: I think we can avoid adding a new utility:
```suggestion
            Matchers.hasModifier(Modifier.FINAL),
```

That will match `interface Iface { Logger log = LoggerFactory.getLogger(Iface.class); }` as `final` which is why we had to implement the custom `MoreMatchers.hasExplicitModifier` for our `RedundantModifier` check.

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/ConsistentLoggerNameTest.java
```diff
@@ -0,0 +1,100 @@
...
+
+class ConsistentLoggerNameTest {
+
+    @Test
+    void testFix() {
+        fix().addInputLines(
+                        "Test.java",
+                        "import org.slf4j.*;",
+                        "class Test {",
+                        "    private static final Logger LOG = LoggerFactory.getLogger(Test.class);",
```
uschi2000, line 29: how about `logger` -> `log`? or do you think that's cool?

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/ConsistentLoggerNameTest.java
```diff
@@ -0,0 +1,100 @@
...
+
+    @Test
+    void ignores_local_variables() {
+        fix().addInputLines(
+                        "Test.java",
+                        "import org.slf4j.*;",
+                        "class Test {",
+                        "    private Logger LOG = LoggerFactory.getLogger(Test.class);",
+                        "}")
+                .addOutputLines(
```
carterkozak: Perhaps `.expectUnchanged()`

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/ConsistentLoggerNameTest.java
```diff
@@ -0,0 +1,100 @@
...
+
+    @Test
+    void ignores_non_final_methods() {
+        fix().addInputLines(
+                        "Test.java",
+                        "import org.slf4j.*;",
+                        "class Test {",
+                        "    private static Logger LOG = LoggerFactory.getLogger(Test.class);",
+                        "}")
+                .addOutputLines(
```
carterkozak: `.expectUnchanged()`

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/ConsistentLoggerNameTest.java
ferozco: Everything goes to `log`. There can only be 1 name for loggers

carterkozak (approved):
thanks!

PR #1654, opened 2021-02-23
Fix false positive ClassInitializationDeadlock on `.class` access

carterkozak:
## Before this PR
False positive preventing baseline upgrades in sls-packaging: https://github.com/palantir/sls-packaging/pull/1062

## After this PR
Accessing the class field does not initialize a class.
==COMMIT_MSG==
Fix false positive ClassInitializationDeadlock on `.class` access
==COMMIT_MSG==

## Possible downsides?
There are operations one can invoke upon the `Class` object which will initialize
the referenced class, but those are relatively rare and out of scope for this check.


fawind (approved):
Makes sense to allow this!

PR #1668, opened 2021-03-05
Exclude internal ImmutablesStyle annotations from auto-imports and completion

pkoenig10:
These `ImmutablesStyle` annotations should not be public, but doing so would be a breaking change.

To prevent continued accidental usage, prevent them from being automatically imported or appearing in completion results.

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
@@ -441,7 +441,7 @@ class BaselineIdea extends AbstractBaselinePlugin {
                 <name>org.inferred.freebuilder.shaded</name>
                 <name>org.immutables.value.internal</name>
                 <name>com.palantir.conjure.java.client.config.ImmutablesStyle</name>
-                <name>com.palantir.sls.versionsImmutablesStyle</name>
+                <name>com.palantir.sls.versions.ImmutablesStyle</name>
                 <name>com.palantir.tokens.auth.ImmutablesStyle</name>
               </excluded-names>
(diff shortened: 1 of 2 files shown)


CRogers (approved):
Nice, thanks.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
```diff
@@ -440,6 +440,9 @@ class BaselineIdea extends AbstractBaselinePlugin {
                 <name>autovalue.shaded</name>
                 <name>org.inferred.freebuilder.shaded</name>
                 <name>org.immutables.value.internal</name>
+                <name>com.palantir.conjure.java.client.config.ImmutablesStyle</name>
+                <name>com.palantir.sls.versionsImmutablesStyle</name>
```
CRogers: shouldn't this be `com.palantir.sls.versions.ImmutablesStyle`?

carterkozak:
Can we excavate folks to the canonical internal annotation?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
pkoenig10: Oops, fixed

pkoenig10:
Agree that excavating this change would be ideal.

However it isn't trivial for a couple reasons:
- It may required adding a dependency on our internal `immutables-style` project
- The annotations are not quite equivalent. It probably won't matter in most cases, but it can result in some unintended side-effects.
    - Most notably the `overshadowImplementation` and `get` options are different on the different annotations.

PR #1673, opened 2021-03-05
New error-prone rule: AutoCloseableMustBeClosed

schlosna:
## Before this PR
Methods and constructors of `AutoCloseable` instances are often not annotated with `@MustBeClosed`, so [MustBeClosedChecker](https://errorprone.info/bugpattern/MustBeClosedChecker) does not always do complete analysis to identify potential resource leaks.

## After this PR
==COMMIT_MSG==
A new suggested error-prone rule `AutoCloseableMustBeClosed` annotates methods and constructors that return an `AutoCloseable` type as `@MustBeClosed` to allow for `MustBeClosedChecker` to perform analysis that resources are appropriately closed.

See https://errorprone.info/bugpattern/MustBeClosedChecker
==COMMIT_MSG==

## Possible downsides?
For now this is just a `SUGGESTION` level to test out on several projects.

This may be noisy both in terms of additional annotations as well as requiring manual work to resolve findings flagged by `MustBeClosedChecker` (both valid and false positives where an `AutoCloseable` is passed through layers of abstractions). 

One example that we may wish to exclude is the case of [`java.util.stream.Stream`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/stream/Stream.html) which implements [`AutoCloseable`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/AutoCloseable.html); however, in most cases it is undesirable to clutter stream usages and methods returning `Stream` with try-with-resources, and a limitation of `Stream` terminal . This is limitation is partially covered by the existing [StreamResourceLeak](https://errorprone.info/bugpattern/StreamResourceLeak) checks.

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/AutoCloseableMustBeClosed.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/AutoCloseableMustBeClosed.java
@@ -1,0 +1,77 @@
+/*
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.baseline.errorprone;
+
+import com.google.auto.service.AutoService;
+import com.google.errorprone.BugPattern;
+import com.google.errorprone.BugPattern.SeverityLevel;
+import com.google.errorprone.VisitorState;
+import com.google.errorprone.bugpatterns.BugChecker;
+import com.google.errorprone.bugpatterns.BugChecker.MethodTreeMatcher;
+import com.google.errorprone.fixes.SuggestedFix;
+import com.google.errorprone.fixes.SuggestedFixes;
+import com.google.errorprone.matchers.Description;
+import com.google.errorprone.matchers.Matcher;
+import com.google.errorprone.matchers.Matchers;
+import com.sun.source.tree.MethodTree;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "AutoCloseableMustBeClosed",
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
+        providesFix = BugPattern.ProvidesFix.REQUIRES_HUMAN_ATTENTION,
+        severity = SeverityLevel.SUGGESTION,
+        summary = AutoCloseableMustBeClosed.SUMMARY)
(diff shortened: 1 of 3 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/AutoCloseableMustBeClosed.java
```diff
@@ -0,0 +1,83 @@
...
+        name = "AutoCloseableMustBeClosed",
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
+        providesFix = BugPattern.ProvidesFix.REQUIRES_HUMAN_ATTENTION,
+        severity = SeverityLevel.SUGGESTION,
+        summary = AutoCloseableMustBeClosed.SUMMARY)
+public final class AutoCloseableMustBeClosed extends BugChecker implements MethodTreeMatcher {
+
+    static final String SUMMARY = "If a constructor or method returns an AutoCloseable, "
+            + "it should be annotated @MustBeClosed to ensure callers appropriately close resources";
```
carterkozak: Note: if this is defined on the class-level annotation, it will be automatically used for each finding description unless `.setMessage` is called.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/AutoCloseableMustBeClosed.java
```diff
@@ -0,0 +1,83 @@
...
+    private static final Matcher<MethodTree> methodShouldBeAnnotatedMustBeClosed = Matchers.allOf(
+            Matchers.anyOf(methodReturnsAutoCloseable, constructsAutoCloseable),
+            methodNotAnnotatedMustBeClosed,
+            methodNotAnnotatedIgnoreReturnValue);
+
+    @Override
+    public Description matchMethod(MethodTree tree, VisitorState state) {
+        if (methodShouldBeAnnotatedMustBeClosed.matches(tree, state)) {
+            SuggestedFix.Builder builder = SuggestedFix.builder().addImport(MUST_BE_CLOSED_TYPE);
+            String annotation = SuggestedFixes.qualifyType(state, builder, MUST_BE_CLOSED_TYPE);
```
carterkozak, line 74: `SuggestedFixes.qualifyType(state, builder, MUST_BE_CLOSED_TYPE)` actually adds the import for us, the `addImport` call can be removed.

pkoenig10:
I generally dislike `SUGGESTION` because it clutters the compiler output without forcing me to fix it. Can we make this `WARNING`?

Or do we expect there are repos with `-Werror` that have a large number of occurrences that would require a lot of effort to resolve? In that case I'd prefer they just disable the check until they are ready to address it.

carterkozak:
I think we want to roll this out to a few repos before hitting everything, we can either:

1. Cut an RC to try it out (next baseline release will revert the checck)
2. Disable the check by setting `errorProneOptions.check("AutoCloseableMustBeClosed", CheckSeverity.OFF)` in BaselineErrorProne.java
3. Just roll it out everywhere

I'm curious how it works with one large internal repo in particular, that normally catches the edge cases.

schlosna:
Hey @pkoenig10 as @carterkozak proposed I was mainly thinking of starting it out as a `SUGGESTION` to collect signal/noise ratio on a few projects first, potentially refine exclusions similar to `Stream`, then when its high signal bump it up to `WARNING` so that projects that opt-in to `-Werror` can address findings.

carterkozak (approved):
Lgtm

PR #1685, opened 2021-03-09
AutoCloseableMustBeClosed doesn't match method overrides

carterkozak:
In most cases interactions occur with the base interface, not
the implementation, so we shouldn't flag implementations of
types we cannot control.

## Before this PR
Lots of findings when the check is enabled.

## After this PR
==COMMIT_MSG==
AutoCloseableMustBeClosed doesn't match method overrides
==COMMIT_MSG==

## Possible downsides?
Might miss some resource leaks, however I expect it to substantially improve the snr for rollout.


schlosna (approved):
👍 nice, thanks!

PR #1690, opened 2021-03-17
Baseline applies the `-parameters` javac option for method parameter metadata

carterkozak:
## Before this PR
No parameter metadata.

## After this PR
This produces bytecode with method parameter metadata to support
more powerful reflection features and stronger automated migrations.
==COMMIT_MSG==
Baseline applies the `-parameters` javac option for method parameter metadata
==COMMIT_MSG==

## Possible downsides?
Not sure how to test this. The change is pretty simple though, and we don't need to worry about applying the `-parameters` option twice because `javac` doesn't mind.

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineJavaParameters.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineJavaParameters.java
@@ -1,0 +1,35 @@
+/*
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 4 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineJavaParameters.java
```diff
@@ -0,0 +1,35 @@
...
+ * Applies the {@code -parameters} compiler option to include additional metadata for reflection on method parameters.
+ */
+public final class BaselineJavaParameters implements Plugin<Project> {
+
+    @Override
+    public void apply(Project project) {
+        project.getTasks()
+                .withType(JavaCompile.class)
+                .configureEach(javaCompileTask ->
+                        javaCompileTask.getOptions().getCompilerArgs().add("-parameters"));
```
CRogers, line 33: I looked to see if there was any downsides of doing this and found these: https://stackoverflow.com/a/44075684/139766. None of them seem that bad.

CRogers:
:+1:

PR #1712, opened 2021-04-01
Feature: Add ConsistentOverrides errorprone rule

ferozco:
## Before this PR
We've encountered a few cases where folks would implement an interface with methods with multiple parameters of the same type and inadvertently switch the variable names around resulting in unexpected behaviour. 

## After this PR
==COMMIT_MSG==
Add ConsistentInterfaceImplementation errorprone rule
==COMMIT_MSG==

The rule will catch cases where a sequence of parameters of the same type have the same variable names as the interface but in a different order


carterkozak (dismissed):
As I said earlier: "we should discuss first to iron out edge cases"

ferozco:
For sure! I just put a quick scaffold to start the discussion :)

ferozco:
@carterkozak think this is ready for a look

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ConsistentInterfaces.java
```diff
@@ -0,0 +1,111 @@
...
+                        UNKNOWN_ARG_NAME.matcher(symbol.name.toString()).matches());
+    }
+
+    /**
+     * Returns the first super method of the given method that neither comes from a parameterized type or is itself
+     * parameterized.
+     */
+    private static Optional<MethodSymbol> getNonParameterizedSuperMethod(MethodSymbol methodSymbol, Types types) {
+        ClassSymbol owner = methodSymbol.enclClass();
+        return types.closure(owner.type).stream()
```
carterkozak: Can we simplify this using `ASTHelpers.findSuperMethods`?

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ConsistentInterfaces.java
```diff
@@ -0,0 +1,111 @@
...
+
+                    return buildDescription(tree).addFix(fix).build();
+                })
+                .orElse(Description.NO_MATCH);
+    }
+
+    private static boolean hasMeaningfulArgNames(MethodSymbol methodSymbol) {
+        return !methodSymbol.params().stream()
+                .allMatch(symbol ->
+                        UNKNOWN_ARG_NAME.matcher(symbol.name.toString()).matches());
```
carterkozak: iirc Name is a CharSequence, is toString necessary?
```suggestion
                        UNKNOWN_ARG_NAME.matcher(symbol.name).matches());
```

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ConsistentInterfaces.java
```diff
@@ -0,0 +1,111 @@
...
+                                String superParam =
+                                        superMethod.params.get(i).name.toString();
+                                if (param.equals(superParam)) {
+                                    return Stream.<SuggestedFix>empty();
+                                }
+                                VariableTree paramTree = tree.getParameters().get(i);
+                                return Stream.of(SuggestedFixes.renameVariable(paramTree, superParam, state));
+                            })
+                            .flatMap(Function.identity())
+                            .reduce(SuggestedFix.builder(), SuggestedFix.Builder::merge, SuggestedFix.Builder::merge)
```
carterkozak: I think this would be easier to follow if we used `state.reportMatch` on each parameter rather the complex stream-based implementation. See the `BracesRequired` check as an example.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ConsistentInterfaces.java
```diff
@@ -0,0 +1,111 @@
...
+        providesFix = BugPattern.ProvidesFix.REQUIRES_HUMAN_ATTENTION,
+        severity = SeverityLevel.ERROR,
+        summary = "Sub-types should have variable names consistent with the super-type")
+public final class ConsistentInterfaces extends BugChecker implements MethodTreeMatcher {
+    private static final Pattern UNKNOWN_ARG_NAME = Pattern.compile("^args\\d+$");
+
+    @Override
+    public Description matchMethod(MethodTree tree, VisitorState state) {
+        Symbol sym = ASTHelpers.getSymbol(tree);
+        if (!(sym instanceof MethodSymbol) || sym.isStatic()) {
```
carterkozak: I think we might as well check for private here as well.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ConsistentInterfaces.java
```diff
@@ -0,0 +1,111 @@
...
+            return Description.NO_MATCH;
+        }
+
+        MethodSymbol methodSymbol = (MethodSymbol) sym;
+        return getNonParameterizedSuperMethod(methodSymbol, state.getTypes())
+                .filter(ConsistentInterfaces::hasMeaningfulArgNames)
```
carterkozak: We need to apply this check to the local code (the `methodSymbol` ref) as well, there are still projects which don't apply the javac `-parameters` flag.

carterkozak (commented):
I thought we decided only to require variable names to match the contract for any two parameters that can be switched without breaking compilation, I think that requiring every name to match the interface is unnecessarily strict. You were worried about the StdDeserializer names `p` and `ctxt` which would be flagged by this PR in its current state.

Note that this will also need to retain leading underscores as to allow StrictUnusedVariable to work correctly.

carterkozak:
Any updates, @ferozco?

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/ConsistentOverridesTest.java
```diff
@@ -0,0 +1,167 @@
...
+    void overridden_name() {
+        fix().addInputLines(
+                        "Test.java",
+                        "class Test {",
+                        "  interface Foo {",
+                        "    void doStuff(String foo, String bar);",
+                        "  }",
+                        "  class DefaultFoo implements Foo {",
+                        "    @Override",
+                        "    public void doStuff(String foo, String baz) {}",
```
carterkozak, line 205: Can we add a pair of tests for underscore prefixed names? For example `void doStuff(String foo, String bar)` should be implementable as `@Override public void doStuff(String foo, String _bar)` if `bar` isn't used, while the implementation `@Override public void doStuff(String foo, String _baz)` should be updated to `@Override public void doStuff(String foo, String _bar)`

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ConsistentOverrides.java
```diff
@@ -0,0 +1,122 @@
...
+                        }
+                    });
+                });
+
+        return Description.NO_MATCH;
+    }
+
+    private static boolean hasMeaningfulArgNames(MethodSymbol methodSymbol) {
+        return !methodSymbol.params().stream()
+                .allMatch(symbol -> UNKNOWN_ARG_NAME.matcher(symbol.name).matches());
```
carterkozak: Should we consider single-character names non-meaningful? Otherwise we'll fight checkstyle.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ConsistentOverrides.java
```diff
@@ -0,0 +1,122 @@
...
+        List<? extends VariableTree> paramTrees = tree.getParameters();
+        getNonParameterizedSuperMethod(methodSymbol, state.getTypes())
+                .filter(ConsistentOverrides::hasMeaningfulArgNames)
+                .ifPresent(superMethod -> {
+                    Map<Type, List<ParamEntry>> superParamsByType = IntStream.range(0, paramTrees.size())
+                            .mapToObj(i -> ParamEntry.of(superMethod.params().get(i), i))
+                            .collect(Collectors.groupingBy(ParamEntry::type));
+
+                    superParamsByType.values().forEach(params -> {
+                        if (params.size() <= 1) {
```
carterkozak: Can we verify this on `methodSymbol` before traversing ancestors?

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ConsistentOverrides.java
```diff
@@ -0,0 +1,122 @@
...
+    public Description matchMethod(MethodTree tree, VisitorState state) {
+        Symbol sym = ASTHelpers.getSymbol(tree);
+        if (!(sym instanceof MethodSymbol) || sym.isStatic()) {
+            return Description.NO_MATCH;
+        }
+
+        MethodSymbol methodSymbol = (MethodSymbol) sym;
+        List<? extends VariableTree> paramTrees = tree.getParameters();
+        getNonParameterizedSuperMethod(methodSymbol, state.getTypes())
+                .filter(ConsistentOverrides::hasMeaningfulArgNames)
```
carterkozak: We need to verify the implementation has `-parameters` enabled as well.

ferozco: I'm not sure that's exactly what you want though since it would let terrible names slip by the errorprone rule. I think we should assume that the local build has been configured "correctly" since if it hasn't the user still has agency to fix it

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/ConsistentOverridesTest.java
ferozco: I added a test to demonstrate that it plays nicely with `StrictUnusedVariable`. I avoided actually making `_baz` be changed to `_bar` since it effectively would be duplicating the logic of `StrictUnusedVariable`. Perhaps we say that if a variable is unused then it doesn't really matter what its name is?

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ConsistentOverrides.java
carterkozak: The error message won't give them anything to work with though. If their names are correct, the'll get something along the lines of "You wrote 'void foo(int first, int second)', did you mean 'void foo(int first, int second)'?

Enforcing that the feature is enabled is reasonable, but I don't think this check should be responsible for that enforcement.

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/ConsistentOverridesTest.java
carterkozak: Sounds good to me, in that case I suppose we still enforce that the other variable matches the supertype? Things might get weird if we override a class that defines a method with an unused arg, but we want to use it?

ferozco: yup we still enforce that the first parameter is `foo` regardless of what the second one is. Ya thats a bit funky. I've made it so that if the super param is `_bar` you can name the param `_bar` or `bar`

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ConsistentOverrides.java
ferozco: ok, lets follow up with something in baseline to ensure that `-parameters` is enabled?

carterkozak: Sounds good, #1690 handles most folks who don't opt into individual plugins, but we don't have anything to warn stragglers to enable the plugin.

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/ConsistentOverridesTest.java
carterkozak: :+1:

ferozco:
Builds are breaking because of a bintray brown out that should hopefully end soon

carterkozak:
:+1:

carterkozak:
needs one more +1

ferozco:
👍

PR #1727, opened 2021-04-14
Require meta-annotations with source rentetion for Immutables styles

pkoenig10:
Update `ImmutablesStyleCollision` check to make it stricter by requiring meta-annotations with source retention.

Use inline annotations or meta-annotations with other retention policies forces a consumers to add an Immutables annotations compile dependency to avoid compile warnings.

For more details see https://github.com/immutables/immutables/issues/291 - specifically the recommendation not to use inline styles.
> Also the recommended way for the styles to be applied is via package (possibly even some parent package) or via a custom style annotation (which also may be attached to class or to a package). Inline styles should be used only for experimentation.

We manually made this change in our repos, some examples:
- https://github.com/palantir/tracing-java/pull/691
- https://github.com/palantir/tritium/pull/1006
- https://github.com/palantir/dialogue/pull/1145


pkoenig10:
Would still like a review of this.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ImmutablesStyle.java
```diff
@@ -18,36 +18,51 @@
...
+        name = "ImmutablesStyle",
         linkType = LinkType.CUSTOM,
         link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
-        severity = BugPattern.SeverityLevel.ERROR,
-        summary = "Immutables @Value.Style inline annotation should not be present alongside a Style "
-                + "meta-annotation, as there is no Style merging. You should either modify the "
-                + "meta-annotation, or add the meta-annotation's fields to your inline @Value.Style "
-                + "declaration.")
-public final class ImmutablesStyleCollision extends BugChecker implements BugChecker.ClassTreeMatcher {
+        severity = SeverityLevel.WARNING,
```
pkoenig10, line 38: Changed to WARNING so users have an escape hatch

fawind (approved):
Looks good. Yeah, makes sense to use WARNING initially. Can always raise this in the future if necessary.

PR #1731, opened 2021-04-15
DangerousIdentityKey identifies unsafe map and set keys

schlosna:
## Before this PR
Using a `java.util.regex.Pattern` as a `Map` or `Set` key could lead to unexpected behavior as `Pattern` does not override `equals` & `hashCode`.

## After this PR

==COMMIT_MSG==
DangerousIdentityKey warns where a collection key type does not override equals and hashCode, so comparisons will be done on reference equality only and is likely not what one expects.
==COMMIT_MSG==

Similar to https://errorprone.info/bugpattern/ArrayAsKeyOfSetOrMap & https://errorprone.info/bugpattern/ProtosAsKeyOfSetOrMap

## Possible downsides?

This is not fully exhaustive of all potential cases, but identifies most of the common JDK, Guava, Caffeine types of hash-based Map, Set, Multimap, Multiset, and cache collections.


baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/PatternAsKeyOfSetOrMapTest.java
```diff
@@ -0,0 +1,78 @@
...
+    @Test
+    public void testValidMap() {
+        compilationHelper
+                .addSourceLines(
+                        "Test.java",
+                        "import java.util.*;",
+                        "import java.util.regex.Pattern;",
+                        "class Test {",
+                        "    private Object test() {",
+                        "        return new HashMap<String, Pattern>();",
```
carterkozak: Does/should this pass if we use an IdentityHashMap?

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PatternAsKeyOfSetOrMap.java
```diff
@@ -0,0 +1,42 @@
...
+        summary = "Pattern does not override equals() or hashCode, so comparisons will be done on"
+                + " reference equality only. If neither deduplication nor lookup are needed,"
+                + " consider using a List instead. Otherwise, use IdentityHashMap/Set,"
+                + " or an Iterable/List of pairs.",
+        severity = SeverityLevel.WARNING)
+public final class PatternAsKeyOfSetOrMap extends AbstractAsKeyOfSetOrMap {
+
+    @Override
+    protected boolean isBadType(Type type, VisitorState state) {
+        return ASTHelpers.isSubtype(type, state.getTypeFromString("java.util.regex.Pattern"), state);
```
carterkozak: I wonder if we should introspect `type` to detect any final class which doesn't override equals/hashcode from Object, solving the more general problem?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineErrorProneExtension.java
```diff
@@ -48,6 +48,7 @@
             "ObjectsHashCodeUnnecessaryVarargs",
             "OptionalFlatMapOfNullable",
             "OptionalOrElseMethodInvocation",
+            "PatternAsKeyOfSetOrMap",
```
carterkozak: This line can be removed, it's only necessary for checks we want to automatically apply suggested fixes for. Mind adding a line to the readme describing the check instead?

carterkozak (commented):
One question about whether we want to detect the general problem, no worries if we'd rather punt that for another time.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PatternAsKeyOfSetOrMap.java
schlosna: yeah, solving the general problem would be interesting and we could test against some larger projects to assess signal/noise

carterkozak (approved):

changelog/@unreleased/pr-1731.v2.yml
```diff
@@ -0,0 +1,5 @@
+type: improvement
+improvement:
+  description: PatternAsKeyOfSetOrMap identifies regex Pattern used as map or set key
```
carterkozak:
```suggestion
  description: DangerousIdentityKey identifies key types which do not override equals or hashCode thus rely on reference comparison.
```

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DangerousIdentityKey.java
```diff
@@ -0,0 +1,57 @@
...
+ * Warns that users should not have a {@link java.util.regex.Pattern} as a key to a Set or Map.
+ */
+@BugPattern(
+        name = "DangerousIdentityKey",
+        summary = "Key type does not override equals() and hashCode, so comparisons will be done on"
+                + " reference equality only. If neither deduplication nor lookup are needed,"
+                + " consider using a List instead. Otherwise, use IdentityHashMap/Set,"
+                + " or an Iterable/List of pairs.",
+        severity = SeverityLevel.WARNING)
+public final class DangerousIdentityKey extends AbstractAsKeyOfSetOrMap {
```
schlosna: Should we try to push this upstream to error-prone proper? I'm curious if there is a reason they haven't added something of this shape (though this is the type of thing drilled fairly well by `Effective Java`, I'd like to have 🤖 compiler assistance). 

I can add a test case to confirm that this catches the construction in places like stream collectors, as that's where I noticed one example.

Also worth noting that since we're building on `AbstractAsKeyOfSetOrMap` it is specific to the actual constructed map/set type (not the apparent type of say `Map<Pattern, String>`) and analyzes the typical JDK & Guava (minus `WeakHash(Map|Set)`) hash based maps & sets where `hashCode` and `equals` are relevant. Not sure if we need/want a similar check for Caffeine caches & sorted/navigable map & set types. Reminds me of some places I've seen [`java.net.URL` as keys](http://errorprone.info/bugpattern/URLEqualsHashCode) 😢 .

https://github.com/google/error-prone/blob/c601758e81723a8efc4671726b8363be7a306dce/core/src/main/java/com/google/errorprone/bugpatterns/AbstractAsKeyOfSetOrMap.java#L43-L91

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/DangerousIdentityKeyTest.java
```diff
@@ -0,0 +1,241 @@
...
+                .addSourceLines(
+                        "Test.java",
+                        "import java.util.*;",
+                        "import java.util.regex.Pattern;",
+                        "import java.util.stream.*;",
+                        "class Test {",
+                        "    private Map<Pattern, String> test() {",
+                        "        // BUG: Diagnostic contains: does not override",
+                        "        return Stream.of(\".\").collect(",
+                        "                Collectors.toMap(Pattern::compile, s -> s));",
```
schlosna: right now we don't actually catch this case

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DangerousIdentityKey.java
carterkozak: That's a good point, this should be a strong candidate for an upstream contribution, although I've never had much luck with error-prone contributions, I can reach out again.

We could extend AbstractAsKeyOfSetOrMap to include caches (guava + caffeine) as well as collectors. I can PR something here shortly, and make a similar contribution to the abstract class upstream.

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/DangerousIdentityKeyTest.java
schlosna: https://github.com/palantir/gradle-baseline/pull/1733 will handle this

schlosna:
👍

PR #1733, opened 2021-04-16
Handle additional hashing cases

carterkozak:
==COMMIT_MSG==
Handle additional hashing cases
==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/MoreAbstractAsKeyOfSetOrMap.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/MoreAbstractAsKeyOfSetOrMap.java
@@ -1,0 +1,139 @@
+/*
+ * (c) Copyright 2021 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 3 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/MoreAbstractAsKeyOfSetOrMap.java
```diff
@@ -0,0 +1,139 @@
...
+            Matchers.anyOf(GUAVA_CACHE_BUILDER, CAFFEINE_CACHE_BUILDER);
+
+    private static final Matcher<MethodInvocationTree> HASH_KEYED_COLLECTOR_METHODS = Matchers.anyOf(
+            SET_COLLECTOR,
+            UNMODIFIABLE_SET_COLLECTOR,
+            IMMUTABLE_SET_COLLECTOR,
+            MAP_COLLECTOR,
+            CONCURRENT_MAP_COLLECTOR,
+            UNMODIFIABLE_MAP_COLLECTOR,
+            IMMUTABLE_MAP_COLLECTOR);
```
schlosna, line 91: maybe overkill, but do we want to cover StreamEx `toImmutableMap` & `toImmutableSet` as well (`PreferImmutableStreamExCollections` pushes these when using `StreamEx`)?

schlosna (approved):

schlosna:
@carterkozak defer to you on when you want to merge into the PR branch

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/MoreAbstractAsKeyOfSetOrMap.java
carterkozak: Let's punt those cases for now, I think they're less common than the stdlib/guava/cache cases, and I'm less familiar with StreamEx.

schlosna: Sounds good. We can also check the JDK 9+ `Map.of` & `Set.of` in a follow up .

PR #1750, opened 2021-05-03
Add baseline-immutables plugin to enable incremental compilation for Immutables

pkoenig10:
## Before this PR
Most of our internal projects use the [Immutables](http://immutables.github.io/) annotation processor to generate immutable value objects.

Unfortunately, the Immutables processor does not support incremental compilation by default. This means that even trivial changes to subprojects that use the Immutables processor trigger a full rebuild.

For more details, see https://g.p.b/foundry/multipass/pull/11849.

## After this PR
Fortunately, Immutables has opt-in support for enabling incremental compilation. See https://github.com/immutables/immutables/issues/804.

This PR adds the `baseline-immutables` plugin to enable incremental compilation for source sets using the Immutables processor.

In my testing, I haven't encountered any issues and it dramatically improves compilation times during incremental work, which is fairly typical in dev workflows.

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineImmutables.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineImmutables.java
@@ -1,0 +1,61 @@
+/*
+ * (c) Copyright 2021 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.baseline.plugins;
+
+import java.util.Objects;
+import org.gradle.api.Plugin;
+import org.gradle.api.Project;
+import org.gradle.api.artifacts.Dependency;
+import org.gradle.api.plugins.JavaPluginConvention;
+import org.gradle.api.tasks.SourceSet;
+import org.gradle.api.tasks.compile.JavaCompile;
+
+public final class BaselineImmutables implements Plugin<Project> {
+
+    @Override
+    public void apply(Project project) {
+        project.getPluginManager().withPlugin("java", unused -> {
+            project.afterEvaluate(proj -> {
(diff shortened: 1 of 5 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineImmutables.java
```diff
@@ -0,0 +1,61 @@
...
+import org.gradle.api.tasks.compile.JavaCompile;
+
+public final class BaselineImmutables implements Plugin<Project> {
+
+    @Override
+    public void apply(Project project) {
+        project.getPluginManager().withPlugin("java", unused -> {
+            project.afterEvaluate(proj -> {
+                proj.getConvention().getPlugin(JavaPluginConvention.class).getSourceSets().stream()
+                        .filter(sourceSet -> hasImmutablesProcessor(project, sourceSet))
```
pkoenig10, line 34: This is necessary to avoid compiler warnings about unrecognized annotation processor options.
```
warning: The following options were not recognized by any processor: '[immutables.gradle.incremental]'
1 error
1 warning
```

carterkozak:
Have you verified this internally?

pkoenig10:
Yes, see the internal PR link above.

I've also testing this locally and have not seen any unexpected behavior.

pkoenig10:
Ah sorry, you meant the plugin. Will verify that also.

carterkozak:
Yep, I just want to make sure the plugin works as intended before we ship it.

pkoenig10:
I've verified locally that this works as expected.

carterkozak (approved):

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineImmutables.java
```diff
@@ -0,0 +1,61 @@
...
+import org.gradle.api.plugins.JavaPluginConvention;
+import org.gradle.api.tasks.SourceSet;
+import org.gradle.api.tasks.compile.JavaCompile;
+
+public final class BaselineImmutables implements Plugin<Project> {
+
+    @Override
+    public void apply(Project project) {
+        project.getPluginManager().withPlugin("java", unused -> {
+            project.afterEvaluate(proj -> {
```
CRogers, line 32: Thanks for tracking down this immutables behaviour and building this plugin! I'm excited for the speedup it should give repos.

Just made a follow up PR to remove this use of afterEvaluate - it's technically deprecated (although I don't think Gradle will ever be able to remove it) and the proliferation of afterEvaluates has caused us issues in the past to do with the order they run in. Luckily, we can use some of the newer lazy APIs to avoid this: https://github.com/palantir/gradle-baseline/pull/1752

pkoenig10: Ah sorry, thanks for fixing this! Didn't know that `afterEvaluate` was deprecated. Wasn't sure when this should be applied so I just copied code from an existing plugin.

PR #1752, opened 2021-05-04
Remove afterEvaluate from baseline immutables

CRogers:
## Before this PR
Thanks to @pkoenig10's hard work, we have enabled incremental compilation of immutables classes in #1750! However, it uses `afterEvaluate`, which in my experience nearly always leads to pain down the road when gradle scripts get complicated and multiple `afterEvaluate`s are run with no defined order.

## After this PR
==COMMIT_MSG==
Build compiler args for `com.palantir.baseline-immutables` plugin lazily to avoid `afterEvaluate` ordering issues.
==COMMIT_MSG==

Additionally add a test.

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineImmutables.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineImmutables.java
@@ -16,12 +16,13 @@
 
 package com.palantir.baseline.plugins;
 
+import java.util.Collections;
 import java.util.Objects;
 import org.gradle.api.Plugin;
 import org.gradle.api.Project;
 import org.gradle.api.artifacts.Dependency;
-import org.gradle.api.plugins.JavaPluginConvention;
 import org.gradle.api.tasks.SourceSet;
+import org.gradle.api.tasks.SourceSetContainer;
 import org.gradle.api.tasks.compile.JavaCompile;
 
 public final class BaselineImmutables implements Plugin<Project> {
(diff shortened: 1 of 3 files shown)


carterkozak (approved):
Thanks for adding at test!

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineImmutables.java
```diff
@@ -16,23 +16,33 @@
...
+import org.gradle.api.tasks.SourceSetContainer;
 import org.gradle.api.tasks.compile.JavaCompile;
 
 public final class BaselineImmutables implements Plugin<Project> {
 
     @Override
     public void apply(Project project) {
         project.getPluginManager().withPlugin("java", unused -> {
-            project.afterEvaluate(proj -> {
-                proj.getConvention().getPlugin(JavaPluginConvention.class).getSourceSets().stream()
```
CRogers, line 33: Calling `.stream()` on a `DomainObjectCollection` is pretty much always a bug, as it doesn't include objects added later (and encourages people to use afterEvaluate). I wonder if we should make this an error prone check?

carterkozak: Sounds like a great idea

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineImmutables.java
```diff
@@ -16,23 +16,33 @@
...
-                        .filter(sourceSet -> hasImmutablesProcessor(project, sourceSet))
-                        .forEach(sourceSet -> addImmutablesIncrementalCompilerArg(project, sourceSet));
+            project.getExtensions().getByType(SourceSetContainer.class).configureEach(sourceSet -> {
```
pkoenig10, line 33: Can we use `.matching(...)` here to keep this fluent?

CRogers: Hmm, what would we match on here?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineImmutables.java
CRogers: That said, we do use a `.stream()` below correctly....

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineImmutables.java
pkoenig10: `hasImmutablesProcessor`

pkoenig10: This is what we do in some other plugins.

CRogers: I don't think this will be as lazy as it can be, as we currently check for immutables the moment the compiler args are requested, where as with `.matching` on the source sets would check for immutables the earliest of when each source set is created or when `configureEach` is called.

Also just be aware that there's a lot of really old stuff in this repo and a lot of it is a bit dodgy (but works well enough)

carterkozak: I tend to avoid fluent code because it makes the behavior less obvious to readers, especially as ordering is concerned.

pkoenig10: Cool, make sense. Thanks for the explanation!

PR #1764, opened 2021-05-17
Feature: Allow other plugins to register tasks to produce junit reports

ferozco:
## Before this PR
It was very difficult for custom tasks to leverage gradle-baesline to produce junit XML resulting in errors not being surfaced cleanly at CI.

## After this PR

==COMMIT_MSG==
Allow other plugins to register tasks to produce junit reports
==COMMIT_MSG==

The idea is that another plugin could use `JunitTaskResultExtension` to register a failure reporter for a task they have created


gradle-junit-reports/src/main/java/com/palantir/gradle/junit/JunitTaskResultExtension.java
```diff
@@ -0,0 +1,58 @@
...
+
+    public JunitTaskResultExtension(ObjectFactory objects) {
+        this.taskEntries = objects.namedDomainObjectSet(TaskEntry.class);
+    }
+
+    public final NamedDomainObjectSet<TaskEntry> getTaskEntries() {
+        return taskEntries;
+    }
+
+    public final void registerTask(String taskName, FailuresSupplier failuresSupplier) {
```
CRogers, line 57: Please can we have test for this PR - would also make it more obvious how it's meant to be used/what it will produced.

gradle-junit-reports/src/main/java/com/palantir/gradle/junit/JunitReportsFinalizer.java
```diff
@@ -24,34 +24,38 @@
...
-        finalizer.getReportDir().set(reportDir);
-
-        task.finalizedBy(finalizer);
+            Project project,
+            String taskName,
+            TaskTimer timer,
+            FailuresSupplier failuresSupplier,
+            Provider<Directory> reportDir) {
+        TaskProvider<JunitReportsFinalizer> finalizer = project.getTasks()
+                .register(GUtil.toLowerCamelCase(taskName + " CircleFinalizer"), JunitReportsFinalizer.class, task -> {
```
CRogers: It's probably fine to use this method, but Gradle regards these `G` prefixed classes as internal (despite not being in the `.internal` package - `GFileUtils` is another example) and regularly breaks/removes methods from them in point releases 😱

gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineCircleCiJavaIntegrationTests.java
```diff
@@ -93,7 +93,7 @@ public void junitIntegrationTest() throws IOException {
                 .buildAndFail();
         assertThat(result.getOutput()).contains("2 tests completed, 1 failed");
 
-        File report = new File(reportsDir, "junit/test/TEST-com.example.MyClassTests.xml");
+        File report = new File(reportsDir, "junit/test/TEST-MyClassTests.xml");
```
robert3005: are file names meaningful here? Does this mean that you have a chance of conflict now?

gradle-junit-reports/build.gradle
```diff
@@ -5,12 +5,18 @@ dependencies {
...
 
-  testCompile 'com.google.guava:guava'
-  testCompile 'junit:junit'
-  testCompile 'org.assertj:assertj-core'
-  testCompile 'org.mockito:mockito-core'
+  testImplementation 'com.google.guava:guava'
   testImplementation 'com.netflix.nebula:nebula-test'
+  testImplementation 'org.assertj:assertj-core'
+  testImplementation 'org.junit.jupiter:junit-jupiter'
+  testImplementation 'org.mockito:mockito-core'
```
robert3005, line 18: you have this twice

gradle-junit-reports/src/main/java/com/palantir/gradle/junit/JunitReportsFinalizer.java
```diff
@@ -21,67 +21,64 @@
...
+                            task.getWrappedTaskName().set(taskName);
+                            task.getDurationNanos()
+                                    .set(project.provider(() -> taskTimer.getTaskTimeNanos(wrappedTask.get())));
+                            task.setFailuresSupplier(failuresSupplier);
+                            task.getTargetFile()
+                                    .set(reportDir.map(dir -> dir.file(project.getName() + "-" + taskName + ".xml")));
+                            task.getReportDir().set(reportDir);
+                        });
+
+        wrappedTask.get().finalizedBy(finalizer);
```
robert3005: why both this and the expression below?

gradle-junit-reports/src/main/java/com/palantir/gradle/junit/JunitTaskResultExtension.java
ferozco: done!

gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineCircleCiJavaIntegrationTests.java
ferozco: The names aren't meaningful, but I've fixed it to avoid conflicts

ferozco:
@CRogers @robert3005 this is ready for another look

robert3005:
👍

PR #1771, opened 2021-05-21
WIP Avoid new HashMap(int)

davidscottcohen:
## Before this PR
HashMap(int) has confusing behavior, use Maps.newHashMapWithExpectedSize(int) instead.

## After this PR

==COMMIT_MSG==
==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/AvoidNewHashMapInt.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/AvoidNewHashMapInt.java
@@ -32,7 +32,7 @@ import com.sun.source.tree.NewClassTree;
 @AutoService(BugChecker.class)
 @BugPattern(
         name = "AvoidNewHashMapInt",
-        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks", // TODO (DCohen)
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
         linkType = BugPattern.LinkType.CUSTOM,
         severity = BugPattern.SeverityLevel.WARNING,
         summary = "The HashMap(int) constructor is misleading, use Maps.newHashMapWithExpectedSize(int) instead.")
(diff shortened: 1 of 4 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/AvoidNewHashMapInt.java
```diff
@@ -0,0 +1,73 @@
...
+                ? ""
+                : state.getSourceForNode(tree.getArguments().get(0));
+        String replacement = newType + ".newHashMapWithExpectedSize(" + arg + ")";
+        return buildDescription(tree)
+                .setMessage("The HashMap(int) constructor is misleading: once the HashMap reaches 3/4 of the supplied"
+                        + " size, it will double its internal storage array. Instead use Maps"
+                        + ".newHashMapWithExpectedSize which behaves as expected."
+                        + "See"
+                                + " https://github.com/palantir/gradle-baseline/blob/develop/docs/best-practices/java-coding-guidelines/readme.md#avoid-generics-clutter-where-possible"
+                                + " for more information.")
```
carterkozak: we can move this message to the `@BugPattern` annotation summary, and it will be applied to the fix automatically when we don't call `setMessage`.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/AvoidNewHashMapInt.java
```diff
@@ -0,0 +1,73 @@
...
+    @Override
+    public Description matchNewClass(
+            NewClassTree tree, VisitorState state) {
+        if (!NEW_HASH_MAP.matches(tree, state)) {
+            return Description.NO_MATCH;
+        }
+
+        SuggestedFix.Builder fixBuilder = SuggestedFix.builder();
+        String newType = SuggestedFixes.qualifyType(state, fixBuilder, Maps.class.getName());
+        String arg = tree.getArguments().isEmpty()
```
carterkozak: isEmpty should never evaluate to true for a constructor which takes an `int` argument --  we can either throw, or return `Description.NO_MATCH`. I tend to prefer the latter because we don't want to fail build sin cases that we haven't considered or don't understand.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/AvoidNewHashMapInt.java
```diff
@@ -0,0 +1,73 @@
...
+import com.google.errorprone.fixes.SuggestedFixes;
+import com.google.errorprone.matchers.Description;
+import com.google.errorprone.matchers.Matcher;
+import com.google.errorprone.matchers.method.MethodMatchers;
+import com.sun.source.tree.ExpressionTree;
+import com.sun.source.tree.NewClassTree;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "AvoidNewHashMapInt",
```
carterkozak, line 34: We should add this to the list of checks we automatically fix in `BaselineErrorProneExtension`, that way humans only have to review the change.

carterkozak:
The changelog robot doesn't know how to push to forks, you'll need to add the changelog manually.

Looks like the formatter needs to run as well, you can install the plugin for IntelliJ to update on save, https://plugins.jetbrains.com/plugin/13180-palantir-java-format, or run `./gradlew format`

davidscottcohen:
Closing in favor of branch on the main repo

PR #1773, opened 2021-05-21
Feature/avoid hashmapint

davidscottcohen:
## Before this PR
new HashMap(int) is confusing

## After this PR

==COMMIT_MSG==
==COMMIT_MSG==


carterkozak:
Recreated from #1771

carterkozak (approved):
lgtm, thanks!

pkoenig10:
The intention of this check conflicts with `PreferCollectionConstructors`. It's a bit confusing for devs to have JDK constructors preferred in some cases and Guava static factory methods preferred in others.

carterkozak:
> The intention of this check conflicts with PreferCollectionConstructors. It's confusing for devs to have JDK constructors preferred in some cases and Guava static factory methods preferred in others.

The `hashmap(int)` ctor itself is confusing, and it can cause serious problems when it's used expecting to insert the given number of elements when the value is used to size the hash table instead. That's why the `PreferCollectionConstructors` was implemented specifically to avoid flagging `Maps.newHashMapWithExpectedSize(int)`

PR #1785, opened 2021-06-11
Checkstyle 8.43 JavadocMethod uses accessModifiers instead of scope

schlosna:
## Before this PR

Excavator https://github.com/palantir/gradle-baseline/pull/1762 is [failing on checkstyle](https://app.circleci.com/pipelines/github/palantir/gradle-baseline/2231/workflows/d448475c-df27-43f0-a03f-0015b98bd090/jobs/26202) upgrade because `JavadocMethod` `scope` was removed in favor of `accessModifiers` in Checkstyle 8.42 per https://checkstyle.org/releasenotes.html#Release_8.42 & https://github.com/checkstyle/checkstyle/issues/7417

## After this PR

==COMMIT_MSG==
Checkstyle 8.43 JavadocMethod uses accessModifiers instead of scope

See https://checkstyle.org/releasenotes.html#Release_8.42 & https://github.com/checkstyle/checkstyle/issues/7417
==COMMIT_MSG==

### DIFF
--- a/.baseline/checkstyle/checkstyle.xml
+++ b/.baseline/checkstyle/checkstyle.xml
@@ -405,7 +405,7 @@
             <property name="ignoredAnnotations" value="ParameterizedTest, Test, Before, BeforeEach, After, AfterEach, BeforeClass, BeforeAll, AfterClass, AfterAll"/>
         </module>
         <module name="JavadocMethod"> <!-- Java Style Guide: Where Javadoc is used -->
-            <property name="scope" value="public"/>
+            <property name="accessModifiers" value="public"/>
             <property name="allowMissingParamTags" value="true"/>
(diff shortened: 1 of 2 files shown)


carterkozak (approved):
Thanks!

I may have left the `.baseline` config as is and let the excavator update it, but it's fine either way.

carterkozak:
:robot:

schlosna:
> Thanks!
> 
> I may have left the `.baseline` config as is and let the excavator update it, but it's fine either way.

I had to update the `.baseline` on as its a bit of catch-22 where its using the new checkstyle version, but since the applied config is not compatible, checkstyle fails

carterkozak:
That makes sense, thanks!

PR #1788, opened 2021-06-14
Ban the var keyword via error-prone `VarUsage`

carterkozak:
## Before this PR
`var`?

## After this PR
This will automatically migrate existing uses to the
implementation type used by the `var` keyword.

==COMMIT_MSG==
Ban the var keyword via error-prone `VarUsage`
==COMMIT_MSG==

## Possible downsides?
Folks might use and enjoy the `var` keyword.

I didn't update the readme to enumerate the new error-prone rule. It's already missing a few, and we should borrow the upstream html generator instead.

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/VarUsage.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/VarUsage.java
@@ -1,0 +1,68 @@
+/*
+ * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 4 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/VarUsage.java
```diff
@@ -0,0 +1,68 @@
...
+        for (ErrorProneToken token : state.getOffsetTokensForNode(tree)) {
+            if (token.kind() == TokenKind.IDENTIFIER
+                    && token.hasName()
+                    && token.name().contentEquals("var")) {
+                SuggestedFix.Builder fix = SuggestedFix.builder();
+                return buildDescription(typeTree)
+                        .addFix(fix.replace(
+                                        token.pos(),
+                                        token.endPos(),
+                                        SuggestedFixes.prettyType(state, fix, ASTHelpers.getType(typeTree)))
```
ferozco, line 61: this will suggest using the concrete type instead of the interface which is typically prefered. It would be a ton of work to find the minimum type required of the variable though so its probably fine

ferozco: Maybe we just force the human to input the right type instead of a fix?

carterkozak: I'd prefer to give them a starting point, the reviewer is responsible for ensuring the quality of the replacement. If the replacement quality is too low we can remove the suggestion.

If we don't automate any of it, they're more likely to turn off the check instead of fixing it themselves.

ferozco:
👍

PR #1799, opened 2021-06-17
Implement ProxyNonConstantType

carterkozak:
Proxy instances should be created using constant types known at compile time to allow native-image
behavior to match hotspot. Methods which build proxies should take a 
`Function<InvocationHandler, ? extends T>` instead of arbitrary class references.

The proxy annotation processor can make this process much easier: https://github.com/palantir/proxy-processor

See https://www.graalvm.org/reference-manual/native-image/DynamicProxy/#automatic-detection

==COMMIT_MSG==
Implement ProxyNonConstantType
==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ProxyNonConstantType.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ProxyNonConstantType.java
@@ -1,0 +1,82 @@
+/*
+ * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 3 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ProxyNonConstantType.java
```diff
@@ -0,0 +1,82 @@
...
+import com.sun.source.tree.NewArrayTree;
+import java.lang.reflect.InvocationHandler;
+import java.lang.reflect.Proxy;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "ProxyNonConstantType",
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = LinkType.CUSTOM,
+        severity = SeverityLevel.SUGGESTION,
```
carterkozak, line 41: currently a non-enforcing suggestion

jkozlowski (approved):

carterkozak:
I thought this had merged! :+1:

PR #1810, opened 2021-06-23
ban assert

carterkozak:
## Before this PR
`assert` was allowed.

## After this PR

==COMMIT_MSG==
Introduce `BadAssert` to ban `assert` statements in favor of better alternatives.
==COMMIT_MSG==

## Possible downsides?
May block builds? That's rather the point.

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/BadAssert.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/BadAssert.java
@@ -1,0 +1,86 @@
+/*
+ * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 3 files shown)


carterkozak:
I've intentionally opted against automatically applying the suggested fixes given how rare they are, and some subtle uses (relying on side-effects based on whether or not asserts are enabled) could be broken by automatic changes.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/BadAssert.java
```diff
@@ -0,0 +1,86 @@
...
+                            state.getSourceForNode(condition),
+                            state.getSourceForNode(detail)));
+        } else {
+            String message = isString(ASTHelpers.getType(detail), state)
+                    ? state.getSourceForNode(detail)
+                    : "String.valueOf(" + state.getSourceForNode(detail) + ")";
+            fix.replace(
+                    tree,
+                    String.format(
+                            "if (!%s) { throw new IllegalStateException(%s); }",
```
carterkozak, line 77: These can result in funny code when the input is `assert !value` -> `if (!!value) { throw...`. Not worth optimizing here imo because we have other tools to catch and resolve those, and the fixes are only for documentation, not enabled by default.

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/BadAssertTest.java
```diff
@@ -0,0 +1,112 @@
...
+                        "  }",
+                        "}")
+                .doTest();
+    }
+
+    @Test
+    void testNonConstantStringDescription() {
+        fix().addInputLines(
+                        "Test.java",
+                        // format-hint
```
carterkozak, line 91: formatter puts all these onto one line without a well-placed comment. Some day we'll have text blocks.

CRogers (approved):
Nice!

PR #1816, opened 2021-06-29
Upgrade ProxyNonConstantType to WARNING (from suggestion)

carterkozak:
==COMMIT_MSG==
Upgrade ProxyNonConstantType to WARNING (from suggestion)
==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ProxyNonConstantType.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ProxyNonConstantType.java
@@ -38,7 +38,7 @@ import java.lang.reflect.Proxy;
         name = "ProxyNonConstantType",
         link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
         linkType = LinkType.CUSTOM,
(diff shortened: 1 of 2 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ProxyNonConstantType.java
```diff
@@ -38,7 +38,7 @@
         name = "ProxyNonConstantType",
         link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
         linkType = LinkType.CUSTOM,
-        severity = SeverityLevel.SUGGESTION,
+        severity = SeverityLevel.WARNING,
```
jkozlowski, line 41: Do you know how many repos have hit this?

carterkozak: A few will hit it hard, most don't hit it at all. I suppressed the warning in WC where we hit it on an old retrofit codepath.

jkozlowski (approved):

PR #1817, opened 2021-06-30
Add LogsafeRid error-prone check

IlanaRadinsky:
## Before this PR
`ResourceIdentifier` are not safe to log if their locators contain unsafe data. However, there are many instances where `ResourceIdentifier`s are logged as safe, as `ResourceIdentifier`s are _usually_ safe. We should warn developers before they safe-log their rids.

## After this PR
==COMMIT_MSG==
`LogsafeRid` errorprone check flags instances of `com.palantir.ri.ResourceIdentifier`s being logged as safe.
==COMMIT_MSG==

This is just a `SeverityLevel.WARNING` right now; we'll change the level to `SeverityLevel.ERROR` on the repos we wish to enforce this.

## Possible downsides?
Ideally, the fix would be to replace `SafeArg` with `RidArg` (instead of `UnsafeArg`, as we do in this PR). We can't do this yet, but will see if we can start publishing `RidArg` in a more easily available library so we can recommend usage of it here.

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/LogsafeRid.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/LogsafeRid.java
@@ -30,13 +30,14 @@ import com.google.errorprone.matchers.Matchers;
 import com.google.errorprone.util.ASTHelpers;
 import com.sun.source.tree.ExpressionTree;
 import com.sun.source.tree.MethodInvocationTree;
+import com.sun.tools.javac.code.Type;
 
 @AutoService(BugChecker.class)
 @BugPattern(
         name = "LogsafeRid",
         link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
         linkType = BugPattern.LinkType.CUSTOM,
-        severity = SeverityLevel.SUGGESTION,
+        severity = SeverityLevel.WARNING,
         summary = "Prevent rids from being logged as safe.")
 public final class LogsafeRid extends BugChecker implements MethodInvocationTreeMatcher {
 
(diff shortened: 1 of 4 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/LogsafeRid.java
```diff
@@ -0,0 +1,66 @@
...
+import com.google.errorprone.util.ASTHelpers;
+import com.sun.source.tree.ExpressionTree;
+import com.sun.source.tree.MethodInvocationTree;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "LogsafeRid",
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
+        severity = SeverityLevel.SUGGESTION,
```
pkoenig10: We almost never should use `SUGGESTION` since it creates compiler output noise and doesn't let repo owners opt-in to enforcement. See https://github.com/palantir/gradle-baseline/pull/1380.

`WARNING` is the right level to use here.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/LogsafeRid.java
```diff
@@ -0,0 +1,66 @@
...
+
+    public LogsafeRid() {}
+
+    @Override
+    public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
+        if (!SAFE_ARG_OF.matches(tree, state)) {
+            return Description.NO_MATCH;
+        }
+
+        if (ASTHelpers.getReturnType(tree).toString().contains("com.palantir.ri.ResourceIdentifier")) {
```
pkoenig10: This is a hack, can we check for the argument type instead?

PR #1824, opened 2021-07-12
Add Gradle 7 support

fawind:
## After this PR

Add support for Gradle 7. This raises the minimum required Gradle version to 6.7.

==COMMIT_MSG==
Add Gradle 7 support
==COMMIT_MSG==


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineExactDependencies.java
```diff
@@ -89,15 +89,13 @@ private static void configureSourceSet(
             TaskProvider<CheckImplicitDependenciesParentTask> checkImplicitDependencies) {
         Configuration implementation =
                 project.getConfigurations().getByName(sourceSet.getImplementationConfigurationName());
-        Configuration compile = project.getConfigurations().getByName(sourceSet.getCompileConfigurationName());
```
fawind, line 92: This break is hard to work around. Ideally we would still get the compile configuration as an optional if present. However with Gradle 7, the `getCompileConfigurationName` method no longer exists and the method for deriving the configuration name based on the task is private ([code ref](https://github.com/gradle/gradle/blob/a2ea10896cbdca85a4fb7b0e30d638f57dbe1533/subprojects/plugins/src/main/java/org/gradle/api/internal/tasks/DefaultSourceSet.java#L143-L145)).

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineReleaseCompatibility.java
```diff
@@ -56,8 +56,7 @@ private ReleaseFlagProvider(JavaCompile javaCompile) {
 
         @Override
         public Iterable<String> asArguments() {
-            JavaVersion jdkVersion =
-                    JavaVersion.toVersion(javaCompile.getToolChain().getVersion());
```
fawind, line 60: The `JavaCompile#getToolChain` method got removed and I am not sure what a good replacement is.

I replaced it with `targetCompat` for now but they are not equivalent. Maybe we can use `JavaVersion#current` here?

fawind:
One more issue is
https://github.com/palantir/gradle-baseline/blob/bb93e1c4dff02d51578ff75e7977f4c7d82dd951/gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckExplicitSourceCompatibilityTask.java#L95-L98

The internal API got moved to a different class. Naively, we could fix this with reflection based on the runtime Gradle version but that doesn't sound sustainable. Maybe this is a good time to lean into toolchains?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckExplicitSourceCompatibilityTask.java
```diff
@@ -92,12 +95,7 @@ public final void setShouldFix(boolean value) {
 
     @TaskAction
     public final void taskAction() throws IOException {
-        // We're doing this naughty casting because we need access to the `getRawSourceCompatibility` method.
-        org.gradle.api.plugins.internal.DefaultJavaPluginConvention convention =
```
fawind, line 96: The internal API got moved to a different class. Naively, we could fix this with reflection based on the runtime Gradle version but that doesn't sound sustainable. Maybe this is a good time to lean into toolchains?

robert3005: doesn't java plugin extension exist since gradle 4.10? Maybe we can standardise on that?

fawind: The extension exists since then but the internal method `getRawSourceCompatibility` that we abuse here got moved from the convention to the extension with Gradle 7. So we can't use the extension for versions before that.

I'll do a bit more digging. Still hoping we find a non-internal way of figuring out if sourceCompat is explicitly set. Another alternative would be to actually look into projects gradle files and fail of sourceCompat is not set, similar to the baseline error-prone rules.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckExplicitSourceCompatibilityTask.java
```diff
@@ -127,4 +125,24 @@ public final void taskAction() throws IOException {
...
+            org.gradle.api.plugins.internal.DefaultJavaPluginConvention convention =
+                    (org.gradle.api.plugins.internal.DefaultJavaPluginConvention)
+                            getProject().getConvention().getPlugin(JavaPluginConvention.class);
+            return convention.getRawSourceCompatibility();
+        }
+
+        DefaultJavaPluginExtension extension =
+                (DefaultJavaPluginExtension) getProject().getExtensions().getByType(JavaPluginExtension.class);
+        try {
+            Method rawSourceCompat = DefaultJavaPluginExtension.class.getMethod("getRawSourceCompatibility");
```
CRogers: Grim, but I can't immediately think of way to do this better (since the gradle jar we build against is 6.x, this API does not exist, right?). When this repo upgrades to 7, we'll probably need to switch this up and make the convention version use reflection, right?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/Baseline.java
```diff
@@ -50,11 +50,6 @@ public void apply(Project project) {
             proj.getPluginManager().apply(BaselineTestHeap.class);
             proj.getPluginManager().apply(BaselineJavaParameters.class);
             proj.getPluginManager().apply(BaselineImmutables.class);
-
-            // TODO(dsanduleac): enable this when people's idea{} blocks no longer reference things like
-            //    configurations.integrationTestCompile
-            // proj.getPluginManager().apply(BaselineFixGradleJava.class);
```
CRogers, line 56: lol

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineReleaseCompatibility.java
CRogers: Does something like:

```java
Optional<JavaVersion> taskTarget = Optional.ofNullable(javaCompile
                .getProject()
                .getExtensions()
                .getByType(JavaPluginExtension.class)
                .getToolchain()
                .getLanguageVersion()
                .getOrNull())
        .map(JavaLanguageVersion::asInt)
        .map(JavaVersion::toVersion);
```

replace it exactly? Not sure if it's an exact replacement, but tests seem to pass at least.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineExactDependencies.java
CRogers: Can we just reimplement that logic while we support Gradle 6?

changelog/@unreleased/pr-1824.v2.yml
```diff
@@ -0,0 +1,5 @@
+type: improvement
+improvement:
+  description: Add Gradle 7 support
```
CRogers: Do we require a new minimum gradle version now? If so, should probably make it a break.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineExactDependencies.java
fawind: Yep, its actually quite the simple method. Added it for now.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineReleaseCompatibility.java
fawind: I think `JavaPluginExtension#getToolchain` is just always empty unless your explicitly configure a toolchain in your build.gradle

fawind: I think what we can do is use `getToolchain().getLanguageVersion().orElseGet(JavaVersion::current)`. In this case we would use the toolchain jdk version if specified and otherwise fall back to the global java version. Pushed up this version for now.

Only thing is that the new `getToolchain` methods only exists since Gradle 6.7. We can either use Gradle 6.7 as the new minimum required Gradle version for baseline or do something like this again:
```
if (Gradle 6.7+) { get new toolchain through reflection }
else { get old toolchain through reflection }
```

gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckExplicitSourceCompatibilityTask.java
fawind: Yep, then we have to do reflection for the old method. But this should fail loudly once we upgrade to build with Gradle 7 as the `JavaPluginConvention#getRawSourceCompatibility` can no longer be found.

changelog/@unreleased/pr-1824.v2.yml
fawind: If we decide with [this one](https://github.com/palantir/gradle-baseline/pull/1824/files#r670421739) then we would require Gradle 6.7.

If we avoid this compile dependency e.g. through reflection, I don't think we introduce a new min version.

CRogers:
:+1:

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineFixGradleJava.java
```diff
@@ -1,62 +0,0 @@
...
-import org.gradle.api.tasks.SourceSet;
-import org.gradle.api.tasks.SourceSetContainer;
-
-/**
- * Fixes up java configurations that are left resolvable and consumable for legacy reasons, in order to prepare for
- * Gradle 7 and to showcase code that resolves these when it shouldn't.
- *
- * <p>This will probably break some plugins in the wild, but we'd rather deal with that now than later.
- */
-public final class BaselineFixGradleJava implements Plugin<Project> {
```
schlosna, line 34: we should remove this in the [readme.md](https://github.com/palantir/gradle-baseline#compalantirbaseline-fix-gradle-java-off-by-default) as it is still mentioned and removal of this plugin caused excavator failure https://github.com/palantir/tritium/pull/1147

```
* Where:
Build file '/home/circleci/project/build.gradle' line: 36

* What went wrong:
A problem occurred evaluating root project 'tritium'.
> Plugin with id 'com.palantir.baseline-fix-gradle-java' not found.
```

https://github.com/palantir/gradle-baseline/blob/d2b734a0c6127f710930417d24b615c5cfb1d02f/README.md#L414-L426

fawind: Good flag, will follow up with a PR!

fawind: https://github.com/palantir/gradle-baseline/pull/1833

PR #1838, opened 2021-07-22
Fix formatting of baseline-class-uniqueness.lock files

### LINKED ISSUE #1289: baseline class uniqueness locks are missing a newline between configurations
## What happened?

On a project configured with:

```
checkClassUniqueness {
    configurations.add project.configurations.testRuntimeClasspath
}
```
```
## runtimeClasspath
[com.a:a, com.b:b]
  - com.same.Class## testRuntimeClasspath
[com.a:a, com.b:b]
  - com.same.Class
```

```diff
## runtimeClasspath
[com.a:a, com.b:b]
-  - com.same.Class## testRuntimeClasspath
+  - com.same.Class
+## testRuntimeClasspath
[com.a:a, com.b:b]
  - com.same.Class
```

AlexLandau:
Fixes #1289 

## Before this PR
When the baseline class uniqueness check found issues in more than one configuration, a newline was missing between the previous configuration and the comment naming the next one. See: #1289

## After this PR

==COMMIT_MSG==
Fix the formatting of `baseline-class-uniqueness.lock` files when more than one configuration is listed. A newline was missing. This may require running `./gradlew checkClassUniqueness --write-locks` to update the files.
==COMMIT_MSG==

## Possible downsides?
Requires running the command to write the locks in repos that are affected. (By default only one configuration is set up for the uniqueness check, so the number of affected repos may be small.) I'm not sure if this would be considered a break.

### DIFF
--- a/changelog/@unreleased/pr-1838.v2.yml
+++ b/changelog/@unreleased/pr-1838.v2.yml
@@ -1,0 +1,7 @@
+type: improvement
+improvement:
+  description: Fix the formatting of `baseline-class-uniqueness.lock` files when more
+    than one configuration is listed. A newline was missing. This may require running
+    `./gradlew checkClassUniqueness --write-locks` to update the files.
+  links:
+  - https://github.com/palantir/gradle-baseline/pull/1838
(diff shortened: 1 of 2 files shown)


carterkozak (approved):
Might require write-locks on upgrade, but we already do that.

PR #1843, opened 2021-07-22
fix NPE on unused lambda param in static initializer

dwyand:
## Before this PR
If a lambda with an unused parameter is declared in a static initializer, the `StrictUnusedVariable` check throws an NPE and halts gradle. `Symbol.MethodSymbol#params` is null on static initializers instead of an empty list.

## After this PR
==COMMIT_MSG==
fix NPE on unused lambda param in static initializer
==COMMIT_MSG==

## Possible downsides?
None.


fawind (approved):
Thanks for fixing!

PR #1846, opened 2021-07-23
Implement automatic fixes for `ImmutablesStyle`

carterkozak:
## Before this PR
Lots of blocked baseline-upgrade excavators, which in turn block Gradle 7 rollout.

## After this PR
Autofix the `ImmutablesStyle` check in a naive way. I'm not thrilled with the code we produce, but I think it satisfies the classpath cleanliness concern without requiring manual intervention.
==COMMIT_MSG==
Implement automatic fixes for `ImmutablesStyle`
==COMMIT_MSG==

## Possible downsides?
The error-prone fix api doesn't allow us to create new files, so I've added suppressions for our `OuterTypeFilename` and `OneTopLevelClass` checkstyle rules.
Ideally we'd detect style configurations that match our style library and use that instead.
Ideally we'd aggregate identical style configurations into a single meta-annotation instead of creating one per class.

### DIFF
--- a/baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/ImmutablesStyleTest.java
+++ b/baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/ImmutablesStyleTest.java
@@ -32,6 +32,30 @@ public class ImmutablesStyleTest {
     }
 
     @Test
+    public void fixInlineAnnotation() {
+        fix().addInputLines(
+                        "Person.java",
+                        "import org.immutables.value.Value;",
+                        "@Value.Style(visibility = Value.Style.ImplementationVisibility.PUBLIC)",
+                        "public interface Person {}")
+                .addOutputLines(
+                        "Person.java",
+                        "import java.lang.annotation.ElementType;",
+                        "import java.lang.annotation.Retention;",
+                        "import java.lang.annotation.RetentionPolicy;",
+                        "import java.lang.annotation.Target;",
+                        "import org.immutables.value.Value;",
+                        "@PersonStyle",
+                        "public interface Person {}",
+                        "@Target(ElementType.TYPE)",
+                        "@Retention(RetentionPolicy.SOURCE)",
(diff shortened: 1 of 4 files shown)


pkoenig10:
Another option is to just fix by suppressing warnings for this check.

This check is really targeted at libraries - it matters less for servers. And most of our widely used libraries are well maintained and not blocked on this check.

The check will still prevent new occurrences because we don't apply fixes by default, so that seems like a reasonable compromise. Especially given that this check has been out for almost 2 months now.

carterkozak:
That's fair, I've updated the top-level-class case to suppress, while nested classes still get a new meta-annotation

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/ImmutablesStyleTest.java
```diff
@@ -31,6 +31,49 @@ public void testInlineAnnotation() {
...
+                        "import java.lang.annotation.ElementType;",
+                        "import java.lang.annotation.Retention;",
+                        "import java.lang.annotation.RetentionPolicy;",
+                        "import java.lang.annotation.Target;",
+                        "import org.immutables.value.Value;",
+                        "public class Enclosing {",
+                        "  @Target(ElementType.TYPE)",
+                        "  @Retention(RetentionPolicy.SOURCE)",
+                        "  @Value.Style(visibility = Value.Style.ImplementationVisibility.PUBLIC)",
+                        "  @interface PersonStyle {}",
```
pkoenig10, line 70: Can this be private? Or does that not work with immutables?

pkoenig10 (approved):

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/ImmutablesStyleTest.java
carterkozak: It would fail to compile when encapsulated within an interface, in that case the annotation will be public. I’m not terribly worried about that case because it won’t have any runtime impact, only build time.

PR #1848, opened 2021-07-23
auto-apply MethodCanBeStatic

iamdanfox:
## Before this PR

I frequently make a code-review suggestion that a method could be static. It occurred to me that this should probably be automated, and indeed Google's error-prone already includes https://errorprone.info/bugpattern/MethodCanBeStatic - but this is a SUGGESTION by default.

I tried it out on an internal project (AS) by running `./gradlew classes -PerrorProneApply=MethodCanBeStatic` and it produced a +135-137 line diff. https://github/foundry/a-s/pull/2915

## After this PR
==COMMIT_MSG==
Automatically apply MethodCanBeStatic
==COMMIT_MSG==

## Possible downsides?
- some lambda references become slightly more verbose, which might trip one-liners onto multiple lines:

```diff
- .map(this::asPredicate)
+.map(InMemoryTestCatalog::asPredicate)
```
- making lots of java changes would mean baseline excavators require approval, which might slow down adoption
- I don't want people to think that static methods are preferred to instance methods in _all_ cases (because they don't really allow dependency injection, interfaces, encapsulation etc)

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/CompileTimeConstantViolatesLiskovSubstitution.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/CompileTimeConstantViolatesLiskovSubstitution.java
@@ -95,7 +95,7 @@ public final class CompileTimeConstantViolatesLiskovSubstitution extends BugChec
         return Description.NO_MATCH;
     }
 
-    private boolean anySuperMethodsMissingParameterAnnotation(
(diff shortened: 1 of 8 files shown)


carterkozak:
I think this is reasonable, although I think it's best if we roll out a temporary excavator to apply these suggested fixes prior to coupling it with baseline upgrades.

robert3005:
I think the excavator has been out in the wild for a while - how about merging this?

carterkozak:
The excavator was reverted because it broke things. See excavator-gradle#2559

PR #1851, opened 2021-07-23
Fix PreferSafeLogger edge case that produced non-compiling code

carterkozak:
In most cases we wouldn't go this far to ensure green builds, these
are easy enough to manually update. However, it's important that
automation can roll out the new facade without blocking for devs
to make changes, and we can follow-up with a strict check later
on to migrate the rest.

==COMMIT_MSG==
Fix PreferSafeLogger edge case that produced suggested fixes that didn't compile without human interaction.
==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/PreferSafeLoggerTest.java
+++ b/baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/PreferSafeLoggerTest.java
@@ -210,4 +210,48 @@ class PreferSafeLoggerTest {
                 .expectUnchanged()
                 .doTest();
     }
+
+    @Test
+    void testGetNameInArgUnmodified() {
+        // getName is not supported by SafeLogger, so we shouldn't make changes that won't compile.
+        fix().addInputLines(
+                        "Test.java",
+                        "import com.palantir.logsafe.*;",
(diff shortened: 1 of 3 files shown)


baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/PreferSafeLoggerTest.java
```diff
@@ -210,4 +210,48 @@ void testPassedLoggerUnmodified() {
...
+    void testGetNameInArgUnmodified() {
+        // getName is not supported by SafeLogger, so we shouldn't make changes that won't compile.
+        fix().addInputLines(
+                        "Test.java",
+                        "import com.palantir.logsafe.*;",
+                        "import org.slf4j.*;",
+                        "class Test {",
+                        "  private static final Logger log = LoggerFactory.getLogger(Test.class);",
+                        "  void action() {",
+                        "    log.info(\"foo\", SafeArg.of(\"name\", log.getName()));",
```
carterkozak, line 224: Previously the `log.getName()` method invocation hid in the shadow of the parent log.info node, so the recommended fix wouldn't compile.

tpetracca (approved):

PR #1863, opened 2021-08-06
recommend witchcraft logging

carterkozak:
==COMMIT_MSG==
Recommend the Witchcraft logging plugin when the sls-packaging distribution plugin is detected
==COMMIT_MSG==

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
@@ -132,6 +132,7 @@ class BaselineIdea extends AbstractBaselinePlugin {
                 Node node = provider.asNode()
                 recommendWitchcraftLoggingPlugin(node)
             }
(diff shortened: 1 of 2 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
```diff
@@ -119,6 +120,29 @@ class BaselineIdea extends AbstractBaselinePlugin {
...
+                Node node = provider.asNode()
+                recommendWitchcraftLoggingPlugin(node)
+            }
+            recommendWitchcraftLoggingPluginForIntellijImport(rootProject)
+        }
+    }
+
+    private static boolean hasJavaServiceDistributionPluginLoaded() {
+        try {
+            Class.forName('com.palantir.gradle.dist.service.JavaServiceDistributionPlugin')
```
carterkozak, line 141: Won't work using the `plugins {}` dsl, but we don't use that at this point.

carterkozak:
I've manually verified that this functions as expected for both gradle integration and ipr workflows.

fawind (approved):

PR #1865, opened 2021-08-09
Add class to generate list of default errorprone checks.

NickHolden:
## Before this PR
In pgdev we had to manually list all of the baseline errorprone checks that we wanted to enable. This meant that we were not automatically using new checks as they are developed.

## After this PR
We now have a way to populate a list of al baseline errorprone checks by not relying on a static list.

==COMMIT_MSG==
==COMMIT_MSG==


carterkozak:
How will this be called? I believe `gradle-baseline-java` (the plugin) doesn't directly depend on `baseline-error-prone`, so this utility wouldn't be reachable without a buildscript dependency onto the error-prone checks jar.

bjlaub:
> How will this be called? I believe `gradle-baseline-java` (the plugin) doesn't directly depend on `baseline-error-prone`, so this utility wouldn't be reachable without a buildscript dependency onto the error-prone checks jar.

@carterkozak yeah, it's not great. For our use case, we were already doing this with `error_prone_core` as well so I was willing to accept that as an inevitability, but it's not ideal for the general use case.

Thoughts on exposing `DefaultChecks.getEnabledErrors` via the extra properties extension?

PR #1879, opened 2021-08-15
JavaDoc uses UTF-8 encoding

schlosna:
## Before this PR

See https://github.com/palantir/tritium/issues/1192 where JavaDoc is failing to generate on system using default US-ASCII encoding

## After this PR

==COMMIT_MSG==
JavaDoc uses UTF-8 encoding by default
==COMMIT_MSG==


carterkozak (approved):
lgtm, thanks!

PR #1884, opened 2021-08-18
Fix missing setProperty error when using Gradle 6

fawind:
## Before this PR
With https://github.com/palantir/gradle-baseline/pull/1875, baseline is build using Gradle 7 and transitively Groovy 3.x.
As a result, projects that still use Gradle 6 and thus Groovy 2.x fail with the following error when applying the `BaselinEclipse` or `BaselineIdea` plugin (both are the only plugins that are still Groovy and not Java classes) ([example](https://app.circleci.com/pipelines/github/palantir/palantir-java-format/1637/workflows/66e85fa7-da35-47cf-8ec0-5a74b6f09e12/jobs/9879)):

```
Caused by: java.lang.AbstractMethodError: Receiver class com.palantir.baseline.plugins.BaselineEclipse does not define or inherit an implementation of the resolved method 'abstract void setProperty(java.lang.String, java.lang.Object)' of interface groovy.lang.GroovyObject.
	at com.palantir.baseline.plugins.BaselineEclipse.apply(BaselineEclipse.groovy:69)
	at com.palantir.baseline.plugins.BaselineEclipse.apply(BaselineEclipse.groovy)
	at org.gradle.api.internal.plugins.ImperativeOnlyPluginTarget.applyImperative(ImperativeOnlyPluginTarget.java:43)
```

Haven't gotten completely on the bottom of this but the issue seems to be that going from Groovy 2.x to 3.x `GroovyObjectSupport` is no longer provided per default.

Relevant issue: https://issues.apache.org/jira/browse/GROOVY-4855
 
## After this PR

Adding `extends GroovyObjectSupport` adds default implementations of Groovy methods including `setProperty`.

Tested it on two internal repos and compiled again.

==COMMIT_MSG==
Fix missing setProperty error when using Gradle 6
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-1884.v2.yml
+++ b/changelog/@unreleased/pr-1884.v2.yml
@@ -1,0 +1,5 @@
+type: fix
+fix:
+  description: Fix missing setProperty error when using Gradle 6
+  links:
+  - https://github.com/palantir/gradle-baseline/pull/1884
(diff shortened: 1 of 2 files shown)


CRogers (approved):

fawind:
~~Want to sanity double check this on CI quickly~~ 👍

PR #1886, opened 2021-08-18
Implement LoggerInterpolationConsumesThrowable

carterkozak:
Flag as an ERROR when logging statements fail to produce a
stack trace due to excessive slf4j-style interpolation
placeholders.

==COMMIT_MSG==
error-prone LoggerInterpolationConsumesThrowable: Flag as an ERROR when logging statements fail to produce a stack trace due to excessive slf4j-style interpolation placeholders.
==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/LoggerInterpolationConsumesThrowable.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/LoggerInterpolationConsumesThrowable.java
@@ -1,0 +1,109 @@
+/*
+ * (c) Copyright 2017 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/LoggerInterpolationConsumesThrowable.java
```diff
@@ -0,0 +1,109 @@
...
+        }
+
+        String literalMessage = (String) ((LiteralTree) messageArg).getValue();
+        int stringPlaceholders = countPlaceholders(literalMessage);
+        int nonThrowableParameters = throwableIndex - messageIndex - 1;
+        if (stringPlaceholders <= nonThrowableParameters) {
+            return Description.NO_MATCH;
+        }
+        int extraPlaceholders = stringPlaceholders - nonThrowableParameters;
+        return buildDescription(tree)
```
carterkozak, line 85: There are a few common patterns we could likely auto-fix, but enough cases we can't that I haven't attempted to implement one.

If we find a lot of failures ending with ` {}` or `: {}` it may be reasonable to automatically update them.

CRogers: Might be interesting to track how many excavators fail to upgrade with this check, given we can't autofix them

CRogers (approved):
nice!

PR #1890, opened 2021-08-19
Java compilation warnings and errors are no longer limited to the first 100

carterkozak:
==COMMIT_MSG==
Java compilation warnings and errors are no longer limited to the first 100
==COMMIT_MSG==

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineJavaCompilerDiagnostics.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineJavaCompilerDiagnostics.java
@@ -1,0 +1,52 @@
+/*
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineJavaCompilerDiagnostics.java
```diff
@@ -0,0 +1,52 @@
...
+
+    private static final String MAX_WARNINGS_ARG = "-Xmaxwarns";
+    private static final String MAX_ERRORS_ARG = "-Xmaxerrs";
+
+    @Override
+    public void apply(Project project) {
+        project.getTasks().withType(JavaCompile.class).configureEach(javaCompileTask -> {
+            List<String> compilerArgs = javaCompileTask.getOptions().getCompilerArgs();
+            // Avoid overriding options that have already been set
+            if (!compilerArgs.contains(MAX_WARNINGS_ARG)) {
```
CRogers: Would we want this to work if people override after applying the plugin too? If so, we'd probably want to put this in a `project.afterEvaluate` to run after people have had a chance to change it.

carterkozak: I think it does, but only due to the fact later arguments take precedence over earlier instances. The afterEvaluate will be cleaner :-)
Note that the same potential duplication can occur if commandlineargumentproviders are used because I'm not checking those, we take advantage of knowing they're applied after the fixed args. I figured we don't want to evaluate providers unnecessarily as part of this check due to potential side-effects given the worst case result still has the desired behavior.

CRogers (approved):

PR #1904, opened 2021-09-02
allow trailing comma in annotation arrays

rzpt:
## Before this PR
```
@JsonIgnoreProperties({
         "foo",
         "bar", // This comma is not allowed
})
```
It's better to allow it to minimize diffs.

## After this PR
==COMMIT_MSG==
allow trailing comma in annotation arrays
==COMMIT_MSG==

## Possible downsides?

None.

### DIFF
--- a/changelog/@unreleased/pr-1904.v2.yml
+++ b/changelog/@unreleased/pr-1904.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: allow trailing comma in annotation arrays
+  links:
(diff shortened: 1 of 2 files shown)


carterkozak:
I think this is reasonable, it would certainly make diffs in excavator-gradle cleaner (roomba-lint additions come to mind). I'm going to give folks on the team an hour or two to opine before I merge.

carterkozak (approved):
discussed, all in favor.

PR #1911, opened 2021-09-10
Remove baseline-scalastyle plugin

fawind:
## Before this PR

We were using the [ngbinh/gradle-scalastyle-plugin](https://github.com/ngbinh/gradle-scalastyle-plugin) for scalastyle integration which has not been updated for multiple years. In addition to that, the original scalastyle plugin has also not been maintained in a while (https://github.com/scalastyle/scalastyle/issues/360).

The plugin is also blocking the Gradle 7 adoption of our scala repos due to incompatibilities.

Note that there are some more recently maintained forks (e.g. [alisiikh/gradle-scalastyle-plugin](https://github.com/alisiikh/gradle-scalastyle-plugin) and [beautiful-scala/scalastyle](https://github.com/beautiful-scala/scalastyle)). However internally, only a few projects are using it and we are getting limited value out of scalastyle.

Hence we are dropping scalastyle plugin for now. 

## After this PR
==COMMIT_MSG==
Remove baseline-scalastyle plugin
==COMMIT_MSG==

* Marking this as a break because consumers could apply the plugin explicitly. However I couldn't find any explicit usage in Palantir owner repos.
* As a follow up, I'll update the "update-baseline" excavator to also remove the scalastyle config when updating to this version.

cc @robert3005, @rshkv


fawind:
Tagging 'do-not-merge' as I want to get the excavator lined up before merging.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineScalastyle.java
```diff
@@ -1,111 +0,0 @@
...
-import org.gradle.api.tasks.scala.ScalaCompile;
-import org.gradle.plugins.ide.idea.model.IdeaModel;
-
-public final class BaselineScalastyle extends AbstractBaselinePlugin {
-    private static final String SCALA_TARGET_VERSION = "jvm-1.8";
-
-    @Override
-    public void apply(Project project) {
-        this.project = project;
-        project.getPluginManager().withPlugin("scala", plugin -> {
```
robert3005, line 42: I think you still want the intellij config here :/

fawind: You're right. Only skimmed this class :/

Let's continue with your PR: https://github.com/palantir/gradle-baseline/pull/1912

PR #1912, opened 2021-09-10
Feature: Remove scalastyle and rename BaselineScalastyle to BaselineScala

robert3005:
## After this PR

==COMMIT_MSG==
Scalastyle is not maintained anymore and config we had wasn't that useful anyway.
==COMMIT_MSG==


fawind (approved):
Look good, thanks for adding!

PR #1920, opened 2021-09-16
BaselineJavaVersions consistent JDK control via gradle toolchains

carterkozak:
## Goal

Root of each project declares java versions in this form:

```gradle
javaVersions {
    libraryTarget = 11
    distributionTarget = 15
}
```

or to target 15 but run using the new 17 release:
```gradle
javaVersions {
    libraryTarget = 11
    distributionTarget = 15
    runtime = 17
}
```

As opposed to setting `sourceCompatibility` and `targetCompatibility` on a per-project basis. We can detect which modules are libraries vs distributions and do the right things. Rolling out new java releases becomes much simpler as we can update this single block instead of hunting through every gradle file. Once this is backed by resolvable toolchains we'll no longer need to sequence CI template updates with services trying new releases.

Needs test coverage.
Eventually will use toolchains, unsure if we should implement that in the initial implementation, or later (likely later so implementation can occur in parallel to config rollout)
Needs support for toolchain configuration (custom jdk repositories for internal things?).

Currently the `runtime` version is unused. Eventually it will be used for JavaExec/Test/distributions.

## After this PR

==COMMIT_MSG==
BaselineJavaVersions consistent JDK control via gradle toolchains on an opt-in basis
==COMMIT_MSG==

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineJavaVersionExtension.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineJavaVersionExtension.java
@@ -1,0 +1,54 @@
+/*
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.baseline.extensions;
+
+import javax.inject.Inject;
+import org.gradle.api.Project;
+import org.gradle.api.provider.Property;
+import org.gradle.jvm.toolchain.JavaLanguageVersion;
+
+public class BaselineJavaVersionExtension {
+
+    private final Property<JavaLanguageVersion> target;
+    private final Property<JavaLanguageVersion> runtime;
+
+    @Inject
+    public BaselineJavaVersionExtension(Project project) {
+        target = project.getObjects().property(JavaLanguageVersion.class);
+        runtime = project.getObjects().property(JavaLanguageVersion.class);
+        // Runtime defaults to the target value
+        runtime.set(target);
+    }
+
+    /** Target {@link JavaLanguageVersion} for compilation. */
+    public final Property<JavaLanguageVersion> target() {
+        return target;
+    }
+
+    public final void setTarget(JavaLanguageVersion value) {
+        target.set(value);
+    }
+
+    /** Runtime {@link JavaLanguageVersion} for testing and distributions. */
+    public final Property<JavaLanguageVersion> runtime() {
+        return runtime;
+    }
+
+    public final void setRuntime(JavaLanguageVersion value) {
+        runtime.set(value);
+    }
+}
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineJavaVersionExtension.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineJavaVersionExtension.java
@@ -21,6 +21,10 @@ import org.gradle.api.Project;
 import org.gradle.api.provider.Property;
 import org.gradle.jvm.toolchain.JavaLanguageVersion;
 
+/**
+ * Extension named {@code javaVersion} used to set the
+ * target and runtime java versions used for a single project.
+ */
 public class BaselineJavaVersionExtension {
 
     private final Property<JavaLanguageVersion> target;
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineJavaVersionExtension.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineJavaVersionExtension.java
@@ -34,8 +34,6 @@ public class BaselineJavaVersionExtension {
     public BaselineJavaVersionExtension(Project project) {
         target = project.getObjects().property(JavaLanguageVersion.class);
         runtime = project.getObjects().property(JavaLanguageVersion.class);
-        // Runtime defaults to the target value
-        runtime.convention(target);
     }
 
     /** Target {@link JavaLanguageVersion} for compilation. */
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineJavaVersionExtension.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineJavaVersionExtension.java
@@ -31,7 +31,7 @@ public class BaselineJavaVersionExtension {
         target = project.getObjects().property(JavaLanguageVersion.class);
         runtime = project.getObjects().property(JavaLanguageVersion.class);
         // Runtime defaults to the target value
-        runtime.set(target);
+        runtime.convention(target);
     }
 
     /** Target {@link JavaLanguageVersion} for compilation. */
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineJavaVersionExtension.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineJavaVersionExtension.java
@@ -43,8 +43,8 @@ public class BaselineJavaVersionExtension {
         return target;
     }
 
-    public final void setTarget(JavaLanguageVersion value) {
-        target.set(value);
+    public final void setTarget(int value) {
+        target.set(JavaLanguageVersion.of(value));
     }
 
     /** Runtime {@link JavaLanguageVersion} for testing and distributions. */
@@ -52,7 +52,7 @@ public class BaselineJavaVersionExtension {
         return runtime;
     }
 
-    public final void setRuntime(JavaLanguageVersion value) {
-        runtime.set(value);
+    public final void setRuntime(int value) {
+        runtime.set(JavaLanguageVersion.of(value));
     }
 }
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineJavaVersionsExtension.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineJavaVersionsExtension.java
@@ -1,0 +1,88 @@
+/*
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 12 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineJavaVersionExtension.java
```diff
@@ -0,0 +1,54 @@
...
+ */
+
+package com.palantir.baseline.extensions;
+
+import javax.inject.Inject;
+import org.gradle.api.Project;
+import org.gradle.api.provider.Property;
+import org.gradle.jvm.toolchain.JavaLanguageVersion;
+
+public class BaselineJavaVersionExtension {
```
carterkozak, line 28: naming is hard: plural 'versions' applies to the root project and sets up other projects, singular 'version' is applied to all java projects.

carterkozak:
Discussed elsewhere, documenting here for visibility: I think my attempt to apply the new configuration to both `sourceCompatibility`/`targetCompatibility` _and_ toolchains was a mistake, it's proving difficult to support both modes at once.

New plan is to provide a new non-default plugin which converts our config into toolchain configurations across all modules. When this is applied we can require that there are no `sourceCompatibility` or `targetCompatibility` declarations.

carterkozak:
todo: lock in toolchain jdk build numbers through another dependency for reproducible builds.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineJavaVersionExtension.java
fawind: Feel ya. Don't have a concrete better name. Would add a small Javadoc to both extensions regarding the difference.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineJavaVersionsExtension.java
```diff
@@ -0,0 +1,82 @@
...
+
+    @Inject
+    public BaselineJavaVersionsExtension(Project project) {
+        this.libraryTarget = project.getObjects().property(JavaLanguageVersion.class);
+        this.distributionTarget = project.getObjects().property(JavaLanguageVersion.class);
+        this.runtime = project.getObjects().property(JavaLanguageVersion.class);
+        // distribution defaults to the library value
+        distributionTarget.convention(libraryTarget);
+        // runtime defaults to the distribution value
+        runtime.convention(distributionTarget);
```
fawind, line 42: For both extensions, I think we should add some preconditions to limit possible configuration options. I.e. `libraryTarget` is not allowed to be larger than `distribution` or `runtime`?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineJavaVersion.java
```diff
@@ -0,0 +1,113 @@
...
+
+            // Set the default project toolchain
+            javaPluginExtension.toolchain(new Action<JavaToolchainSpec>() {
+                @Override
+                public void execute(JavaToolchainSpec javaToolchainSpec) {
+                    javaToolchainSpec.getLanguageVersion().set(extension.runtime());
+                }
+            });
+
+            project.getTasks().withType(JavaCompile.class, new Action<JavaCompile>() {
```
fawind: We might also need to configure the compile tasks for groovy and potentially scala. But can also track this as a flup.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineJavaVersions.java
```diff
@@ -0,0 +1,77 @@
...
+        project.allprojects(proj -> proj.getPluginManager().withPlugin("java", unused -> {
+            proj.getPluginManager().apply(BaselineJavaVersion.class);
+            BaselineJavaVersionExtension projectVersions =
+                    proj.getExtensions().getByType(BaselineJavaVersionExtension.class);
+            projectVersions
+                    .target()
+                    .set(proj.provider(() -> isLibrary(proj)
+                            ? rootExtension.libraryTarget().get()
+                            : rootExtension.distributionTarget().get()));
+        }));
```
fawind, line 53: Can contribute or do a FLUP but I imagine it would be helpful for debugging if we had a task that prints the final configured java versions for each project.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineJavaVersion.java
carterkozak: Good shout, might as well include it in this change

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineJavaVersions.java
carterkozak: +1, as well as whether or not the component is considered a library

carterkozak: I don't know as much as I probably should about gradle plugins, I think the `set(` on line 47 should probably be `convention(` to allow it to be explicitly overridden in the subproject, right?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineJavaVersionsExtension.java
carterkozak: Agreed, how is that normally handled within an extension? I don't want to force an order that the two must be updated in config. I think it's only when `get` is called on a property that we need consistency.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineJavaVersions.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.Objects;
+import org.gradle.api.GradleException;
+import org.gradle.api.Plugin;
+import org.gradle.api.Project;
+import org.gradle.api.publish.Publication;
+import org.gradle.api.publish.PublishingExtension;
+import org.gradle.api.publish.ivy.IvyPublication;
+import org.gradle.api.publish.maven.MavenPublication;
+
+public final class BaselineJavaVersions implements Plugin<Project> {
```
fawind, line 32: I might be missing it but we somehow also still need to configure some of the `source/target/release` fields, right as these are still used by other consumers, e.g. sls-packaging here:

https://github.com/palantir/sls-packaging/blob/c44f84cc236e3e78b94f26d3b2e30bbe9c86ae27/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/JavaServiceDistributionExtension.java#L61-L63

Not super sure though wether this plugin should set them or if we should update downstream plugins to get the java version from this new extension if present.

carterkozak: I should re-verify as I might have this a bit wrong, but I believe the `getSourceCompatibility()` and `getTargetCompatibility()` methods are computed -- we're no longer setting the raw input which is checked by `getRawSourceCompatibility`, but I _think_ the getters should read from our the configured toolchain.

carterkozak: I've put together a test to verify that this works. One oddity is that due to the way I've applied toolchains using the `runtime` toolchain as the project default, `getTargetCompatibility()` returns the runtime version, not the `libraryTarget` or `distributionTarget`. For the case you've linked, that's actually exactly what we want, but I'm not sure if that's always true.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineJavaVersion.java
carterkozak: Updated with code for groovy and scala, lacks test coverage for now though

carterkozak:
I need to figure out something for environments where os detection isn't supported (groovy ci) and I imagine the solution will have overlap with setting more specific internal jdk resolvers.

Still need to add test coverage for running tests across multiple java versions.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineJavaVersions.java
carterkozak: Logging the final value is difficult because the properties are meant to be lazy, it's not clear that the value at any time I log it is the value used to execute tasks.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineJavaVersionsExtension.java
carterkozak: I've figured this one out: I added a new task which `check` depends on, to verify java versions at task execution time.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineJavaVersions.java
carterkozak: @CRogers This is the thing we were discussing this morning.

This snippet in sls-packaging will return the java feature version (e.g. 11) of the project-level default toolchain:
https://github.com/palantir/sls-packaging/blob/c44f84cc236e3e78b94f26d3b2e30bbe9c86ae27/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/JavaServiceDistributionExtension.java#L61-L63

We need to add some code to JavaServiceDistributionExtension to detect the runtime version in some way, perhaps with a direct dependency into baseline to read the extension?

carterkozak:
I think this is ready to start trying out. Mind giving it a review?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineJavaVersionsExtension.java
fawind: Nice, thats a good solution!

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineJavaVersions.java
fawind: That's true :/ Maybe we can add some (debug) logging to the check task? Haven't played around with this much so not sure how good observability is when you retroactively want to figure out what java version a task in CI was run with. But don't think its super important as this should be visible by looking at the gradle build files and we can add more verbose logging once we actually run into issue that we can't easily debug.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineReleaseCompatibility.java
```diff
@@ -58,14 +58,16 @@ private ReleaseFlagProvider(JavaCompile javaCompile) {
 
         @Override
         public Iterable<String> asArguments() {
-            JavaVersion jdkVersion = getJdkVersion(javaCompile);
-
-            if (!supportsReleaseFlag(jdkVersion)) {
+            if (javaCompile.getProject().getPlugins().hasPlugin(BaselineJavaVersion.class)) {
+                return Collections.emptyList();
```
fawind, line 67: nit: Can we add a log here that this is also a no-op because BaselineJavaVersion is used?

fawind (approved):
Very nice! Lets try it out on some repos 🚀

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineJavaVersions.java
carterkozak: Good idea

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineReleaseCompatibility.java
carterkozak: :+1: updated

PR #1921, opened 2021-09-16
[tests only] Migrate tests off JCenter to Maven Central

carterkozak:
==COMMIT_MSG==
Migrate tests off JCenter to Maven Central
==COMMIT_MSG==


fawind (approved):

fawind:
I thought we had an excavator for this but maybe its skipping or not finding occurences in groovy code? Will do some digging!

PR #1929, opened 2021-09-30
Ensure spock 2/nebula-test 10 tests actually run

CRogers:
## Before this PR
[Nebula test 10 upgraded Spock to version 2](https://github.com/nebula-plugins/nebula-test/releases/tag/v10.0.0), which is required for Gradle 7/Groovy 3 support.

Spock 2 has a breaking change that means it no longer runs using JUnit 4, but instead uses the JUnit Platform. In Gradle, you need to opt into the JUnit Platform by setting some configuration:

```gradle
test {
    useJUnitPlatform()
}
```

If you don't do this, your tests will be silently not be run 😢 . For many of our repos that use nebula-test, they got upgraded to 10 automatically and the tests just stopped running.

## After this PR
==COMMIT_MSG==
Ensure nebula-test 10/Spock 2 test run by automatically setting `useJUnitPlatform()` on test tasks. Add extra verification to the `checkJUnitDependencies` task for nebula-test 10/Spock 2 tests.
==COMMIT_MSG==

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineTesting.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineTesting.java
@@ -16,18 +16,20 @@
 
 package com.palantir.baseline.plugins;
 
+import com.google.common.base.Predicate;
 import com.google.common.collect.ImmutableSet;
 import com.palantir.baseline.tasks.CheckJUnitDependencies;
 import com.palantir.baseline.tasks.CheckUnusedDependenciesTask;
+import com.palantir.baseline.util.VersionUtils;
 import java.util.Objects;
 import java.util.Optional;
 import org.gradle.api.Plugin;
 import org.gradle.api.Project;
 import org.gradle.api.Task;
-import org.gradle.api.artifacts.Dependency;
+import org.gradle.api.artifacts.component.ModuleComponentIdentifier;
+import org.gradle.api.artifacts.result.ResolvedComponentResult;
 import org.gradle.api.plugins.JavaPlugin;
 import org.gradle.api.plugins.JavaPluginConvention;
-import org.gradle.api.specs.Spec;
 import org.gradle.api.tasks.SourceSet;
 import org.gradle.api.tasks.TaskContainer;
 import org.gradle.api.tasks.TaskProvider;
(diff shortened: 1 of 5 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineTesting.java
```diff
@@ -112,20 +120,32 @@ public void apply(Project project) {
...
-                .matching(spec)
-                .stream()
-                .findAny()
-                .isPresent();
+    private static void ifHasResolvedCompileDependenciesMatching(
+            Project project, SourceSet sourceSet, Predicate<ModuleComponentIdentifier> spec, Runnable runnable) {
+        project.getConfigurations()
+                .getByName(sourceSet.getRuntimeClasspathConfigurationName())
+                .getIncoming()
+                .afterResolve(deps -> {
```
CRogers, line 128: Since we need to distinguish between Spock 1 and Spock 2, we need to actually resolve the configuration to get a version number. It's actually quite tricky to resolve here, as it fails foul of the checks we have to prevent [resolving configuration at configuration time](https://docs.gradle.org/current/userguide/performance.html#dont_resolve_dependencies_at_configuration_time), which destroys perf.

Instead, we wait for this configuration to resolved, which will happen before the test is run, by the `checkJUnitDependencies` task if nothing else...

(This would be so much easier if `useJUnitPlatform` was a lazy property)

gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineTestingIntegrationTest.groovy
```diff
@@ -16,26 +16,27 @@
...
 package com.palantir.baseline
 
-import org.gradle.testkit.runner.BuildResult
-import org.gradle.testkit.runner.TaskOutcome
+import nebula.test.IntegrationSpec
+import nebula.test.functional.ExecutionResult
 import spock.lang.Unroll
 
 @Unroll
-class BaselineTestingIntegrationTest extends AbstractPluginTest {
```
CRogers, line 24: Fun fact, this homemade `AbstractPluginTest` framework from 2017 wasted nearly 2 hours of my time because it doesn't run changed code in the test from IntelliJ unless you first recompile the code through gradle 😱 😱 😱 😱 😱 😱 😱

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineTesting.java
```diff
@@ -61,38 +63,44 @@ public void apply(Project project) {
...
-                                tasks.named(
-                                        BaselineExactDependencies.checkUnusedDependenciesNameForSourceSet(ss),
-                                        CheckUnusedDependenciesTask.class,
-                                        task -> task.ignore("org.junit.jupiter", "junit-jupiter"));
-                            });
+            TaskProvider<CheckJUnitDependencies> checkJUnitDependencies =
+                    project.getTasks().register("checkJUnitDependencies", CheckJUnitDependencies.class);
+
+            project.getConvention()
+                    .getPlugin(JavaPluginConvention.class)
```
fawind, line 70: I think we should use the `JavaPluginExtension` as conventions are deprecated and to be removed with Gradle 8.

fawind (commented):
Scary stuff but looks good, just one smaller comment.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineTesting.java
fawind: I was initially thinking that it might be sufficient to just go through the explicitly declared deps which would not require resolution. However most of the time Spock is brought in as a transitive and so you have to do resolution :(

CRogers: I initially did that and checked for nebula-test too, but until resolution happens you don't know the version number, so can't decide whether or not to do the right thing

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineTesting.java
CRogers: I tried, but Gradle 6 doesn't have the `getSourceSets` method we need https://github.com/palantir/gradle-baseline/pull/1929/commits/118cb972f812a11e3b781f90e29064c10efc4439

fawind: Ah I remember encountering this before. I guess Gradle 7 is the bridge version then at with Gradle 8 we have to cut support for Gradle 6. Fine for now then!

fawind (approved):

PR #1932, opened 2021-10-04
Fix MalformedInputException in checkJUnitDependencies

CRogers:
## Before this PR
In #1929, I had originally changed searching in `sourceSet.getAllJava()` to `sourceSet.getAllSource()` to make sure we were looking in `.groovy` files as well as java. I left it in there by accident. However, this has had the downside of searching through the resources too, which may contain non-utf8 encoded files. This leads to this kind of error:

```
java.io.UncheckedIOException: java.nio.charset.MalformedInputException: Input length = 1
```

as it tries to decode non-utf8 files into utf8.

Additionally, I never proved out whether groovy files are included in `getAllSource` and not `getAllJava`. Looking at the `GroovyBasePlugin` it actually puts the groovy source into the java source sets:

```java
sourceSet.getAllJava().source(groovySource);
sourceSet.getAllSource().source(groovySource);
```

## After this PR
==COMMIT_MSG==
Fix `MalformedInputException` when checking non-utf8 files for correct junit dependencies.
==COMMIT_MSG==

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckJUnitDependencies.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckJUnitDependencies.java
@@ -163,6 +163,7 @@ public class CheckJUnitDependencies extends DefaultTask {
     }
 
     private boolean sourceSetMentionsJUnit4(SourceSet ss) {
+        // getAllJava() includes groovy sources too
         return !ss.getAllJava()
                 .filter(file -> fileContainsSubstring(
                         file,
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckJUnitDependencies.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckJUnitDependencies.java
@@ -164,7 +164,7 @@ public class CheckJUnitDependencies extends DefaultTask {
     }
 
     private boolean sourceSetMentionsJUnit4(SourceSet ss) {
-        return !ss.getAllSource()
+        return !ss.getAllJava()
                 .filter(file -> fileContainsSubstring(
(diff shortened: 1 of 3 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckJUnitDependencies.java
```diff
@@ -174,18 +173,16 @@ private boolean sourceSetMentionsJUnit4(SourceSet ss) {
...
 
     private boolean fileContainsSubstring(File file, Predicate<String> substring) {
         try (Stream<String> lines = Files.lines(file.toPath())) {
-            boolean hit = lines.anyMatch(substring::test);
-            getProject().getLogger().debug("[{}] {}", hit ? "hit" : "miss", file);
-            return hit;
-        } catch (IOException e) {
-            throw new RuntimeException("Unable to check file " + file, e);
+            return lines.anyMatch(substring);
+        } catch (Exception e) {
```
CRogers, line 185: `MalformedInputException` is thrown as an `UncheckedIOException`.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/tasks/CheckJUnitDependencies.java
```diff
@@ -164,7 +163,7 @@ private boolean hasDep(Set<ResolvedComponentResult> deps, Predicate<ModuleVersio
     }
 
     private boolean sourceSetMentionsJUnit4(SourceSet ss) {
-        return !ss.getAllSource()
+        return !ss.getAllJava()
```
fawind, line 167: Might be worth adding a small comment here that `getAllJava` is also including groovy files.

fawind (approved):
Looks good!

PR #1944, opened 2021-10-27
Plumb Add-Exports through to execution and annotation processing sites

carterkozak:
## Before this PR
Unable for jars to define the `--add-exports` options they need.
We added this to support safepoint metrics, however it would become unnecessary if we add a way to collect manifest entries in sls-packaging: https://github.com/palantir/sls-packaging/blob/7ac5d3bef6528fe36e8dc5a249c2540b7a7023c4/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.java#L63-L67

## After this PR
==COMMIT_MSG==
Plumb Add-Exports through to execution sites
==COMMIT_MSG==

sls-packaging piece here: https://github.com/palantir/sls-packaging/pull/1215
Ideas how we can deduplicate that code?

## Possible downsides?
It's a bit funky, we don't have a way to pass this through to IDEs aside from intellij idea with gradle integration.
We're re-purposing the `Add-Exports` manifest entry in a way that's slightly different from how it was intended.


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
```diff
@@ -0,0 +1,204 @@
...
+                                            .debug(
+                                                    "BaselineModuleJvmArgs compiling {} on {} with exports: {}",
+                                                    javaCompile.getName(),
+                                                    project,
+                                                    arguments);
+                                    return arguments;
+                                }
+                            });
+                });
+            }
```
carterkozak, line 89: block is ignored for now, hopefully something becomes of https://github.com/gradle/gradle/issues/18824

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
```diff
@@ -0,0 +1,204 @@
...
+            project.getTasks().withType(JavaExec.class, new Action<JavaExec>() {
+
+                @Override
+                public void execute(JavaExec javaExec) {
+                    javaExec.getJvmArgumentProviders().add(new CommandLineArgumentProvider() {
+
+                        @Override
+                        public Iterable<String> asArguments() {
+                            ImmutableList<String> arguments =
+                                    collectClasspathExports(project, extension, javaExec.getClasspath());
```
carterkozak, line 122: I think I'll need to duplicate this logic in sls-packaging, or figure out some other way to plumb args through

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
```diff
@@ -0,0 +1,214 @@
...
+            if (project.hasProperty("add.exports.to.javac")) {
+                project.getExtensions().getByType(SourceSetContainer.class).configureEach(sourceSet -> {
+                    JavaCompile javaCompile = project.getTasks()
+                            .named(sourceSet.getCompileJavaTaskName(), JavaCompile.class)
+                            .get();
+                    javaCompile
+                            .getOptions()
+                            .getCompilerArgumentProviders()
+                            // Use an anonymous class because tasks with lambda inputs cannot be cached
+                            .add(new CommandLineArgumentProvider() {
```
fawind, line 73: As a nitty suggestion, all three `CommandLineArgumentProvider` are quite similar except for the `arguments = ...` line. To make this more concise, we might we can define this as a single static class that gets a `Provider<List<String>> arguments` injected in its constructor?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineExportsExtension.java
```diff
@@ -0,0 +1,40 @@
...
+    private final SetProperty<String> exports;
+
+    @Inject
+    public BaselineExportsExtension(Project project) {
+        exports = project.getObjects().setProperty(String.class);
+    }
+
+    /** Target {@link JavaLanguageVersion} for compilation. */
+    public final SetProperty<String> exports() {
+        return exports;
```
fawind: The format of a single value here is the full flag value, e.g. `"javac.util=ALL-UNNAMED"`?

fawind: Got it from the tests, its just the value, so only `"javac.util"`

fawind: Have to read through the JEP again, but there is no other export that you can set besides `ALL-UNNAMED`?

fawind (approved):
Looks good! Just some minor optional comments.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineExportsExtension.java
carterkozak: It's the module-package tuple e.g. `java.management/sun.management`, the `=ALL-UNNAMED` suffix is implied. I'll add documentation describing this :-)

gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineExportsExtension.java
```diff
@@ -0,0 +1,40 @@
...
+public class BaselineExportsExtension {
+
+    private final SetProperty<String> exports;
+
+    @Inject
+    public BaselineExportsExtension(Project project) {
+        exports = project.getObjects().setProperty(String.class);
+    }
+
+    /** Target {@link JavaLanguageVersion} for compilation. */
```
carterkozak: Well, this is javadoc for a completely different extension.... updating.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineExportsExtension.java
```diff
@@ -0,0 +1,40 @@
...
+    public final SetProperty<String> exports() {
```
carterkozak: @fawind Extension accessor methods are a bit magical to me, is there a standard way to name this so that I can do something like:
```groovy
moduleJvmArgs {
  exports += ['java.management/sun.management']
}
```

fawind: Let me quickly check. I think there is some fancy groovy syntax you can use...

carterkozak: I updated this with a validating `setExports(Iterable<String>)` which allows us to do `exports = ['java.management/sun.management']`, so that may be sufficient

fawind: You could do something like:
```groovy
exports() addAll 'java.management/sun.management', 'java.management/sun.other'
```

We could also add some helper methods to the extension, e.g.:
```java
public final void addExports(String... values) {
    exports.addAll(values);
}
```

and then we can use it from gradle:
```groovy
moduleJvmArgs {
    addExports 'java.management/sun.management', 'java.management/sun.other'
}
```

fawind: Ah, didn't see your previous comment. That works very well, didn't think of overloading the setter!

carterkozak: I like your varargs example, I'm going to swap to that!

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
carterkozak: Thoughts on punting this? It would mean plumbing around logging args to differentiate the types of tasks as well

fawind: Yeah, no strong opinion on that one!

Issue #1946, opened 2021-10-28
Zero warm-up Guava RateLimiter


kristofarkas:
## What happened?

The default Guava RateLimiter initializer creates a rate limiter with 0 permits, and starts acquiring them. This causes an unexpected behaviour, where at the beginning of the rate limiters lifetime the rate limit is much lower that specified. 
 
```java
// rate limits way more aggressively than `permitsPerSecond` at start
RateLimiter.create(permitsPerSecond)
```
## What did you want to happen?

Setting the warm up time to zero means that max permits are available at start.

```java
RateLimiter.create(permitsPerSecond, Duration.ZERO)
```

PR #1950, opened 2021-11-02
Feature: Zero Warmup Guava RateLimiter

kevin-yu-0602:
## Before this PR

This addresses #1946
## After this PR

==COMMIT_MSG==
 Adds a new rule NoWarmupRateLimiter that catches `RateLimite.create ` calls with no explicit warmup arguments.
==COMMIT_MSG==

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,7 +206,7 @@ Safe Logging can be found at [github.com/palantir/safe-logging](https://github.c
 - `PreferImmutableStreamExCollections`: It's common to use toMap/toSet/toList() as the terminal operation on a stream, but would be extremely surprising to rely on the mutability of these collections. Prefer `toImmutableMap`, `toImmutableSet` and `toImmutableList`. (If the performance overhead of a stream is already acceptable, then the `UnmodifiableFoo` wrapper is likely tolerable).
 - `DangerousIdentityKey`: Key type does not override equals() and hashCode, so comparisons will be done on reference equality only.
 - `ConsistentOverrides`: Ensure values are bound to the correct variables when overriding methods
-- 'NoWarmupRateLimiter': Ensures Guava RateLimiter initializers specify a warm-up duration.
+- `NoWarmupRateLimiter`: Ensures Guava RateLimiter initializers specify a warm-up duration.
 
 ### Programmatic Application
 
--- a/README.md
+++ b/README.md
@@ -206,7 +206,7 @@ Safe Logging can be found at [github.com/palantir/safe-logging](https://github.c
 - `PreferImmutableStreamExCollections`: It's common to use toMap/toSet/toList() as the terminal operation on a stream, but would be extremely surprising to rely on the mutability of these collections. Prefer `toImmutableMap`, `toImmutableSet` and `toImmutableList`. (If the performance overhead of a stream is already acceptable, then the `UnmodifiableFoo` wrapper is likely tolerable).
 - `DangerousIdentityKey`: Key type does not override equals() and hashCode, so comparisons will be done on reference equality only.
 - `ConsistentOverrides`: Ensure values are bound to the correct variables when overriding methods
-- `NoWarmupRateLimiter`: Ensure Guava RateLimiter initializers specifies a warm-up duration.
+- `NoWarmupRateLimiter`: Ensure Guava RateLimiter initializers specify a warm-up duration.
 
 ### Programmatic Application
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -206,6 +206,7 @@ Safe Logging can be found at [github.com/palantir/safe-logging](https://github.c
 - `PreferImmutableStreamExCollections`: It's common to use toMap/toSet/toList() as the terminal operation on a stream, but would be extremely surprising to rely on the mutability of these collections. Prefer `toImmutableMap`, `toImmutableSet` and `toImmutableList`. (If the performance overhead of a stream is already acceptable, then the `UnmodifiableFoo` wrapper is likely tolerable).
 - `DangerousIdentityKey`: Key type does not override equals() and hashCode, so comparisons will be done on reference equality only.
 - `ConsistentOverrides`: Ensure values are bound to the correct variables when overriding methods
+- 'NoWarmupRateLimiter': Ensures Guava RateLimiter initializers specify a warm-up duration.
```
carterkozak:
```suggestion
- `NoWarmupRateLimiter`: Ensures Guava RateLimiter initializers specify a warm-up duration.
```

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/NoWarmupRateLimiterTest.java
```diff
@@ -0,0 +1,114 @@
...
+                        "Test.java",
+                        "import com.google.common.util.concurrent.RateLimiter;",
+                        "import java.time.Duration;",
+                        "class Test {",
+                        "  void f() {",
+                        "    RateLimiter.create(10, Duration.ZERO);",
+                        "  }",
+                        "",
+                        "}")
+                .doTest(BugCheckerRefactoringTestHelper.TestMode.TEXT_MATCH);
```
carterkozak: Some of our older tests specify text match rather than ast, I wouldn't worry about that here -- I believe the text-match arg used to provide better debugging info on failure but that's no longer the case.
```suggestion
                .doTest();
```

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/NoWarmupRateLimiterTest.java
```diff
@@ -0,0 +1,114 @@
...
```
carterkozak: here as well

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/NoWarmupRateLimiter.java
```diff
@@ -0,0 +1,72 @@
...
+import com.google.errorprone.matchers.Description;
+import com.google.errorprone.matchers.Matcher;
+import com.google.errorprone.matchers.method.MethodMatchers;
+import com.sun.source.tree.ExpressionTree;
+import com.sun.source.tree.MethodInvocationTree;
+import java.util.List;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "NoWarmupRateLimiter",
```
carterkozak, line 36: Let's register this in `BaselineErrorProneExtension.java` so that the baseline upgrade excavator will automatically apply the fix

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/NoWarmupRateLimiterTest.java
kevin-yu-0602: Done

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/NoWarmupRateLimiterTest.java
kevin-yu-0602: Got it

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/NoWarmupRateLimiter.java
kevin-yu-0602: I believe I did add this to `BaselineErrorProneExtension.java`, do let me know if I'm doing this wrong

carterkozak: Doh, sorry!

carterkozak (approved):
Great stuff, thanks for the contribution!

pkoenig10:
This PR introduced a behavior change. After this change, we have unit tests that began failing.
https://c.p.b/gh/foundry/multipass/321545

I don't think it's reasonable to assume that everyone wants this behavior.

carterkozak:
> I don't think it's reasonable to assume that everyone wants this behavior.

Right, it requires human intervention to merge. Many consumers do want this behavior, the idea is that a duration should always be provided to explicitly acknowledge the behavior either way. We suggest the value we think most folks want.

pkoenig10:
I think there's something else going on here. For example, this test will pass:
```java
@ParameterizedTest
@ValueSource(ints = {1, 10, 100, 1000})
void testRateLimiter(int permitsPerSecond) {
    RateLimiter rateLimiterZero = RateLimiter.create(permitsPerSecond, Duration.ZERO);
    assertThat(Stream.generate(rateLimiterZero::tryAcquire)
                    .limit(10 * permitsPerSecond)
                    .allMatch(v -> v))
            .isTrue();
}
```

carterkozak:
Ouch, good catch.

PR #1958, opened 2021-11-04
Replace out instances of `RateLimiter.create(rate, Duration.ZERO)`

carterkozak:
==COMMIT_MSG==
Replace out instances of `RateLimiter.create(rate, Duration.ZERO)` which do not rate limit at all. See https://github.com/google/guava/issues/2730
==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/ZeroWarmupRateLimiterTest.java
+++ b/baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/ZeroWarmupRateLimiterTest.java
@@ -1,0 +1,87 @@
+/*
+ * (c) Copyright 2021 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.baseline.errorprone;
+
+import org.junit.jupiter.api.Test;
+
+class ZeroWarmupRateLimiterTest {
+
+    @Test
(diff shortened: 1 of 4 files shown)


baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/ZeroWarmupRateLimiterTest.java
```diff
@@ -0,0 +1,87 @@
...
+                        "import java.time.Duration;",
+                        "class Test {",
+                        "  void f() {",
+                        "    RateLimiter.create(10, Duration.ZERO);",
+                        "  }",
+                        "}")
+                .addOutputLines(
+                        "Test.java",
+                        "import com.google.common.util.concurrent.RateLimiter;",
+                        "import java.time.Duration;",
```
kevin-yu-0602, line 37: No action required. Wondering if it is common/feasible to do import removal in cases like these. Or is there no way of guaranteeing the import isn't being used elsewhere?

carterkozak: Great question: We could search the CompilationUnit for uses of any given import, but the fun part is that we don't have to :-) We always follow up automated refactors with our code formatter, that way the results become pretty, and unused imports are removed automatically!

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/ZeroWarmupRateLimiterTest.java
```diff
@@ -0,0 +1,87 @@
...
+                        "class Test {",
+                        "  void f() {",
+                        "    RateLimiter.create(10, Duration.ofMillis(100));",
+                        "  }",
+                        "}")
+                .expectUnchanged()
+                .doTest();
+    }
+
```
kevin-yu-0602, line 83: nit: Could benefit from a sanity check that tests double arguments, i.e: `RateLimiter.create(10, 100, TimeUnit.SECONDS);`

kevin-yu-0602 (approved):

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/ZeroWarmupRateLimiterTest.java
carterkozak: I figure we can add that when we fix the constant-zero case for the value-and-unit overload -- at the moment the check doesn't look at other overloads.

PR #1960, opened 2021-11-04
Improve Zero Warmup Rate Limiter to Catch Int Literals

kevin-yu-0602:
## Before this PR
Follow up to https://github.com/palantir/gradle-baseline/pull/1958 to allow the rule to catch int literals passing in 0 as the warmup time.

## After this PR

==COMMIT_MSG==
==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ZeroWarmupRateLimiter.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ZeroWarmupRateLimiter.java
@@ -92,7 +92,7 @@ public final class ZeroWarmupRateLimiter extends BugChecker implements BugChecke
     }
 
     private static boolean isIntLiteralZero(ExpressionTree expressionTree, VisitorState state) {
-        Integer warmupTime = ASTHelpers.constValue(expressionTree, Integer.class);
-        return warmupTime != null && warmupTime.equals(0);
+        Number warmupTime = ASTHelpers.constValue(expressionTree, Number.class);
+        return warmupTime != null && (warmupTime.equals(0) || warmupTime.equals(0L));
     }
 }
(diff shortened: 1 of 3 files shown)


baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/ZeroWarmupRateLimiterTest.java
```diff
@@ -66,6 +66,29 @@ public void should_remove_duration_zero_static_import() {
...
 
+    @Test
+    public void should_remove_long_literal_zero() {
+        fix().addInputLines(
+                        "Test.java",
+                        "import com.google.common.util.concurrent.RateLimiter;",
+                        "import java.util.concurrent.TimeUnit;",
+                        "class Test {",
+                        "  void f() {",
+                        "    RateLimiter.create(10, 0, TimeUnit.SECONDS);",
```
carterkozak, line 77: Might be helpful to add a test using a long literal `0L`

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ZeroWarmupRateLimiter.java
```diff
@@ -81,4 +90,9 @@ private static boolean isDurationZero(ExpressionTree expressionTree, VisitorStat
         }
         return false;
     }
+
+    private static boolean isIntLiteralZero(ExpressionTree expressionTree, VisitorState state) {
+        Integer warmupTime = ASTHelpers.constValue(expressionTree, Integer.class);
+        return warmupTime != null && warmupTime.equals(0);
+    }
```
carterkozak: I wonder if this would be more robust using `Number` (entirely untested):

```suggestion
    private static boolean isLiteralZero(ExpressionTree expressionTree, VisitorState state) {
        Number warmupTime = ASTHelpers.constValue(expressionTree, Number.class);
        return warmupTime != null && warmupTime.longValue() == 0L;
    }
```

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/ZeroWarmupRateLimiterTest.java
kevin-yu-0602: Good idea.

carterkozak (approved):
Thanks!

PR #1966, opened 2021-11-09
Add error-prone check to prefer hasNoArgs

pkoenig10:
Error prone check to switch to the methods added in https://github.com/palantir/safe-logging/pull/607 and https://github.com/palantir/conjure-java-runtime-api/pull/702.

Probably want to wait a while before merging this so repos have upgraded to releases with the `hasNoArgs` method.


baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/AssertNoArgsTest.java
```diff
@@ -0,0 +1,202 @@
...
+        fix().addInputLines(
+                        "Test.java",
+                        "import com.palantir.logsafe.Arg;",
+                        "import com.palantir.logsafe.testing.LoggableExceptionAssert;",
+                        "public class Test {",
+                        "  void f(LoggableExceptionAssert a, Arg<?> arg) {",
+                        "    a.hasArgs(arg);",
+                        "  }",
+                        "}")
+                .addOutputLines(
```
carterkozak: expectUnchanged?

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/AssertNoArgsTest.java
```diff
@@ -0,0 +1,202 @@
...
+                        "  void f(LoggableExceptionAssert a) {",
+                        "    a.containsArgs();",
+                        "  }",
+                        "}")
+                .addOutputLines(
+                        "Test.java",
+                        "import com.palantir.logsafe.testing.LoggableExceptionAssert;",
+                        "public class Test {",
+                        "  void f(LoggableExceptionAssert a) {",
+                        "    a.hasNoArgs();",
```
carterkozak, line 146: I think this one could produce failures, but I can't imagine anyone would write `.containsArgs()` and expect it to be a no-op.

pkoenig10: Searching our internal repos, I only see 2 uses of `.containsArgs()` without any arguments and they are both in a repo I own. And the intention in both cases was to assert that the exception has no args.

It's probably more likely that this change is a bug fix.

carterkozak: Yep :+1:

carterkozak (approved):

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/AssertNoArgs.java
```diff
@@ -0,0 +1,59 @@
...
+import com.sun.source.tree.MethodInvocationTree;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "AssertNoArgs",
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
+        severity = BugPattern.SeverityLevel.WARNING,
+        summary = "Prefer hasNoArgs when asserting that an exception has no args.")
+public final class AssertNoArgs extends BugChecker implements BugChecker.MethodInvocationTreeMatcher {
```
carterkozak, line 37: @pkoenig10 we probably should have added this to the BaselineErrorProneExtension so it's automatically fiexed

carterkozak: #1967

PR #1974, opened 2021-11-12
Fix errors setting test framework due to finalized property

fawind:
## Before this PR

Using baseline on Gradle 7.3 would fail with ([circle](https://app.circleci.com/pipelines/github/palantir/gradle-baseline/2911/workflows/917953bc-bfe6-4b90-9f67-45e3d1e10429/jobs/30225/steps), [PR](https://github.com/palantir/gradle-baseline/pull/1969)):
```
The value for task ':test' property 'testFrameworkProperty' is final and cannot be changed any further.
```

This is because starting with Gradle 7.3, the test framework property is now getting finalized after getting set initially ([ref](https://github.com/gradle/gradle/blob/a520acd9912c9d5f0e548338b32ef77d66f76058/subprojects/testing-jvm/src/main/java/org/gradle/api/tasks/testing/Test.java#L917-L934)).

Finalizing the property actually happens in our precondition where we call `Test#getOptions`:
https://github.com/palantir/gradle-baseline/blob/8013b3becc5ef623b08cd7e361137111e0be6ee8/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineTesting.java#L179-L180

This is because the public getters all initialize and finalize the property with a default if nothing is set yet ([ref](https://github.com/gradle/gradle/blob/a520acd9912c9d5f0e548338b32ef77d66f76058/subprojects/testing-jvm/src/main/java/org/gradle/api/tasks/testing/Test.java#L913-L915)). Thus, when we later explicitly try to set the test framework ([ref](https://github.com/palantir/gradle-baseline/blob/8013b3becc5ef623b08cd7e361137111e0be6ee8/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineTesting.java#L153)), Gradle fails because the property is finalized now.

## After this PR

Unfortunately before Gradle 7.3, I couldn't find a public method that allows us to figure out if and what kind of test framework is configured without finalizing a test framework.

However with Gradle 7.3, they added a public getter for getting the `Property<TestFramework>` which we can use to check if a test framework is configured without finalizing it.

==COMMIT_MSG==
Fix errors setting test framework due to finalized property
==COMMIT_MSG==

## Possible downsides?
`Test#getTestFrameworkProperty` is marked as incubating so not sure how stable it is.

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineTesting.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineTesting.java
@@ -21,6 +21,8 @@ import com.google.common.collect.ImmutableSet;
 import com.palantir.baseline.tasks.CheckJUnitDependencies;
 import com.palantir.baseline.tasks.CheckUnusedDependenciesTask;
 import com.palantir.baseline.util.VersionUtils;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 import java.util.Objects;
 import java.util.Optional;
 import org.gradle.api.Plugin;
@@ -30,12 +32,14 @@ import org.gradle.api.artifacts.component.ModuleComponentIdentifier;
 import org.gradle.api.artifacts.result.ResolvedComponentResult;
 import org.gradle.api.plugins.JavaPlugin;
 import org.gradle.api.plugins.JavaPluginConvention;
+import org.gradle.api.provider.Property;
 import org.gradle.api.tasks.SourceSet;
 import org.gradle.api.tasks.TaskContainer;
 import org.gradle.api.tasks.TaskProvider;
 import org.gradle.api.tasks.testing.Test;
 import org.gradle.api.tasks.testing.junitplatform.JUnitPlatformOptions;
 import org.gradle.api.tasks.testing.logging.TestLogEvent;
+import org.gradle.util.GradleVersion;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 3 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineTesting.java
```diff
@@ -176,7 +180,34 @@ private static void enableJunit5ForTestTask(Test task) {
...
+    }
+
+    @SuppressWarnings("IllegalImports")
+    private static Property<org.gradle.api.internal.tasks.testing.TestFramework> getTestFrameworkWithReflection(
+            Test task) {
+        try {
+            Method getTestFrameworkProperty = Test.class.getMethod("getTestFrameworkProperty");
+            return (Property<org.gradle.api.internal.tasks.testing.TestFramework>)
+                    getTestFrameworkProperty.invoke(task);
+        } catch (NoSuchMethodException | InvocationTargetException | IllegalAccessException e) {
```
carterkozak: no change requested: Could catch [ReflectiveOperationException](https://docs.oracle.com/javase/8/docs/api/java/lang/ReflectiveOperationException.html) here

carterkozak (approved):
thanks!

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineTesting.java
fawind: Nice, updated!

PR #2009, opened 2021-12-07
Gradle plugins don't enforce PublicConstructorForAbstractClass

carterkozak:
It could break gradle magic injection

==COMMIT_MSG==
Gradle plugins don't enforce PublicConstructorForAbstractClass which can break gradle injection
==COMMIT_MSG==


fawind (approved):
Nice!

xRuiAlves (approved):

xRuiAlves:
I'm late

PR #2010, opened 2021-12-07
Improvement: Disable errorprone in intellij

ferozco:
## Before this PR
Running test/compilation from intellij with gradle integration was pretty painful since it would cause all errorprone rules to run. This slowed down overall compilation but also made it hard for you to quickly iterate on a test, or hack together a proof of concept

## After this PR
==COMMIT_MSG==
Disable errorprone in intellij
==COMMIT_MSG==

This _will not_ impact CI in any way and should only make the gradle integration's behaviour closer to working with a `*ipr` file

cc @iamdanfox

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/IntellijSupport.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/IntellijSupport.java
@@ -1,0 +1,27 @@
+/*
+ * (c) Copyright 2021 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.baseline;
+
+public final class IntellijSupport {
+    private static final String INTELLIJ_ACTIVE = "idea.active";
+
+    public static boolean isRunningInIntellij() {
+        return System.getProperties().getProperty(INTELLIJ_ACTIVE) != null;
+    }
+
+    private IntellijSupport() {}
+}
--- a/changelog/@unreleased/pr-2010.v2.yml
+++ b/changelog/@unreleased/pr-2010.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Disable errorprone in intellij
+  links:
+  - https://github.com/palantir/gradle-baseline/pull/2010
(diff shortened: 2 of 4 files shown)


iamdanfox:
I _think_ this would probably be net positive, but just trying to consider possible downsides:

 - I think we would no longer get cache hits from compilation inside IntelliJ vs in a terminal/on CI
 - a few people might actually _want_ to see all the ErrorProne rules as squiggly lines in their IDE 🤷  ??

One alternative might be downgrade things from error->warning or even info in IntelliJ?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/IntellijSupport.java
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.baseline;
+
+public final class IntellijSupport {
+    private static final String INTELLIJ_ACTIVE = "idea.active";
+
+    public static boolean isRunningInIntellij() {
```
carterkozak, line 22: Mind consolidating the `idea.active` checks from `BaselineIdea` as well?

carterkozak:
> a few people might actually want to see all the ErrorProne rules as squiggly lines in their IDE shrug ??

Does this actually happen? I didn't think idea picked up error-prone warnings, but most of the projects I work on fail in that case, so I may not have seen it. If so, I'd be down for opting out of `-Werror` when intellij gradle integration is used.

Re caching: It's unfortunate to lose gradle caching, but it's a bit difficult to prototype things with StrictUnusedVariable enabled, so I think the trade-off makes sense.

iamdanfox:
Yeah StrictUnusedVariable is savage. :+1: from me.

ferozco:
I don't believe that intellij shows squiggly for errorprone OOTB

gradle-baseline-java/src/main/groovy/com/palantir/baseline/IntellijSupport.java
ferozco: of course! done

carterkozak (approved):
tyvm!

pkoenig10:
While I understand the motivation for this change, it breaks what is, IMO, one of the biggest benefits of using the native Gradle integration - shared build infrastructure and caching.

When switching between my IDE and the command line, I now need to compile everything twice. In my experience, being able to avoid expensive recompilations is far more valuable that avoiding error-prone warnings when iterating.

pkoenig10:
If we care about disabling error-prone warnings in dev environments, then we should do it consistently - for both Intellij and the command line. The only place we truly care about these checks is in CI, so we can just enable all checks explicitly in CI.

pkoenig10:
I'm also not convinced that disabling error-prone is the correct tradeoff. It can be very confusing if my code doesn't pass CI due to an error I cannot see when I build locally. It now becomes very time-consuming to get passing CI build, especially if I have errors in multiple projects that depend on each other transitively because I will only see one error at a time. Avoiding this requires knowing that compiling on the command line will show these errors.

I would much rather know that my code won't pass CI _before_ I push (since I'm almost certainly compiling locally anyways) instead of pushing thinking it's good, and then finding out 5-10 minutes later than it was not good.

pkoenig10:
At the very least, I'd appreciate if we gave devs an option to avoid this "feature". AFAICT there's no way for me to prevent IntelliJ from setting the `idea.active` property in my builds, so I'm stuck with this behavior.

PR #2015, opened 2021-12-09
configure codeStyles for intellij import

carterkozak:
Based in part upon #1231 (Thanks @dansanduleac!)

## Before this PR
Import `.ipr`, then migrate to gradle integration

## After this PR
==COMMIT_MSG==
It's no longer necessary to import an ipr before using gradle integration, code styles are imported correctly out of the box.
==COMMIT_MSG==

## Possible downsides?
Maybe I messed up?

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
@@ -183,10 +183,13 @@ class BaselineIdea extends AbstractBaselinePlugin {
                     def codeScheme = GroovyXmlUtils.matchOrCreateChild(it, "code_scheme", [name: 'Project'])
                     codeScheme.attributes().putIfAbsent("version", 173)
                     def javaCodeStyleSettings = GroovyXmlUtils.matchOrCreateChild(codeScheme, "JavaCodeStyleSettings")
-                    // Just add the default configuration nodes on top of whatever nodes already exist
-                    // We could be better about this, but IDEA will mostly resolve the duplicates here for us
+                    // Avoid re-adding duplicate options to the project. This allows users to override settings based
+                    // on preference.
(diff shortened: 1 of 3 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
```diff
@@ -176,10 +181,15 @@ class BaselineIdea extends AbstractBaselinePlugin {
                 project.file(".idea/codeStyles/Project.xml"),
                 {
                     def codeScheme = GroovyXmlUtils.matchOrCreateChild(it, "code_scheme", [name: 'Project'])
-                    // Just add the default configuration nodes on top of whatever nodes already exist
-                    // We could be better about this, but IDEA will mostly resolve the duplicates here for us
-                    ideaStyleSettings.value.option.forEach {
-                        codeScheme.append(it)
```
carterkozak, line 182: Turns out `Project.xml` used to grow indefinitely over time....

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
```diff
@@ -176,10 +181,15 @@ class BaselineIdea extends AbstractBaselinePlugin {
...
+                    codeScheme.attributes().putIfAbsent("version", 173)
```
fawind, line 184: Do we need to keep this version updated going forwards?

fawind (approved):
Let's test it out!

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
carterkozak: I'm not entirely sure, but the property names may not work correctly on other versions

PR #2021, opened 2021-12-15
Suggest putting jupiter on runtime only

jakobjuelich:
## Before this PR
checkUnusedDependenciesTest fails complaining org.junit.jupiter:junit-jupiter is unused and should be replaced by org.junit.jupiter:junit-jupiter-api. If one does that, checkJUnitDependencies fails and tells to put it back.

With this PR, a better solution is proposed in the error.

## After this PR

==COMMIT_MSG==
Suggest putting jupiter on runtime only in error
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-2021.v2.yml
+++ b/changelog/@unreleased/pr-2021.v2.yml
@@ -1,0 +1,6 @@
+type: improvement
+improvement:
+  description: Upon missing jupiter on classpath, suggest putting it on runtime rather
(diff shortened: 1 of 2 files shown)


jakobjuelich:
Go away stalebot!

Friendly ping @iamdanfox and @xRuiAlves – this one needs your approval.

carterkozak (approved):
lgtm, thanks!

PR #2022, opened 2021-12-23
Only enforce constant strings when log format has arguments.

josephw:
If a single string is passed to SLF4J's logging messages, either with or
without a Marker before or a Throwable after, then it's a literal message,
rather than a format. Permit that usage without complaint.

## Before this PR
I'm looking to identify cases in a codebase where an SLF4J format is not hardcoded, and this checker is referenced as a solution to https://github.com/google/error-prone/issues/565 . However, I found too many false positives for the case where the log message was just a message, rather than being used as a format string.

## After this PR
Cases with or without a `Marker`, with or without a `Throwable`, but with no arguments on top of the message no longer cause warnings.

## Possible downsides?
Non-constant messages aren't wrong, but it's still a very reasonable to want prevent them as part of a style guide. This change treats them as generally acceptable, with no way to preserve the current behaviour.

### DIFF
--- a/baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/Slf4jConstantLogMessageTests.java
+++ b/baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/Slf4jConstantLogMessageTests.java
@@ -51,16 +51,17 @@ public final class Slf4jConstantLogMessageTests {
 
     @Test
     public void testNonCompileTimeConstantExpression() throws Exception {
-        test("log.trace(\"constant\" + param);");
-        test("log.debug(\"constant\" + param);");
-        test("log.info(\"constant\" + param);");
-        test("log.warn(\"constant\" + param);");
-        test("log.error(\"constant\" + param);");
-        test("log.trace(new DummyMarker(), \"constant\" + param);");
-        test("log.debug(new DummyMarker(), \"constant\" + param);");
-        test("log.info(new DummyMarker(), \"constant\" + param);");
-        test("log.warn(new DummyMarker(), \"constant\" + param);");
-        test("log.error(new DummyMarker(), \"constant\" + param);");
+        test("log.trace(\"constant\" + param, \"argument\");");
+        test("log.debug(\"constant\" + param, \"argument\");");
+        test("log.info(\"constant\" + param, \"argument\");");
+        test("log.warn(\"constant\" + param, \"argument\");");
+        test("log.error(\"constant\" + param, \"argument\");");
+        test("log.trace(new DummyMarker(), \"constant\" + param, \"argument\");");
+        test("log.debug(new DummyMarker(), \"constant\" + param, \"argument\");");
+        test("log.info(new DummyMarker(), \"constant\" + param, \"argument\");");
+        test("log.warn(new DummyMarker(), \"constant\" + param, \"argument\");");
+        test("log.error(new DummyMarker(), \"constant\" + param, \"argument\");");
+        test("log.error(new DummyMarker(), \"constant\" + param, \"argument\", new Throwable());");
     }
 
     @Test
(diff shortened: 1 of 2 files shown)


palantirtech:
Thanks for your interest in [palantir/gradle-baseline](https://github.com/palantir/gradle-baseline), @josephw! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/Slf4jConstantLogMessageTests.java
```diff
@@ -121,6 +122,22 @@ public void negative() {
                         "    log.info(new DummyMarker(), \"constant\");",
                         "    log.warn(new DummyMarker(), \"constant\");",
                         "    log.error(new DummyMarker(), \"constant\");",
+                        "}",
+                        "",
+                        "void f(String param) {",
+                        "    // Logging of a non-constant, but with no arguments so it's a message",
+                        "    log.trace(\"constant\" + param);",
```
carterkozak, line 129: This is inefficient code which results in a great deal of overhead in hot paths! When `TRACE` level is disabled, `log.trace("constant{}", param);` is a simple volatile read (level check) where `log.trace("constant" + param);` evaluates the string concatenation before the level check!

carterkozak (changes requested):
I've added a comment describing one of your test cases describing the intent of this check, and why I believe the test case should not be valid.

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/Slf4jConstantLogMessageTests.java
josephw: You're absolutely right about this being terrible code for performance reasons. Maybe I should take a step back and ask what the goal of this bug checker is?

I was looking at it as a way to detect incorrect construction of an SLF4J format at runtime: especially, mixing of string concatenation and passing arguments. So, correctness, rather than performance. For that use case, this is fine.

Would it be fair to say you see this as a more general enforcement that also requires that the logging call is very cheap, if it doesn't result in output?

If so, I'll decline this PR: it (intentionally!) allows inefficient code through.

Either way, the example you give is a perfect candidate for adding to this file's `@BugPattern(explanation="...")`, for more context on why the code it catches is a bug.

carterkozak: > You're absolutely right about this being terrible code for performance reasons. Maybe I should take a step back and ask what the goal of this bug checker is?

The goal is to ensure that only constant-string messages are logged, as the description states ;-)
It's light on details describing _why_ that's the case, as you have highlighted. Performance is one reason, but also due to the unique way we handle log data, separating user-data from internal diagnostic info. Fully describing these in the error-prone check would be difficult because the reasons don't all apply broadly, although I'm of the opinion that the string component should _always_ be a compile-time constant!

> I was looking at it as a way to detect incorrect construction of an SLF4J format at runtime: especially, mixing of string concatenation and passing arguments. So, correctness, rather than performance. For that use case, this is fine.

I think you'd actually want a separate check -- if the format values is a constant string value, you can read it, count the slf4j placeholders, and match them with non-throwable args to verify that data will be logged. This check doesn't count placeholders at all :-)
In practice we tend to log the raw format string and the parameters separately without any interpolation, thus we don't have such a check here.

Stephan202: > I think you'd actually want a separate check -- if the format values is a constant string value, you can read it, count the slf4j placeholders, and match them with non-throwable args to verify that data will be logged. This check doesn't count placeholders at all :-)

We have such a check internally, and while I've made such a claim in the past, it looks like this time we might _actually_ be working towards open-sourcing all our Error Prone code. If you could wait for that (exact timeline TBD, but let's say 2-3 months) then this might save you some time :) (Our check also auto-fixes cases where the developer used `%s` instead of `{}`.)

josephw: Thanks for the details; I think a separate check is exactly what I'm after here. For now, an internal copy with these changes covers my requirements, and I'll look to pick up either google/error-prone#565 or a separate project later.

The mismatched `{}` count would be ideal; I've seen a few cases where a final error argument is assumed to be a `Throwable`, but isn't.

PR #2024, opened 2022-01-05
Add FilterOutputStreamSlowMultibyteWrite

schlosna:
## Before this PR

Subtypes of `FilterOutputStream` that do not implement `write(byte[], int, int)` would be inefficient and fall back to single byte writes. 

## After this PR
==COMMIT_MSG==
If a subclass of FilterOutputStream implements `void write(int)`, they
should also override `void write(byte[], int, int)`, otherwise the
performance of the stream is likely to be slow.

See https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/io/FilterOutputStream.html#write(byte%5B%5D,int,int)
> Subclasses of FilterOutputStream should provide a more efficient implementation of this method.

Similar in concept to https://errorprone.info/bugpattern/InputStreamSlowMultibyteRead
==COMMIT_MSG==

An example of this being fixed: https://github.com/palantir/hadoop-crypto/pull/42

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineErrorProneExtension.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineErrorProneExtension.java
@@ -41,7 +41,6 @@ public class BaselineErrorProneExtension {
             "ExecutorSubmitRunnableFutureIgnored",
             "ExtendsErrorOrThrowable",
             "FinalClass",
-            "FilterOutputStreamSlowMultibyteWrite",
             "ImmutablesStyle",
             "ImplicitPublicBuilderConstructor",
             "JavaTimeDefaultTimeZone",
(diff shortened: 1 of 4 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineErrorProneExtension.java
```diff
@@ -41,6 +41,7 @@
             "ExecutorSubmitRunnableFutureIgnored",
             "ExtendsErrorOrThrowable",
             "FinalClass",
+            "FilterOutputStreamSlowMultibyteWrite",
```
carterkozak: This is only necessary to automatically apply suggested fixes.
```suggestion
```

carterkozak (approved):
Otherwise lgtm, thanks! Feel free to re-add merge-when-ready after fixing the BaselineErrorProneExtension.

PR #2028, opened 2022-01-07
Add BugCheckerAutoService

schlosna:
## Before this PR
If a BugChecker is not annotated as `@AutoService(BugChecker.class)` it will not be registered with error-prone as an available bug pattern (see #2024 & #2026).

## After this PR
==COMMIT_MSG==
Add BugCheckerAutoService

Concrete BugChecker implementations should be annotated
`@AutoService(BugChecker.class)` for auto registration with error-prone.
==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/BugCheckerAutoService.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/BugCheckerAutoService.java
@@ -1,0 +1,83 @@
+/*
+ * (c) Copyright 2022 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/BugCheckerAutoService.java
```diff
@@ -0,0 +1,83 @@
...
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "BugCheckerAutoService",
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
+        summary = "Concrete BugChecker implementations should be annotated "
+                + "`@AutoService(BugChecker.class)` for auto registration with error-prone.",
+        severity = SeverityLevel.ERROR,
+        tags = StandardTags.PERFORMANCE)
```
carterkozak: I don’t think this is a perf issue :-)

carterkozak (approved):
Lgtm, feel free to merge after removing the perf tag. Thanks!

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/BugCheckerAutoService.java
schlosna: doh, copy pasted that block

PR #2042, opened 2022-01-14
Add property to explicitly enable Error Prone

pkoenig10:
See https://github.com/palantir/gradle-baseline/pull/2010#issuecomment-1012612199

## Before this PR
Error Prone is always disabled when running in IntelliJ.

## After this PR
Users can explicitly enable Error Prone, even when running in IntelliJ, by setting the `com.palantir.baseline-error-prone.disable=false` Gradle property.

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
@@ -66,6 +66,7 @@ public final class BaselineErrorProne implements Plugin<Project> {
     private static final String PROP_ERROR_PRONE_APPLY = "errorProneApply";
     private static final String PROP_REFASTER_APPLY = "refasterApply";
     private static final String DISABLE_PROPERTY = "com.palantir.baseline-error-prone.disable";
+    private static final String ENABLE_PROPERTY = "com.palantir.baseline-error-prone.enable";
 
     @Override
     public void apply(Project project) {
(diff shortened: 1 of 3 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
```diff
@@ -209,11 +210,11 @@ private static void configureErrorProneOptions(
             JavaCompile javaCompile,
             ErrorProneOptions errorProneOptions) {
 
-        if (project.hasProperty(DISABLE_PROPERTY) || IntellijSupport.isRunningInIntellij()) {
-            log.info("Disabling baseline-error-prone for {} due to {}", project, DISABLE_PROPERTY);
-            errorProneOptions.getEnabled().set(false);
-        } else {
+        if (project.hasProperty(ENABLE_PROPERTY)) {
             errorProneOptions.getEnabled().set(true);
```
pkoenig10: We can just rely on the plugin to have an appropriate default value here:
https://github.com/tbroyer/gradle-errorprone-plugin/blob/8d1f44dcef0ee68b5712ab1d319acfd2826db2a6/src/main/kotlin/net/ltgt/gradle/errorprone/ErrorProneOptions.kt#L24-L25

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
```diff
@@ -66,6 +66,7 @@
     private static final String PROP_ERROR_PRONE_APPLY = "errorProneApply";
     private static final String PROP_REFASTER_APPLY = "refasterApply";
     private static final String DISABLE_PROPERTY = "com.palantir.baseline-error-prone.disable";
+    private static final String ENABLE_PROPERTY = "com.palantir.baseline-error-prone.enable";
```
iamdanfox: seems a tiny bit funky to have two flags here, how would you feel about using it as 'key=value' manner rather than just testing for the presence of 'key'?

I'm thinking we'd just have the `com.palantir.error-prone.disable` flag and say it defaults to be `runningInIntellij`, but if you wanted to provide it explicitly you can??

pkoenig10:
@carterkozak @ferozco Any thoughts on this?

carterkozak (approved):
lgtm

PR #2061, opened 2022-01-26
improvement: Remove java 8 errorprone support

robert3005:
## After this PR

==COMMIT_MSG==
Remove support for running errorprone witth java 8 target compatibility
==COMMIT_MSG==


carterkozak:
What prompted this?

robert3005:
I originally wrote this code and since we have increased source compatibility across the board to 11 I don't think it's no longer necessary. Thought of this as a bit of a cleanup

carterkozak:
oof it looks like java 9 support broke around 4.59.0 based on https://github.com/palantir/jvm-diagnostics/pull/96

robert3005:
Looks like as of 2.11 error-prone requires java 11 https://github.com/google/error-prone/issues/2730

carterkozak (approved):

PR #2076, opened 2022-02-10
enable annotation processing in eclipse by default

rzpt:
## Before this PR
Annotation processing does not work in eclipse by default. It is not a feature in regular gradle, but there is a maintained plugin https://github.com/diffplug/goomph that can enable it.

## After this PR
==COMMIT_MSG==
enable annotation processing in eclipse by default
==COMMIT_MSG==

## Possible downsides?
None.


fawind (approved):
Thanks!

PR #2089, opened 2022-02-23
Use compileOnly configuration for immutables annotations

pkoenig10:


robert3005 (approved):

robert3005:
now we need to fix the tests to include the annotations explicitly?

PR #2108, opened 2022-03-08
baseline-java-versions uses library presets for projects using shadow-jar

carterkozak:
## Before this PR

```
Cannot change dependencies of dependency configuration ':PROJECT:runtimeElements' after it has been included in dependency resolution
```
## After this PR
==COMMIT_MSG==
baseline-java-versions uses library presets for projects using shadow-jar to prevent resolution/mutation ordering failures
==COMMIT_MSG==

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineJavaVersions.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineJavaVersions.java
@@ -40,6 +40,8 @@ public final class BaselineJavaVersions implements Plugin<Project> {
     public static final String EXTENSION_NAME = "javaVersions";
 
     public static final GradleVersion MIN_GRADLE_VERSION = GradleVersion.version("7.0");
+    // 'nebula.maven-publish' and 'com.palantir.shadow-jar' create publications lazily which cause inconsistencies
+    // based on ordering.
     private static final ImmutableSet<String> LIBRARY_PLUGINS =
             ImmutableSet.of("nebula.maven-publish", "com.palantir.shadow-jar");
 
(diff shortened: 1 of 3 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineJavaVersions.java
```diff
@@ -39,6 +40,10 @@
     public static final String EXTENSION_NAME = "javaVersions";
 
     public static final GradleVersion MIN_GRADLE_VERSION = GradleVersion.version("7.0");
+    // 'nebula.maven-publish' and 'com.github.johnrengelman.shadow' create publications lazily which cause
+    // inconsistencies based on ordering.
+    private static final ImmutableSet<String> LIBRARY_PLUGINS =
+            ImmutableSet.of("nebula.maven-publish", "com.github.johnrengelman.shadow");
```
CRogers: Is it actually the `johnrengelman` shadow or is it our `gradle-shadow-jar` which does this?

CRogers: Hmm, our one doesn't create publications - I'd be surprised if the `johnrengelman` did as well, as you have normally have to fiddle about with publications when using it.

CRogers:
:+1:

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineJavaVersions.java
carterkozak: You're right, I think this is on us due to ShadowJarPlugin `ensureShadowJarIsOnlyArtifactOnJavaConfigurations` + `dependOnJarTaskInOrderToTriggerTasksAddingManifestAttributes`

PR #2113, opened 2022-03-16
Prevent instance methods of auth tokens from getting called in logging args

fawind:
## Before this PR

Something like this would not get flagged by the existing error-prone check:

```java
void myMethod(AuthHeader header) {
    log.info("msg", SafeArg.of("h", header.toString()));
}
``` 

## After this PR

Instead of just checking that the type of the safe arg parameter is not an auth token, we can also verify that it is not the invocation of a method of an auth token.

Note that this only catches simple cases and can't handle indirections, e.g. `header.toString()` being called elsewhere and passed down to the arg. 

==COMMIT_MSG==
Prevent instance methods of auth tokens from getting called in logging args
==COMMIT_MSG==


carterkozak (approved):
thanks!

fawind:
👍

PR #2127, opened 2022-03-23
Implement `IllegalSafeLoggingArgument` check

carterkozak:
Prevent unsafe data from being passed as safe.

==COMMIT_MSG==
Implement `IllegalSafeLoggingArgument` check
==COMMIT_MSG==


fawind (approved):
Very nice!

PR #2133, opened 2022-03-25
improvement: suggested fixes for IllegalSafeLoggingArgument check

tpetracca:
Follow-up to #2127.
In cases where validation would fail specifically due to calls to `SafeArg#of` where the argument was marked as `Unsafe`, suggest using `UnsafeArg#of` instead

==COMMIT_MSG==
suggested fixes for IllegalSafeLoggingArgument check
==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/IllegalSafeLoggingArgument.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/IllegalSafeLoggingArgument.java
@@ -20,6 +20,8 @@ import com.google.auto.service.AutoService;
 import com.google.errorprone.BugPattern;
 import com.google.errorprone.VisitorState;
 import com.google.errorprone.bugpatterns.BugChecker;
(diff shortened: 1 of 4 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/IllegalSafeLoggingArgument.java
```diff
@@ -90,13 +97,25 @@ public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState
                             .setMessage(String.format(
                                     "Dangerous argument value: arg is '%s' but the parameter requires '%s'.",
                                     argumentSafety, parameterSafety))
+                            .addFix(getSuggestedFix(tree, state, argumentSafety))
```
carterkozak, line 98: We can remove the "future work" item from the class-level javadoc describing this feature :-)

carterkozak (approved):
lgtm, thanks!

PR #2136, opened 2022-03-28
Compilation supports add-exports/add-opens with toolchains

carterkozak:
Using workaround suggested by:
https://github.com/gradle/gradle/issues/18824#issuecomment-1026909824

==COMMIT_MSG==
Compilation supports add-exports/add-opens with toolchains
==COMMIT_MSG==


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
```diff
@@ -68,32 +68,31 @@ public void apply(Project project) {
 
             // javac isn't provided `--add-exports` args for the time being due to
             // https://github.com/gradle/gradle/issues/18824
-            if (project.hasProperty("add.exports.to.javac")) {
```
carterkozak, line 71: Large diff is the result of removing this conditional, the rest is formatting

fawind (approved):
Looks good! Let's see for how long this workaround will continue to work ^^

PR #2137, opened 2022-03-28
Javadoc tasks supporty BaselineModuleJvmArgs

carterkozak:
This unblocks the error-prone 2.11 upgrade.

==COMMIT_MSG==
Javadoc tasks supporty BaselineModuleJvmArgs
==COMMIT_MSG==

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
@@ -64,6 +64,7 @@ public final class BaselineModuleJvmArgs implements Plugin<Project> {
             Splitter.on(' ').trimResults().omitEmptyStrings();
 
     @Override
+    @SuppressWarnings("checkstyle:MethodLength")
     public void apply(Project project) {
         project.getPluginManager().withPlugin("java", unused -> {
             BaselineModuleJvmArgsExtension extension =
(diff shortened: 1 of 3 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
```diff
@@ -92,6 +96,62 @@ public void apply(Project project) {
...
+                                                "BaselineModuleJvmArgs building {} on {} "
+                                                        + "with exports: {} and opens: {}",
+                                                javadoc.getName(),
+                                                project,
+                                                exportValues,
+                                                opens);
+                                if (!exportValues.isEmpty()) {
+                                    coreOptions
+                                            // options are automatically prefixed with '-' internally
+                                            .addMultilineStringsOption("-add-exports")
```
fawind, line 136: Was wondering why we can't use `collectClasspathArgs` here which does a very similar colelction of the export and open args. Is it because of the `--` vs `-` distinction?
https://github.com/palantir/gradle-baseline/blob/159bb3811ffb9aa053e10a202d43c0e7827cb8e1/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java#L261

fawind (approved):

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
carterkozak: Sort of, `add-exports` only occurs once, and the values are collected on the right-hand-side.

PR #2139, opened 2022-03-28
prevent using logsafe Args with guava Preconditions

rzpt:
## Before this PR
Sometimes the wrong Preconditions is imported. This prevents one such case.

## After this PR
==COMMIT_MSG==
prevent using logsafe Args with guava Preconditions
==COMMIT_MSG==

## Possible downsides?
None?


carterkozak:
Mind linking me where this came up internally?

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferSafeLoggingPreconditions.java
```diff
@@ -77,23 +81,54 @@ public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState
...
+        List<? extends ExpressionTree> args = tree.getArguments();
+        if (args.size() > 2) {
+            if (PRECONDITIONS_MATCHER.matches(tree, state)) {
+                Description result = checkGuavaPreconditionsAndLogsafeArgMixing(tree, state, args);
+                if (result != null) {
+                    return result;
+                }
+            }
+            return Description.NO_MATCH;
+        } else if (args.size() == 2) {
```
carterkozak: It might be cleanest if we merge these branches, `allMatch` for zero additional args (`args.size() == 2`) appears to do the right thing

carterkozak (commented):
one suggestion, otherwise lgtm. Thanks!

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferSafeLoggingPreconditions.java
rzpt: I tried to do that but I couldn't find a way to do it without making the rule more aggressive.

carterkozak: Thoughts on this? #2166

carterkozak (approved):

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferSafeLoggingPreconditions.java
rzpt: Thanks! I thought I tried that and many tests failed.

PR #2140, opened 2022-03-28
prevent exceptions from being passed into Args

rzpt:
## Before this PR
Exceptions can be passed into SafeArg/UnsafeArg, which loses the stacktrace.

## After this PR
==COMMIT_MSG==
prevent exceptions from being passed into Args
==COMMIT_MSG==

## Possible downsides?
None.


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/LogsafeArgument.java
```diff
@@ -0,0 +1,63 @@
...
+import com.sun.source.tree.ExpressionTree;
+import com.sun.source.tree.MethodInvocationTree;
+import java.util.List;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "LogsafeArgument",
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
+        severity = SeverityLevel.ERROR,
```
carterkozak: Let's make this one a warning, at least for initial rollout
```suggestion
        severity = SeverityLevel.WARNING,
```

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/LogsafeArgument.java
```diff
@@ -0,0 +1,63 @@
...
+        summary = "Prevent certain arguments from being logged.")
+public final class LogsafeArgument extends BugChecker implements MethodInvocationTreeMatcher {
+    static final String UNSAFE_ARG_NAMES_FLAG = "LogsafeArgName:UnsafeArgNames";
+
+    private static final Matcher<ExpressionTree> SAFE_ARG_OF =
+            Matchers.staticMethod().onClass("com.palantir.logsafe.SafeArg").named("of");
+    private static final Matcher<ExpressionTree> UNSAFE_ARG_OF =
+            Matchers.staticMethod().onClass("com.palantir.logsafe.UnsafeArg").named("of");
+    private static final Matcher<ExpressionTree> THROWABLE = MoreMatchers.isSubtypeOf("java.lang.Throwable");
```
carterkozak: I think we can simplify this using a couple obscure error-prone utilities:

```suggestion
    private static final Matcher<ExpressionTree> MATCHER = Matchers.staticMethod()
            .onClassAny("com.palantir.logsafe.SafeArg", "com.palantir.logsafe.UnsafeArg")
            .named("of")
            .withParameters(String.class.getName(), Object.class.getName());
```

This way we don't need to do anything clever in the `matchMethodInvocation` body

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/LogsafeArgument.java
```diff
@@ -0,0 +1,63 @@
...
+    public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
+        if (!SAFE_ARG_OF.matches(tree, state) && !UNSAFE_ARG_OF.matches(tree, state)) {
+            return Description.NO_MATCH;
+        }
+
+        List<? extends ExpressionTree> args = tree.getArguments();
+        if (THROWABLE.matches(args.get(1), state)) {
+            return buildDescription(tree)
+                    .setMessage("Args with type Throwable are not allowed.")
+                    .build();
```
carterkozak: When there's a single failure message, we can set it in the top-level `@BugPattern` annotation `summary` field, then we can `return describeMatch(tree);` on failure.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineErrorProneExtension.java
```diff
@@ -50,6 +50,7 @@
             "LambdaMethodReference",
             "LoggerEnclosingClass",
             "LogsafeArgName",
+            "LogsafeArgument",
```
carterkozak: It's only necessary to update the extension with names of checks that are automatically fixed by default, no need to register this one since it doesn't have an automatic-fix.
```suggestion
```

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/LogsafeArgument.java
```diff
@@ -0,0 +1,63 @@
...
+import com.google.errorprone.matchers.Description;
+import com.google.errorprone.matchers.Matcher;
+import com.google.errorprone.matchers.Matchers;
+import com.sun.source.tree.ExpressionTree;
+import com.sun.source.tree.MethodInvocationTree;
+import java.util.List;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        name = "LogsafeArgument",
```
carterkozak: Sorry for the delay, I think we'll need to rebase and remove this line (due to the error_prone 2.11 upgrade)
```suggestion
```

carterkozak (commented):
generally lgtm, just a couple suggestions!

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/LogsafeArgument.java
```diff
@@ -0,0 +1,63 @@
+/*
+ * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
```
carterkozak:
```suggestion
 * (c) Copyright 2022 Palantir Technologies Inc. All rights reserved.
```

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/LogsafeArgumentTest.java
```diff
@@ -0,0 +1,57 @@
+/*
+ * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
```
carterkozak:
```suggestion
 * (c) Copyright 2022 Palantir Technologies Inc. All rights reserved.
```

carterkozak (approved):

PR #2142, opened 2022-03-28
Simplify errorprone version constraint

pkoenig10:
Possible because we no longer pull in `com.google.errorprone:javac` or `com.google.errorprone:javac-shaded`.

### DIFF
--- a/versions.props
+++ b/versions.props
@@ -1,6 +1,6 @@
 com.diffplug.spotless:spotless-plugin-gradle = 6.4.0
 com.google.auto.service:auto-service = 1.0.1
-com.google.errorprone:error_prone_* = 2.11.0
+com.google.errorprone:* = 2.11.0
 com.google.guava:guava = 31.1-jre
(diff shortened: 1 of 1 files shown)


versions.props
```diff
@@ -1,6 +1,6 @@
 com.diffplug.spotless:spotless-plugin-gradle = 6.4.0
 com.google.auto.service:auto-service = 1.0.1
-com.google.errorprone:error_prone_* = 2.11.0
+com.google.errorprone:* = 2.11.0
```
carterkozak, line 3: What does this buy us? It may confuse our version upgrade excavators

pkoenig10: How so?

pkoenig10: As long as our excavators don't try to upgrade `com.google.errorprone:javac` (which I don't think they do), this should be fine.

carterkozak: Why introduce the potential problem?

PR #2143, opened 2022-03-29
Safety data-flow

carterkozak:
==COMMIT_MSG==
Implement Safety flow checks
==COMMIT_MSG==

This is likely fairly slow, and probably needs optimization.
Needs more test coverage.

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/safety/SafetyPropagationTransfer.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/safety/SafetyPropagationTransfer.java
@@ -34,6 +34,7 @@ import com.sun.tools.javac.code.Flags;
 import com.sun.tools.javac.code.Symbol;
 import com.sun.tools.javac.code.Symbol.MethodSymbol;
 import com.sun.tools.javac.code.Symbol.VarSymbol;
+import com.sun.tools.javac.code.Type;
 import com.sun.tools.javac.processing.JavacProcessingEnvironment;
 import java.io.Closeable;
 import java.util.HashMap;
@@ -636,7 +637,8 @@ public final class SafetyPropagationTransfer implements ForwardTransferFunction<
     public TransferResult<Safety, AccessPathStore<Safety>> visitFieldAccess(
             FieldAccessNode node, TransferInput<Safety, AccessPathStore<Safety>> input) {
         Safety fieldSafety = SafetyAnnotations.getSafety(ASTHelpers.getSymbol(node.getTree()), state);
-        Safety typeSafety = SafetyAnnotations.getSafety(ASTHelpers.getType(node.getTree()).tsym, state);
+        Type fieldType = ASTHelpers.getType(node.getTree());
+        Safety typeSafety = fieldType == null ? Safety.UNKNOWN : SafetyAnnotations.getSafety(fieldType.tsym, state);
         VarSymbol symbol = (VarSymbol) ASTHelpers.getSymbol(node.getTree());
         AccessPath maybeAccessPath = AccessPath.fromFieldAccess(node);
         Safety flowSafety = fieldSafety(symbol, maybeAccessPath, input.getRegularStore());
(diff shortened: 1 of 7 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/safety/SafetyPropagationTransfer.java
```diff
@@ -0,0 +1,854 @@
...
+        if (target instanceof LocalVariableNode) {
+            updates.trySet(target, safety);
+        }
+        if (target instanceof ArrayAccessNode) {
+            updates.trySet(((ArrayAccessNode) target).getArray(), safety);
+        }
+        if (target instanceof FieldAccessNode) {
+            FieldAccessNode fieldAccess = (FieldAccessNode) target;
+            updates.set(fieldAccess, safety);
+        }
```
fawind, line 532: Is this a complete list or should we handle the case when target is not an instance of these three types?

fawind: I guess with the current impl, we just update the store with an empty update and just stop propagating safety information which sounds fine!

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/safety/SafetyPropagationTransfer.java
```diff
@@ -0,0 +1,854 @@
...
+                return Safety.mergeAssumingUnknownIsSame(symbolAndTypeSafety, initializer);
+            }
+        }
+        return symbolAndTypeSafety;
+    }
+
+    private Safety fieldInitializerSafetyIfAvailable(VarSymbol accessed) {
+        if (!traversed.add(accessed)) {
+            // Initializer circularity
+            return Safety.UNKNOWN;
```
fawind, line 616: Can you elaborate a bit on this traversed guard? Is it to avoid loops when running the flow on the field initializer?

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/safety/SafetyPropagationTransfer.java
```diff
@@ -0,0 +1,854 @@
...
+        Safety valueSafety = input.getValueOfSubNode(node.getOperand());
+        Type widenTarget = ASTHelpers.getType(node.getTree());
+        Safety widenTargetSafety =
+                widenTarget == null ? Safety.UNKNOWN : SafetyAnnotations.getSafety(widenTarget.tsym, state);
+        Safety resultSafety = Safety.mergeAssumingUnknownIsSame(valueSafety, widenTargetSafety);
+        return noStoreChanges(resultSafety, input);
+    }
+
+    @Override
+    public TransferResult<Safety, AccessPathStore<Safety>> visitNarrowingConversion(
```
fawind, line 717: nit: `visitWideningConversion` and `visitNarrowingConversion` uses the same code. Maybe we can extract or delegate?

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/safety/SafetyPropagationTransfer.java
```diff
@@ -0,0 +1,854 @@
...
+        }
+        if (target instanceof FieldAccessNode) {
+            FieldAccessNode fieldAccess = (FieldAccessNode) target;
+            updates.set(fieldAccess, safety);
+        }
+        return updateRegularStore(safety, input, updates);
+    }
+
+    @SuppressWarnings("checkstyle:CyclomaticComplexity")
+    private Safety fieldSafety(
```
fawind, line 575: nit: These two methods are a bit out-of-place. Maybe we move them further down below the method that uses them?

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/safety/SafetyPropagationTransfer.java
```diff
@@ -0,0 +1,854 @@
...
+        }
+    }
+
+    private static TransferResult<Safety, AccessPathStore<Safety>> noStoreChanges(
+            Safety value, TransferInput<?, AccessPathStore<Safety>> input) {
+        return new RegularTransferResult<>(value, input.getRegularStore());
+    }
+
+    @CheckReturnValue
+    private static TransferResult<Safety, AccessPathStore<Safety>> updateRegularStore(
```
fawind, line 176: This took me a while to grok so think a comment on when we need to update the store and when we don't would be helpful!

fawind (commented):
Generally looks good, left some smaller comments. But happy to roll this out and iterate based on what we see on the excavators!

fawind:
👍

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/safety/SafetyPropagationTransfer.java
carterkozak: Yep, it's an exhaustive list:
https://github.com/typetools/checker-framework/blob/090d02424e6d825d45406b959010dd0ac81fbac2/dataflow/src/main/java/org/checkerframework/dataflow/cfg/node/AssignmentNode.java#L37-L39

We might as well fail in the case of an unknown target, that way we're forced to update the code to handle new types if they're added.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/safety/SafetyPropagationTransfer.java
carterkozak: :+1: I'll include `visitTypeCast` as well

PR #2148, opened 2022-03-30
Fix BaselineModuleJvmArgs without toolchains

carterkozak:
Otherwise the plugin interacts poorly with
`BaselineReleaseCompatibility` because `--release` and internal
module exports/opens cannot be used at the same time.

==COMMIT_MSG==
Fix BaselineModuleJvmArgs without toolchains
==COMMIT_MSG==

### DIFF
--- a/gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineModuleJvmArgsIntegrationTest.groovy
+++ b/gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineModuleJvmArgsIntegrationTest.groovy
@@ -131,6 +131,44 @@ class BaselineModuleJvmArgsIntegrationTest extends IntegrationSpec {
         result.standardOutput.contains('--add-exports=java.management/sun.management=ALL-UNNAMED')
     }
 
+    def 'Runs with locally defined exports with the release plugin, not toolchains'() {
+        when:
+        buildFile.text = '''
+        plugins {
+            id 'java-library'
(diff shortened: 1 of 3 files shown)


gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineModuleJvmArgsIntegrationTest.groovy
```diff
@@ -131,6 +131,44 @@ class BaselineModuleJvmArgsIntegrationTest extends IntegrationSpec {
...
+            }
+        }
+        '''.stripIndent(true))
+
+        then:
+        ExecutionResult result = runTasksSuccessfully('run')
+        // Gradle appears to normalize args, joining '--add-exports java.management/sun.management=ALL-UNNAMED'
+        // with an equals.
+        result.standardOutput.contains('--add-exports=java.management/sun.management=ALL-UNNAMED')
+    }
```
carterkozak, line 170: At the moment this test does not reproduce the problem because we run baseline with a java 11 toolchain.
I verified the fix using #2149 locally, but I didn't include it in this change as it's likely to impact other tests.

fawind: Merged in develop to include running tests on Java 15 now.

fawind (approved):

PR #2151, opened 2022-03-30
Throwable.getMessage is unsafe by default

carterkozak:
==COMMIT_MSG==
Throwable.getMessage is unsafe by default
==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/safety/SafetyPropagationTransfer.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/safety/SafetyPropagationTransfer.java
@@ -142,6 +142,9 @@ public final class SafetyPropagationTransfer implements ForwardTransferFunction<
             .namedAnyOf("getMessage", "getLocalizedMessage")
             .withNoParameters();
 
+    private static final Matcher<ExpressionTree> STRING_FORMAT =
+            MethodMatchers.staticMethod().onClass(String.class.getName()).named("format");
+
     private VisitorState state;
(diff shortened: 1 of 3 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/safety/SafetyPropagationTransfer.java
```diff
@@ -137,6 +137,11 @@
     private static final Matcher<ExpressionTree> TO_STRING =
             MethodMatchers.instanceMethod().anyClass().named("toString").withNoParameters();
 
+    private static final Matcher<ExpressionTree> THROWABLE_GET_MESSAGE = MethodMatchers.instanceMethod()
+            .onDescendantOf(Throwable.class.getName())
+            .named("getMessage")
```
schlosna: we should also treat [`Throwable#getLocalizedMessage()`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/Throwable.html#getLocalizedMessage()) as unsafe as well

carterkozak: Great idea, updated! Let me know if you think of other types/methods which are always unsafe.

`Throwable.toString()` is unsafe as well, but I'd like to handle that differently later on in a way that captures string covnersion e.g. `"exception: " + myException;`

fawind (approved):
Looks good! Can follow up if we think of more types/methods that are always unsafe.

PR #2152, opened 2022-03-30
Ban java deserialization

carterkozak:
https://cwe.mitre.org/data/definitions/502.html

==COMMIT_MSG==
Ban java deserialization
==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DangerousJavaDeserialization.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DangerousJavaDeserialization.java
@@ -1,0 +1,65 @@
+/*
+ * (c) Copyright 2022 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


Cosmin1490 (approved):
👍

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DangerousJavaDeserialization.java
```diff
@@ -0,0 +1,65 @@
...
+
+    private static final Matcher<ExpressionTree> LANG3_SERIALIZATION_UTILS_DESERIALIZE = MethodMatchers.instanceMethod()
+            .onExactClassAny(
+                    "org.apache.commons.lang.SerializationUtils",
+                    "org.apache.commons.lang3.SerializationUtils",
+                    "org.springframework.util.SerializationUtils")
+            .named("deserialize");
+
+    private static final Matcher<ExpressionTree> DESERIALIZE =
+            Matchers.anyOf(OBJECT_INPUT_READ_OBJECT, LANG3_SERIALIZATION_UTILS_DESERIALIZE);
```
schlosna, line 63: should we also ban `ObjectOutput#writeObject` and the `SerializationUtils#serialize` permutations as well under the assumptions that if someone is writing the bytes, someone somewhere will read them

carterkozak: Perhaps, but I'd like to make sure this doesn't cause issues with spark code that we expect to use java serialization before expanding the breadth of the check.

tpetracca (approved):

PR #2157, opened 2022-04-01
Fix switch expression handling

carterkozak:
This upgrades checkerframework to handle new language features.
baseline-error-prone tests are run on java 15. In a flup I plan
to bump this to java 17.

==COMMIT_MSG==
Fix switch expression handling
==COMMIT_MSG=


baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/IllegalSafeLoggingArgumentTest.java
```diff
@@ -723,6 +723,24 @@ public void testArraySafety() {
...
+                        "  void f(Nums value) {",
+                        "    fun(switch (value) {",
+                        "        case ONE -> 1;",
+                        "        case TWO -> 2;",
+                        "    });",
+                        "  }",
+                        "  private static void fun(@Safe Object obj) {}",
+                        "}")
+                .doTest();
+    }
```
carterkozak, line 742: This test successfully reproduced the failure

PR #2159, opened 2022-04-01
Add known safety-passthrough methods

carterkozak:
Methods which result in safety based on their arguments or
receiver. Certainly a lot more we can add here, but I tried
to hit the highlights!

==COMMIT_MSG==
Add known safety-passthrough methods
==COMMIT_MSG==


fawind (approved):
Nice!

PR #2160, opened 2022-04-01
Validate annotated variable and field assignment safety

carterkozak:
==COMMIT_MSG==
Validate annotated variable and field assignment safety
==COMMIT_MSG==


fawind (approved):
Looks good!

PR #2161, opened 2022-04-01
Validate that safety var annotations and type annotations agree

carterkozak:
==COMMIT_MSG==
Validate that safety var annotations and type annotations agree
==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/safety/SafetyPropagationTransfer.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/safety/SafetyPropagationTransfer.java
@@ -191,7 +191,7 @@ public final class SafetyPropagationTransfer implements ForwardTransferFunction<
     private static final Matcher<ExpressionTree> OPTIONAL_ACCESSORS = Matchers.anyOf(
             MethodMatchers.instanceMethod()
                     .onDescendantOf(Optional.class.getName())
-                    .namedAnyOf("get", "orElseThrow", "stream"),
+                    .namedAnyOf("filter", "get", "orElseThrow", "stream"),
             MethodMatchers.instanceMethod()
                     .onDescendantOf(OptionalInt.class.getName())
                     .namedAnyOf("getAsInt", "orElseThrow"),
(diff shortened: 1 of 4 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/safety/SafetyPropagationTransfer.java
```diff
@@ -173,9 +177,30 @@
...
+
     // These methods do not take the receiver (generally a static class) into account, only the inputs.
     private static final Matcher<ExpressionTree> RETURNS_SAFETY_COMBINATION_OF_ARGS =
-            Matchers.anyOf(STRING_FORMAT, OBJECTS_TO_STRING, IMMUTABLE_COLLECTION_FACTORY);
+            Matchers.anyOf(STRING_FORMAT, OBJECTS_TO_STRING, IMMUTABLE_COLLECTION_FACTORY, OPTIONAL_FACTORIES);
+
+    private static final Matcher<ExpressionTree> OPTIONAL_ACCESSORS = Matchers.anyOf(
+            MethodMatchers.instanceMethod()
+                    .onDescendantOf(Optional.class.getName())
+                    .namedAnyOf("get", "orElseThrow", "stream"),
```
schlosna: Do we need to handle `filter`, `orElse`, `orElseGet`?

Do we need to handle `Stream`-like `map` and `flatMap`?

carterkozak: I'd like to, but I haven't included those in the initial optional support -- I'll need to implement something to determine safety of map/flatmap transformer results, and merge that with the safety of the receiver.
I want to avoid considering something like this safe:
```java
@Safe Object f(@Safe @Nullable String input, @DoNotLog password) {
    return Optional.ofNullable(input).orElseGet(() -> password);
}
```
It wouldn't be the end of the world to consider it safe rather than unknown since we don't currently suggest replacing UnsafeArg with SafeArg, but I'd like to try handling it based on all the data available first if I can.

schlosna: makes sense as a separate follow up. I think we should be able to handle filter as-is since its not widening or transforming the types.

```suggestion
                    .namedAnyOf("filter", "get", "orElseThrow", "stream"),
```

schlosna (approved):
looks good, had one follow up question/suggestion

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/safety/SafetyPropagationTransfer.java
carterkozak: Great idea, thanks!

PR #2167, opened 2022-04-01
Compilation uses 'opens' values as 'exports'

carterkozak:
Opens are ignored at compilation time.

==COMMIT_MSG==
Compilation uses 'opens' values as 'exports'
==COMMIT_MSG==

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
@@ -128,7 +128,7 @@ public final class BaselineModuleJvmArgs implements Plugin<Project> {
                                 CoreJavadocOptions coreOptions = (CoreJavadocOptions) options;
                                 ImmutableList<JarManifestModuleInfo> info = collectClasspathInfo(project, sourceSet);
                                 List<String> exportValues = Stream.concat(
-                                                // Compilation only supports exports, so we with opens.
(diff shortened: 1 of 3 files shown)


carterkozak:
This unblocks jdk-17 testing in #2165

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
```diff
@@ -130,39 +128,29 @@ public void execute(Task task) {
...
-                                                extension.exports().get().stream(),
-                                                info.stream().flatMap(item -> item.exports().stream()))
-                                        .distinct()
-                                        .sorted()
-                                        .map(item -> item + "=ALL-UNNAMED")
-                                        .collect(ImmutableList.toImmutableList());
-                                List<String> opens = Stream.concat(
-                                                extension.opens().get().stream(),
-                                                info.stream().flatMap(item -> item.opens().stream()))
+                                                // Compilation only supports exports, so we with opens.
```
schlosna:
```suggestion
                                                // Compilation only supports exports, so we union with opens.
```

schlosna (approved):
One comment nit suggestion

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
carterkozak: 🤦 thanks!

carterkozak:
👍

PR #2169, opened 2022-04-01
Avoid nulls in safety analysis

carterkozak:
==COMMIT_MSG==
Avoid nulls in safety analysis
==COMMIT_MSG==


schlosna (approved):
Looks reasonable. Should we consider annotating `@Nullable` args & methods?

PR #2174, opened 2022-04-04
Support safety passthrough on several streamlike methods

carterkozak:
==COMMIT_MSG==
Support safety passthrough on several streamlike methods
==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/safety/SafetyPropagationTransfer.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/safety/SafetyPropagationTransfer.java
@@ -49,6 +49,7 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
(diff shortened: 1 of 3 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/safety/SafetyPropagationTransfer.java
```diff
@@ -202,6 +212,28 @@
...
+                            "limit",
+                            "skip",
+                            "takeWhile",
+                            "dropWhile",
+                            "toArray",
+                            "toList",
+                            "min",
+                            "max",
+                            "findFirst",
+                            "findAny"));
```
fawind, line 235: Shouldn't this be every instance method on `Stream` that returns another `Stream` or even 'Collection'? Maybe we can list those methods using reflection instead of listing them manually?

fawind (approved):
Looks good!

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/safety/SafetyPropagationTransfer.java
carterkozak: Not quite, the list does not include `map`/`flatMap`/`mapTo*` because those methods may modify the safety of a stream. e.g. `streamOfUnsafeObject.map(unsafeObject -> unsafeObject.getSafeField())` is safe. Lambda handling for this sort of thing is super duper tough though, and I'm still trying to figure out the best way to do it.

PR #2187, opened 2022-04-08
Support safety annotations on type-use e.g. `Collection<@Safe String>`

carterkozak:
==COMMIT_MSG==
Support safety annotations on type-use e.g. `Collection<@Safe String>`
==COMMIT_MSG==

Depends on https://github.com/palantir/safe-logging/pull/699

## Possible downsides?
This can get tricky -- I haven't implemented anything yet to fail when a `List<@Unsafe String>` is passed to a location that requires `@Safe`. It may be confusing to have the ability to describe a piece of data as `@Unsafe List<String>` or `List<@Unsafe String>`, however the upside this feature provides is the ability to handle more complex types cleanly. Consider a map of userId to logins, both of which are strings: `Map<@Safe String, @Unsafe String>` could be used such that `map.get(key)` is known to be unsafe, while `map.keySet().iterator().next()` is safe.


fawind (approved):
Nifty!

PR #2193, opened 2022-04-12
Configurable java toolchains

CRogers:
## Before this PR
We use Gradle's [in built java toolchain support ](https://docs.gradle.org/current/userguide/toolchains.html). However, it has many downsides:
1. Autoprovisioning of JDKs downloads via AdoptOpenJDK/Adoptium, which is not our preferred JDK vendor (Azul is).
    * We use Azul, not Adoptium, in prod.
    * Adoptium does not provide aarch64 macos builds for Java 11 and 15.
3. There is no way to specify the exact version of the JDK that is downloaded
   * JDKs are downloaded once per major version then never again.
   * There is no way to ensure 100% reproducible builds by ensure the same JDKs are used each time.
   * We can't line up the JDKs used by our test tasks with the JDKs we actually deploy in our prod docker images - this means for JDK bugs or behaviour changes across minor versions unit tests do not accurately test the same stuff as the deployed asset.
4. There is no ability or extension point to add custom CA certs in sensible way, for interacting with a corporate firewall that does TLS inception for example.
5. The Adoptium servers are used for downloading rather than being able to specify a custom internal mirror.
    * Firewalls can slow down large downloads from outside corporate networks.
    * Downloads over public internet is slow compared to using an internal mirror.

## After this PR
Add a configuration point so plugins can supply their own JDKs per java major version:

```java
MapProperty<JavaLanguageVersion, JavaInstallationMetadata> getJdks()
```

Another plugin I am yet to write will download JDKs based on some configuration from various different sources eg Azul Zulu, Amazon Corretto etc


gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineJavaVersionExtension.java
```diff
@@ -26,19 +31,33 @@
...
+        overrideLibraryAutoDetection = project.getObjects().property(Boolean.class);
         overrideLibraryAutoDetection.finalizeValueOnRead();
+
+        zuluVersions = project.getObjects().mapProperty(JavaLanguageVersion.class, String.class);
+        zuluVersions.putAll(project.provider(() -> parseOutVersions(project, ZULU_VERSION_PATTERN)));
+        zuluVersions.finalizeValueOnRead();
+
+        javaVersions = project.getObjects().mapProperty(JavaLanguageVersion.class, String.class);
+        javaVersions.putAll(project.provider(() -> parseOutVersions(project, JAVA_VERSION_PATTERN)));
+        javaVersions.finalizeValueOnRead();
```
carterkozak: Do you think we could define these at most once per repo instead of per-module? (possibly `BaselineJavaVersionsExtension.java`, which only exists on the root project)
I don't think we would want the values to differ across modules.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/javaversions/JdkManager.java
```diff
@@ -0,0 +1,76 @@
...
+                FileUtils.deleteDirectory(temporaryJdkPath.toFile());
+            } catch (IOException e) {
+                throw new RuntimeException("Failed to delete temporary downloaded jdk " + temporaryJdkPath, e);
+            }
+        }
+    }
+
+    private Path findJavaHome(Path temporaryJdkPath) {
+        try (Stream<Path> files = Files.walk(temporaryJdkPath)) {
+            return files.filter(file -> Files.isDirectory(file) && file.endsWith(Paths.get("Contents/Home")))
```
carterkozak: I think only the mac jdks use this directory structure. We may want to  find `bin/java` and walk back a directory

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/javaversions/JdkManager.java
```diff
@@ -0,0 +1,76 @@
...
+                jdkPath + ".in-progress-" + UUID.randomUUID().toString().substring(0, 8));
+        try {
+            try {
+                archiver.extract(jdkArchive.toFile(), temporaryJdkPath.toFile());
+            } catch (IOException e) {
+                throw new RuntimeException("Failed to extract jdk to directory", e);
+            }
+
+            try {
+                Files.move(findJavaHome(temporaryJdkPath), jdkPath, StandardCopyOption.ATOMIC_MOVE);
```
carterkozak: atomic move is great, however in some cases it's not possible (depending on the filesystem, or the src and destination are on different physical disks). Unsure if that will be an issue here.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/javaversions/JdkManager.java
```diff
@@ -0,0 +1,76 @@
...
+
+        Archiver archiver = ArchiverFactory.createArchiver(jdkArchive.toFile());
+        Path temporaryJdkPath = Paths.get(
+                jdkPath + ".in-progress-" + UUID.randomUUID().toString().substring(0, 8));
+        try {
+            try {
+                archiver.extract(jdkArchive.toFile(), temporaryJdkPath.toFile());
+            } catch (IOException e) {
+                throw new RuntimeException("Failed to extract jdk to directory", e);
+            }
```
carterkozak: I wonder if it would be possible/beneficial to leverage [gradle cacheable artifact transforms](https://docs.gradle.org/current/userguide/artifact_transforms.html)?
That way we could reuse the same unpacked jdk across projects/checkouts instead of extracting in each repo.

CRogers: I was actually just going to extract to `~/.gradle/gradle-baseline/jdks` to reuse jdks across repos and depend on the using a hash to keep them the same - but this is worth investigating, particularly as I wouldn't have to write all this caching code myself!

gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineJavaVersionExtension.java
CRogers: Heh actually in the process of doing this right now. The `BaselineJavaVersion` vs `BaselineJavaVersions` difference was lost on me for a little bit.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/javaversions/JdkManager.java
CRogers: Filesystems are scary

gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineJavaVersionExtension.java
carterkozak: I could have named things much better :disappointed: Sorry!

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/javaversions/JdkManager.java
CRogers: Mmmm, I actually only had to do this because the commons-compress or jarchivelib don't support symlinks, maybe with an artifact transform I wouldn't need to do this

gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineJavaVersionsExtension.java
```diff
@@ -40,9 +42,13 @@ public BaselineJavaVersionsExtension(Project project) {
         distributionTarget.convention(libraryTarget);
         // runtime defaults to the distribution value
         runtime.convention(distributionTarget);
+        runtime.convention(distributionTarget);
```
carterkozak: I think this duplicate line is a typo

gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineJavaVersionsExtension.java
```diff
@@ -18,18 +18,20 @@
...
 /**
  * Extension named {@code javaVersions} on the root project used to configure all java modules
  * with consistent java toolchains.
  */
 public class BaselineJavaVersionsExtension {
-
     private final Property<JavaLanguageVersion> libraryTarget;
     private final Property<JavaLanguageVersion> distributionTarget;
     private final Property<JavaLanguageVersion> runtime;
+    private final MapProperty<JavaLanguageVersion, JavaInstallationMetadata> jdks;
```
carterkozak, line 34: I wonder if the map will come back to haunt us? I suspect our jdk repository layouts will be predictable as new language versions are released, we may be able to provide something like a function from `JavaLanguageVersion` -> `Provider<JavaInstallationMetadata>` without requiring custom/internal toolchain plugins to be updated for every release.

carterkozak (commented):
I like it :-]

gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineJavaVersionsExtension.java
CRogers: A map is a function!

But seriously, this is effectively the same as registering a function with this class then calling it to work out the metadata. Pretty sure that's how `MapProperty.put(All)` works under the scenes.

carterkozak (approved):
lgtm!

PR #2205, opened 2022-04-18
LoggingDoNotLog check ensures credentials aren't logged in any form

carterkozak:
Loggers themselves are obvious, however exception messages are less
obvious and equally dangerous.

==COMMIT_MSG==
LoggingDoNotLog check ensures credentials aren't logged in any form
==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/LoggingDoNotLog.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/LoggingDoNotLog.java
@@ -1,5 +1,5 @@
 /*
- * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ * (c) Copyright 2022 Palantir Technologies Inc. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/LoggingDoNotLog.java
```diff
@@ -0,0 +1,119 @@
...
+            // MDC interactions result in data in the logs
+            MethodMatchers.staticMethod().onClass("org.slf4j.MDC"),
+            MethodMatchers.staticMethod().onClass("org.apache.log4j.MDC"),
+            MethodMatchers.staticMethod().onClass("org.apache.logging.log4j.ThreadContext"));
+
+    // Result in indirect logging
+    private static final Matcher<ExpressionTree> PRECONDITIONS_METHODS = Matchers.anyOf(
+            MethodMatchers.staticMethod()
+                    .onClass("com.google.common.base.Preconditions")
+                    .namedAnyOf("checkArgument", "checkNotNull", "checkState"),
```
fawind, line 77: Should this just be any method? E.g. there is also `checkElementIndex` and `checkPositionIndex`.

fawind (approved):
Looks good!

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/LoggingDoNotLog.java
carterkozak: I tried to cover the most common methods, supporting others requires a different argument index offset. Could be a future improvement to this check.

PR #2210, opened 2022-04-19
Add check for Immutables reference equality

pkoenig10:
The `ReferenceEquality` check does not apply when using an Immutables definition because the definition does not [implement equals](https://github.com/google/error-prone/blob/f174a804a74b7ef413f946cd44977ad3059c2eee/core/src/main/java/com/google/errorprone/bugpatterns/ReferenceEquality.java#L68-L70).

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ImmutablesReferenceEquality.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ImmutablesReferenceEquality.java
@@ -1,0 +1,49 @@
+/*
+ * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.palantir.baseline.errorprone;
+
+import com.google.auto.service.AutoService;
+import com.google.errorprone.BugPattern;
+import com.google.errorprone.BugPattern.LinkType;
+import com.google.errorprone.BugPattern.SeverityLevel;
+import com.google.errorprone.VisitorState;
+import com.google.errorprone.bugpatterns.AbstractReferenceEquality;
+import com.google.errorprone.bugpatterns.BugChecker;
+import com.google.errorprone.util.ASTHelpers;
+import com.sun.source.tree.ExpressionTree;
+import com.sun.tools.javac.code.Symbol.ClassSymbol;
+import com.sun.tools.javac.code.Type;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        linkType = LinkType.CUSTOM,
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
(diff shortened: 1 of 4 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ImmutablesReferenceEquality.java
```diff
@@ -0,0 +1,49 @@
+/*
+ * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
```
carterkozak:
```suggestion
 * (c) Copyright 2022 Palantir Technologies Inc. All rights reserved.
```

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/ImmutablesReferenceEqualityTest.java
```diff
@@ -0,0 +1,42 @@
+/*
+ * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
```
carterkozak:
```suggestion
 * (c) Copyright 2022 Palantir Technologies Inc. All rights reserved.
```

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/ImmutablesReferenceEqualityTest.java
```diff
@@ -0,0 +1,42 @@
...
+                        "class Test {",
+                        "  static boolean f(Foo foo1, Foo foo2) {",
+                        "    // BUG: Diagnostic contains: foo1.equals(foo2)",
+                        "    return foo1 == foo2;",
+                        "  }",
+                        "  @Value.Immutable",
+                        "  interface Foo {}",
+                        "}")
+                .doTest();
+    }
```
carterkozak, line 37: Mind adding a test case that shows the automatic fix? Should we apply suggested fixes by default?

pkoenig10: > Should we apply suggested fixes by default?

We should be consistent with the `ReferenceEquality` check. We don't currently apply that check by default, which is probably the right choice.

I would hope that there aren't (m)any existing places where we're using reference equality on immutables values, so I didn't apply this by default.

pkoenig10: So it's a bit weird to test the automatic fixes here because the `AbstractReferenceEquality` check actually suggests two fixes.
```
[ImmutablesReferenceEquality] Comparison of Immutables value using reference equality instead of value equality.
Did you mean 'return Objects.equals(foo1, foo2);' or 'return foo1.equals(foo2);'?
```

When applying fixes, Error Prone will [select the first one](https://github.com/google/error-prone/blob/f174a804a74b7ef413f946cd44977ad3059c2eee/check_api/src/main/java/com/google/errorprone/apply/DescriptionBasedDiff.java#L91-L94). In this case, that's `Objects.equals(foo1, foo2);` which probably isn't the one we want.

In tests you can [specify a `FixChooser`](https://github.com/google/error-prone/blob/f174a804a74b7ef413f946cd44977ad3059c2eee/test_helpers/src/main/java/com/google/errorprone/BugCheckerRefactoringTestHelper.java#L323) to pick a specific fix, but using that feels a bit gross here.

FWIW the Error Prone `ReferenceEquality` tests don't test the automatic fixes, probably for this reason. I don't think it's that important to test the automatic fixes here, since the automatic fixes aren't part of baseline code.

carterkozak: I like Objects.equals, but no strong opinion. Just don't want baseline upgrades to get stuck, but it's not likely many folks have this issue.

carterkozak (approved):

PR #2215, opened 2022-04-21
Fix: Include mocktio-errorprone after it has picked up new errorprone

robert3005:
## After this PR

==COMMIT_MSG==
==COMMIT_MSG==


baseline-error-prone/build.gradle
```diff
@@ -7,8 +7,7 @@ dependencies {
     implementation 'com.google.errorprone:error_prone_core'
     // Ensure a new enough version of dataflow-errorprone is available
     implementation 'org.checkerframework:dataflow-errorprone'
-    // bring back mockito-errorprone once it is updated for 2.11.0+
-    //implementation 'org.mockito:mockito-errorprone'
+    implementation 'org.mockito:mockito-errorprone'
```
carterkozak: Do we want `runtimeOnly` here?

robert3005: yes, I blindly uncommented but that's right

carterkozak (approved):

PR #2218, opened 2022-04-22
Require safety propagation based on getters

carterkozak:
==COMMIT_MSG==
Require log-safety annotation propagation based detected getters and superclasses/superinterfaces
==COMMIT_MSG==

We combine the safety of superclasses, as well as getters (non-static methods with zero parameters and non-void return type) to estimate the safety of a class. We only apply unsafe/do-not-log annotations as this heuristic is not exhaustive, so we bias toward cases in which we're confident we can be safer.


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/SafeLoggingPropagation.java
```diff
@@ -0,0 +1,115 @@
...
+import com.sun.source.tree.Tree;
+import com.sun.tools.javac.code.Symbol.ClassSymbol;
+import javax.lang.model.element.Modifier;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
+        // This will be promoted after an initial rollout period
+        severity = BugPattern.SeverityLevel.SUGGESTION,
```
carterkozak, line 43: This is currently `SUGGESTION` so I can excavate fixes in a way that's decoupled from baseline upgrades.

fawind (approved):
Nice!

PR #2230, opened 2022-04-26
Propagate additional safety information

carterkozak:
==COMMIT_MSG==
Propagate additional safety information in the `SafeLoggingPropagation` check and automated fixes:
1. Method return statements are analyzed to determine safety of unmarked methods
2. Types are annotated based on the safety of their `toString` method, which is a reasonable heuristic for value types that may be logged.
==COMMIT_MSG==

These two checks work together nicely when types define a `toString`: the inputs to the toString (assuming string concatenation, we haven't implemented anything special for StringBuilder accumulation) are combined into an annotation on the toString method, which is applied at the type level in the next run.

## Possible downsides?
This may result in a lot of annotations.
Our automation can restrict safety, but i never eases safety -- we can mark a method `@Unsafe` or `@DoNotLog`, but never `@Safe`.


fawind (approved):
Makes sense!

PR #2237, opened 2022-04-28
Safety propagation ignores immutables helpers

carterkozak:
For example:
```
@Value.Immutable
interface Context {
  AuthService authService();

  // not a field, only a utility method
  default BearerToken token() {
    return authService().currentToken();
  }
}
```

==COMMIT_MSG==
Safety propagation ignores utility methods on immutables definitions
==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/SafeLoggingPropagationTest.java
+++ b/baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/SafeLoggingPropagationTest.java
@@ -16,6 +16,7 @@
 
 package com.palantir.baseline.errorprone;
 
+import org.immutables.value.Value;
 import org.junit.jupiter.api.Test;
 
 class SafeLoggingPropagationTest {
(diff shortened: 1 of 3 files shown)


baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/SafeLoggingPropagationTest.java
```diff
@@ -249,13 +249,284 @@ void testIgnoresStaticMethods() {
...
+                        "interface Test {",
+                        "  @DoNotLog",
+                        "  @Value.Lazy",
+                        "  default String token() { return \"\"; }",
+                        "}")
+                .doTest();
+    }
+
+    @Test
+    void testIgnoresHelperMethods_iface() {
```
tpetracca, line 413: Is this just the same as the `ignoresDefaultMethod` test on line 275?

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/SafeLoggingPropagationTest.java
```diff
@@ -249,13 +249,284 @@ void testIgnoresStaticMethods() {
...
+                        "abstract class Test {",
+                        "  @DoNotLog",
+                        "  @JsonProperty",
+                        "  String token() { return \"\"; }",
+                        "}")
+                .doTest();
+    }
+
+    @Test
+    void testIgnoresJsonIgnoreHelperMethod() throws Exception {
```
tpetracca: this test would have been an ignore even without the JsonIgnore annotation. Is it worth a version of this test that would actually propagate the method safety annotaiton to the class if the JsonIgnore wasn't there?

carterkozak: I'll add another test with a jsonignored field, however we _do_ want to include those in safety because they're included in the toString.

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/SafeLoggingPropagationTest.java
carterkozak: It is, good catch! Deduplicated.

tpetracca (approved):

PR #2242, opened 2022-05-01
Fix safety evaluation requiring lhs assignment to be safe unnecessarily

carterkozak:
The type parameter passthrough code would evaluate to SAFE rather
as opposed to UNKNOWN when type parameters were unknown/erased.

==COMMIT_MSG==
Fix safety evaluation requiring lhs of an assignment to be safe unnecessarily
==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/IllegalSafeLoggingArgumentTest.java
+++ b/baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/IllegalSafeLoggingArgumentTest.java
@@ -1096,6 +1096,26 @@ class IllegalSafeLoggingArgumentTest {
     }
 
     @Test
+    public void testThrowableFieldAssignment() {
+        helper().addSourceLines(
+                        "Test.java",
+                        "import com.palantir.logsafe.*;",
+                        "import java.util.*;",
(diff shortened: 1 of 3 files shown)


baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/IllegalSafeLoggingArgumentTest.java
```diff
@@ -1095,6 +1095,26 @@ public void testFieldAnnotationAssignment() {
                 .doTest();
     }
 
+    @Test
+    public void testThrowableFieldAssignment() {
+        helper().addSourceLines(
+                        "Test.java",
+                        "import com.palantir.logsafe.*;",
+                        "import java.util.*;",
+                        "class Test<OutputT> {",
```
carterkozak, line 1104: This test exposed the issue, but only when the (unused) type parameter is present. `JCTree.type` of the field when `<OutputT>` is present is `java.util.Set` and without it, `JCTree.type` is `java.util.Set<java.lang.Throwable>`!

I've updated the check to both:
1. avoid returning type-info based safety in the case that there's no type info
2. check both the symbol type and tree type for safety info

xRuiAlves (approved):

PR #2254, opened 2022-05-04
Implement class-level type variable safety validation

carterkozak:
==COMMIT_MSG==
Implement class-level type variable safety validation
==COMMIT_MSG==
In the future, I'll implement method type variable validation
as well, but this is fairly complex.

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/IllegalSafeLoggingArgument.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/IllegalSafeLoggingArgument.java
@@ -34,11 +34,13 @@ import com.sun.source.tree.CompoundAssignmentTree;
 import com.sun.source.tree.ExpressionTree;
 import com.sun.source.tree.MethodInvocationTree;
 import com.sun.source.tree.MethodTree;
+import com.sun.source.tree.NewClassTree;
 import com.sun.source.tree.ReturnTree;
 import com.sun.source.tree.StatementTree;
 import com.sun.source.tree.Tree;
 import com.sun.source.tree.VariableTree;
 import com.sun.source.util.TreePath;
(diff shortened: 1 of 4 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/IllegalSafeLoggingArgument.java
```diff
@@ -144,12 +160,42 @@ public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState
         return Description.NO_MATCH;
     }
 
-    private static SuggestedFix getSuggestedFix(MethodInvocationTree tree, VisitorState state, Safety argumentSafety) {
+    private void handleTypeArguments(ExpressionTree tree, VisitorState state) {
+        Type type = ASTHelpers.getResultType(tree);
+        if (type != null && !type.getTypeArguments().isEmpty()) {
+            List<Type> resultTypeArguments = type.getTypeArguments();
+            List<TypeVariableSymbol> parameterTypes = type.tsym.getTypeParameters();
+            if (parameterTypes.size() == resultTypeArguments.size()) {
```
fawind, line 168: Do you have an example of where these sizes would not be equal?

fawind, line 168: Do you have an example of where these sizes would not be equal?

fawind (approved):

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/IllegalSafeLoggingArgument.java
carterkozak: I do not, but I’ve run into cases where generic type info is missing due to compiler optimizations, so I figured it was best not to risk it.

PR #2263, opened 2022-05-10
Lazily acquire JDK storage locations

CRogers:
## Before this PR
You must know about all the JDK versions you need to set before the baseline-java-versions plugin uses them. This can make being a lazy as possible harder, and complicates the logic slightly.

## After this PR

==COMMIT_MSG==
More lazily acquire JDK storage locations
==COMMIT_MSG==


gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineJavaVersionsExtension.java
```diff
@@ -31,7 +33,8 @@ public class BaselineJavaVersionsExtension {
     private final Property<JavaLanguageVersion> libraryTarget;
     private final Property<JavaLanguageVersion> distributionTarget;
     private final Property<JavaLanguageVersion> runtime;
-    private final MapProperty<JavaLanguageVersion, JavaInstallationMetadata> jdks;
+    private final LazilyConfiguredMapping<JavaLanguageVersion, AtomicReference<JavaInstallationMetadata>> jdks =
```
CRogers: Why use an `AtomicReference` here? I wanted to make a general type I could [reuse in gradle-jdks](https://github.com/palantir/gradle-jdks/pull/19/files#r866173898) to configure the versions/distributions, which would involve modifying extension objects in the regular gradle style. So we use an `AtomicReference` here just to make it act like a box you can store data in.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/javaversions/LazilyConfiguredMapping.java
```diff
@@ -0,0 +1,89 @@
...
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.Supplier;
+import org.gradle.api.Action;
+
+public final class LazilyConfiguredMapping<K, V> {
```
CRogers: I would hope to reuse this in gradle-jdks to have everything be ETE lazy.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/javaversions/LazilyConfiguredMapping.java
```diff
@@ -0,0 +1,89 @@
...
+    private final Supplier<V> valueFactory;
+    private final List<LazyValues<K, V>> values = new ArrayList<>();
```
CRogers: It might have been easier to set this as just a single function you have to replace - but I'm concerned that just leads to everything becoming confusing if you "override" the value in custom gradle for example to test a new JDK major version that hasn't been fully released yet.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineJavaVersionsExtension.java
```diff
@@ -80,7 +80,20 @@ public final void setRuntime(int value) {
...
+        jdks.put(javaLanguageVersion, ref -> ref.set(javaInstallationMetadata));
+    }
+
+    public final void jdks(LazyJdks lazyJdks) {
+        jdks.put(javaLanguageVersion -> lazyJdks.jdkFor(javaLanguageVersion)
+                .map(javaInstallationMetadata -> ref -> ref.set(javaInstallationMetadata)));
+    }
+
+    public interface LazyJdks {
+        Optional<JavaInstallationMetadata> jdkFor(JavaLanguageVersion javaLanguageVersion);
```
CRogers: This is basically making it fully lazy.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/javaversions/LazilyConfiguredMapping.java
carterkozak: perhaps move to the extensions package and make package private?

carterkozak (approved):

PR #2268, opened 2022-05-12
Fix concurrency issue in LazilyConfiguredMapping

carterkozak:
==COMMIT_MSG==
Fix concurrency issue in LazilyConfiguredMapping
==COMMIT_MSG==

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/javaversions/LazilyConfiguredMapping.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/javaversions/LazilyConfiguredMapping.java
@@ -23,25 +23,32 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Supplier;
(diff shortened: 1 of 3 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/javaversions/LazilyConfiguredMapping.java
```diff
@@ -23,25 +23,32 @@
...
 import org.gradle.api.Action;
 
 final class LazilyConfiguredMapping<K, V, A> {
     private final Supplier<V> valueFactory;
+
+    @GuardedBy("this")
     private final List<LazyValues<K, V, A>> values = new ArrayList<>();
+
+    @GuardedBy("this")
     private final Map<K, Optional<V>> computedValues = new HashMap<>();
```
carterkozak, line 36: Considered swapping to a ConcurrentHashmap which handles nested mutations more gracefully, however that's not expected to happen (and isn't related to this issue).

pkoenig10 (approved):

PR #2271, opened 2022-05-13
Safety analysis detects annotations on superclasses and their interfaces

carterkozak:
Previously it would detect direct interfaces and supertypes, but
not interfaces on supertypes.

## After this PR
==COMMIT_MSG==
Safety analysis detects annotations on superclasses and their interfaces
==COMMIT_MSG==

## Possible downsides?
More work at build time!

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/safety/SafetyAnnotations.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/safety/SafetyAnnotations.java
@@ -49,6 +49,13 @@ public final class SafetyAnnotations {
     public static final String UNSAFE = "com.palantir.logsafe.Unsafe";
     public static final String DO_NOT_LOG = "com.palantir.logsafe.DoNotLog";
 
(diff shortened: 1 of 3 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/safety/SafetyAnnotations.java
```diff
@@ -122,6 +124,30 @@ public static Safety getSafety(@Nullable Type type, VisitorState state) {
         return Safety.UNKNOWN;
     }
 
+    public static Safety getDirectSafety(@Nullable Symbol symbol, VisitorState state) {
```
carterkozak, line 127: Extracted to avoid duplicating effort between `ASTHelpers.hasAnnotation` and our annotation search. `ASTHelpers.hasAnnotation` searches supertypes, but not their interfaces, and there's no sense doing it twice.

I took the opportunity to use an error-prone supplier for names, which should reduce overhead a bit.

fawind (approved):

PR #2291, opened 2022-06-06
Disallow usage of .collapseKeys() in EntryStream(s)

a-k-g:
## Before this PR
The `collapseKeys()` API of `EntryStream` is dangerous and should be avoided. In internal products the API is frequently used as a grouping operation but its not suitable for that use case. The contract requires duplicate keys to be adjacent to each other in the stream, which is rarely the case in production code paths. When this constraint is violated, it leads to a duplicate key error at runtime.

A work around for the issue is to sort the keys prior to running the collapse operation. Since the sort operation is surprising, a comment is often added to explain. Overall the usage of `collapseKeys()` leads to code that is error prone or surprising.

Hence in place of `collapseKeys()` we should just use grouping operations which may require creation of intermediate maps but should avoid surprising code.

Notional example of buggy code:

    List<Pair<Integer, Integer>> nodesAndPriorities = List.of
            Pair.of(1, 5),
            Pair.of(1, 3),
            Pair.of(4, 2),
            Pair.of(1, 9));

    Map<Integer, Set<Integer>> nodesByPriority = EntryStream.of(nodesAndPriorities)
            .mapValues(Pair::getRight)
            .collapseKeys(Collectors.toSet())
            .toMap();

Notional example of applying sort work around:

    Map<Integer, Set<Integer>> nodesByPriority = EntryStream.of(nodesAndPriorities)
            .mapValues(Pair::getRight)
            // collapseKeys() requires duplicate keys to be adjacent
            .sorted()
            .collapseKeys(Collectors.toSet())
            .toMap();

Notional example with proposed alternative of grouping:

    Map<Integer, Set<Integer>> nodesByPriority = EntryStream.of(nodesAndPriorities)
            .mapValues(Pair::getRight)
            .grouping(Collectors.toSet());

## After this PR

==COMMIT_MSG==
Added `DangerousCollapseKeysUsage` error prone check to disallow usage of `collapseKeys()` API of `EntryStream`.
==COMMIT_MSG==

## Possible downsides?
Will need manual action for existing usages.


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DangerousCollapseKeysUsage.java
```diff
@@ -0,0 +1,60 @@
...
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
+        severity = SeverityLevel.ERROR,
+        summary = "Disallow usage of .collapseKeys() in EntryStream(s).")
+public final class DangerousCollapseKeysUsage extends BugChecker implements BugChecker.MethodInvocationTreeMatcher {
+    private static final long serialVersionUID = 1L;
+    private static final String ERROR_MESSAGE = "The collapseKeys API of EntryStream should be avoided. The "
```
carterkozak: "should" implies info/warning, not ` SeverityLevel.ERROR`.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DangerousCollapseKeysUsage.java
```diff
@@ -0,0 +1,60 @@
...
+            + "API is frequently used as a grouping operation but its not suitable for that use case. The contract "
+            + "requires duplicate keys to be adjacent to each other in the stream, which is rarely the case in "
+            + "production code paths. When this constraint is violated, it leads to a duplicate key error at runtime.\n"
+            + "A work around for the issue is to sort the keys prior to running the collapse operation. Since the "
+            + "sort operation is surprising, a comment is often added to explain. Overall the usage of collapseKeys() "
+            + "leads to code that is error prone or surprising.\nHence in place of collapseKeys() we recommend using "
+            + "grouping operations which may require creation of intermediate maps but should avoid surprising code.";
+
+    private static final Matcher<ExpressionTree> COLLAPSE_KEYS_CALL = MethodMatchers.instanceMethod()
+            .onExactClass("one.util.streamex.EntryStream")
```
carterkozak, line 48:
```suggestion
            .onDescendantOf("one.util.streamex.EntryStream")
```

a-k-g: Is this just out of caution? `EntryStream` is a final class at the moment.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DangerousCollapseKeysUsage.java
a-k-g: Would you suggest changing the error level, or the language? I'm inclined to change this to `The collapseKeys API of EntryStream must be avoided`.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DangerousCollapseKeysUsage.java
carterkozak: Ah, if it's final, exact-class is sufficient

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DangerousCollapseKeysUsage.java
carterkozak: I'd prefer setting this to WARNING to reduce friction of rollout. I might suggest adding a check for a sort called prior to this method, to avoid flagging cases that are safe.

a-k-g: I've removed most of the internal usage so rollout should be friction free and we can keep it at ERROR. I've changed the `should` to `must`.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DangerousCollapseKeysUsage.java
```diff
@@ -0,0 +1,60 @@
...
+            .named("collapseKeys");
+
+    @Override
+    public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
+        if (!COLLAPSE_KEYS_CALL.matches(tree, state)) {
+            return Description.NO_MATCH;
+        }
+
+        // Fail on any 'parallel(...)' implementation, regardless of how many parameters it takes
```
carterkozak, line 57: `collapseKeys`, not `parallel`

carterkozak (approved):

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DangerousCollapseKeysUsage.java
a-k-g: Ah, fixing in https://github.com/palantir/gradle-baseline/pull/2510.

PR #2295, opened 2022-06-07
Fix class hierarchy safe-logging error message agreement

carterkozak:
Previously the safety referenced in the message did not agree
with the failure.

==COMMIT_MSG==
Fix class hierarchy safe-logging error message agreement
==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/IllegalSafeLoggingArgumentTest.java
+++ b/baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/IllegalSafeLoggingArgumentTest.java
@@ -1768,7 +1768,8 @@ class IllegalSafeLoggingArgumentTest {
                         "import com.palantir.logsafe.*;",
                         "class Test {",
                         "  @Unsafe interface UnsafeClass {}",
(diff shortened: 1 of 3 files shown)


baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/IllegalSafeLoggingArgumentTest.java
```diff
@@ -1768,7 +1768,8 @@ public void testSubclassWithLenientSafety() {
                         "import com.palantir.logsafe.*;",
                         "class Test {",
                         "  @Unsafe interface UnsafeClass {}",
-                        "  // BUG: Diagnostic contains: Dangerous type: annotated 'SAFE' but ancestors declare 'SAFE'.",
```
carterkozak, line 1771: apparently I didn't read this very well and copy/pasted :-(

fawind: Woops, also slipped through review 🙈

fawind (approved):

PR #2299, opened 2022-06-11
gradle task configuration avoidance

rzpt:
also fix an issue in BaselineJavaVersion

## Before this PR
We are unnecessarily creating and configuring tasks during gradle configuration. See https://docs.gradle.org/current/userguide/task_configuration_avoidance.html for more details.

## After this PR
==COMMIT_MSG==
Avoid creating and configuring gradle tasks
==COMMIT_MSG==

## Possible downsides?
None.

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/javaversions/BaselineJavaVersion.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/javaversions/BaselineJavaVersion.java
@@ -30,6 +30,7 @@ import org.gradle.api.tasks.CacheableTask;
 import org.gradle.api.tasks.Input;
 import org.gradle.api.tasks.JavaExec;
 import org.gradle.api.tasks.TaskAction;
+import org.gradle.api.tasks.TaskProvider;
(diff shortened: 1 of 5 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/javaversions/BaselineJavaVersion.java
```diff
@@ -71,15 +72,15 @@ public void execute(JavaToolchainSpec javaToolchainSpec) {
...
 
             // Validation
-            project.getTasks()
+            TaskProvider<CheckJavaVersionsTask> checkJavaVersions = project.getTasks()
                     .register("checkJavaVersions", CheckJavaVersionsTask.class, new Action<CheckJavaVersionsTask>() {
                         @Override
                         public void execute(CheckJavaVersionsTask task) {
                             task.getTargetVersion().set(extension.target());
                             task.getRuntimeVersion().set(extension.runtime());
-                            project.getTasks().getByName("check").dependsOn(task);
```
rzpt, line 80: This is a problem because check would only depend on checkJavaVersions if checkJavaVersions was configured.

CRogers (approved):

PR #2300, opened 2022-06-13
Move all AtlasDB persisters to the new reusable type

raiju:
## Before this PR

Non-reusable persisters are created for each cell read/write. This causes a lot of object creation, which is particularly problematic when a new object mapper is created,

## After this PR

==COMMIT_MSG==
Enforce all AtlasDB persisters to be marked as reusable
==COMMIT_MSG==

## Possible downsides?
- Should applying this fix be broken out into a separate excavator to ensure enough attention is paid to it? I haven't found a single non-reusable persister yet, but I'm sure they exist.
- @jeremyk-91 @Jolyon-S Are there reasons to not do this?


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/NonReusableAtlasdbPersister.java
```diff
@@ -0,0 +1,67 @@
...
+        if (!persisterClass.matches(tree, state)) {
+            return Description.NO_MATCH;
+        }
+
+        if (reusableAnnotationMatcher.matches(tree, state)) {
+            return Description.NO_MATCH;
+        }
+
+        return buildDescription(tree)
+                .addFix(SuggestedFix.prefixWith(tree, "@com.palantir.atlasdb.annotation.Reusable\n"))
```
schlosna: nit: qualify type to generate import and `@Reusable`:

```suggestion
        String annotation = SuggestedFixes.qualifyType(state, builder, "com.palantir.atlasdb.annotation.Reusable");
        return buildDescription(tree)
                .addFix(SuggestedFix.prefixWith(tree, "@" + annotation + "\n"))
```

carterkozak: I'd suggest using a space instead of a newline for simplicity, the formatter knows what to do either way.

carterkozak:
Given that we own the persister API, I would prefer if we updated the API in such a way that it's difficult to get wrong rather than relying on static analysis.

raiju:
I'd prefer that too; @jeremyk-91 is there a world where we could flip the expectations/always enforce this?

@carterkozak Would you be ok merging this with a plan to get persisters to be safe by default?

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/NonReusableAtlasdbPersister.java
raiju: Ah, perfect. Done!

jeremyk-91:
No objections to changing the API in the long term (in fact that's preferable, I remember this bit of defensiveness being somewhat annoying to add) - the main thing is how to get there. If this is likely to have people touch each of their persisters, I'm thinking if we should instead do:

- add `@StatefulPersister` annotation to AtlasDB (or something like that)
- require users to specify *at least one*
- (not 100% safe, but probably >99%) after some point in time, switch the default behaviour over

raiju:
Refactored this to support the new setup (only `ReusablePersister`s are supported, `Persister`s are deprecated)

gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineErrorProneExtension.java
```diff
@@ -50,6 +50,7 @@ public class BaselineErrorProneExtension {
             "LambdaMethodReference",
             "LoggerEnclosingClass",
             "LogsafeArgName",
+            "NonReusableAtlasdbPersister",
```
raiju, line 53: Is this appropriate for this check? We want to move everyone to this new persister given the risks of the old persister

PR #2304, opened 2022-06-16
Don't count identifier expressions as variable usages in `StrictUnusedVariable` check

IlanaRadinsky:
## Before this PR
If you have a class like this
```
public class Test {
    private final String myField;

    public Test(String myField) {
        this.myField = myField;
    }
}
```
```
    private final String myField;
                         ^
```

```
public class Test {

    public Test(String myField) {
        this._myField = myField;
    }
}
```
```
                         ^
```

See https://github.com/palantir/gradle-baseline/pull/2304/commits/a1b70a50a849b374bf590009c112265e09fb2525 https://app.circleci.com/pipelines/github/palantir/gradle-baseline/4165/workflows/756c72ee-8297-4da5-837a-40a3b010ad09/jobs/34598 for example of failing test before change.

## After this PR
==COMMIT_MSG==
Don't count identifier expressions as variable usages in `StrictUnusedVariable` check
==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StrictUnusedVariable.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StrictUnusedVariable.java
@@ -91,6 +91,7 @@ import com.sun.source.tree.MethodTree;
 import com.sun.source.tree.ReturnTree;
 import com.sun.source.tree.StatementTree;
 import com.sun.source.tree.Tree;
+import com.sun.source.tree.Tree.Kind;
 import com.sun.source.tree.TryTree;
 import com.sun.source.tree.UnaryTree;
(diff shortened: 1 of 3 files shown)


carterkozak:
Hmm, in this case I think we'd prefer to suggest removing the variable and assignment entirely rather than allowing underscores. That can be a bit tricky to auto-suggest, though.

IlanaRadinsky:
@carterkozak Would it be preferable to update the error message without the auto-suggest, or just to close this PR?

carterkozak:
Yep, ideally we would fail due to unused fields, but we wouldn't suggest renaming anything.

IlanaRadinsky:
Tracking in https://github.com/palantir/gradle-baseline/issues/2361 since I haven't had a chance to come back to this yet.

PR #2307, opened 2022-06-17
Promote ReturnsNullCollection and ReturnMissingNullable to error

robert3005:
## Before this PR
We would not alter on null return values from methods. We generally though never return null values and it should be considered an error if one does so. Errorprone has checks for those but they're just suggestions

## After this PR

==COMMIT_MSG==
Promote ReturnsNullCollection and ReturnMissingNullable to error
==COMMIT_MSG==

## Possible downsides?
The recommended resolution is to add Nullable while we would want the author to never return null


carterkozak:
Nullaway tends to provide more accurate and more performant analysis, should we focus our efforts there?

robert3005:
Ah, I totally forgot about nullaway

PR #2313, opened 2022-07-05
Roll back spotless auto-bump

carterkozak:
Fails with
```
GradleException: Not allowed to resolve configuration ':palantirJavaFormat'
at configuration time. Please upgrade your plugins and double-check your
gradle scripts
```

## After this PR

==COMMIT_MSG==
Roll back spotless auto-bump
==COMMIT_MSG==


carterkozak:
Rolling this into #2312

PR #2319, opened 2022-07-11
[draft] `com.palantir.baseline-enable-preview-flag` works with IntelliJ integration correctly

iamdanfox:
## Before this PR

The plugin to add the `--enable-preview` flags isn't _actually_ usable because when you open IntelliJ, the `languageLevel` / `LANGUAGE_LEVEL` fields aren't configured correctly.

The motivation for this PR is I'm trying to build conjure switch expressions during hack week.

## After this PR

==COMMIT_MSG==
==COMMIT_MSG==


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineEnablePreviewFlag.java
```diff
@@ -29,26 +31,28 @@
...
     public void apply(Project project) {
         // The idea behind registering a single 'extra property' is that other plugins (like
-        // sls-packaging) can easily detect this and also also add the --enable-preview jvm arg
-        Provider<Boolean> enablePreview = project.provider(() -> {
-            JavaVersion jvmExecutingGradle = JavaVersion.current();
-            JavaPluginConvention javaConvention = project.getConvention().findPlugin(JavaPluginConvention.class);
-            if (javaConvention == null) {
-                return false;
+        // sls-packaging) can easily detect this and also also add the --enable-preview jvm arg.
+        project.getExtensions().getExtraProperties().set(ENABLE_PREVIEW_PROPERTY, project.provider(() -> {
```
carterkozak, line 43: This section has a lot in common with the jvm module options plugin, including that we'll need to plumb it through to sls-packaging in some way

iamdanfox: Yeah - I think stuff has evolved since I first made these commits... this is very much just so that I can get my local machine working, happy to rework to integrate with [that plugin!](https://github.com/palantir/gradle-baseline/blob/develop/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java).

I also wanted to figure out how explicit the opt-in to enable-preview _should_ be... I'd feel a bit weird about this magically being transitively applied to a project

carterkozak: > I'd feel a bit weird about this magically being transitively applied to a project

That's fair, however if it's required by features used within a jar, we don't want to fail at runtime. The other side of this coin is that enable-previews locks us into a specific jdk version, as these features _may_ change between releases in incompatible ways.

iamdanfox: Perhaps a middle ground could be:
- the `--enable-preview` flag is automatically propagated to the launcher-static args used at runtime
- it is _not_ propagated to the compiler args of projects, so that each gradle project which wants to use fancy new features will need to explicitly opt-in.

iamdanfox:
Closing out in favour of https://github.com/palantir/gradle-baseline/pull/2322 which integrates with the `com.palantir.baseline-java-versions` plugin

PR #2322, opened 2022-07-14
baseline-java-versions allows opting in to `--enable-preview`

iamdanfox:
## Before this PR

I had previously made a half-baked attempt to support `--enable-preview` via gradle-baseline (https://github.com/palantir/gradle-baseline/pull/1549) but unfortunately it doesn't fully configure Gradle's IntelliJ integration. This isn't really good enough, because it means folks would open a project and see compilation errors.

Also note that _literally zero people were using it_.

Rather than investing more in this plugin, we've decided to throw it away and merge the functionality into the existing `com.palantir.baseline-java-versions` plugin.

## After this PR
==COMMIT_MSG==
Users of the `com.palantir.baseline-java-versions` plugin can now set `javaVersions { distributionTarget = '17_PREVIEW' }` to opt-in to Java's `--enable-preview` flag at compile time.
==COMMIT_MSG==

After merging this, I'll also need to fix-up the sls-packaging integration. If someone tries to flip this switch _before_ sls-packaging is updated, then their java will fail at startup. This would be caught by dockerized tests (if people have them).

https://github.com/palantir/sls-packaging/blob/1b4c98ce9652d515618ccb426830e14f3df8f34a/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/JavaServiceDistributionPlugin.java#L82-L93

🌶🌶🌶 This functionality is intentionally NOT ALLOWED for projects that publish libraries, because I think the special 'enable preview minor bytecode' is too much of a footgun for consumers to deal with (it requires _exactly_ the same version of java that it was compiled with, so wouldn't allow consumers to _upgrade_ their own java and still use the library)!!!

## Possible downsides?

- The IntelliJ integration works if you use `open *.ipr` workflow, and it works if you use the IntelliJ-Gradle integration and write features in your main source set, but it does NOT automatically tell IntelliJ to enable-preview on test sourcesets :(. I spent hours trying to make this work, and have not succeeded.
- The ABI of the extensions is changing, so I'll need to fix up a couple of internal consumers:
```java
            BaselineJavaVersionExtension extension = project.getExtensions().getByType(BaselineJavaVersionExtension.class)
            extension.target().set(JavaLanguageVersion.of(targetVersion))
            extension.runtime().set(JavaLanguageVersion.of(runtimeVersion))
```


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
```diff
@@ -220,22 +218,21 @@ class BaselineIdea extends AbstractBaselinePlugin {
...
-            if (version != null) {
-                int featureRelease = version.target().get().asInt()
-                JavaVersion javaVersion = JavaVersion.toVersion(featureRelease)
+            BaselineJavaVersionExtension versionExtension = currentProject.extensions.findByType(BaselineJavaVersionExtension.class)
+            if (versionExtension != null) {
+                ChosenJavaVersion chosenJavaVersion = versionExtension.target().get()
                 Node node = provider.asNode()
                 Node newModuleRootManager = node.component.find { it.'@name' == 'NewModuleRootManager' }
-                newModuleRootManager.attributes().put("LANGUAGE_LEVEL", new IdeaLanguageLevel(javaVersion).getLevel())
+                newModuleRootManager.attributes().put("LANGUAGE_LEVEL", chosenJavaVersion.asIdeaLanguageLevel())
```
iamdanfox, line 235: this is a crucial change for Intellij usability - previously it would say something like `JDK_17`, but now it knows how to say `JDK_17_PREVIEW`.

I found this by committing all my intellij-related generated files (which are normally gitignored),

then writing some new code:

image

then hitting the button 
image

then seeing what changed in my git diff:

image

robert3005: you probably want to delete this module and ban running `./gradlew idea` just like gradle/gradle does. There's too many things that go wrong when you're not using gradle integration in intellij these days

iamdanfox: @robert3005 I agree the `open *.ipr` thing seems to have got more fragile (I haven't personally used it in a while)... but with this PR I was specifically trying to make the IntelliJ _gradle integration_ work out of the box!

When I accepted IntelliJ's prompt to enable preview, it added two lines of seemingly non-functional gradle!!

image

image

iamdanfox: [Here's the JavaVersion](https://github.com/gradle/gradle/blob/4b7ef9a72f5ddd0da071577d5c68c0914404ef97/subprojects/base-services/src/main/java/org/gradle/api/JavaVersion.java#L79) class... I don't understand how the `JavaVersion.VERSION_17_PREVIEW` thing is supposed to work!

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/javaversions/BaselineJavaVersionsExtension.java
```diff
@@ -31,20 +31,22 @@
  */
 public class BaselineJavaVersionsExtension {
     private final Property<JavaLanguageVersion> libraryTarget;
```
iamdanfox, line 33: I'm choosing to literally not support the --enable-preview functionality in this top level `javaVersions {}` extension, because I think the fact that the generated bytecode CANNOT be run by newer versions of java is such a massive footgun that I don't want folks to ever publish jars with this.

if people _really_ need to break glass, they can still conceivably use the subproject-level `javaVersion {}` extensions

carterkozak (approved):
Lgtm. Thanks, @iamdanfox!

PR #2331, opened 2022-07-25
fix #2328 Handle safety flow through instanceof pattern matching

### LINKED ISSUE #2328: Extremely slow compilation for pattern matching instanceof method
## What happened?

I wrote code which looks roughly like:

```java
        @SuppressWarnings("CyclomaticComplexity")
        private static long getRoutingId(BlockId block) {
            if (block instanceof RDDBlockId b) {
                return b.rddId();
            } else if (block instanceof ShuffleBlockId b) {
                return b.shuffleId();
            } else if (block instanceof ShuffleBlockBatchId b) {
                return b.shuffleId();
            } else if (block instanceof ShuffleBlockChunkId b) {
                return b.shuffleId();
            } else if (block instanceof ShuffleDataBlockId b) {
                return b.shuffleId();
            } else if (block instanceof ShuffleIndexBlockId b) {
                return b.shuffleId();
            } else if (block instanceof ShuffleChecksumBlockId b) {
                return b.shuffleId();
            } else if (block instanceof ShufflePushBlockId b) {
                return b.shuffleId();
            } else if (block instanceof ShuffleMergedBlockId b) {
                return b.shuffleId();
            } else if (block instanceof ShuffleMergedDataBlockId b) {
                return b.shuffleId();
            } else if (block instanceof ShuffleMergedIndexBlockId b) {
                return b.shuffleId();
            } else if (block instanceof ShuffleMergedMetaBlockId b) {
                return b.shuffleId();
            } else if (block instanceof BroadcastBlockId b) {
                return b.broadcastId();
            } else {
                throw new SafeIllegalStateException("unhandled");
            }
```

where the underlying type of BlockId is a Scala case class with many subtypes. Not my favourite code, but surely close to as good as can be done.

Compilation is currently 9 minutes in. Stack traces are implying that the expensive code relates to the Safety Propagation code. JFR can be provided on request.

Compilation in a reasonable amount of time.

carterkozak:
==COMMIT_MSG==
Handle safety flow through instanceof pattern matching
==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/build.gradle
+++ b/baseline-error-prone/build.gradle
@@ -3,6 +3,11 @@ import net.ltgt.gradle.errorprone.CheckSeverity
 apply plugin: 'java-library'
 apply plugin: 'com.palantir.external-publish-jar'
 
+javaVersions {
+    libraryTarget = 11
+    runtime = 17
+}
+
 dependencies {
     implementation 'com.google.errorprone:error_prone_core'
     // Ensure a new enough version of dataflow-errorprone is available
--- a/baseline-error-prone/build.gradle
+++ b/baseline-error-prone/build.gradle
@@ -3,8 +3,8 @@ import net.ltgt.gradle.errorprone.CheckSeverity
 apply plugin: 'java-library'
 apply plugin: 'com.palantir.external-publish-jar'
 
-javaVersions {
-    libraryTarget = 11
+javaVersion {
+    target = 11
     runtime = 17
 }
 
--- a/baseline-refaster-rules/build.gradle
+++ b/baseline-refaster-rules/build.gradle
@@ -34,4 +34,5 @@ moduleJvmArgs {
             'jdk.compiler/com.sun.tools.javac.tree',
             'jdk.compiler/com.sun.tools.javac.util'
     ]
+    opens = ['jdk.compiler/com.sun.tools.javac.comp']
 }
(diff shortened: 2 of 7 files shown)


baseline-error-prone/build.gradle
```diff
@@ -3,6 +3,11 @@ import net.ltgt.gradle.errorprone.CheckSeverity
 apply plugin: 'java-library'
 apply plugin: 'com.palantir.external-publish-jar'
 
+javaVersions {
+    libraryTarget = 11
+    runtime = 17
+}
```
carterkozak: We'll see what this breaks

baseline-refaster-rules/build.gradle
```diff
@@ -34,4 +34,5 @@ moduleJvmArgs {
             'jdk.compiler/com.sun.tools.javac.tree',
             'jdk.compiler/com.sun.tools.javac.util'
     ]
+    opens = ['jdk.compiler/com.sun.tools.javac.comp']
```
fawind, line 37: For my understanding, why do we need the additional "opens" declaration now? I don't see any new usage in the diff here.

fawind (approved):
👍

baseline-refaster-rules/build.gradle
carterkozak: This allows this module to pass tests using jdk17. Not required for the build, but initially I tried using jdk17 everywhere, but some tests are more difficult to fix.

PR #2338, opened 2022-07-29
Fix javadoc when enable preview flag is added

iamdanfox:
## Before this PR

We'd successfully add the `-enable-preview` flag to the `javadoc` task, but since we don't have the `-source` or `--release` flags, it fails with the following:

```
> Task :witchcraft-example-api:witchcraft-example-api-undertow:javadoc FAILED
error: --enable-preview must be used with either -source or --release
1 error
```

We did not have an automated test that actually invokes `javadoc`, hence why this bug slipped in.

## After this PR
==COMMIT_MSG==
javadoc tasks are now properly configured when `--enable-preview` is used
==COMMIT_MSG==

We now have a failing test that runs javadoc, and the last commit fixes it!

### DIFF
--- a/changelog/@unreleased/pr-2338.v2.yml
+++ b/changelog/@unreleased/pr-2338.v2.yml
@@ -1,0 +1,6 @@
+type: fix
+fix:
+  description: javadoc tasks are now properly configured when `--enable-preview` is
+    used
+  links:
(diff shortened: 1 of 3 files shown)


carterkozak (approved):

iamdanfox:
Sorry I didn't include a test for this first time round!!

PR #2343, opened 2022-08-02
Add DefaultLocale check

pkoenig10:
Related to https://github.com/google/error-prone/issues/632

Adds a `DefaultLocale` check that replaces uses of `String.toLowerCase()` and `String.toUpperCase()` with the overloads that take a `Locale`, using `Locale.ROOT`.

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DefaultLocale.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DefaultLocale.java
@@ -22,6 +22,7 @@ import com.google.errorprone.BugPattern.SeverityLevel;
 import com.google.errorprone.VisitorState;
 import com.google.errorprone.bugpatterns.BugChecker;
 import com.google.errorprone.fixes.SuggestedFix;
+import com.google.errorprone.fixes.SuggestedFixes;
 import com.google.errorprone.matchers.Description;
 import com.google.errorprone.matchers.Matcher;
 import com.google.errorprone.matchers.method.MethodMatchers;
@@ -34,7 +35,7 @@ import com.sun.source.tree.MethodInvocationTree;
         linkType = BugPattern.LinkType.CUSTOM,
         summary = "Implicit use of the platform default locale, which can result in differing behaviour between JVM"
                 + " executions.",
-        severity = SeverityLevel.WARNING)
+        severity = SeverityLevel.SUGGESTION)
 public final class DefaultLocale extends BugChecker implements BugChecker.MethodInvocationTreeMatcher {
 
     private static final long serialVersionUID = 1L;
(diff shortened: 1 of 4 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DefaultLocale.java
```diff
@@ -0,0 +1,63 @@
...
+            .onExactClass("java.lang.String")
+            .namedAnyOf("toLowerCase", "toUpperCase")
+            .withNoParameters();
+
+    @Override
+    public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
+        if (MATCHER.matches(tree, state)) {
+            return buildDescription(tree)
+                    .addFix(SuggestedFix.builder()
+                            .addImport("java.util.Locale")
```
carterkozak: A bit safer not to add imports directly, but use SuggestedFixes.qualifyType

gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineErrorProneExtension.java
```diff
@@ -38,6 +38,7 @@ public class BaselineErrorProneExtension {
             "CompileTimeConstantViolatesLiskovSubstitution",
             "ConsistentLoggerName",
             "ConsistentOverrides",
+            "DefaultLocale",
```
carterkozak: Initially, let’s make this check a suggestion, and not add it to this list. We can spin up an excavator to roll out the suggested fix asynchronously, otherwise some baseline upgrades will be blocked on a way that we aren’t currently prepared to handle.

carterkozak: On us to build a better ratcheting rollout mechanism in the medium term

pkoenig10: Yeah it feels a bit pointless to write these error-prone checks if we're too worried about blocking upgrades to actually enforce them.

carterkozak: It depends on the criticality of the finding. I suspect this will find a lot of things that aren’t going to cause issues in practice. I agree we should fix them, but we have other work in flight that isn’t worth risking.

carterkozak (approved):
Thanks!

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DefaultLocale.java
```diff
@@ -0,0 +1,64 @@
...
+import com.sun.source.tree.MethodInvocationTree;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
+        summary = "Implicit use of the platform default locale, which can result in differing behaviour between JVM"
+                + " executions.",
+        severity = SeverityLevel.SUGGESTION)
+public final class DefaultLocale extends BugChecker implements BugChecker.MethodInvocationTreeMatcher {
```
findepi, line 39: Does this add a check withi same name as, similar spirit to, but different implementation from https://errorprone.info/bugpattern/DefaultLocale ?

findepi: i see it's removed -- https://github.com/palantir/gradle-baseline/pull/2839

PR #2348, opened 2022-08-04
Update error-prone configuration for gradle 7 up-to-date correctness

bellybuttonbackwards:
## Before this PR
The suggested error-prone configuration syntax results in tasks never being up-to-date in gradle 7+

## After this PR
Swapping from lambda notation to an explicit interface implementation causes these tasks to be correctly considered up-to-date, following the instructions from https://docs.gradle.org/current/userguide/validation_problems.html#implementation_unknown

==COMMIT_MSG==

==COMMIT_MSG==

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,9 +127,11 @@ Error-prone rules can be suppressed on a per-line or per-block basis just like C
 Rules can be suppressed at the project level, or have their severity modified, by adding the following to the project's `build.gradle`:
 
 ```gradle
-tasks.withType(JavaCompile).configureEach {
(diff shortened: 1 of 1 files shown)


carterkozak (approved):
Thanks!

schlosna:
Do we have any mechanisms to identify these problematic lambdas & auto-fix them?

carterkozak:
> Do we have any mechanisms to identify these problematic lambdas & auto-fix them?

We have error-prone checks for java-based gradle plugins, but I don't think we have similar static analysis for gradle/groovy code

PR #2351, opened 2022-08-08
Set the java launcher for Checkstyle tasks, too

### LINKED ISSUE #93: Gradle provisioning is applied if the user does not have the relevant JDK already installed
## What happened?

I tried to `./gradlew check` on a repo that used JDKs. It failed because Adoptium didn't include the specific version of Java that I cared about, for my architecture. I fixed this by setting `org.gradle.java.installations.auto-download=false` in gradle.properties.

I think it'd be great if this plugin warned if the Gradle downloader is turned on. I'm not sure of exact behaviour, but IIRC it will always prefer the Adoptium version if it's present? There were loads of edge cases. But in any case, people using this plugin probably don't want the gradle one in-frame.

j-baker:
The gradle-jdks plugin aims to download JDKs from a source and use them
at runtime. In theory, this means that in order to develop using such a
library, one only needs a JDK installed that can run Gradle (there are
some limitations but this is the theory).

Inside the Checkstyle task is a Java launcher property. When this is
resolved, it will query the Java version service, which will query
the configured JDKs. This means that the baseline-java-versions
configured JDK is not used, and if the gradle-jdks approach is used, the
Checkstyle task will fail.

I _think_ that this change is enough to correctly override the Java
launcher used.

Fixes https://github.com/palantir/gradle-jdks/issues/93

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/javaversions/BaselineJavaVersion.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/javaversions/BaselineJavaVersion.java
@@ -132,6 +132,13 @@ public final class BaselineJavaVersion implements Plugin<Project> {
             }
         });
 
+        project.getTasks().withType(Checkstyle.class).configureEach(new Action<Checkstyle>() {
+            @Override
+            public void execute(Checkstyle checkstyle) {
+                checkstyle.getJavaLauncher().set(javaToolchain.flatMap(BaselineJavaToolchain::javaLauncher));
+            }
+        });
+
         project.getTasks().withType(GroovyCompile.class).configureEach(new Action<GroovyCompile>() {
             @Override
             public void execute(GroovyCompile groovyCompileTask) {
@@ -178,13 +185,6 @@ public final class BaselineJavaVersion implements Plugin<Project> {
             @Override
             public void execute(ScalaDoc scalaDoc) {
                 scalaDoc.getJavaLauncher().set(javaToolchain.flatMap(BaselineJavaToolchain::javaLauncher));
-            }
-        });
-
-        project.getTasks().withType(Checkstyle.class).configureEach(new Action<Checkstyle>() {
-            @Override
-            public void execute(Checkstyle checkstyle) {
-                checkstyle.getJavaLauncher().set(javaToolchain.flatMap(BaselineJavaToolchain::javaLauncher));
             }
         });
     }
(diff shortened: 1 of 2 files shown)


j-baker:
I verified that this fixed my problem locally. I wasn't sure how to write a test that isn't overengineered. Concretely my problem is that it's failing on CI, where my Java version is not installed by default.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/javaversions/BaselineJavaVersion.java
```diff
@@ -198,6 +199,13 @@ public void execute(Test test) {
                 test.getJvmArgumentProviders().add(new EnablePreviewArgumentProvider(runtime));
             }
         });
+
+        project.getTasks().withType(Checkstyle.class).configureEach(new Action<Checkstyle>() {
+            @Override
+            public void execute(Checkstyle checkstyle) {
+                checkstyle.getJavaLauncher().set(javaToolchain.flatMap(BaselineJavaToolchain::javaLauncher));
+            }
+        });
```
carterkozak: Mind adding a test that this resolves?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/javaversions/BaselineJavaVersion.java
```diff
@@ -198,6 +199,13 @@ public void execute(Test test) {
                 test.getJvmArgumentProviders().add(new EnablePreviewArgumentProvider(runtime));
             }
         });
+
+        project.getTasks().withType(Checkstyle.class).configureEach(new Action<Checkstyle>() {
+            @Override
+            public void execute(Checkstyle checkstyle) {
+                checkstyle.getJavaLauncher().set(javaToolchain.flatMap(BaselineJavaToolchain::javaLauncher));
```
carterkozak: I think we want to use the compile target for checkstyle, not the runtime java version

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/javaversions/BaselineJavaVersion.java
j-baker: I'm not entirely sure what this means. What sort of test would you like me to write? I'm not really sure how to write the Checkstyle equivalent of the Java one.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/javaversions/BaselineJavaVersion.java
j-baker: you are right. Moved into the compile section

carterkozak (approved):

pkoenig10:
Just want to flag that this bumps the minimum Gradle version to 7.5.

https://github.com/gradle/gradle/blob/af40a01f65c321846765e40500a5271a868ffd15/subprojects/code-quality/src/main/groovy/org/gradle/api/plugins/quality/Checkstyle.java#L180-L189

We likely have quite a few internal repos that are not yet on Gradle 7.5.

carterkozak:
Good flag, let's add a check rather than increasing the minimum supported gradle release: https://github.com/palantir/gradle-baseline/pull/2360

PR #2356, opened 2022-08-10
Remove gradle-junit-reports

pkoenig10:
FLUP to https://github.com/palantir/gradle-baseline/pull/2355


iamdanfox:
I'm gonna defer CR here to a current member (e.g. @CRogers perhaps) as I no longer have powers on this repo.

One thing to note is that with the new CircleCI 3 UI, I have found it more annoying to find compilation failures in the CI output... previously it would be hoisted to the top using this junit-xml approach, but now I have to scroll in a small window within the CI page.

pkoenig10:
Note that you can open the CI output in a separate full size window that is much more usable. This is what I typically do when looking at errors in the output.

Screen Shot 2022-09-14 at 5 56 02 AM

CRogers:
We've actually had complaints internally about not getting the compile errors right at the top any more - I didn't realise we had turned it off from baseline itself! I thought it was because we messed up setting the `CIRCLE_ARITFACTS` env var correctly.

CRogers:
Yeah, I actually think we should re-enable this feature - using the junit test reports a place to expose failures was a thing we did even on circle 2 (I don't think it's that big a deal) - and the pain of having to trawl through the output for some gradle task failure is very painful. Perhaps we might need to tweak how we report duplicate failures on Circle 3 with high parallelism but I still think this feature is very very useful and saves engineers time.

CRogers:
Actually re-enabling the functionality here: https://github.com/palantir/gradle-baseline/pull/2391

ash211:
@pkoenig10 I think this is a dupe of the now-merged https://github.com/palantir/gradle-baseline/pull/2734, so we can close

PR #2359, opened 2022-08-10
Update minimum required Gradle version

pkoenig10:
See https://github.com/palantir/gradle-baseline/pull/2351#issuecomment-1211113305

Not sure if we'd rather bump the version of revert the change. We likely have quite a few internal repos that are not yet on Gradle 7.5.


carterkozak:
Let's do this instead: #2360

PR #2363, opened 2022-08-18
Make baseline-exact-dependencies compatible with configure-on-demand

AlexLandau:
## Before this PR
With configure-on-demand enabled, these configurations get resolved
before projectsEvaluated triggers are run. (I haven't dug in fully,
but I think this is happening during task graph expansion.)

Make this check more lenient if configure-on-demand is enabled.

## After this PR
==COMMIT_MSG==
Make the `checkUnusedDependencies` tasks added by `baseline-exact-dependencies` compatible with Gradle's configure-on-demand feature.
==COMMIT_MSG==

## Possible downsides?
Unlikely


carterkozak:
I've tagged some folks with more historical context than myself to review

CRogers (approved):

PR #2364, opened 2022-08-18
Apply more task config laziness to gradle-junit-reports

AlexLandau:
## Before this PR
Applying the junit-reports plugin always causes JavaCompile and Checkstyle tasks to get instantiated.

## After this PR
==COMMIT_MSG==
Make task initialization lazier in the `junit-reports` plugin.
==COMMIT_MSG==

## Possible downsides?
None


carterkozak (approved):
Thanks!

PR #2366, opened 2022-08-24
Add an errorprone check and typed annotation for Javax -> Jakarta

mglazer:
## Before this PR

Legacy javax.ws.rs types could be submitted to methods which only accept jakarta.ws.rs types at runtime.

## After this PR

==COMMIT_MSG==
Add an errorprone check and typed annotation for Javax -> Jakarta

There is a certain class of very problematic cases whereby if you have
a method such as the following:

```
myJerseyResource.register(/* this is of type Object */ object);
```

Then if you supply a resource which includes any `javax.ws.rs`
annotations on it, then those will not be registered if your Jersey
version is 3.x or later (and you'll only find this out at runtime).

The opposite is also true if you try to supply resources annotated
with `jakarta.ws.rs` to Jersey 2.x.

To address this, this commit attempts to add an errorprone check
which lets implementors add an annotation `@ForbidJavax` to methods
which have been knowingly migrated to Jakarta EE9 and cannot
accept legacy javax types.
==COMMIT_MSG==


baseline-error-prone-annotations/src/main/java/com/palantir/errorprone/ForbidJavax.java
```diff
@@ -0,0 +1,36 @@
...
+ * or <code>jakarta.ws.rs.core.Feature</code> instead of the legacy <code>javax</code>
+ * versions of those.
+ *
+ * This sprinkling is required since many methods take a simple <code>Object</code> type
+ * and let Jersey determine what to do with the object at runtime. In order to obtain better
+ * compile time checking, this method can be applied.
+ */
+@Retention(RetentionPolicy.CLASS)
+@Target(ElementType.PARAMETER)
+public @interface ForbidJavax {}
```
carterkozak: I think this would fit better in [conjure-java-runtime `conjure-java-annotations`](https://github.com/palantir/conjure-java-runtime/tree/develop/conjure-java-annotations/src/main/java/com/palantir/conjure/java/annotations), since we already define similar enforcement annotations there.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ForbidJavaxParameterType.java
```diff
@@ -0,0 +1,133 @@
...
+                + "untyped Object. There is no auto-fix for this check, you must fix it manually")
+public final class ForbidJavaxParameterType extends BugChecker implements BugChecker.MethodInvocationTreeMatcher {
+
+    private static final Matcher<ClassTree> HAS_JAXRS_ANNOTATIONS = Matchers.hasMethod(Matchers.anyOf(
+            Matchers.hasAnnotation("javax.ws.rs.GET"),
+            Matchers.hasAnnotation("javax.ws.rs.POST"),
+            Matchers.hasAnnotation("javax.ws.rs.DELETE"),
+            Matchers.hasAnnotation("javax.ws.rs.PUT"),
+            Matchers.hasAnnotation("javax.ws.rs.HEAD"),
+            Matchers.hasAnnotation("javax.ws.rs.OPTIONS")));
```
carterkozak: iirc these are all annotated with something along the lines of `javax.ws.rs.Method`, we could search for any annotation from `javax.ws.rs.*` or annotated with the jaxrs meta-annotations

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ForbidJavaxParameterType.java
```diff
@@ -0,0 +1,133 @@
...
+        for (int i = 0; i < typeParameters.size(); ++i) {
+            VarSymbol parameter = typeParameters.get(i);
+            ExpressionTree argument = arguments.get(i);
+
+            Annotation maybeRequireJakarta = parameter.getAnnotation(ForbidJavax.class);
+            if (maybeRequireJakarta == null) {
+                continue;
+            }
+
+            if (hasJavaxInclusions(ASTHelpers.getSymbol(argument), state)) {
```
carterkozak: This might be a bit easier if we use `ASTHelpers.getResultType(argument)` which handles the method/ctor return type indirection

baseline-error-prone-annotations/src/main/java/com/palantir/errorprone/ForbidJavax.java
mglazer: So what I'm attempting by putting this here is to guarantee that this check lives on every classpath for "free", without having to push out a second check which adds the `errorprone` configuration to every build containing a separate errorprone project.

Is your suggestion simply to take this annotation and move it there, and then have baseline depend on that separate project, or move this entire check there?

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ForbidJavaxParameterType.java
mglazer: I haven't been able to find any clean method of doing such a thing....I could probably write something custom though, it'll just be a little messy.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ForbidJavaxParameterType.java
mglazer: 👍

baseline-error-prone-annotations/src/main/java/com/palantir/errorprone/ForbidJavax.java
carterkozak: Yep, sorry I was unclear! The check would live in this repo and apply at ERROR, but the annotation would live in the other repo

mglazer: https://github.com/palantir/conjure-java-runtime/pull/2442

mglazer:
When the CJR change goes through it should also resolve the error that I'm currently getting where I'm not publishing the annotations correctly somehow.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ForbidJavaxParameterType.java
```diff
@@ -0,0 +1,122 @@
...
+            return false;
+        }
+
+        return hasJavaxInclusionsOnType(resultType.asElement(), state);
+    }
+
+    private boolean hasJavaxInclusionsOnType(TypeSymbol symbol, VisitorState state) {
+        if (symbol instanceof ClassSymbol) {
+            ClassSymbol classType = (ClassSymbol) symbol;
+            ClassTree classTree = ASTHelpers.findClass(classType, state);
```
carterkozak: I suspect we won't be able to find ClassTrees for most types that are registered as they're defined elsewhere and haven't been parsed by the compiler. Can we use the `Type` object directly with errorprone TypePredicates?

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/ForbidJavaxParameterTypeTest.java
```diff
@@ -0,0 +1,131 @@
...
+        helper().addSourceLines(
+                        "Test.java",
+                        "package test;",
+                        "import jakarta.ws.rs.Path;",
+                        "import jakarta.ws.rs.GET;",
+                        "import com.palantir.errorprone.ForbidJavax;",
+                        "class Test {",
+                        "  public void requiresJakarta(@ForbidJavax Object o) {}",
+                        "  ",
+                        "  public static class Inner {",
```
carterkozak, line 206: Concretely, I think if you move the `Inner` definition to a nested class within this test class, the helper should still be able to reference it, but I don't think it would be detected due to the ClassTree search

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ForbidJavaxParameterType.java
mglazer: Hrm...TypePredicates only give me DescendentOf, so I can get `Feature` but likely could not get anything else.

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/ForbidJavaxParameterTypeTest.java
mglazer: Just added a test where it's moved to a separate file...seems to work? Or is there another case I'm not thinking of?

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ForbidJavaxParameterType.java
carterkozak: I've created a PR with an example that fails: #2367

carterkozak: Operating directly on the result type symbol will give us what we want, that provides the set of class+runtime retained annotations, list of methods and their annotations/etc

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/ForbidJavaxParameterTypeTest.java
carterkozak: I've created a PR with an example that fails: https://github.com/palantir/gradle-baseline/pull/2367

(accidentally replied to the wrong thread a moment ago, fixed)

mglazer:
Added a couple of more tests, one to handle the case where you're supplying a resource which extends from a JaxRS annotated interface, and another to assert that this errorprone check is not perfect, and can't handle the case where there is no type information at all on the ingress types.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ForbidJavaxParameterType.java
```diff
@@ -0,0 +1,131 @@
...
+                || arguments.isEmpty()
+                || arguments.size() != typeParameters.size()) {
+            return Description.NO_MATCH;
+        }
+
+        for (int i = 0; i < typeParameters.size(); ++i) {
+            VarSymbol parameter = typeParameters.get(i);
+            ExpressionTree argument = arguments.get(i);
+
+            Annotation maybeRequireJakarta = parameter.getAnnotation(ForbidJavax.class);
```
carterkozak: Thoughts on keeping `com.palantir.conjure.java.runtime:conjure-java-annotations` as a `testRuntimeOnly` dependency, and using the fqcn string to reference this annotation? In general I try to avoid adding dependencies to baseline for this sort of check because it can weigh down the build process.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/MoreMatchers.java
```diff
@@ -74,6 +76,19 @@ static <T extends ClassTree> Matcher<T> classEnclosingClass(Matcher<ClassTree> m
...
+            Name packageName = state.getName(prefix);
+            Symbol sym = ASTHelpers.getSymbol(classTree);
+            for (Compound a : sym.getRawAttributes()) {
+                if (a.type.tsym.flatName().startsWith(packageName)) {
+                    return true;
+                }
+            }
+            return false;
+        };
+    }
```
carterkozak: I think this is unused, only `SymbolPredicates.hasAnnotationWithPackage` appears to be referenced

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ForbidJavaxParameterType.java
```diff
@@ -0,0 +1,131 @@
...
+    private static final Predicate<Symbol> HAS_JAXRS_ANNOTATION =
+            SymbolPredicates.hasAnnotationWithPackage("javax.ws.rs");
+
+    private static final TypePredicate IMPLEMENTS_FEATURE = TypePredicates.isDescendantOf("javax.ws.rs.core.Feature");
+
+    @Override
+    public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
+        List<? extends ExpressionTree> arguments = tree.getArguments();
+        MethodSymbol methodSymbol = ASTHelpers.getSymbol(tree);
+        List<VarSymbol> typeParameters = methodSymbol.getParameters();
```
carterkozak: nit: Naming is a bit confusing, the method symbol has a `getTypeParameters` method, which refers to the generic type params. Calling this `parameters` may be easier to understand

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ForbidJavaxParameterType.java
mglazer: I'm good with that. And this check would just...not run if that annotation wasn't present?

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/MoreMatchers.java
mglazer: Yup, latent code.

carterkozak (approved):
Thanks!

mglazer:
👍

PR #2367, opened 2022-08-24
[stacked] Failing test example for ForbidJavaxParameterType

carterkozak:
==COMMIT_MSG==
Failing test example for ForbidJavaxParameterType
==COMMIT_MSG==


baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/ForbidJavaxParameterTypeTest.java
```diff
@@ -127,6 +127,28 @@ void stillFailsIfInnerIsSeparateClass() {
                 .doTest();
     }
 
+    public interface JaxrsAnnotatedIface {
```
mglazer, line 130: ooooh, that's what you meant....

PR #2368, opened 2022-09-01
Workaround to IDEA-301084

j-baker:
See https://github.com/JetBrains/intellij-community/pull/2135 for
more details. Even if JetBrains takes this PR, it'll be months
before it's common that our devs have a fixed version (it's common
for devs to be a version or two behind) so it seems sensible to
just workaround in baseline for hte next year or two.

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/javaversions/BaselineJavaVersion.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/javaversions/BaselineJavaVersion.java
@@ -97,7 +97,11 @@ public final class BaselineJavaVersion implements Plugin<Project> {
             @Override
             public void execute(JavaCompile javaCompileTask) {
                 javaCompileTask.getJavaCompiler().set(javaToolchain.flatMap(BaselineJavaToolchain::javaCompiler));
-                enablePreviewOnCompileTask(target, javaCompileTask.getOptions());
+                javaCompileTask
+                        .getOptions()
+                        .getCompilerArgumentProviders()
+                        .add(new EnablePreviewArgumentProvider(target));
+                hackTryToMakePreviewFeaturesWorkInIntellij(target, javaCompileTask.getOptions());
 
                 // Set sourceCompatibility to opt out of '-release', allowing opens/exports to be used.
                 javaCompileTask.doFirst(new Action<Task>() {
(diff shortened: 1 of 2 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/javaversions/BaselineJavaVersion.java
```diff
@@ -267,6 +265,16 @@ public final void checkJavaVersions() {
         }
     }
 
+    /**
+     * <a href="https://github.com/JetBrains/intellij-community/pull/2135">Further details available here.</a>
+     */
+    private static void enablePreviewOnCompileTask(
+            Provider<ChosenJavaVersion> provider, CompileOptions compileOptions) {
+        if (provider.get().enablePreview()) {
```
j-baker, line 282: This is eager and not lazy, obviously. @iamdanfox would this be a problem?

iamdanfox: The downside of this 'eager' approach is:

If the first line of your build.gradle says `apply plugin: 'com.palantir.baseline-java-version'` and then your second line says: `javaVersions { runtime = '19_PREVIEW' }`, then we will already have configured this `compileOptions` task and _not_ added the `--enable-preview` compiler arg.

CRogers: Yeah this should be lazy. You can use `getCompilerArgumentProviders().add` to give an argument provider that can return a list of size 0 or 1 to be lazy.

iamdanfox: @CRogers the compilerArgumentProviders approach is what's already on develop, but sadly when you use IntelliJ's native integration, they literally check `javaCompileTask.options.compilerArgs.contains("--enable-preview")` which does _not_ invoke my beloved command line argument provider. 

James has a fix up here: https://github.com/JetBrains/intellij-community/pull/2135

CRogers: Yeah I probably should have read that first 🤦🏻

j-baker: ok folks. I've updated this PR to actually use both methods (lazy and eager). Dan and I think that this will _most likely_ solve the problem in all practical cases because in any case where code is in subprojects, the javaVersions closure will have been set on the root and so evaluated before the java library declaration, avoiding the problem.

However, just in case laziness is an issue (e.g. you have only a root project) this will maintain the fact that --enable-preview is set, with the only downside being it's most likely set twice. This should be a no-op in all cases, should guarantee no semantic changes from present _except_ avoiding the IntelliJ bug.

CRogers (approved):

iamdanfox:
Unfortunately this seems to not work internally in our normal multi-project repos:

```
Cannot query the value of this property because it has no value available.
	at org.gradle.api.internal.provider.AbstractMinimalProvider.get(AbstractMinimalProvider.java:85)
	at com.palantir.baseline.plugins.javaversions.BaselineJavaVersions.lambda$apply$0(BaselineJavaVersions.java:66)
```

PR #2373, opened 2022-09-07
create GradleTaskConfigurationAvoidance error prone check

rzpt:
## Before this PR

There is no error prone check to prevent usages of these gradle apis which can make gradle commands unnecessarily slow. Some of the apis are also very likely to be mistakenly used.

## After this PR

==COMMIT_MSG==
Create GradleTaskConfigurationAvoidance error prone check to prevent
eager task creation.
==COMMIT_MSG==

## Possible downsides?

None.


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/GradleTaskConfigurationAvoidance.java
```diff
@@ -0,0 +1,97 @@
...
+import com.google.errorprone.matchers.Description;
+import com.google.errorprone.matchers.Matcher;
+import com.google.errorprone.matchers.Matchers;
+import com.google.errorprone.matchers.method.MethodMatchers;
+import com.sun.source.tree.ExpressionTree;
+import com.sun.source.tree.MethodInvocationTree;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
```
j-baker, line 33: drive-by: unless i'm missing something the README also needs to be updated with the addition of this check

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/GradleTaskConfigurationAvoidance.java
```diff
@@ -0,0 +1,97 @@
...
+        linkType = BugPattern.LinkType.CUSTOM,
+        severity = SeverityLevel.ERROR,
+        summary = "Forbid gradle apis that eagerly create tasks.")
```
j-baker: can this have more detail here on what the user should do instead?

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/GradleTaskConfigurationAvoidance.java
```diff
@@ -0,0 +1,98 @@
...
+import com.google.errorprone.matchers.Matchers;
+import com.google.errorprone.matchers.method.MethodMatchers;
+import com.sun.source.tree.ExpressionTree;
+import com.sun.source.tree.MethodInvocationTree;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
+        severity = SeverityLevel.ERROR,
```
ash211: can we start this at a lower severity level that shows just warnings before making this fail builds?  We have to give projects some time between getting a new baseline version that shows warnings and making those warnings fail the build.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/GradleTaskConfigurationAvoidance.java
```diff
@@ -0,0 +1,98 @@
...
+        summary = "Forbid gradle apis that eagerly create tasks. Learn more at"
+                + " https://docs.gradle.org/current/userguide/task_configuration_avoidance.html")
```
ash211, line 37: is there more specific guidance devs should follow how to fix any of these warnings that will be thrown?  Is https://docs.gradle.org/current/userguide/task_configuration_avoidance.html#sec:task_configuration_avoidance_pitfalls the best guidance that's out there and changing from eager task creation to lazy task creation?

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/GradleTaskConfigurationAvoidance.java
```diff
@@ -0,0 +1,98 @@
...
+@SuppressWarnings("deprecation")
+public final class GradleTaskConfigurationAvoidance extends BugChecker
+        implements BugChecker.MethodInvocationTreeMatcher {
```
ash211, line 40: does this also work on `.gradle` files written in Groovy?  The tests are only in Java, so I'm not sure if this would run on `.gradle` files.  Or is this only meant to run on gradle plugins written in Java?

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/GradleTaskConfigurationAvoidance.java
rzpt: Sure. Does anyone look at warnings though? Most projects that are affected are gradle plugins, which are probably not actively developed.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/GradleTaskConfigurationAvoidance.java
rzpt: As someone who doesn't really work on gradle plugins, I don't know the answer to this.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/GradleTaskConfigurationAvoidance.java
rzpt: I believe it does not because error prone only runs on java.

PR #2383, opened 2022-09-15
Enable switchBlockAsSingleDecisionPoint

pkoenig10:
See https://github.com/palantir/gradle-baseline/pull/2365#discussion_r971528780

PR #2386, opened 2022-09-15
Revert "Enable switchBlockAsSingleDecisionPoint (#2383)"

carterkozak:
This reverts commit e132e2ddb3b50fd35c29190f7766780455f70cb4.

==COMMIT_MSG==
Revert "Enable switchBlockAsSingleDecisionPoint (#2383)" due to configuration parsing failures.
==COMMIT_MSG==

### DIFF
--- a/gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
+++ b/gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
@@ -420,9 +420,7 @@
             <property name="tagOrder" value="@param, @return, @throws, @deprecated"/>
             <property name="target" value="CLASS_DEF, INTERFACE_DEF, ENUM_DEF, METHOD_DEF, CTOR_DEF, VARIABLE_DEF"/>
         </module>
-        <module name="CyclomaticComplexity"> <!-- Java Coding Guidelines: Reduce Cyclomatic Complexity -->
-            <property name="switchBlockAsSingleDecisionPoint">true</property>
-        </module>
+        <module name="CyclomaticComplexity"/> <!-- Java Coding Guidelines: Reduce Cyclomatic Complexity -->
         <module name="DesignForExtension"> <!-- Java Coding Guidelines: Design for extension -->
             <property name="ignoredAnnotations" value="ParameterizedTest, Test, Before, BeforeEach, After, AfterEach, BeforeClass, BeforeAll, AfterClass, AfterAll"/>
         </module>


CRogers (approved):

pkoenig10:
@carterkozak What exactly is failing here? Can we just fix the config instead of reverting?

gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
```diff
@@ -420,9 +420,7 @@
             <property name="tagOrder" value="@param, @return, @throws, @deprecated"/>
             <property name="target" value="CLASS_DEF, INTERFACE_DEF, ENUM_DEF, METHOD_DEF, CTOR_DEF, VARIABLE_DEF"/>
         </module>
-        <module name="CyclomaticComplexity"> <!-- Java Coding Guidelines: Reduce Cyclomatic Complexity -->
-            <property name="switchBlockAsSingleDecisionPoint">true</property>
-        </module>
+        <module name="CyclomaticComplexity"/> <!-- Java Coding Guidelines: Reduce Cyclomatic Complexity -->
```
schlosna, line 423: Revert seems fine, but should we just fix this?

```suggestion
        <module name="CyclomaticComplexity"> <!-- Java Coding Guidelines: Reduce Cyclomatic Complexity -->
            <property name="switchBlockAsSingleDecisionPoint" value="true"/>
        </module>
```

schlosna: also surprised a test didn't catch this, and excavator bump https://github.com/palantir/gradle-baseline/pull/2385 had to catch the [failure](https://app.circleci.com/pipelines/github/palantir/gradle-baseline/4450/workflows/25cc584a-2ef8-4a5a-9e5c-3888a1ef9531/jobs/35168)

```
Caused by: org.xml.sax.SAXParseException; systemId: file:/home/circleci/project/.baseline/checkstyle/checkstyle.xml; lineNumber: 406; columnNumber: 63; Attribute "value" is required and must be specified for element type "property".
```

carterkozak:
> Can we just fix the config instead of reverting?

My preference is to get back to a known good state asap, unrevert+fix isn't substantially more difficult after the revert has been rolled out, but it means if the second attempt doesn't work, we don't have to do commit archeology to get ourselves back to a stable place.

I didn't do enough testing when I approved the earlier PR, so I'd like to ensure I have the time to thoroughly test before we fix forward, and I don't have availability in the next couple hours. I'd really like to see an integration test that prevents this sort of thing moving forward.

PR #2389, opened 2022-09-16
Roll back nullaway+checkerframework for jdk15 support

carterkozak:
Newer release support 11 and 17, but do not support the now-sunset jdk-15.

==COMMIT_MSG==
Roll back nullaway+checkerframework for jdk15 support
==COMMIT_MSG==

## Possible downsides?
Unfortunately we can't use the latest release of checkerframework libs until folks have migrated away from jdk15 MTS builds.


gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineNullAwayIntegrationTest.groovy
```diff
@@ -85,4 +85,24 @@ class BaselineNullAwayIntegrationTest extends AbstractPluginTest {
...
+
+    def 'compileJava succeeds when null-away finds no errors on jdk15'() {
+        when:
+        buildFile << standardBuildFile.replace('libraryTarget = 17', 'libraryTarget = 15')
+        file('src/main/java/com/palantir/test/Test.java') << validJavaFile
+
+        then:
+        BuildResult result = with('compileJava').build()
+        result.task(":compileJava").outcome == TaskOutcome.SUCCESS
+    }
```
carterkozak, line 97: This test fails without the change.

xRuiAlves (approved):
👍

PR #2390, opened 2022-09-16
Exclude Spark UnusedStubClass from class uniqueness

j-baker:
Spark UnusedStubClass exists for mechanical reasons and is as described, never used. With this PR it is skipped, like module-infos.

Potential badness: It's possible that this change would block excavators on certain projects by changing the set of duplicated classes. I think it's likely that this would simply require approval from a human, depending on how the excavator was written.


carterkozak (approved):
Looks reasonable to me. At the limit this would become a large central list, but if it grows we can consider moving to an extension model for archetype-specific values

PR #2393, opened 2022-09-19
lazily configure NullAway dependencies

carterkozak:
## Before this PR
Plugin failed to apply in some cases

## After this PR
==COMMIT_MSG==
lazily configure NullAway dependencies to successfully apply NullAway without ordering issues
==COMMIT_MSG==

## Possible downsides?
~~Unable to repro the failure which prompted this~~ Solved

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineNullAway.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineNullAway.java
@@ -22,9 +22,11 @@ import net.ltgt.gradle.errorprone.ErrorProneOptions;
 import org.gradle.api.Action;
 import org.gradle.api.Plugin;
 import org.gradle.api.Project;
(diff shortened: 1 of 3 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineNullAway.java
```diff
@@ -60,11 +75,18 @@ public void execute(ErrorProneOptions options) {
         });
     }
 
-    private static void configureErrorProneOptions(Project project, Action<ErrorProneOptions> action) {
-        project.getTasks().withType(JavaCompile.class).configureEach(new Action<JavaCompile>() {
+    private static void configureErrorProneOptions(Project proj, Action<ErrorProneOptions> action) {
+        proj.afterEvaluate(new Action<Project>() {
```
carterkozak, line 79: This matches our internal plugin configuration

pkoenig10 (approved):

PR #2397, opened 2022-09-20
Disable NullAway for Test Source

bmarcaur:
## Before this PR
`NullAway` was interacting poorly with mockito and other testing patterns

## After this PR
`NullAway` should no longer fail on tests

## Possible downsides?
Difficult to enable if you care about that thing


bmarcaur:
Struggling to get this to publish locally so I tested this by putting this:

```
tasks.withType(JavaCompile) {
    options.errorprone.disable 'NullAway'
}
```

In an example project and the failures went away (poor test but was proving out that errorprone is enough to disable this).

carterkozak:
Did you test that the value of `errorProneOptions.getCompilingTestOnlyCode()` was what we expect?

bmarcaur:
> Did you test that the value of `errorProneOptions.getCompilingTestOnlyCode()` was what we expect?

Yup! Just followed through with a local maven publish. Tested that adding this line preserves the failures on main sourcesets, but also no longer fails on test sourcesets.

carterkozak (approved):

PR #2399, opened 2022-09-20
Ignore fields annotated with `@Mock`

ntrinquier:
## Before this PR
```java
@Mock
private Object foo;
```

would yield a false positive. See https://github.com/uber/NullAway/issues/280 for more context on the problem and the fix.

## After this PR

==COMMIT_MSG==
Ignore fields annotated with `@Mock`
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-2399.v2.yml
+++ b/changelog/@unreleased/pr-2399.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Ignore fields annotated with `@Mock`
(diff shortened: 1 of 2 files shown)


carterkozak:
This shouldn't be necessary in the latest release, where we don't use NullAway in any test sources.

ntrinquier:
> This shouldn't be necessary in the latest release, where we don't use NullAway in any test sources.

Ah great, I'll wait to pick up the latest then!

PR #2400, opened 2022-09-20
Allow projects which don't use jdk-15 to resolve latest nullaway

carterkozak:
## Before this PR
Crashes due to bugs resolved in newer nullaway releases

## After this PR

==COMMIT_MSG==
Allow projects which don't use jdk-15 to resolve latest nullaway
==COMMIT_MSG==

## Possible downsides?
Bit of a hack, but likely better than alternatives?

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineNullAway.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineNullAway.java
@@ -16,10 +16,13 @@
 
 package com.palantir.baseline.plugins;
 
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
+import com.palantir.baseline.plugins.javaversions.BaselineJavaVersions;
 import java.util.Optional;
 import net.ltgt.gradle.errorprone.ErrorProneOptions;
 import org.gradle.api.Action;
+import org.gradle.api.JavaVersion;
 import org.gradle.api.Plugin;
 import org.gradle.api.Project;
 import org.gradle.api.artifacts.Configuration;
@@ -73,6 +76,7 @@ public final class BaselineNullAway implements Plugin<Project> {
                 options.option("NullAway:CheckOptionalEmptiness", "true");
             }
         });
+        newerNullAwayInNonJdk15Projects(project);
     }
 
     private static void configureErrorProneOptions(Project proj, Action<ErrorProneOptions> action) {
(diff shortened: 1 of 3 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineNullAway.java
```diff
@@ -90,4 +102,43 @@ public void execute(JavaCompile javaCompile) {
             }
         });
     }
+
+    // Workaround for nullaway bugs in the last release to support jdk-15. Projects that don't use jdk-15
+    // resolve a newer nullaway version with bug-fixes. This may be deleted after we've rolled everything
+    // off jdk-15 MTS.
+    private static void newerNullAwayInNonJdk15Projects(Project project) {
+        project.getRootProject().getPlugins().hasPlugin(BaselineJavaVersions.class);
+
```
carterkozak:
```suggestion
```

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineNullAway.java
```diff
@@ -90,4 +102,41 @@ public void execute(JavaCompile javaCompile) {
...
+                    @Override
+                    public void execute(Configuration _files) {
+                        for (String dep : NON_15_DEPS) {
+                            project.getDependencies().addProvider("errorprone", project.provider(() -> {
+                                if (anyProjectUsesJava15(project)) {
+                                    // Cannot upgrade dependencies in this case because newer nullaway/checkerframework
+                                    // are not compatible with java 15, but fully support jdk11 and jdk17.
+                                    return null;
+                                }
+                                return dep;
```
carterkozak: Simplify
```suggestion
                                // Cannot upgrade dependencies in this case because newer nullaway/checkerframework
                                // are not compatible with java 15, but fully support jdk11 and jdk17.
                                return anyProjectUsesJava15(project) ? null : dep;
```

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineNullAway.java
```diff
@@ -90,4 +101,38 @@ public void execute(JavaCompile javaCompile) {
...
+                    }
+                })
+                .configureEach(new Action<Configuration>() {
+                    @Override
+                    public void execute(Configuration _files) {
+                        for (String dep : NON_15_DEPS) {
+                            project.getDependencies().addProvider("errorprone", project.provider(() -> {
+                                // Cannot upgrade dependencies in this case because newer nullaway/checkerframework
+                                // are not compatible with java 15, but fully support jdk11 and jdk17.
+                                return anyProjectUsesJava15(project) ? null : dep;
```
fawind, line 123: I wonder if this should be the other way around? Per default, we use the latest versions but downgrade to the old compatible version when using j15? That way, j15 is the special case and usage of this hack will phase out naturally once projects roll to j17 (plus excavator will continue to update our used versions)?

carterkozak: How would you suggest doing that? Part of the issue is that it would bump errorprone deps as well to a version that’s no longer compatible with jdk15. Gcv makes upgrades much easier than downgrades :-/

fawind: Hmm yeah, I was thinking of using `resolutionStrategy.force("java-15-dep-coordinate")` for all deps that we need to pin to a specific version when using j15. But yeah, we would have to enumerate all the error-prone deps that are also not compatible and this might cause issues when other deps depend on them.

fawind: Hmm yeah, this is becoming more complicated. I guess its fine to put this in for now, given that jdk 15 is EOL soon and we can remove the special casing.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineNullAway.java
```diff
@@ -36,6 +38,14 @@ public final class BaselineNullAway implements Plugin<Project> {
     /** We may add a gradle extension in a future release allowing custom additional packages. */
     private static final ImmutableSet<String> DEFAULT_ANNOTATED_PACKAGES = ImmutableSet.of("com.palantir");
 
+    private static final ImmutableList<String> NON_15_DEPS = ImmutableList.of(
+            "com.uber.nullaway:nullaway:0.10.1",
+            // align 'org.checkerframework' dependency versions on
+            // the current latest version.
+            "org.checkerframework:dataflow-errorprone:3.25.0",
+            "org.checkerframework:dataflow-nullaway:3.25.0",
+            "org.checkerframework:checker-qual:3.25.0");
```
fawind, line 47: We should probably also move nullaway and checkerframework to this section in versions.props to prevent excavator updating it directly:

https://github.com/palantir/gradle-baseline/blob/c1552fa1c5eb551aa6b4316aa07521251c17077e/versions.props#L42-L48

carterkozak: I could go either way, we do want to upgrade it eventually, and our test coverage will prevent version-bumps from going through (jdk-15 nullaway integration test won't work!)

fawind: Ah nice, just saw that you added regression tests in https://github.com/palantir/gradle-baseline/pull/2389

fawind (approved):

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineNullAway.java
carterkozak: Baseline-java-versions is super helpful for this style of testing :-]

PR #2401, opened 2022-09-20
Revert "Disable NullAway for Test Source"

bmarcaur:
Turns out that this breaks if you do not have the NullAway plugin installed. Will move this to that plugin instead.

Reverts palantir/gradle-baseline#2397


bmarcaur:
Fail forward https://github.com/palantir/gradle-baseline/pull/2402 🥳

PR #2405, opened 2022-09-22
Error-prone is enabled in idea for uniformity with CLI compilation

carterkozak:
error-prone requires several javac options to access compiler innards, which have the side-effect of enabling immutables to function correctly. Given this spooky interaction and our migration toward jdk17 source, it's best to provide uniform behavior.

==COMMIT_MSG==
Error-prone is enabled in idea for uniformity with CLI compilation
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-2405.v2.yml
+++ b/changelog/@unreleased/pr-2405.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Error-prone is enabled in idea for uniformity with CLI compilation
+  links:
+  - https://github.com/palantir/gradle-baseline/pull/2405
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
@@ -22,7 +22,6 @@ import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.MoreCollectors;
-import com.palantir.baseline.IntellijSupport;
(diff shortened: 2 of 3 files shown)


pkoenig10:
Should we just remove this logic? The whole point of adding it in the first place is to disable it in IntelliJ. If we're not doing that then it seems a bit unnecessary to keep this.

Unless we think there are people that want to opt-in?

pkoenig10 (approved):

fawind:
Think the test fails to handle the build result: https://output.circle-artifacts.com/output/job/d51da10e-b56d-4384-8d83-fcbf7e70ac8b/artifacts/0/~/artifacts/junit/gradle-baseline-java/test/classes/com.palantir.baseline.BaselineErrorProneIntegrationTest.html#error-prone%20is%20not%20disabled%20in%20IntelliJ

carterkozak:
@pkoenig10 I'm making a minimal change for now, the property itself predates our intellij logic, so I'm leaving that alone for the time being. Figure it's best to avoid behavior changes in edge cases outside of intellij as that's easier than verifying nobody has used the boolean value in unexpected ways.

Medium term, I'd like to see how folks react to errorprone failures in the IDE before making larger changes, we may want to filter out certain checks from idea based on feedback (which would unfortunately bust caches)

fawind:
> Should we just remove this logic? The whole point of adding it in the first place is to disable it in IntelliJ. If we're not doing that then it seems a bit unnecessary to keep this.
> Unless we think there are people that want to opt-in?

We have some usage here and there:
* https://github.com/palantir/jvm-diagnostics/blob/eace9eb294f36658228a59671c1062fe426f1b85/gradle.properties#L2
* And the unit-test-11 template uses it: https://github.com/palantir/human-readable-types/blob/154eb5695268a58347a66b55b3cccce72d2f2944/.circleci/config.yml#L111

Might not be necessary anymore but probably lower lift to just leave it around.

PR #2414, opened 2022-09-28
Exclude javadoc @link from linelength

proggeo:
`@link` in javadoc should not break into multiple lines

## Before this PR
Using `@link` with a long reference would trigger LineLength violation.

## After this PR
`@link` javadoc comment lines are exempted from LineLength check

### DIFF
--- a/gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
+++ b/gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
@@ -49,7 +49,7 @@
     </module>
     <module name="LineLength"> <!-- Java Style Guide: No line-wrapping -->
         <property name="max" value="120"/>
-        <property name="ignorePattern" value="^package.*|^import.*|a href|href|http://|https://|ftp://"/>
+        <property name="ignorePattern" value="^package.*|^import.*|a href|href|http://|https://|ftp://|\* \{@link"/>
     </module>
     <module name="TreeWalker">
         <module name="SuppressionCommentFilter"/> <!-- baseline-gradle: README.md -->
--- a/gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
+++ b/gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
@@ -49,7 +49,7 @@
     </module>
     <module name="LineLength"> <!-- Java Style Guide: No line-wrapping -->
         <property name="max" value="120"/>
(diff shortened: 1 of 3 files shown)


palantirtech:
Thanks for your interest in [palantir/gradle-baseline](https://github.com/palantir/gradle-baseline), @proggeo! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
```diff
@@ -49,7 +49,7 @@
     </module>
     <module name="LineLength"> <!-- Java Style Guide: No line-wrapping -->
         <property name="max" value="120"/>
-        <property name="ignorePattern" value="^package.*|^import.*|a href|href|http://|https://|ftp://"/>
+        <property name="ignorePattern" value="^package.*|^import.*|a href|href|http://|https://|ftp://|\* \{@link"/>
```
carterkozak: I think just the `{@link` piece is sufficient, no need to match `* {@link`, this way we match `a href` handling.

proggeo: imo we should still encourage shorter lines. i.e., something like 
```
* My amazing function runs the eternal engine. See {@link com.superlongcorpname.thebestteamever.AmazingEternalEngine#runThisSuperDuperAmazingEngine}
```

should still be forced to look like

```
* My amazing function runs the eternal engine. See 
* {@link com.superlongcorpname.thebestteamever.AmazingEternalEngine#runThisSuperDuperAmazingEngine}
```

carterkozak: I get that, however consistency is more important with this sort of thing than optimality.

For example, how would this work in the context of a list in javadoc (or other nested structure used in high-quality documentation).

If we want to bike-shed on optimality, the following results in the smallest maximum line length, but isn't covered by either suppression (which is fine imo because it matches the behavior of `a href`!)

```java
/**
 * My amazing function runs the eternal engine. See {@link
 * com.superlongcorpname.thebestteamever.AmazingEternalEngine#runThisSuperDuperAmazingEngine
 * }
 */
```

carterkozak (approved):
lgtm, thanks!

PR #2416, opened 2022-09-30
Implement ImmutableMapDuplicateKeyStrategy for obvious map construction

carterkozak:
==COMMIT_MSG==
Implement ImmutableMapDuplicateKeyStrategy for obvious map construction. By default we suggest replacing `build()` with the more descriptive `buildOrThrow()`, however `buildKeepingLast()` may be the original intent in some cases.
==COMMIT_MSG==

## Possible downsides?
Code churn


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ImmutableMapDuplicateKeyStrategy.java
```diff
@@ -0,0 +1,56 @@
+/*
+ * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
```
carterkozak:
```suggestion
 * (c) Copyright 2022 Palantir Technologies Inc. All rights reserved.
```

schlosna (approved):
Thanks for adding this. I'm curious to see the generated excavators as this rolls out.

PR #2421, opened 2022-10-03
adding errorprone rule for inclusive language

lsingh123:
## After this PR

==COMMIT_MSG==
==COMMIT_MSG==


carterkozak:
Checkstyle may be a better option for this because it isn't limited to non-generated, compiled java files. Rollout strategy will be tricky, perhaps a new plugin that acts as a marker for baseline to apply a slightly modified checkstyle configuration (we do something similar with palantir-java-format is applied, removing rules that the formatter may not satisfy).

PR #2425, opened 2022-10-04
adding inclusive code checks to checkstyle.xml

lsingh123:
Add checkstyle rules for inclusive code.


gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
```diff
@@ -18,6 +18,67 @@
     <module name="NewlineAtEndOfFile"> <!-- Java Style Guide: Line ending: LF -->
         <property name="lineSeparator" value="lf"/>
     </module>
+    <!-- Inclusive Code checks: https://wiki.palantir.tech/display/DIVERSITY/Inclusive+Language+Glossary -->
```
carterkozak: Please avoid internal links from public repositories. We should **absolutely** include our inclusive language requirements in the documentation portion of this repository, any chance you would be interested in contributing that as a separate PR?

gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
```diff
@@ -18,6 +18,67 @@
...
+        <property name="format" value="ablebodied"/>
+        <property name="message" value="To convey the same idea and be more inclusive, consider nondisabled instead of ablebodied."/>
+    </module>
+    <module name="RegexpSingleline"> 
+        <property name="format" value="blackout"/>
+        <property name="message" value="To convey the same idea and be more inclusive, consider downtime instead of blackout."/>
+    </module>
+    <module name="RegexpSingleline"> 
+        <property name="format" value="blacklist"/>
+        <property name="message" value="To convey the same idea and be more inclusive, consider denylist instead of blacklist."/>
```
carterkozak: I like the wording and in general appreciate the relaxed tone, but I think we can take a stronger stance than "consider" here. "Please use ", perhaps?

I don't want to bikeshed on replacements, but we may be best served by providing a list of alternatives -- I tend toward "banned" or "disallowed" rather than "denylist", but it depends on the context.

gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
```diff
@@ -18,6 +18,67 @@
...
+    <module name="RegexpSingleline"> 
+        <property name="format" value="ablebodied"/>
+        <property name="message" value="To convey the same idea and be more inclusive, consider nondisabled instead of ablebodied."/>
+    </module>
+    <module name="RegexpSingleline"> 
+        <property name="format" value="blackout"/>
+        <property name="message" value="To convey the same idea and be more inclusive, consider downtime instead of blackout."/>
+    </module>
+    <module name="RegexpSingleline"> 
+        <property name="format" value="blacklist"/>
```
carterkozak, line 33: Let's use case-insensitive matches
```suggestion
        <property name="format" value="(?i)blacklist"/>
```

gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
lsingh123: Apologies for getting to this so late - just saw this review. Yes, happy to! I'll put that up as a separate PR.

gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
lsingh123: cool, I can offer a list! i agree that it's hard to provide a single correct replacement without context

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineConfig.java
```diff
@@ -102,41 +107,83 @@ public void execute(Task task) {
...
-                    DOMSource source = new DOMSource(document);
-                    StreamResult result = new StreamResult(new FileWriter(checkstyleXml.toFile()));
-                    transformer.transform(source, result);
+                    writeEditedCheckstyle(document, checkstylePath);
+                } catch (Exception e) {
+                    throw new RuntimeException("Unable to patch " + checkstylePath, e);
+                }
+            }
+
+            if (!project.hasProperty(inclusiveCodeCheckOn)) {
```
carterkozak, line 131: Thoughts on using a simple gradle plugin instead of a gradle.properties line? That way enablement works via `apply plugin: 'com.palantir.baseline-inclusive-language'`, where it's easier for us to leverage existing machinery (and state lives in the root build.gradle rather than being split between build.gradle and gradle.properties).

The palantir-java-format enablement has some legacy gradle properties, but in practice it's enabled by using `apply plugin: 'com.palantir.java-format'` at this point

gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineConfigIntegrationTest.groovy
```diff
@@ -138,4 +138,53 @@ class BaselineConfigIntegrationTest extends AbstractPluginTest {
...
+        """.stripIndent()
+
+        when:
+        with('baselineUpdateConfig').build()
+
+        then:
+        new File(projectDir, '.baseline/checkstyle/checkstyle.xml').readLines().any {
+            it.contains 'To convey the same idea and be more inclusive,'
+        }
+    }
```
carterkozak, line 189: great tests!

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineConfig.java
```diff
@@ -77,6 +78,10 @@ private class BaselineUpdateConfigAction implements Action<Task> {
         private final Configuration configuration;
         private final Project rootProject;
 
+        private final String inclusiveCodeCheckOn = "inclusive-language";
+
+        private final String checkstylePathString = "checkstyle/checkstyle.xml";
```
carterkozak, line 83: style nit: let's make these `private static final` constants and update to SCREAMING_SNAKE_CASE

gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
```diff
@@ -18,6 +18,97 @@
...
         <property name="lineSeparator" value="lf"/>
     </module>
+    <!-- Inclusive Code checks: https://github.com/palantir/gradle-baseline#inclusive-language -->
+    <module name="RegexpSingleline"> 
+        <property name="format" value="ablebodied"/>
+        <property name="ignoreCase" value="true"/>
+        <property name="message" value="To convey the same idea and be more inclusive, please use nondisabled instead of ablebodied."/>
+    </module>
+    <module name="RegexpSingleline"> 
+        <property name="format" value="blackout"/>
```
carterkozak, line 28: Once this is applied, will the inclusive language checks fail on this repo due to this file?
On projects which apply this check, will the verbiage in `.baseline/checkstyle/checkstyle.xml` cause failures on otherwise passing projects?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineConfig.java
lsingh123: That sounds good! A plugin felt like a heavier lift than just a property, but if that's the standard then I will do that!

gradle-baseline-java-config/resources/checkstyle/checkstyle.xml
lsingh123: Oh yikes good point. I'll exclude files in `.baseline/checkstyle/checkstyle.xml`. For this repo, I can add a repo-specific checkstyle suppression for the specific files involved in the rule.

carterkozak: I took a quick google (not exhaustive) and didn't see any easy ways to obfuscate text, ideally `git grep <disallowed-phrase>` wouldn't turn up any results

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineConfig.java
carterkozak: To clarify, since this bit probably isn't obvious, the logic can be the same (since it's the update-config task that we're mutating in practice), but we'll use `hasPlugin(BaselineInclusiveLanguage.class)` instead of `hasProperty`.

lsingh123:
I unfortunately won't have time to finish this out, remaining work is addressing Carter's comments and creating a gradle plugin to roll this out.

Issue #2413, opened 2022-09-28
ConcurrentModificationException from BaselineNullAway when running ./gradlew classes


ash211:
## What happened?

On an internal repo, I ran `./gradlew classes --stacktrace` and got this stacktrace:

```
FAILURE: Build failed with an exception.

* What went wrong:
Could not determine the dependencies of task ':compileJava'.
> Could not resolve all dependencies for configuration ':annotationProcessor'.
   > java.util.ConcurrentModificationException (no error message)

* Try:
> Run with --info or --debug option to get more log output.

* Exception is:
org.gradle.api.internal.tasks.TaskDependencyResolveException: Could not determine the dependencies of task ':compileJava'.
        at org.gradle.api.internal.tasks.CachingTaskDependencyResolveContext.getDependencies(CachingTaskDependencyResolveContext.java:68)
        at org.gradle.execution.plan.TaskDependencyResolver.resolveDependenciesFor(TaskDependencyResolver.java:49)
        at org.gradle.execution.plan.LocalTaskNode.getDependencies(LocalTaskNode.java:147)
        at org.gradle.execution.plan.LocalTaskNode.resolveDependencies(LocalTaskNode.java:121)
        at org.gradle.execution.plan.DefaultExecutionPlan.doAddNodes(DefaultExecutionPlan.java:209)
        at org.gradle.execution.plan.DefaultExecutionPlan.addEntryTasks(DefaultExecutionPlan.java:174)
        at org.gradle.execution.plan.DefaultExecutionPlan.addEntryTasks(DefaultExecutionPlan.java:158)
        at org.gradle.execution.TaskNameResolvingBuildConfigurationAction.configure(TaskNameResolvingBuildConfigurationAction.java:47)
        at org.gradle.execution.DefaultBuildConfigurationActionExecuter.configure(DefaultBuildConfigurationActionExecuter.java:49)
        at org.gradle.execution.DefaultBuildConfigurationActionExecuter.access$000(DefaultBuildConfigurationActionExecuter.java:25)
        at org.gradle.execution.DefaultBuildConfigurationActionExecuter$1.proceed(DefaultBuildConfigurationActionExecuter.java:62)
        at org.gradle.execution.DefaultTasksBuildExecutionAction.configure(DefaultTasksBuildExecutionAction.java:48)
        at org.gradle.execution.DefaultBuildConfigurationActionExecuter.configure(DefaultBuildConfigurationActionExecuter.java:49)
[...]

PR #2427, opened 2022-10-06
Fix ConcurrentModificationExceptions thrown from BaselineNullAway

ash211:
Addresses https://github.com/palantir/gradle-baseline/issues/2413

## Before this PR
`java.util.ConcurrentModificationException` thrown from this plugin:

```
Caused by: java.util.ConcurrentModificationException
        at org.gradle.api.internal.collections.FilteredCollection$FilteringIterator.findNext(FilteredCollection.java:121)
        at org.gradle.api.internal.collections.FilteredCollection$FilteringIterator.next(FilteredCollection.java:140)
        at org.gradle.api.internal.DefaultDomainObjectCollection$IteratorImpl.next(DefaultDomainObjectCollection.java:485)
        at com.palantir.baseline.plugins.BaselineNullAway.lambda$anyProjectUsesJava15$4(BaselineNullAway.java:138)
        at com.palantir.baseline.plugins.BaselineNullAway.anyProjectUsesJava15(BaselineNullAway.java:137)
        at com.palantir.baseline.plugins.BaselineNullAway$5.lambda$execute$0(BaselineNullAway.java:128)
```

## After this PR
==COMMIT_MSG==
Fix ConcurrentModificationExceptions thrown from BaselineNullAway
==COMMIT_MSG==

## Possible downsides?
I think this exception indicates that the project's task list is being modified/resolved at the same time as we iterate over it in this class. 

By making a copy before iterating, I think we'll reduce the CMEs being thrown, but not actually fix the underlying issue of concurrent changes happening.  I'm not super familiar with Gradle internals, but think we may instead need to do something where BaselineNullAway runs in a `doLast` so it doesn't interfere with project resolution.

This fix was confirmed on an internal failing repro, but we don't have a test case for it as part of this PR.

~~By using the TaskCollection's `.anyMatch()` directly, we hope that it is a bit smarter about modification during iteration versus converting to a Stream and using `Stream#anyMatch()`.~~

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineNullAway.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineNullAway.java
@@ -134,7 +134,7 @@ public final class BaselineNullAway implements Plugin<Project> {
 
     private static boolean anyProjectUsesJava15(Project proj) {
         return proj.getAllprojects().stream()
-                .anyMatch(project -> project.getTasks().withType(JavaCompile.class).stream()
+                .anyMatch(project -> ImmutableList.copyOf(project.getTasks().withType(JavaCompile.class)).stream()
                         .anyMatch(comp -> {
                             JavaVersion javaVersion = JavaVersion.toVersion(comp.getTargetCompatibility());
                             return javaVersion == JavaVersion.VERSION_15;
--- a/changelog/@unreleased/pr-2427.v2.yml
+++ b/changelog/@unreleased/pr-2427.v2.yml
@@ -1,0 +1,5 @@
+type: fix
+fix:
+  description: Fix ConcurrentModificationExceptions thrown from BaselineNullAway
+  links:
+  - https://github.com/palantir/gradle-baseline/pull/2427


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineNullAway.java
```diff
@@ -134,7 +134,7 @@ public void execute(Configuration _files) {
 
     private static boolean anyProjectUsesJava15(Project proj) {
         return proj.getAllprojects().stream()
-                .anyMatch(project -> project.getTasks().withType(JavaCompile.class).stream()
+                .anyMatch(project -> ImmutableList.copyOf(project.getTasks().withType(JavaCompile.class)).stream()
```
carterkozak, line 137: I wonder if we would be better off replacing the `Collection.stream()` here using something along these lines:
```java
project.getTasks().withType(JavaCompile.class).matching(comp -> {
  JavaVersion javaVersion = JavaVersion.toVersion(comp.getTargetCompatibility());
  return javaVersion == JavaVersion.VERSION_15;
}).isEmpty()
```

I suspect the TaskContainer may be a bit smarter about modification within iteration when it controls the iteration rather than returning a stream wrapping a spliterator, but it's difficult to say without a reliable reproducer.

ash211: I tried this by testing `gw classes --stacktrace` on my internal failing repo, and that command:

- fails on `com.palantir.baseline:gradle-baseline-java:4.180.0`
- succeeds on `com.palantir.baseline:gradle-baseline-java:4.180.0-8-ga477298`
- fails on `com.palantir.baseline:gradle-baseline-java:4.180.0-9-g05114a5`

So I think the ImmutableList copy approach works and the TaskCollection one doesn't.  At least it gets these runs to pass.

carterkozak (approved):
Thanks!

PR #2441, opened 2022-10-31
upgrade nullaway

carterkozak:
This nullaway release includes a fix for an NPE when `CheckOptionalEmptiness` is enabled, which resulted in us disabling optional emptiness checks initially. I have not re-enabled optional emptiness checks in this PR because I'd like to avoid causing excavators to get stuck.
Also, improved support for boolean constraints on `@Contract` annotations to enable better analysis of conjure generated error factory methods.
==COMMIT_MSG==
upgrade nullaway
==COMMIT_MSG==


versions.props
```diff
@@ -47,4 +46,8 @@ com.palantir.javaformat:gradle-palantir-java-format = 1.1.0
 com.diffplug.spotless:spotless-plugin-gradle = 6.6.0
 # Groovy versions must be compatible with gradle
 org.codehaus.groovy:* = 3.0.10
+# error-prone isn't compatible with checkerframework 3.26.0 yet
+org.checkerframework:* = 3.25.0
+# Allow nullaway to specify its own checkerframework release
+org.checkerframework:dataflow-nullaway = 3.26.0
```
schlosna, line 52: will these version mismatches cause any runtime conflicts / `LinkageError`s?

schlosna (approved):
I think this should be ok, but have a question about safety of checker framework version mismatches on `dataflow-nullaway`

versions.props
carterkozak: Nope, nullaway provides test coverage for this sort of thing, versions are decoupled from checkerframework to allow for broader compatibility with errorprone

PR #2446, opened 2022-11-04
adding inclusive language standards to the README

lsingh123:
For now, this PR simply adds our inclusive standards as a best practice. In the future, we will [enforce these standards as a checkstyle rule.](https://github.com/palantir/gradle-baseline/pull/2425)

DEI and Legal have approved this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -254,32 +254,6 @@ in `.baseline/copyright/*.txt`. The largest file (sorted lexicographically) will
 
 To automatically update all files with mismatching/missing copyrights, run `./gradlew format`.
 
-### Inclusive Language 
-
-Baseline enforces the use of inclusive language in our code, according to the standards developed by our Diversity, 
-Equity, and Inclusion team. Below are some terms that can carry hidden meanings and our recommended replacements.
-
-| Term               | Suggested Replacements |
-|--------------------| ---------------------- |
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -254,6 +254,32 @@ in `.baseline/copyright/*.txt`. The largest file (sorted lexicographically) will
 
 To automatically update all files with mismatching/missing copyrights, run `./gradlew format`.
 
+### Inclusive Language
```
carterkozak: Thoughts on moving this to fit into https://github.com/palantir/gradle-baseline/tree/develop/docs/best-practices/java-coding-guidelines
It's not java-specific, but otherwise I think that's a reasonable place for guidelines to live (where this won't describe a specific check, but rather our guidance which will future checks may enforce)

carterkozak: A new file in that directory that isn't specific to java could also work nicely, though a bit harder for folks to discover. I defer to your judgement :-]

lsingh123: good idea! moved them to a new file in that directory and linked from the best practices readme.

lsingh123: i'll update the top-level readme with information about the checkstyle enforcement in that PR.

carterkozak (approved):
thanks!

PR #2465, opened 2023-01-06
baseline-immutables handles annotationProcessor configuration extendsFrom

CRogers:
## Before this PR
Internally, someone had written the code:

```gradle
configurations {
    testAnnotationProcessor.extendsFrom annotationProcessor
    testCompileOnly.extendsFrom compileOnly
} 
```

in their repo. They then only put the immutables processor in `annotationProcessor`, but not `testAnnotationProcessor`:

```gradle
dependencies {
    annotationProcess 'org.immutables:value'
}
````

The plugin currently only checks dependencies declared directly in the configuration. This means the `test` source set had immutables, but was not configuring it to be incremental, meaning that the iteration loop for tests was a 2 min full compile.

Additionally, the current plugin does not handle the case where immutables would be brought in transitively.

## After this PR

==COMMIT_MSG==
Ensure that `baseline-immutables` configures immutables to work incrementally when the immutables `annotationProcessor` dependency is not a direct dependency (ie it is brought in transitively or by an `extendsFrom`).
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-2465.v2.yml
+++ b/changelog/@unreleased/pr-2465.v2.yml
@@ -1,0 +1,7 @@
+type: fix
+fix:
+  description: Ensure that `baseline-immutables` configures immutables to work incrementally
+    when the immutables `annotationProcessor` dependency is not a direct dependency
+    (ie it is brought in transitively or by an `extendsFrom`).
+  links:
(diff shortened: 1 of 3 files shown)


carterkozak (approved):
Thanks!

PR #2477, opened 2023-01-30
ensure tasks that use moduleJvmArgs have task inputs set correctly

### LINKED ISSUE #2476: Stale Jars when Updating 'moduleJvmArgs' Stanza
## What happened?

When modifying the `moduleJvmArgs` block in my service I noticed that my tests were continuing to fail. Digging in I noticed that the `MANIFEST.MF` did not contain my updated `exports`. To repro:

1. Add the following to your favorite repo `build.gradle`:
```gradle
moduleJvmArgs {
    exports = ['java.base/java.net']
}
```
2. Run `./gradlew :MODULE:jar`
3. Open the resultant jar and note the `Exports` entry in the `MANIFEST.MF`
4. Change the previously added stanza to:
```gradle
moduleJvmArgs {
    exports = ['java.base/java.net123']
}
```
5. Run `./gradlew :MODULE:jar`
6. Open the "resultant jar" (its the same) and note the `Exports` entry in the `MANIFEST.MF` are unchanged.
7. Run `./gradlew :MODULE:clean :MODULE:jar`
8. Open the new jar and note the `Exports` entry in the `MANIFEST.MF` are now updated.

Digging in to the Gradle build scans I see that Gradle is considering the `jar` task as `UP-TO-DATE`. Ping me if you want the links to the Gradle buildscans.

Changing the metadata of the module should invalidate the resulting artifacts and cause Gradle to rebuild the jars.

bjlaub:
The BaselineModuleJvmArgs plugin configures several tasks from the 'java' plugin to use values from the 'moduleJvmArgs' extension (e.g. to update compiler arguments, output values to the jar manifest, etc.). However, when the value on the extension changes, those tasks are not re-run.

To fix that, add input properties for each task that uses moduleJvmArgs so that when the extension value changes, the task will no longer be up-to-date.

Fixes #2476

## Before this PR
The BaselineModuleJvmArgs plugin configures several tasks from the 'java' plugin to use values from the 'moduleJvmArgs' extension (e.g. to update compiler arguments, output values to the jar manifest, etc.). However, when the value on the extension changes, those tasks are not re-run.

## After this PR

==COMMIT_MSG==
add input properties for each task that uses moduleJvmArgs so that when the extension value changes, the task will no longer be up-to-date.
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-2477.v2.yml
+++ b/changelog/@unreleased/pr-2477.v2.yml
@@ -1,0 +1,6 @@
+type: fix
+fix:
+  description: add input properties for each task that uses moduleJvmArgs so that
+    when the extension value changes, the task will no longer be up-to-date.
+  links:
+  - https://github.com/palantir/gradle-baseline/pull/2477
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
@@ -272,12 +272,9 @@ public final class BaselineModuleJvmArgs implements Plugin<Project> {
     }
 
     private static void setTaskInputsFromExtension(Task task, BaselineModuleJvmArgsExtension extension) {
-        task.getInputs()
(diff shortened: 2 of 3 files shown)


carterkozak (approved):
thanks!

PR #2482, opened 2023-02-01
Increase javac heap to 2g by default (up from 512m)

carterkozak:
==COMMIT_MSG==
Increase javac heap to 2g by default (up from 512m). Existing overrides are not impacted.
==COMMIT_MSG==

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineJavaCompilerHeap.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineJavaCompilerHeap.java
@@ -1,0 +1,41 @@
+/*
+ * (c) Copyright 2023 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineJavaCompilerHeap.java
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * Increases the default {@link JavaCompile} task heap from {@code 512m} to {@code 2g}.
+ */
+public final class BaselineJavaCompilerHeap implements Plugin<Project> {
+
+    private static final String JAVAC_HEAP = "2g";
+
+    @Override
+    public void apply(Project proj) {
+        proj.afterEvaluate(
```
fawind, line 33: To clarify, `afterEvaluate` is used because we want to check if something else already set `options.getMemoryMaximumSize`?

fawind: Ah, you even added a test for this! :)

fawind (approved):

PR #2500, opened 2023-02-22
New plugin: com.palantir.baseline-dead-code

iamdanfox:
## Before this PR

Recently I've been tasked with deleting codepaths that have been superseded by newer codepaths. Historically, it's been hard to do this because there is no static analysis that repos run regularly to find dead code. There _is_ some functionality in IntelliJ to identify methods which are unused, but this has a couple of key deficiencies:

- it considers methods that are only used in _tests_ as still used
- a pair of classes might use methods from each other, causing IntelliJ to think both of them are valuable

## After this PR
==COMMIT_MSG==
New plugin: com.palantir.baseline-dead-code
==COMMIT_MSG==


ash211:
I gave this a shot on the foundry/magritte repo internally and found a bunch of dead code! 🎆 

One future ER is that it'd be nice to be able to add another project in a repo as a live set of classes, in addition to the `sls-java-service-distribution` project.  In my case, I want to apply `com.palantir.baseline-dead-code` to my magritte-coordinator project, but also have baseline-dead-code/proguard consider the magritte-sdk project live as well.  This repo contains both a service, and an SDK for use by plugins, and anything part of the SDK or used transitively from the SDK must not be deleted.

PR #2505, opened 2023-03-06
Prefer Guava version of VisibleForTesting annotation

ash211:
## Before this PR
There are multiple versions of the `@VisibleForTesting` annotation, and if a developer isn't careful they can inadvertently import a version of that annotation besides the main one from Guava.

## After this PR
==COMMIT_MSG==
Prefer Guava version of VisibleForTesting annotation
==COMMIT_MSG==

This PR makes importing 

## Possible downsides?
Repos with `-Werror` that trip this warning will now fail to compile.  However, they have opted into this behavior by enabling `-Werror` and their CI checks should prevent gradle-baseline auto-upgrades from breaking their main branches.  However, they would then be stuck on the previous version of gradle-baseline, and not receive future updates to gradle-baseline automatically via excavator auto-bumps.


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferGuavaVisibleForTesting.java
```diff
@@ -0,0 +1,76 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.baseline.errorprone;
+
+// modeled off of
+// https://github.com/google/error-prone/blob/master/core/src/main/java/com/google/errorprone/bugpatterns/checkreturnvalue/UsingJsr305CheckReturnValue.java
+
```
pkoenig10:
```suggestion
```

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferGuavaVisibleForTesting.java
```diff
@@ -0,0 +1,76 @@
...
+            "org.assertj.core.util.VisibleForTesting",
+            "org.apache.flink.annotation.VisibleForTesting",
+            "com.cronutils.utils.VisibleForTesting",
+            "com.google.api.client.repackaged.com.google.common.annotations.VisibleForTesting",
+            "org.sparkproject.guava.annotations.VisibleForTesting",
+            "org.apache.hadoop.classification.VisibleForTesting",
+            "org.jheaps.annotations.VisibleForTesting",
+            "io.debezium.annotation.VisibleForTesting",
+            "graphql.VisibleForTesting",
+            "avro.shaded.com.google.common.annotations.VisibleForTesting");
```
pkoenig10: I'd prefer to avoid explicitly listing these and just forbid any annotation named "VisibleForTesting" that isn't the error-prone one.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferGuavaVisibleForTesting.java
```diff
@@ -0,0 +1,76 @@
...
+import com.sun.tools.javac.code.Type;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+/**
+ * Checker that recommends using Guava's version of {@code @VisibleForTesting} over other copies.
+ */
+@AutoService(BugChecker.class)
+@BugPattern(summary = "Prefer Guava's @VisibleForTesting over other versions.", severity = SeverityLevel.WARNING)
+public final class PreferGuavaVisibleForTesting extends BugChecker implements ImportTreeMatcher {
```
pkoenig10: These are Error-Prone annotations, not Guava

```suggestion
public final class PreferErrorProneAnnotations extends BugChecker implements ImportTreeMatcher {
```

WDYT about reframing/restructuring this as a general `PreferErrorProneAnnotations` check so we can add other annotations here in the future?

carterkozak: I'm pretty sure that `VisibleForTesting` exists in standard guava without error-prone annotation apis.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferGuavaVisibleForTesting.java
```diff
@@ -0,0 +1,76 @@
...
+import com.sun.source.tree.ImportTree;
+import com.sun.tools.javac.code.Type;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+/**
+ * Checker that recommends using Guava's version of {@code @VisibleForTesting} over other copies.
+ */
+@AutoService(BugChecker.class)
+@BugPattern(summary = "Prefer Guava's @VisibleForTesting over other versions.", severity = SeverityLevel.WARNING)
```
carterkozak: Initially this should roll out as a suggestion so that we can have the robots auto-fix before ratcheting to WARNING.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferGuavaVisibleForTesting.java
schlosna: Agree with Patrick, I think we may also want to consider blocking some of these repackaged imports in the baseline checkstyle as well by expanding `IllegalImports`:

https://github.com/palantir/gradle-baseline/blob/f797c67965ecce0d3714ad3a47884e87a898732d/.baseline/checkstyle/checkstyle.xml#L127-L132

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferGuavaVisibleForTesting.java
schlosna: [`@VisibleForTesting` has been in standard Guava since was renamed from Google-collections in 2010.
](https://github.com/google/guava/blame/f5ad01f0326405c321a4d1a7bbafb01b7d4d5074/src/com/google/common/annotations/VisibleForTesting.java)


We may want to consider something like `PreferCommonAnnotations` to steer folks to preferred annotations like cases like this, nullness, etc.

We'll likely be deconflicting between [checker-framework](https://checkerframework.org/), 
[jspecify](http://jspecify.org/docs/start-here/), guava, legacy jsr305 & findbugs/spotbugs, etc. and ideally consolidating on the smallest combination of these.

pkoenig10: Ah sorry, I didn't look closely enough here. I thought this was for an EP annotation.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferGuavaVisibleForTesting.java
ash211: Removed the deny list of imports and instead now look for all imports with the same class name.

Is it a requirement to change Checkstyle at the same time as ErrorProne, or can they be independent?  The EP change seems better to me, because it can suggest and apply fixes.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferGuavaVisibleForTesting.java
ash211: Changed to `PreferCommonAnnotations`.  If there are other annotations like VisibleForTesting we'd like to add now, please lmk.  I don't know of others, and am happy to get just this one in.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferGuavaVisibleForTesting.java
ash211: Changed to SUGGESTION

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferCommonAnnotations.java
```diff
@@ -0,0 +1,66 @@
...
+            Map.of("VisibleForTesting", "com.google.common.annotations.VisibleForTesting");
+
+    @Override
+    public Description matchImport(ImportTree tree, VisitorState state) {
+        Type importType = ASTHelpers.getType(tree.getQualifiedIdentifier());
+        for (String affectedClassName : PREFERRED_IMPORTS.keySet()) {
+            String preferredType = PREFERRED_IMPORTS.get(affectedClassName);
+            if (importType != null
+                    && importType.toString().endsWith(affectedClassName)
+                    && !Objects.equals(importType.toString(), preferredType)) {
```
schlosna: a few perf nits:

```suggestion
        if (importType == null) {
            return Description.NO_MATCH;
        }
        String importName = importType.toString();

        for (Map.Entry<String, String> entry = PREFERRED_IMPORTS.entrySet()) {
            affectedClassName = entry.getName();
            String preferredType = entry.getValue();
            if (importName.endsWith(affectedClassName)
                    && !Objects.equals(importName, preferredType)) {
```

schlosna:
Minor perf suggestion

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferCommonAnnotations.java
ash211: Applied (with fixups to compile).

I see the changes were:
1) early return for `importType == null`
2) map over Map entries, not keys followed by .get()
3) evaluate `importName = importType.toString()` once instead of multiple times

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/PreferCommonAnnotationsTest.java
```diff
@@ -0,0 +1,100 @@
...
+    @Test
+    public void wildcard_import_annotations_are_not_currently_rewritten() {
+        CompilationTestHelper.newInstance(PreferCommonAnnotations.class, getClass())
+                .addSourceLines(
+                        "Client.java",
+                        "package com.google.frobber;",
+                        "import org.jetbrains.annotations.*;",
+                        "public final class Client {",
+                        // NOTE: wildcard-imported annotations are not currently re-written
+                        "  @VisibleForTesting",
```
schlosna, line 89: this seems ok for now given we block wildcard imports

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/PreferCommonAnnotationsTest.java
```diff
@@ -0,0 +1,100 @@
...
+
+    @Test
+    public void fully_qualified_annotations_are_not_currently_rewritten() {
+        CompilationTestHelper.newInstance(PreferCommonAnnotations.class, getClass())
+                .addSourceLines(
+                        "Client.java",
+                        "package com.google.frobber;",
+                        "public final class Client {",
+                        // NOTE: fully-qualified annotations are not currently re-written
+                        "  @org.jetbrains.annotations.VisibleForTesting",
```
schlosna, line 72: seems reasonable to allow fully qualified as method to explicitly use different type

schlosna (approved):
seems reasonable to me

PR #2506, opened 2023-03-07
add JooqBatchWithoutBindArgs check

berler:
## Before this PR
Jooq provides multiple methods to execute queries in batches. Some of these methods end up executing queries without bind args which can cause performance problems.

## After this PR
Disallow jOOQ's batch methods that execute queries without bind args. Batch methods that do use bind args should be used instead.

==COMMIT_MSG==
Add error-prone check JooqBatchWithoutBindArgs
==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/JooqBatchWithoutBindArgs.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/JooqBatchWithoutBindArgs.java
@@ -39,17 +39,22 @@ import java.util.Collection;
 @BugPattern(
         link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
         linkType = BugPattern.LinkType.CUSTOM,
-        severity = SeverityLevel.ERROR,
+        severity = SeverityLevel.WARNING,
         tags = StandardTags.PERFORMANCE,
-        summary = "Disallow jOOQ batch methods that execute queries without bind args.")
+        summary = "jOOQ batch methods that execute without bind args can cause performance problems.",
+        explanation =
+                "When batch queries execute without bind args, each query is sent to the database as a string with all"
+                    + " variables inline. Inline variables cause each query in the batch to be unique, so the database"
+                    + " uses extra CPU and memory to parse and query plan each query. Instead use one of the other"
+                    + " jOOQ batch methods that is documented as executing queries with bind args, as this allows"
+                    + " parsing and query planning the query once and then executing any number of times with"
+                    + " different bind values.")
 public final class JooqBatchWithoutBindArgs extends BugChecker implements MethodInvocationTreeMatcher {
 
     private static final long serialVersionUID = 1L;
 
     private static final String DSL_CONTEXT = "org.jooq.DSLContext";
     private static final String BATCH = "batch";
-    private static final String ERROR_MESSAGE =
-            "Jooq batch methods that execute queries without bind args cause performance problems.";
 
     private static final Supplier<Type> QUERY_TYPE =
             VisitorState.memoize(state -> state.getTypeFromString("org.jooq.Query"));
@@ -75,7 +80,7 @@ public final class JooqBatchWithoutBindArgs extends BugChecker implements Method
     @Override
     public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
         if (BATCH_WITHOUT_BINDS_MATCHER.matches(tree, state)) {
-            return buildDescription(tree).setMessage(ERROR_MESSAGE).build();
+            return describeMatch(tree);
         }
         return Description.NO_MATCH;
     }
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/JooqBatchWithoutBindArgs.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/JooqBatchWithoutBindArgs.java
@@ -1,0 +1,82 @@
+/*
+ * (c) Copyright 2023 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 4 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/JooqBatchWithoutBindArgs.java
```diff
@@ -0,0 +1,82 @@
...
+        tags = StandardTags.PERFORMANCE,
+        summary = "Disallow jOOQ batch methods that execute queries without bind args.")
+public final class JooqBatchWithoutBindArgs extends BugChecker implements MethodInvocationTreeMatcher {
+
+    private static final long serialVersionUID = 1L;
+
+    private static final String DSL_CONTEXT = "org.jooq.DSLContext";
+    private static final String BATCH = "batch";
+    private static final String ERROR_MESSAGE =
+            "Jooq batch methods that execute queries without bind args cause performance problems.";
```
carterkozak: I wonder if @lukaseder has guidance on this?

@berler It would be helpful if you could provide more information. How do we ensure this remains true over time? Do you have public benchmarking data to substantiate the claim?

berler: We don't have public benchmarking data, but we do know that the way this executes is using extra CPU on the database because each query is sent over as a unique string (with all variable values inline as a string in the query). There is CPU overhead parsing and query planning each query in the batch.

Looking at how we we have used jooq internally, it was actually surprising to most of our dev teams how these calls were being executed. People usually use the batch methods to execute the same query hundreds to thousands of times but just with different bind variables. What we want to happen (and what most of our devs thought was actually happening) is it sends the query to the DB once and then just sends the arrays of bind args to execute the same query with each set of bind variables. Internally we've already refactored nearly all of our code to use the other batch methods that execute with bind args.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/JooqBatchWithoutBindArgs.java
```diff
@@ -0,0 +1,82 @@
...
+import com.sun.source.tree.ExpressionTree;
+import com.sun.source.tree.MethodInvocationTree;
+import com.sun.tools.javac.code.Type;
+import java.util.Collection;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
+        severity = SeverityLevel.ERROR,
```
carterkozak: Let's use the warning severity here rather than error

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/JooqBatchWithoutBindArgs.java
carterkozak: Please remove `ERROR_MESSAGE` in favor of using this text in the `summary` field of the class-level annotation. Then you can replace the `return buildDescription(tree).setMessage(ERROR_MESSAGE).build();` below with `return describeMatch(tree);`

carterkozak: The failure message shouldn't tell folks not to do something, but describe why they shouldn't and provide alternatives, otherwise developers will suppress and move on! When an error-prone check fails, it presents us with an opportunity to teach developers without having to go through the same P0s we have :-)

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/JooqBatchWithoutBindArgs.java
CVdV-au: I think error is appropriate; this problem was really hard to track down as a problem in the first instance because it makes these queries invisible in any aggregate stats from the DB. It is easy to fix, and developers _think_ they are batching when calling the 4 methods we want to stop consumption of but they are really just issuing N instances of similar SQL - but force the DB to parse + plan since variables are substituted not parameterised) but in a single network interaction. There are other constructs in Jooq that will still allow that - BatchedConnection.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/JooqBatchWithoutBindArgs.java
```diff
@@ -0,0 +1,87 @@
...
+import com.sun.source.tree.ExpressionTree;
+import com.sun.source.tree.MethodInvocationTree;
+import com.sun.tools.javac.code.Type;
+import java.util.Collection;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
+        severity = SeverityLevel.WARNING,
```
CVdV-au, line 42: I really think error is appropriate here.

CVdV-au: We've done the hard yards to fix _every_ project so this shouldn't cause grief to the code base.

CVdV-au (changes requested):
ERROR please!

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/JooqBatchWithoutBindArgs.java
berler: @carterkozak Is it okay to use error here?

CVdV-au (approved):

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/JooqBatchWithoutBindArgs.java
berler: Discussed offline, I think we're fine leaving it as warn.

carterkozak (approved):

PR #2521, opened 2023-03-15
Remove support for refaster

carterkozak:
Refaster rules, while simpler to write, are substantially more difficult to maintain compared to standard error-prone rules. We no longer write them ourselves, and have recommended against them for the last few years, as it's difficult to tell where they will match, and more difficult to tweak them to avoid edge cases.

We expect that removal of refaster will improve build times, especially for excavators. Failures which are caught and ignored compiling refaster rules are a common red herring which point devs in the wrong direction while debugging other issues.

==COMMIT_MSG==
Remove support for refaster
==COMMIT_MSG==


CRogers (approved):

schlosna (approved):
Overall I'm supportive.

Do we have any plans to convert some/all of these refaster templates into error-prone checker(s)/fixers?

carterkozak:
> Do we have any plans to convert some/all of these refaster templates into error-prone checker(s)/fixers?

It's not something I'm currently tracking, but it would be a great project to ramp somebody up on writing error-prone rules.

cc @mpritham ;-)

ash211:
Clearly this has already happened, but I'm not sure that it's realistic to rewrite refaster rules into error prone rules at scale.  They end up being an order of magnitude larger in errorprone.  Take for example https://github.com/palantir/gradle-baseline/pull/2530 which is ~450 lines in error prone but is [36 lines](https://github.com/palantir/gradle-baseline/pull/2521/files#diff-d763a6fd4a964940c7eb573172be59d553bf9a2890319d3b08ff1ee8381eda74L1-L36) in refaster.

We've got a bunch of rules in https://github.com/palantir/assertj-automation/tree/develop/assertj-refaster-rules/src/main/java/com/palantir/assertj/refaster that I value, but now I'm not sure they're worth the effort to port to errorprone.

carterkozak:
> Take for example https://github.com/palantir/gradle-baseline/pull/2530 which is ~450 lines in error prone but is [36 lines](https://github.com/palantir/gradle-baseline/pull/2521/files#diff-d763a6fd4a964940c7eb573172be59d553bf9a2890319d3b08ff1ee8381eda74L1-L36) in refaster.

I'd consider that a feature, not a bug. The biggest problem with refaster is precision: It's tremendously difficult to gain confidence that it will impact precisely what you want without overreaching or under-reaching. Furthermore refaster rules can't be applied in a way that prevents regressions.

PR #2530, opened 2023-03-24
Add check: Use Collection.isEmpty() instead of comparing size() with 0

mpritham:
## Before this PR
We had a refaster template which suggested that `{implementation of Collection}.size() == 0` be changed to `{implementation of Collection}.isEmpty()`. This was removed [in this PR](https://github.com/palantir/gradle-baseline/pull/2521).

## After this PR

==COMMIT_MSG==
Add check: Use Collection.isEmpty() instead of comparing size() with 0
==COMMIT_MSG==


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/CollectionsIsEmpty.java
```diff
@@ -0,0 +1,120 @@
...
+        Optional<EqualsZeroExpression> maybeEqualsZeroExpression = getEqualsZeroExpression(tree);
+        if (maybeEqualsZeroExpression.isEmpty()) {
+            return Description.NO_MATCH;
+        }
+        EqualsZeroExpression equalsZeroExpression = maybeEqualsZeroExpression.get();
+        ExpressionTree operand = equalsZeroExpression.operand();
+
+        if (!(operand instanceof MethodInvocationTree) || !COLLECTION_SIZE_METHOD_MATCHER.matches(operand, state)) {
+            return Description.NO_MATCH;
+        }
```
carterkozak: style nit: I think we can invert this to:

```suggestion
        if (COLLECTION_SIZE_METHOD_MATCHER.matches(operand, state)) {
            return describeMatch(/*etc...
        }
        return Description.NO_MATCH;
```

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/CollectionsIsEmpty.java
```diff
@@ -0,0 +1,120 @@
...
+@AutoService(BugChecker.class)
+@BugPattern(
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
+        severity = SeverityLevel.WARNING,
+        summary = "Use the isEmpty method instead of checking collection size")
+public final class CollectionsIsEmpty extends BugChecker implements BugChecker.BinaryTreeMatcher {
+    private static final Matcher<ExpressionTree> COLLECTION_SIZE_METHOD_MATCHER = MethodMatchers.instanceMethod()
+            .onDescendantOf(Collection.class.getName())
+            .named("size");
```
carterkozak: Let's update to include a parameter specifier to avoid matching arbitrary Collection subtypes with `int size(SomeArg value)`
```suggestion
            .named("size")
            .withNoParameters();
```

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/CollectionsIsEmpty.java
```diff
@@ -0,0 +1,120 @@
...
+                ret = ExpressionType.NEQ;
+                break;
+            default:
+                return Optional.empty();
+        }
+        ExpressionTree leftOperand = tree.getLeftOperand();
+        ExpressionTree rightOperand = tree.getRightOperand();
+
+        if (leftOperand.getKind() == Kind.INT_LITERAL
+                && ((LiteralTree) leftOperand).getValue().equals(0)) {
```
carterkozak: We could create a `Matchers.intLiteral(0)` to use for these checks (requires the VisitorState to be passed into this method, but may simplify the implementation a bit)

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/CollectionsIsEmptyTest.java
```diff
@@ -0,0 +1,125 @@
...
+        helper().addSourceLines(
+                        "Test.java",
+                        "import " + List.class.getCanonicalName() + ";",
+                        "class Test {",
+                        "  static boolean f(List<String> foo) {",
+                        "    // BUG: Diagnostic contains: foo.isEmpty()",
+                        "    return foo.size() == 0;",
+                        "  }",
+                        "}")
+                .doTest();
```
carterkozak: Since we have an automated fix implemented, I'd use the refactoring helper and verify teh result instead of using the detection helper

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/CollectionsIsEmptyTest.java
```diff
@@ -0,0 +1,125 @@
...
+    public void test_no_match() {
+        helper().addSourceLines(
+                        "Test.java",
+                        "import " + List.class.getCanonicalName() + ";",
+                        "class Test {",
+                        "  static boolean f(List<String> foo) {",
+                        "    return foo.size() == 1 && foo.size() != 1 && 1 == foo.size() && 1 != foo.size();",
+                        "  }",
+                        "}")
+                .doTest();
```
carterkozak: with the refactoring test, we'd update this to assert no changes

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/CollectionsIsEmptyTest.java
```diff
@@ -0,0 +1,125 @@
...
+ */
+
+package com.palantir.baseline.errorprone;
+
+import com.google.errorprone.CompilationTestHelper;
+import java.util.List;
+import java.util.Set;
+import org.junit.jupiter.api.Test;
+
+public class CollectionsIsEmptyTest {
```
carterkozak: One edge case we may wish to consider:
Sometimes collections implement
```java
@Override
public boolean isEmpty() {
    return this.size() == 0;
}
```
or without the `this` keyword:
```java
@Override
public boolean isEmpty() {
    return size() == 0;
}
```

Could we special case invocations on the current object and avoid producing changes? Otherwise the above examples would recurse infinitely after applying our suggestion!

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/CollectionsIsEmpty.java
```diff
@@ -0,0 +1,120 @@
...
+import com.sun.source.tree.Tree.Kind;
+import java.util.Collection;
+import java.util.Optional;
+import org.immutables.value.Value.Immutable;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
+        severity = SeverityLevel.WARNING,
```
carterkozak: Let's call this a `SUGGESTION`, and eventually have it auto-fix itself

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/CollectionsIsEmpty.java
```diff
@@ -0,0 +1,120 @@
...
+        summary = "Use the isEmpty method instead of checking collection size")
+public final class CollectionsIsEmpty extends BugChecker implements BugChecker.BinaryTreeMatcher {
```
carterkozak: I wonder if we could name this something slightly different, such that we could include support for non-collection items like `map.size() == 0` -> `map.isEmpty()` and `charSequence.length() == 0` -> `charSequence.isEmpty()`

carterkozak: (not suggesting adding those in this PR, but it could be a helpful follow-up!)

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/CollectionsIsEmptyTest.java
mpritham: Good catch! Will add a test for that.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/CollectionsIsEmpty.java
mpritham: Called it `CardinalityEqualsZero`. Keeping the "summary" specific to collections for now. Let me know what you think.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/CardinalityEqualsZero.java
```diff
@@ -0,0 +1,140 @@
...
+
+    private static boolean isInIsEmptyDefinition(VisitorState state) {
+        MethodTree methodTree = ASTHelpers.findEnclosingNode(state.getPath(), MethodTree.class);
+        return methodTree != null && methodTree.getName().contentEquals("isEmpty");
+    }
+
+    private static boolean isExpressionThis(ExpressionTree expressionTree) {
+        return ((expressionTree instanceof IdentifierTree)
+                && ((JCTree.JCIdent) expressionTree).getName().contentEquals("this"));
+    }
```
carterkozak: We may want to draw inspiration from [`ASTHelpers#isSuper`](https://github.com/google/error-prone/blob/c262ba7d329b7fdf6533c0e989cf386af2895d21/check_api/src/main/java/com/google/errorprone/util/ASTHelpers.java#L1623-L1632).

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/CardinalityEqualsZero.java
```diff
@@ -0,0 +1,140 @@
...
+                                    + "isEmpty()"));
+        }
+
+        return Description.NO_MATCH;
+    }
+
+    private static boolean isInIsEmptyDefinition(VisitorState state) {
+        MethodTree methodTree = ASTHelpers.findEnclosingNode(state.getPath(), MethodTree.class);
+        return methodTree != null && methodTree.getName().contentEquals("isEmpty");
+    }
```
carterkozak: I'd probably not include this in the condition due to potential edge cases (method delegation to private functions, etc).

carterkozak: I'd rather allow potential failures through than produce false positives (which erode trust in the check).

mpritham: I've removed this check. Now, the check does not match when the expression looks like `this.size() == 0` or `size() == 0`, regardless of the method.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/CardinalityEqualsZero.java
```diff
@@ -0,0 +1,135 @@
...
+                                    + state.getSourceForNode(collectionInstance)
+                                    + ".isEmpty()"));
+        }
+
+        return Description.NO_MATCH;
+    }
+
+    private static boolean isExpressionThis(ExpressionTree tree) {
+        switch (tree.getKind()) {
+            case IDENTIFIER:
```
carterkozak, line 82: I think the memberselect case is technically necessary here for "qualified this" usages e.g.
```java
class Outer {
  int foo();
  
  Inner getInner() {
    return new Inner();
  }
  class Inner {
    int foo() {
        // 'return foo()' would recurse infinitely
        // this allows us to reference components on an enclosing class from a non-static inner class.
        return Outer.this.foo();
    }
  }
}
```

mpritham: Ah. Have not seen this before. Updated, and added a test.

carterkozak (approved):
Great tests!

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/CardinalityEqualsZero.java
```diff
@@ -0,0 +1,138 @@
...
+    @Override
+    public Description matchBinary(BinaryTree tree, VisitorState state) {
+        Optional<EqualsZeroExpression> maybeEqualsZeroExpression = getEqualsZeroExpression(tree, state);
+        if (maybeEqualsZeroExpression.isEmpty()) {
+            return Description.NO_MATCH;
+        }
+
+        EqualsZeroExpression equalsZeroExpression = maybeEqualsZeroExpression.get();
+        ExpressionTree operand = equalsZeroExpression.operand();
+        ExpressionTree collectionInstance = ASTHelpers.getReceiver(operand);
```
schlosna, line 61: Filed https://github.com/palantir/gradle-baseline/issues/2543

PR https://github.com/palantir/tritium/pull/1688 is failing attempting to upgrade to gradle-baseline with this change on https://github.com/palantir/tritium/blob/13799047a8b715b4ebf395de80deed78a86b734c/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/TagMap.java#L120-L123

```
> Task :tritium-registry:compileJava FAILED
/home/circleci/project/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/TagMap.java:123: error: An unhandled exception was thrown by the Error Prone static analysis plugin.
            if (comparisonResult == 0) {
                                 ^
  
     error-prone version: 2.18.0
     BugPattern: CardinalityEqualsZero
     Stack Trace:
     java.lang.IllegalStateException: Expected expression 'comparisonResult' to be a method invocation or field access, but was IDENTIFIER
  	at com.google.errorprone.util.ASTHelpers.getReceiver(ASTHelpers.java:605)
  	at com.palantir.baseline.errorprone.CardinalityEqualsZero.matchBinary(CardinalityEqualsZero.java:61)
  	at com.google.errorprone.scanner.ErrorProneScanner.processMatchers(ErrorProneScanner.java:449)
  	at com.google.errorprone.scanner.ErrorProneScanner.visitBinary(ErrorProneScanner.java:512)
  	at com.google.errorprone.scanner.ErrorProneScanner.visitBinary(ErrorProneScanner.java:150)
```

PR #2535, opened 2023-03-29
SafeLoggingPropagation uses new ASTHelpers.isRecord function

carterkozak:
No need to declare this ourselves. Introduced in 2.17 by https://github.com/google/error-prone/commit/640bc78870f1100ee38b96566de988d7b433de26

==COMMIT_MSG==
SafeLoggingPropagation uses new ASTHelpers.isRecord function
==COMMIT_MSG==


mpritham (approved):
👍

PR #2537, opened 2023-03-30
StrictUnusedVariable is not suppressed when UnusedVariable is disabled

carterkozak:
This is a workaround for error-prone#3831

Unfortunately, we disable UnusedVariable by default because it duplicates StrictUnusedVariable. I've removed the `UnusedVariable` alternative name from StrictUnusedVariable to avoid this collateral damage, while updating the `isSuppressed` check with an override which handles existing suppressions gracefully.

==COMMIT_MSG==
StrictUnusedVariable is not suppressed when UnusedVariable is disabled
==COMMIT_MSG==

## Possible downsides?
Rollout will catch the places where StrictUnusedVariable has regressed, slowing adoption.

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StrictUnusedVariable.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StrictUnusedVariable.java
@@ -67,7 +67,10 @@ import com.google.errorprone.bugpatterns.BugChecker;
 import com.google.errorprone.bugpatterns.UnusedVariable;
 import com.google.errorprone.fixes.SuggestedFix;
 import com.google.errorprone.fixes.SuggestedFixes;
+import com.google.errorprone.matchers.ChildMultiMatcher.MatchType;
 import com.google.errorprone.matchers.Description;
+import com.google.errorprone.matchers.Matcher;
+import com.google.errorprone.matchers.Matchers;
 import com.google.errorprone.suppliers.Suppliers;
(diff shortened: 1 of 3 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StrictUnusedVariable.java
```diff
@@ -124,7 +128,7 @@
  */
 @AutoService(BugChecker.class)
 @BugPattern(
-        altNames = {"unused", "UnusedVariable"},
+        altNames = "unused",
```
mpritham, line 131: Should we also remove "unused"? [It is an alt name for the UnusedVariable check](https://github.com/google/error-prone/blob/master/core/src/main/java/com/google/errorprone/bugpatterns/UnusedVariable.java#L120), and users may use `-Xep:unused:OFF` thinking they are disabling the `UnusedVariable` check, but they also disabling this?

carterkozak: That should be unlikely as it's not the name of an upstream error-prone check, or the intended use of the enablement options. I'd prefer to minimize out workaround as much as possible since it may not be 100% equivalent to the previous alternate name behavior.

mpritham: Ok, sounds good.

mpritham (approved):

PR #2541, opened 2023-03-31
Add check: Use InvocationTargetException.getCause instead of getTargetException

mpritham:
## Before this PR
One of the refaster templates removed in [this PR](https://github.com/palantir/gradle-baseline/pull/2521), replaced calls to `InvocationTargetException.getTargetException` with `InvocationTargetException.getCause`.

## After this PR
==COMMIT_MSG==
Add check: Use InvocationTargetException.getCause instead of getTargetException. 
==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/CardinalityEqualsZero.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/CardinalityEqualsZero.java
@@ -31,7 +31,9 @@ import com.sun.source.tree.BinaryTree;
 import com.sun.source.tree.ExpressionTree;
 import com.sun.source.tree.IdentifierTree;
 import com.sun.source.tree.MemberSelectTree;
+import com.sun.source.tree.Tree.Kind;
 import java.util.Collection;
+import java.util.Objects;
 import java.util.Optional;
 import org.immutables.value.Value.Immutable;
 
@@ -58,8 +60,11 @@ public final class CardinalityEqualsZero extends BugChecker implements BugChecke
 
         EqualsZeroExpression equalsZeroExpression = maybeEqualsZeroExpression.get();
         ExpressionTree operand = equalsZeroExpression.operand();
-        ExpressionTree collectionInstance = ASTHelpers.getReceiver(operand);
+        if (!Objects.equals(operand.getKind(), Kind.METHOD_INVOCATION)) {
+            return Description.NO_MATCH;
+        }
 
+        ExpressionTree collectionInstance = ASTHelpers.getReceiver(operand);
         if (collectionInstance == null || isExpressionThis(collectionInstance)) {
             return Description.NO_MATCH;
         }
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/CardinalityEqualsZero.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/CardinalityEqualsZero.java
@@ -29,6 +29,8 @@ import com.google.errorprone.matchers.method.MethodMatchers;
 import com.google.errorprone.util.ASTHelpers;
 import com.sun.source.tree.BinaryTree;
 import com.sun.source.tree.ExpressionTree;
+import com.sun.source.tree.IdentifierTree;
+import com.sun.source.tree.MemberSelectTree;
 import java.util.Collection;
 import java.util.Optional;
 import org.immutables.value.Value.Immutable;
@@ -58,7 +60,7 @@ public final class CardinalityEqualsZero extends BugChecker implements BugChecke
         ExpressionTree operand = equalsZeroExpression.operand();
         ExpressionTree collectionInstance = ASTHelpers.getReceiver(operand);
 
-        if (collectionInstance == null || MoreASTHelpers.isExpressionThis(collectionInstance)) {
+        if (collectionInstance == null || isExpressionThis(collectionInstance)) {
             return Description.NO_MATCH;
         }
 
@@ -73,6 +75,17 @@ public final class CardinalityEqualsZero extends BugChecker implements BugChecke
         }
 
         return Description.NO_MATCH;
+    }
+
+    public static boolean isExpressionThis(ExpressionTree tree) {
+        switch (tree.getKind()) {
+            case IDENTIFIER:
+                return ((IdentifierTree) tree).getName().contentEquals("this");
+            case MEMBER_SELECT:
+                return ((MemberSelectTree) tree).getIdentifier().contentEquals("this");
+            default:
+                return false;
+        }
     }
 
     private static Optional<EqualsZeroExpression> getEqualsZeroExpression(BinaryTree tree, VisitorState state) {
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/CardinalityEqualsZero.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/CardinalityEqualsZero.java
@@ -77,7 +77,7 @@ public final class CardinalityEqualsZero extends BugChecker implements BugChecke
         return Description.NO_MATCH;
     }
 
-    public static boolean isExpressionThis(ExpressionTree tree) {
+    private static boolean isExpressionThis(ExpressionTree tree) {
         switch (tree.getKind()) {
             case IDENTIFIER:
                 return ((IdentifierTree) tree).getName().contentEquals("this");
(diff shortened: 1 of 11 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/InvocationTargetExceptionGetCause.java
```diff
@@ -0,0 +1,63 @@
...
+            .named("getTargetException")
+            .withNoParameters();
+
+    @Override
+    public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
+        if (ITE_GET_TARGET_EXCEPTION_MATCHER.matches(tree, state)) {
+            ExpressionTree caller = ASTHelpers.getReceiver(tree.getMethodSelect());
+            if (caller == null || MoreASTHelpers.isExpressionThis(caller)) {
+                return Description.NO_MATCH;
+            }
```
carterkozak: I think we can drop this check for a couple reasons:
1. It's very rare to _extend_ InvocationTargetException
2. InvocationTargetException is a concrete implementation, not just a contract

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/InvocationTargetExceptionGetCause.java
```diff
@@ -0,0 +1,63 @@
...
+
+            return describeMatch(
+                    tree,
+                    SuggestedFix.replace(
+                            tree,
+                            state.getSourceForNode(ASTHelpers.getReceiver(tree.getMethodSelect())) + ".getCause()"));
```
carterkozak: I think you may be able to simplify this using the builtin utility function`SuggestedFixes.renameMethodInvocation`

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/InvocationTargetExceptionGetCause.java
mpritham: Given (1), I think it'd make sense to change the `ITE_GET_TARGET_EXCEPTION_MATCHER` to match on instance methods `onExactClass` of InvocationTargetException, rather than subclasses. What are your thoughts?

carterkozak: It's a concrete implementation which _also_ provides a contract (otherwise subclasses would violate liskov substitution principle). I think it's reasonable to check all subtypes.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/CardinalityEqualsZero.java
```diff
@@ -77,7 +77,7 @@ public Description matchBinary(BinaryTree tree, VisitorState state) {
         return Description.NO_MATCH;
     }
 
-    private static boolean isExpressionThis(ExpressionTree tree) {
+    public static boolean isExpressionThis(ExpressionTree tree) {
```
carterkozak:
```suggestion
    private static boolean isExpressionThis(ExpressionTree tree) {
```

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/InvocationTargetExceptionGetCause.java
```diff
@@ -0,0 +1,58 @@
...
+            .onExactClass(InvocationTargetException.class.getName())
+            .named("getTargetException")
+            .withNoParameters();
+
+    @Override
+    public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
+        ExpressionTree caller = ASTHelpers.getReceiver(tree.getMethodSelect());
+        if (caller == null) {
+            return Description.NO_MATCH;
+        }
```
carterkozak: I don't think this block is necessary

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/CardinalityEqualsZero.java
mpritham: oops.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/InvocationTargetExceptionGetCause.java
```diff
@@ -0,0 +1,58 @@
...
+import com.sun.source.tree.MethodInvocationTree;
+import java.lang.reflect.InvocationTargetException;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
+        severity = SeverityLevel.SUGGESTION,
+        summary = "Use getCause instead of getTargetException")
+public final class InvocationTargetExceptionGetCause extends BugChecker
```
carterkozak: Naming nit: We try to name checks after the bad thing, not the suggested fix. Perhaps `InvocationTargetExceptionGetTargetException`?

changelog/@unreleased/pr-2541.v2.yml
```diff
@@ -0,0 +1,6 @@
+type: feature
+feature:
+  description: 'Add check: Use InvoationTargetException.getCause instead of getTargetException.
+    Add `isExpressionThis` to MoreASTHelpers.'
```
carterkozak: `MoreASTHelpers` is an internal util, not meant as public API -- no need to describe it in changelogs (also, this piece has been reverted ;-)

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/InvocationTargetExceptionGetCause.java
mpritham: hm ok. This was for:

```java
class Test extends InvocationTargetException {
  @Override
  public Throwable getCause() {
    // maybe some logic
    Throwable t = getTargetException();
    // maybe some logic 
    return t;
  }
}
```

We're saying this are very unlikely to exist, so the check for null callers isn't needed. will remove.

changelog/@unreleased/pr-2541.v2.yml
mpritham: Got it. Also fixed the typo in invocation.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/InvocationTargetExceptionGetTargetException.java
```diff
@@ -0,0 +1,52 @@
...
+import com.sun.source.tree.ExpressionTree;
+import com.sun.source.tree.MethodInvocationTree;
+import java.lang.reflect.InvocationTargetException;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
+        severity = SeverityLevel.SUGGESTION,
+        summary = "Use InvocationTargetException.getCause instead of InvocationTargetException.getTargetException")
```
carterkozak: It's helpful to describe the reason rather than state rules. We can link+summarize the javadoc (which is actually a super interesting -- if you're into that sort of thing -- detail):

https://docs.oracle.com/en/java/javase/17/docs/api//java.base/java/lang/reflect/InvocationTargetException.html#getTargetException()
> This method predates the general-purpose exception chaining facility. The [Throwable.getCause()](https://docs.oracle.com/en/java/javase/17/docs/api//java.base/java/lang/Throwable.html#getCause()) method is now the preferred means of obtaining this information.

mpritham: Makes sense, will update.

---
I was deciding between adding more detail (like the API note from the javadoc) or being super concise. Decided on being concise because I thought a dev would just want the fastest resolution, and wouldn't think too much about the difference.

But you're right. Since we have a suggested fix anyway, they will know how to change their code, and we can use the summary to provide context. For times we don't have a suggested fix, maybe we'd favor conciseness over context.

mpritham: Oh. I have also missed updating the README for this check, and the `collections.size() == 0` check... will update.

carterkozak: I haven't been updating hte readme lately -- imo we should codegen the list of checks and use the summary from annotations, but it's not high enough on my list to implement at the moment!

mpritham: edit: posted this before I read your latest comment.

~Also, what are your thoughts on moving the giant list of checks out of the main README, and into a separate md file? Each check can be a heading, so we could link directly to the check in the `link` argument of the BugPattern, and add as much detail as we'd like.~

carterkozak: I like it

mpritham: cool. Made a task in the backlog.

carterkozak (approved):
thanks!

Issue #2545, opened 2023-04-03
gradle-baseline recommends "Save Actions" plugin which is incompatible with IntelliJ 2023.1


ash211:
## What happened?

- gradle-baseline recommends "Save Actions" plugin [here](https://github.com/palantir/gradle-baseline/blob/144edb46c54bcef01825942de8f81edacff7a1b3/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy#L599-L610)
- that plugin is incompatible with the latest version of IntelliJ: https://github.com/dubreuia/intellij-plugin-save-actions/issues/427
- the way this fails is that you can't save files in window panes

As a result, I lost a whole screen full of written code as a result :(

## What did you want to happen?

We should stop suggesting the "Save Actions" plugin, and instead move to IntelliJ's native "Actions on Save" functionality, assuming that meets our needs.  I think just code formatting and optimizing imports.

ash211:
An alternative is to move to this fork, since the original plugin is now un-maintained: https://github.com/dubreuia/intellij-plugin-save-actions/issues/427#issuecomment-1488812298

CRogers:
This is being fixed. Please do not use the fork by some random person.

PR #2547, opened 2023-04-05
Stop suggesting the save actions plugin

CRogers:
## Before this PR
The save actions plugin is broken in IntelliJ 2023.1. It is unmaintained and there is no official fix.

In 2021.2, IntelliJ introduced a [new built in reformat on save feature](https://blog.jetbrains.com/idea/2021/07/intellij-idea-2021-2/#key_updates:~:text=We%E2%80%99ve%20added%20several%20actions%20that%20the%20IDE%20will%20initiate%20when%20you%20save%20the%20project%2C%20including%20reformatting%20code).

## After this PR
==COMMIT_MSG==
No longer suggest the "Save Actions" IntelliJ plugin which does not work in IntelliJ 2023.1 for use with Palantir Java Format. Instead, Palantir Java Format will support the IntelliJ native "Actions on save" reformat capability in a future release.
==COMMIT_MSG==

## Possible downsides?
Why this code is in gradle-baseline rather than the palantir-java-format gradle plugin I _do not_ understand. It's going to make fixing this issue quite a bit harder.

### DIFF
--- a/changelog/@unreleased/pr-2547.v2.yml
+++ b/changelog/@unreleased/pr-2547.v2.yml
@@ -1,0 +1,8 @@
+type: fix
+fix:
+  description: No longer suggest the "Save Actions" IntelliJ plugin which does not
+    work in IntelliJ 2023.1 for use with Palantir Java Format. Instead, Palantir Java
+    Format will support the IntelliJ native "Actions on save" reformat capability
+    in a future release.
+  links:
+  - https://github.com/palantir/gradle-baseline/pull/2547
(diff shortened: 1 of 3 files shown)


ash211:
Makes sense to me. First removing this suggestion, and following up with a replacement that configures native IntelliJ "Actions on save" works nicely.

CRogers:
Related: https://github.com/palantir/palantir-java-format/pull/862

felixdesouza (approved):

ash211:
This addresses https://github.com/palantir/gradle-baseline/issues/2545

PR #2555, opened 2023-04-14
Add check replacing `stream.sorted().findFirst()` with `stream.min()`

mpritham:
## Before this PR
As part of [this PR](https://github.com/palantir/gradle-baseline/pull/2521), the check to recommend using `stream.min()` over `stream.sorted().findFirst()` was removed. This PR adds an error-prone rule with the same effect.

## After this PR

==COMMIT_MSG==
Add check replacing `stream.sorted().findFirst()` with `stream.min()`
==COMMIT_MSG==


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/SortedStreamFirstElement.java
```diff
@@ -0,0 +1,100 @@
...
+            return Description.NO_MATCH;
+        }
+        MethodInvocationTree sortedTree = (MethodInvocationTree) sorted;
+        ExpressionTree stream = ASTHelpers.getReceiver(sorted);
+        if (stream == null) {
+            // Not expected.
+            return Description.NO_MATCH;
+        }
+
+        if (Matchers.receiverOfInvocation(STREAM_SORTED_NO_PARAMS_MATCHER).matches(tree, state)) {
```
carterkozak: Let's extract the `Matchers.receiverOfInvocation(` `Matcher` instances to static fields so they can be reused.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/SortedStreamFirstElement.java
```diff
@@ -0,0 +1,100 @@
...
+            return describeMatch(
+                    tree,
+                    SuggestedFix.builder()
+                            .replace(tree, state.getSourceForNode(stream) + ".min(Comparator.naturalOrder())")
```
carterkozak: it may be slightly cleaner to use an index-based `replace` call from the beginning of `sorted` to the end of `findFirst`. That way we're more likely to preserve comments and such. Doesn't make a great deal of differnce either way though :-]

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/SortedStreamFirstElement.java
```diff
@@ -0,0 +1,100 @@
...
+                .matches(tree, state)) {
+            return describeMatch(
+                    tree,
+                    SuggestedFix.builder()
+                            .replace(
+                                    tree,
+                                    state.getSourceForNode(stream) + ".min("
+                                            + state.getSourceForNode(
+                                                    sortedTree.getArguments().get(0)) + ")")
+                            .addImport(Comparator.class.getCanonicalName())
```
carterkozak: I don't think this `addImport` is used

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/SortedStreamFirstElementTest.java
```diff
@@ -0,0 +1,133 @@
...
+                        "  public Optional<Integer> f(Stream<Integer> s, Comparator c) {",
+                        "    return s.min(c);",
+                        "  }",
+                        "}")
+                .doTest();
+    }
+
+    private RefactoringValidator fix() {
+        return RefactoringValidator.of(SortedStreamFirstElement.class, getClass());
+    }
```
carterkozak, line 132: great tests!

carterkozak (approved):
Looks great, thanks @mpritham!

PR #2563, opened 2023-04-19
fix eclipse classpath generation

rzpt:
Before, generating eclipse classpath will merge new entries into the existing classpath. This will delete the existing classpath, which makes the classpath correct when it changes.

Also add a few eclipse generated files to gitignore.

## Before this PR

Eclipse classpath generation doesn't work when it merges the new classpath with the old classpath. It produces incorrect entries and eclipse will complain that the classpath is invalid.

## After this PR

==COMMIT_MSG==
Fix eclipse classpath generation by deleting the existing classpath, which makes the classpath correct when it changes.
==COMMIT_MSG==

## Possible downsides?

None.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,9 @@
 .java-version
 .project
 .settings
+.factorypath
+.apt_generated_tests/
+bin/
 build
 out/
 docs/node_modules/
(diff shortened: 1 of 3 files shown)


carterkozak (approved):
lgtm

PR #2578, opened 2023-05-16
Don't ignore all directories named build/

alxhill:
## Before this PR
Previously, error prone would ignore all directories with build/ anywhere in the path, even for non-generated source code. This PR switches to only ignoring directories with both `build/` and `generated_src` somewhere in the full path.

## After this PR

==COMMIT_MSG==
Don't ignore all directories named build/
==COMMIT_MSG==

## Possible downsides?
If there are build/ directories containing generated code without a generated directory, gradle baseline will now scan the code and could block merge/release.

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
@@ -186,7 +186,7 @@ public final class BaselineErrorProne implements Plugin<Project> {
         // Error-prone normalizes filenames to use '/' path separator:
         // https://github.com/google/error-prone/blob/c601758e81723a8efc4671726b8363be7a306dce
         // /check_api/src/main/java/com/google/errorprone/util/ASTHelpers.java#L1277-L1285
-        return ".*/(build|generated_.*[sS]rc|src/generated.*)/.*";
+        return ".*/((build/.*)?generated_.*[sS]rc|src/generated.*)/.*";
     }
 
     private static Optional<Stream<String>> getSpecificErrorProneChecks(Project project) {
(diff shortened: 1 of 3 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
```diff
@@ -186,7 +186,7 @@ static String excludedPathsRegex() {
         // Error-prone normalizes filenames to use '/' path separator:
         // https://github.com/google/error-prone/blob/c601758e81723a8efc4671726b8363be7a306dce
         // /check_api/src/main/java/com/google/errorprone/util/ASTHelpers.java#L1277-L1285
-        return ".*/(build|generated_.*[sS]rc|src/generated.*)/.*";
+        return ".*/((build/.*)?generated_.*[sS]rc|src/generated.*)/.*";
```
pkoenig10: This isn't right - we shouldn't be dropping the alternation here.

We want to ignore directories like `generated_src`. That's what some our legacy [gradle-processors](https://github.com/palantir/gradle-processors) plugin uses and some internal repos have unfortunately adopted that as convention.

I propose we make this regex:
```
.*/(build/generated.*|src/generated.*|generated_.*[S]rc)/.*
```

alxhill: sounds good, will switch to your regex.

fwiw i think the regex above does ignore `generated_src` based on this test case:

```        
predicate.test 'tritium-core/build/metricSchema/generated_src/'
````

but `build/generated` without the `_src` would not match. noticed the tests have `generated_src` nested a few levels deeper than build - is it worth matching `build/.../generated/` as well?

carterkozak: We have historical patterns using this format: `tritium-core/generated_src/` (sibling to the `build/` dir)

pkoenig10 (approved):

PR #2586, opened 2023-06-01
Ckozak/alandau/errorprone 2.19.1

carterkozak:
==COMMIT_MSG==
Opt out of potentially disruptive new checks and revert StrictUnusedVariable workaround (#2537)
==COMMIT_MSG==


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
```diff
@@ -133,6 +133,9 @@ private static void configureErrorProneOptions(
                 "CanIgnoreReturnValueSuggester",
                 "InlineMeSuggester",
                 "PreferImmutableStreamExCollections",
+                // StringCaseLocaleUsage duplicates our existing DefaultLocale check which is already
+                // enforced in some places.
+                "StringCaseLocaleUsage",
                 "UnnecessaryTestMethodPrefix",
```
carterkozak, line 139: Getting failures on this, which is unexpected -- `UnnecessaryTestMethodPrefix` is not enabled by default, so I'll remove it from the list.

```suggestion
```

carterkozak:
I've merged develop into this branch -- I believe the test flake is due to newer versions of baseline than the publishToMavenLocal (with the current history/tags) already being pulished from develop, because I didn't review in a timely fashion :-( I've done this to myself, sorry.

carterkozak:
Unless... that's not it?

PR #2591, opened 2023-06-12
Opt out of the 'NotJavadoc' errorprone check

carterkozak:
We often use this style, it's not worth blocking baseline upgrades.
==COMMIT_MSG==
Opt out of the 'NotJavadoc' errorprone check
==COMMIT_MSG==


bjlaub (approved):

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
```diff
@@ -132,6 +132,8 @@ private static void configureErrorProneOptions(
                 "CatchSpecificity",
                 "CanIgnoreReturnValueSuggester",
                 "InlineMeSuggester",
+                // We often use javadoc comments without javadoc parameter information.
+                "NotJavadoc",
```
pkoenig10, line 136: It feels wrong to disable this. We should not be using Javadoc comments (`/**`) in places where Javadoc is not supported. In those places we should using regular multi-line comments (`/*`).

PR #2593, opened 2023-06-14
`StrictUnusedVariable`: update severity to a warning, and remove suggested fixes to prefix variables with underscores

### LINKED ISSUE #2361: `StrictUnusedVariable` check should suggest removing variables assigned in constructor but never read
## What happened?

If you have a class like this
```
public class Test {
    private final String myField;

    public Test(String myField) {
        this.myField = myField;
    }
}
```
```
    private final String myField;
                         ^
```

```
public class Test {

    public Test(String myField) {
        this._myField = myField;
    }
}
```
```
                         ^
```

The check should suggest removing the variable and assignment entirely rather than allowing underscores.

mpritham:
## Before this PR
The `StrictUnusedVariable` check currently produces errors when a variable is unused unless it is prefixed with an underscore. There are cases (albeit rare) where users legitimately need to keep a reference to an object in scope to prevent the GC from cleaning it. For these scenarios, the underscore prefix (or a suppression) is introduced to satisfy this check.

By moving the severity to the `WARNING` level, users will still be alerted of their unused variables, and can decide what action they'd like to take.

## After this PR

==COMMIT_MSG==
Closes #2361. Change the severity to `WARNING`. Remove suggested fixes to prefix unused variables with underscores, but keep the suggested fix which renames used variables prefixed with underscores.
==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/StrictUnusedVariableTest.java
+++ b/baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/StrictUnusedVariableTest.java
@@ -70,11 +70,11 @@ public class StrictUnusedVariableTest {
                         "Test.java",
                         "import java.util.Optional;",
                         "class Test {",
-                        "  // BUG: Diagnostic contains: '_foo', for example",
+                        "  // BUG: Diagnostic contains: Unused",
                         "   Test(String foo) { }",
-                        "  // BUG: Diagnostic contains: '_buggy', for example",
+                        "  // BUG: Diagnostic contains: Unused",
                         "  private static void privateMethod(String buggy) { }",
-                        "  // BUG: Diagnostic contains: '_buggy', for example",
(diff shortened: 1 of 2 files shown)


baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/StrictUnusedVariableTest.java
```diff
@@ -133,63 +133,49 @@ void handles_lambdas_in_static_init() {
 
     @Test
     public void renames_previous_suppression() {
-        refactoringTestHelper
-                .addInputLines(
+        // if it is prefixed with 'unused' don't do anything.
```
mpritham, line 136: This is not the current functionality, but something I was thinking about. It's could be a bit annoying if we show this warning when the user already acknowledged that the variable is unused by prefixing it with `unused` or `ignored`.

PR #2599, opened 2023-06-26
Fix unintentional suppression of StrictUnusedVariable

carterkozak:
StrictUnusedVariable was suppressed due to disablement of the upstream UnusedVariable check matched in alt-names.
==COMMIT_MSG==
Fix unintentional suppression of StrictUnusedVariable
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-2599.v2.yml
+++ b/changelog/@unreleased/pr-2599.v2.yml
@@ -1,0 +1,5 @@
+type: fix
+fix:
+  description: Fix unintentional suppression of StrictUnusedVariable
+  links:
+  - https://github.com/palantir/gradle-baseline/pull/2599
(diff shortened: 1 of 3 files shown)


changelog/@unreleased/pr-2599.v2.yml
```diff
@@ -0,0 +1,5 @@
+type: fix
+fix:
+  description: Fix unintentional suppression of StrictUnusedVariable
```
CRogers, line 3: Did this suppression happen recently or has it been like this for a long time? Just wondering if this is a "breaking" fix.

carterkozak: 5.12, somewhat recent. We do attempt to auto-fix occurrences in the baseline upgrade though: https://github.com/palantir/gradle-baseline/blob/d18d97050f4bef7491cc1f71cc1e09d588e0b55f/gradle-baseline-java/src/main/groovy/com/palantir/baseline/extensions/BaselineErrorProneExtension.java#L73

CRogers:
:+1:

PR #2602, opened 2023-06-27
Fix nullaway checkerframework dependency

carterkozak:
## After this PR

==COMMIT_MSG==
Fix nullaway checkerframework dependency
==COMMIT_MSG==

## Possible downsides?
Ideally I'd have been able to reproduce this in a test. I've verified manually that this works as expected.


pkoenig10 (approved):

versions.props
```diff
@@ -47,4 +46,7 @@ com.palantir.javaformat:gradle-palantir-java-format = 1.1.0
 com.diffplug.spotless:spotless-plugin-gradle = 6.6.0
 # Groovy versions must be compatible with gradle
 org.codehaus.groovy:* = 3.0.10
+# checkerframework dependencies should be limited to errorpropne and nullaway
+# requirements, upgrading prior to those often causes compilation failures.
+org.checkerframework:* = 3.33.0
```
schlosna, line 51: Is there a specific bottom level dependency we want to control the org.checkerframework version (possibly error-prone itself)? Do we need to also pin lower version and disable  dependency-upgrader in other projects, or rely on excavators failing on compilation (e.g. [tritium gradle-baseline bump](https://app.circleci.com/pipelines/github/palantir/tritium/2881/workflows/4ad05675-bfef-4209-a56d-623494ed1d58/jobs/19797?invite=true#step-106-119))?

carterkozak: I wouldn't expect the dependency upgrader to attempt upgrading checkerframework on many other projects, but it's certainly possible.

By excluding the upgrade here, my goal is to allow nullaway and error-prone to upgrade, which will keep checkerframework updated based on their constraints.

PR #2605, opened 2023-06-29
Use a `Proxy` for `JavaInstallationMetadata` so we can work across Gradle 7 and 8

felixdesouza:
## Before this PR
If you try to run this on Gradle 8, this would fail as this interface has been updated in Gradle 8. 

To ensure that we can work across both Gradle 7/8 during this transitional period, we should use a proxy.

Longer term, we should get rid of all of this and do the gradle jdks thing so we don't have to muck around with the internal toolchains stuff.

## After this PR

==COMMIT_MSG==
Use a `Proxy` for `JavaInstallationMetadata` so we can work across Gradle 7 and 8.
==COMMIT_MSG==


felixdesouza:
👍  but you're gonna need another reviewer 😬

CRogers:
:+1:

PR #2608, opened 2023-07-03
Make custom jdks work with gradle 8 😅 

felixdesouza:
## Before this PR
`BaselineJavaCompiler` needs to be of type `DefaultToolchainJavaCompiler` to work correctly in Gradle 8. In Gradle 7, when it came to compile, there was a codepath that meant that the values from the `BaselineJavaCompiler` were used, but when it came to compile, it used the normal `DefaultToolchainJavaCompiler`. This all happened to work, because of some potentially unintended side effect. In Gradle 8, that has been cleaned up, making this approach non-viable as it fails to be cast to `DefaultToolchainJavaCompiler. 

## After this PR

~So the idea here, is to then make `BaselineJavaCompiler` extend `DefaultToolchainJavaCompiler` and that should be okay?~

~However, the method signature for `JavaToolchain` has changed between multiple versions of gradle, so you effectively 
have this massive switch statement. All in all, it's pretty grim, and I hope we can find a better solution here.~

I was in too deep, fortunately @CRogers rescued me and said we could instead just use a `null` toolchain, and just copy-pasta the code that uses it to not use it.

Worked much better + no reflection. Also didn't produce weird stack overflow errors when used by an actual internal product.

==COMMIT_MSG==
Ensure custom jdks work with gradle 8.
==COMMIT_MSG==

## Possible downsides?

This is ~so~  a bit fragile, ~I'm breaking even writing this description~. There is zero protection from any version of gradle higher than 8.2 changing the signature and breaking us. After all, we are heavily using internal properties ~*and* reflection~.

If we are happy with just using internal properties, another potential approach is:
* register our own implementation of `InstallationSupplier` which supplies a `Set<InstallationLocation>`.
* disable auto registration + auto detection, if we've set these up.

Don't think we have to worry about any reflection issues between 7.6 to 8.2, but it's substantially less reflection however, a bit more work to uproot the existing jdks stuff.

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/javaversions/BaselineJavaCompiler.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/javaversions/BaselineJavaCompiler.java
@@ -16,102 +16,23 @@
 
 package com.palantir.baseline.plugins.javaversions;
 
-import java.lang.reflect.Constructor;
-import java.lang.reflect.Proxy;
-import org.gradle.api.JavaVersion;
 import org.gradle.api.file.RegularFile;
-import org.gradle.api.internal.file.FileFactory;
-import org.gradle.internal.jvm.inspection.JvmInstallationMetadata;
-import org.gradle.internal.operations.BuildOperationProgressEventEmitter;
+import org.gradle.api.tasks.WorkResult;
 import org.gradle.internal.service.ServiceRegistry;
 import org.gradle.jvm.toolchain.JavaInstallationMetadata;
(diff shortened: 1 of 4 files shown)


carterkozak (commented):
This would be helpful for https://github.com/palantir/guava-compatibility-agent which is used in cassandra tests that require jdk8 :-)

CRogers:
@carterkozak JDK8 or Gradle 8?

carterkozak:
I… did not read very thoroughly 🤦

CRogers:
Superseded by https://github.com/palantir/gradle-baseline/pull/2641

PR #2616, opened 2023-07-29
Prefer InputStream.transferTo(OutputStream)

### LINKED ISSUE #2615: FR: Prefer InputStream.transferTo(OutputStream)
## What happened?

While reviewing https://github.com/palantir/dialogue/pull/1983 @jkozlowski flagged that it could be useful to have an error-prone check to automate migration to prefer `InputStream.transferTo(OutputStream)` instead of utility methods such as Guava's `com.google.common.io.ByteStreams.copy(InputStream, OutputStream)`.

Allow for optimization when underlying input stream (such as `ByteArrayInputStream`, `ChannelInputStream`) overrides `transferTo(OutputStream)` to avoid extra array allocations and copy larger chunks at a time (e.g. allowing 16KiB chunks via `ApacheHttpClientBlockingChannel.ModulatingOutputStream` from #1790).

When running on JDK 21+, this also enables 16KiB byte chunk copies via `InputStream.transferTo(OutputStream)` per [JDK-8299336](https://bugs.openjdk.org/browse/JDK-8299336), where as on JDK < 21 and when using Guava `ByteStreams.copy` 8KiB byte chunk copies are used.

References:
* https://github.com/palantir/hadoop-crypto/pull/586
* https://bugs.openjdk.org/browse/JDK-8299336
* https://bugs.openjdk.org/browse/JDK-8067661
* https://bugs.openjdk.org/browse/JDK-8265891
* https://bugs.openjdk.org/browse/JDK-8273038
* https://bugs.openjdk.org/browse/JDK-8279283
* https://bugs.openjdk.org/browse/JDK-8296431

baseline could automate use of `InputStream.transferTo(OutputStream)`

schlosna:
## Before this PR

https://github.com/palantir/gradle-baseline/issues/2615

## After this PR

==COMMIT_MSG==
Prefer InputStream.transferTo(OutputStream)

Add error-prone check to automate migration to prefer `InputStream.transferTo(OutputStream)` instead of utility methods such as Guava's `com.google.common.io.ByteStreams.copy(InputStream, OutputStream)`.

Allow for optimization when underlying input stream (such as `ByteArrayInputStream`, `ChannelInputStream`) overrides `transferTo(OutputStream)` to avoid extra array allocations and copy larger chunks at a time (e.g. allowing 16KiB chunks via `ApacheHttpClientBlockingChannel.ModulatingOutputStream` from #1790).

When running on JDK 21+, this also enables 16KiB byte chunk copies via `InputStream.transferTo(OutputStream)` per
[JDK-8299336](https://bugs.openjdk.org/browse/JDK-8299336), where as on JDK < 21 and when using Guava `ByteStreams.copy` 8KiB byte chunk copies are used.

References:
* https://github.com/palantir/hadoop-crypto/pull/586
* https://bugs.openjdk.org/browse/JDK-8299336
* https://bugs.openjdk.org/browse/JDK-8067661
* https://bugs.openjdk.org/browse/JDK-8265891
* https://bugs.openjdk.org/browse/JDK-8273038
* https://bugs.openjdk.org/browse/JDK-8279283
* https://bugs.openjdk.org/browse/JDK-8296431

Closes https://github.com/palantir/gradle-baseline/issues/2615

==COMMIT_MSG==


README.md
```diff
@@ -228,6 +228,7 @@ Safe Logging can be found at [github.com/palantir/safe-logging](https://github.c
 - `DangerousCollapseKeysUsage`: Disallow usage of `EntryStream#collapseKeys()`.
 - `JooqBatchWithoutBindArgs`: jOOQ batch methods that execute without bind args can cause performance problems.
 - `InvocationTargetExceptionGetTargetException`: InvocationTargetException.getTargetException() predates the general-purpose exception chaining facility. The Throwable.getCause() method is now the preferred means of obtaining this information. [(source)](https://docs.oracle.com/en/java/javase/17/docs/api//java.base/java/lang/reflect/InvocationTargetException.html#getTargetException())
+- `PreferInputStreamTransferTo`: Prefer JDK `InputStream.transferTo(OutputStream)` over utility methods such as `com.google.common.io.ByteStreams.copy(InputStream, OutputStream)`, `org.apache.commons.io.IOUtils.copy(InputStream, OutputStream)`, `org.apache.commons.io.IOUtils.copyLong(InputStream, OutputStream)`.
```
schlosna, line 231: I had to try really hard not to sort & clean this up in this PR, but I would like to create a separate PR to sort these and ensure we actually have coverage for all the current checks in baseline.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferInputStreamTransferTo.java
```diff
@@ -0,0 +1,115 @@
...
+            .onDescendantOf("org.apache.commons.io.IOUtils")
+            .namedAnyOf("copy", "copyLarge")
+            .withParameters(INPUT_STREAM, OUTPUT_STREAM);
+    private static final Matcher<ExpressionTree> AWS_BYTE_STREAM_COPY_MATCHER = MethodMatchers.staticMethod()
+            .onDescendantOf("com.amazonaws.util.IOUtils")
+            .namedAnyOf("copy")
+            .withParameters(INPUT_STREAM, OUTPUT_STREAM);
+
+    private static final Matcher<ExpressionTree> METHOD_MATCHER = Matchers.anyOf(
+            GUAVA_BYTE_STREAM_COPY_MATCHER, APACHE_COMMONS_BYTE_STREAM_COPY_MATCHER, AWS_BYTE_STREAM_COPY_MATCHER);
```
schlosna, line 85: I started with these mostly for simplicity and to cover the common usages, and may expand in follow on PR. Hadoop [`IOUtils.copyBytes`](https://hadoop.apache.org/docs/stable/api/org/apache/hadoop/io/IOUtils.html#copyBytes-java.io.InputStream-java.io.OutputStream-org.apache.hadoop.conf.Configuration-) methods are some the worst offenders I've seen as they default to 4KiB chunks 😢

README.md
carterkozak: I toyed with the upstream docs generator to programmatically generate this list, but it added more complexity to the build than I’d really like.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferInputStreamTransferTo.java
```diff
@@ -0,0 +1,115 @@
...
+ * References:
+ *  <ul>
+ *  <li><a href="https://github.com/palantir/hadoop-crypto/pull/586">https://github.com/palantir/hadoop-crypto/pull/586</a></li>
+ *  <li><a href="https://bugs.openjdk.org/browse/JDK-8299336">https://bugs.openjdk.org/browse/JDK-8299336</a></li>
+ *  <li><a href="https://bugs.openjdk.org/browse/JDK-8067661">https://bugs.openjdk.org/browse/JDK-8067661</a></li>
+ *  <li><a href="https://bugs.openjdk.org/browse/JDK-8265891">https://bugs.openjdk.org/browse/JDK-8265891</a></li>
+ *  <li><a href="https://bugs.openjdk.org/browse/JDK-8273038">https://bugs.openjdk.org/browse/JDK-8273038</a></li>
+ *  <li><a href="https://bugs.openjdk.org/browse/JDK-8279283">https://bugs.openjdk.org/browse/JDK-8279283</a></li>
+ *  <li><a href="https://bugs.openjdk.org/browse/JDK-8296431">https://bugs.openjdk.org/browse/JDK-8296431</a></li>
+ *  </ul>
```
carterkozak: Much of this would work nicely in the summary field — error-prone is fantastic at providing relevant information to devs at the moment it’s most useful :-)

README.md
schlosna: We could throw together an annotation processor that just generates a `checks.md` similar to metric-schema. I might try that if I have some coding time during break.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferInputStreamTransferTo.java
schlosna: I can move a lot of this and some of `summary` to [`explanation`](https://errorprone.info/api/latest/com/google/errorprone/BugPattern.html#explanation())

schlosna: updated

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferInputStreamTransferTo.java
```diff
@@ -0,0 +1,111 @@
...
+        if (METHOD_MATCHER.matches(tree, state)) {
+            List<? extends ExpressionTree> args = tree.getArguments();
+            if (args.size() != 2) {
+                return Description.NO_MATCH;
+            }
+
+            ExpressionTree maybeInputStreamArg = args.get(0);
+            ExpressionTree maybeOutputStreamArg = args.get(1);
+            if (INPUT_STREAM_MATCHER.matches(maybeInputStreamArg, state)
+                    && OUTPUT_STREAM_MATCHER.matches(maybeOutputStreamArg, state)) {
```
carterkozak, line 98: The signatures that we match already require these to be an input stream and outputstream, but I suppose future methods we discover may be broader and I don’t mind being especially defensive in this sort of check.
It might be worthwhile to add some defensiveness against updating a transferTo implementation to call itself. We could exclude invocations using ‘this’ on the lhs, with the assumption that inputstream subtypes may require a closer look.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferInputStreamTransferTo.java
```diff
@@ -0,0 +1,111 @@
...
+
+    private static final String ERROR_MESSAGE = "Prefer InputStream.transferTo(OutputStream)";
+
+    public static final String INPUT_STREAM = "java.io.InputStream";
+    private static final Matcher<Tree> INPUT_STREAM_MATCHER = MoreMatchers.isSubtypeOf(INPUT_STREAM);
+    public static final String OUTPUT_STREAM = "java.io.OutputStream";
+    private static final Matcher<Tree> OUTPUT_STREAM_MATCHER = MoreMatchers.isSubtypeOf(OUTPUT_STREAM);
+
+    private static final Matcher<ExpressionTree> GUAVA_BYTE_STREAM_COPY_MATCHER = MethodMatchers.staticMethod()
+            .onDescendantOf("com.google.common.io.ByteStreams")
```
carterkozak: Checking specific target classes may slightly reduce the cost of this check
```suggestion
            .onClass("com.google.common.io.ByteStreams")
```

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferInputStreamTransferTo.java
schlosna: > The signatures that we match already require these to be an input stream and outputstream, but I suppose future methods we discover may be broader and I don’t mind being especially defensive in this sort of check.

Yeah, might be worth simplifying the arg checks now.

schlosna: > It might be worthwhile to add some defensiveness against updating a transferTo implementation to call itself. We could exclude invocations using ‘this’ on the lhs, with the assumption that inputstream subtypes may require a closer look.

I think this would likely still be a valid refactor. I could imagine some  custom InputStream that had a `writeTo(OutputStream out)` method that used say `ByteStreams.copy(this, out)`

carterkozak: The case that I have in mind here is

```diff
@Override
public long transferTo(OutputStream other) throws IOException {
-    ByteStreams.copy(this, other);
+    this.transferTo(other); // Infinite recursion!
}

Comment (schlosna) [reply baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferInputStreamTransferTo.java:98]: Ahh, good call. Would you prefer we leave as-is for `this` LHS for now or try to check the caller method scope is `transferTo(OutputStream)` and swap `super` for `this`?

File: baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferInputStreamTransferTo.java (hunk 5 above)
Comment (schlosna) [reply baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferInputStreamTransferTo.java]: Thanks, nice! I need to dig into the error-prone APIs a bit more (may also be worth comparing to LLM output for matchers)

File: baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferInputStreamTransferTo.java [hunk 6]
```diff
@@ -0,0 +1,111 @@
...
+@BugPattern(
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
+        severity = SeverityLevel.WARNING,
+        summary = "Prefer JDK `InputStream.transferTo(OutputStream)` over utility methods such as "
+                + "`com.google.common.io.ByteStreams.copy(InputStream, OutputStream)`, "
+                + "`org.apache.commons.io.IOUtils.copy(InputStream, OutputStream)`, "
+                + "`org.apache.commons.io.IOUtils.copyLong(InputStream, OutputStream)` "
+                + "see: https://github.com/palantir/gradle-baseline/issues/2615 ",
+        explanation = "Allow for optimization when underlying input stream (such as `ByteArrayInputStream`, "
```
schlosna: - [x] need to fix up this wrapping and formatting

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferInputStreamTransferTo.java
carterkozak: I think either way you prefer is reasonable. If I were writing this, I would probably leave as-is for `this` on the lhs since I expect more edge cases while implementing an InputStream compared to consuming one. My philosophy with this sort of check is that we should catch the common cases where we have the most confidence in our recommendations, and use those to teach folks the intricacies which they can apply to more ambiguous cases. But again, it's a subtle difference, and I'm happy to move forward with either.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferInputStreamTransferTo.java
```diff
@@ -0,0 +1,121 @@
...
+            ExpressionTree maybeOutputStreamArg = args.get(1);
+            if (INPUT_STREAM_MATCHER.matches(maybeInputStreamArg, state)
+                    && OUTPUT_STREAM_MATCHER.matches(maybeOutputStreamArg, state)) {
+                String inputStreamArg = state.getSourceForNode(maybeInputStreamArg);
+                String outputStreamArg = state.getSourceForNode(maybeOutputStreamArg);
+                if (inputStreamArg == null || outputStreamArg == null) {
+                    return Description.NO_MATCH;
+                }
+
+                if (inputStreamArg.contains("this")) {
```
carterkozak: We may want to check something more precise to avoid updating something like `this.getStream()` into `super.getStream()` (this is speculation on my part, I haven't verified that this example will fail)

```java
if (inputStreamArg instanceof IdentifierTree
    && ((IdentifierTree) inputStreamArg).getName().contentEquals("this")) {
    inputStreamArg = "super";
}
```

carterkozak (approved):
Generally looks good to me, sorry for the delay reviewing! One potential nit, but I defer to your judgement on that.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferInputStreamTransferTo.java
schlosna: Thanks @carterkozak ! I added some more test coverage for this case and updated.

carterkozak: Looks great, thank you!

schlosna: I knew I should have split up the tests into individual methods, checkstyle was unhappy

PR #2629, opened 2023-08-25
Improve SafeLoggingPropagation on Immutables

carterkozak:
==COMMIT_MSG==
Improve SafeLoggingPropagation on Immutables, taking into account fields from superinterfaces
==COMMIT_MSG==

This change migrates much of the immutables safety logic from `Tree` analysis to `Symbol` analysis, allowing us to introspect linked code rather than immediate sources. This is required in order to take fields from superinterfaces into account.

Note that because this change will suggest changes that we weren't able to enforce before, I've added `SafeLoggingPropagation` to `BaselineErrorProneExtension` in order to apply the requested changes by default on upgrade instead of failing to compile.

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/SafeLoggingPropagation.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/SafeLoggingPropagation.java
@@ -49,6 +49,7 @@ import com.sun.tools.javac.code.Symbol.ClassSymbol;
 import com.sun.tools.javac.code.Symbol.MethodSymbol;
 import com.sun.tools.javac.code.Symbol.TypeSymbol;
 import com.sun.tools.javac.code.Symbol.VarSymbol;
+import com.sun.tools.javac.code.Type;
 import com.sun.tools.javac.util.Name;
 import java.util.List;
 import javax.lang.model.element.Modifier;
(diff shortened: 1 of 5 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/SafeLoggingPropagation.java
```diff
@@ -195,43 +174,87 @@ private Description matchClassOrInterface(ClassTree classTree, ClassSymbol class
         return matchBasedOnToString(classTree, classSymbol, state);
     }
 
+    private static boolean isImmutablesField(
+            ClassSymbol enclosingClass, MethodSymbol methodSymbol, VisitorState state) {
+        return methodSymbol.getModifiers().contains(Modifier.ABSTRACT)
+                || ASTHelpers.hasAnnotation(methodSymbol, "org.immutables.value.Value.Default", state)
```
mpritham, line 180: Clarification q: We don't have the `Redacted` annotation here, which means `isGetterMethod` would return false for `test` here: 
```
@Value.Redacted
default String test() { return "test'; }
```

I think we expect `isGetterMethod` to evaluate to true here given we have logic to handle/ignore redacted fields in the branch that evaluates when `isGetterMethod` returns true?

carterkozak: It really depends on `immutablesDefaultAsDefault(enclosingClass, state)` below -- by default, `default` methods are considered utility functions, and aren't included as fields in immutables objects. The redacted annotation isn't meaningful on utility methods, and can be ignored.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/SafeLoggingPropagation.java
```diff
@@ -195,43 +174,87 @@ private Description matchClassOrInterface(ClassTree classTree, ClassSymbol class
...
+            if (superIface.tsym instanceof ClassSymbol) {
+                ClassSymbol superIfaceClassSymbol = (ClassSymbol) superIface.tsym;
+                Safety superClassMethodSafety = scanSymbolMethods(superIfaceClassSymbol, state, usesJackson);
+                safety = Safety.mergeAssumingUnknownIsSame(safety, superClassMethodSafety);
+            }
+        }
+        return safety;
+    }
+
+    @SuppressWarnings("checkstyle:CyclomaticComplexity")
```
mpritham: I think this isn't needed anymore?

mpritham (approved):
lgtm! just a nit.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/SafeLoggingPropagation.java
carterkozak: Good catch, thanks!

PR #2636, opened 2023-09-28
Allow incubating method use inside other incubating methods

pkoenig10:
See https://github.com/palantir/conjure-java/pull/2108


carterkozak (approved):
lgtm, thanks!

PR #2639, opened 2023-10-10
Lazy exact dependencies

CRogers:
## Before this PR
Repo upgrades to Gradle 8 are failing because of errors when using `baseline-java-versions` and `baseline-exact-dependencies` together:

```
> Failed to apply plugin class ‘com.palantir.baseline.plugins.javaversions.BaselineJavaVersion’.
The value for property ‘languageVersion’ is final and cannot be changed any further.
```

The [source of the problem is that in Gradle 8.3](https://docs.gradle.org/current/userguide/upgrading_version_8.html#eager_evaluation_of_configuration_attributes), the values of various Java toolchain related attributes are finalised for a `Configuration` the moment that `Configuration` is created. `baseline-exact-dependencies` was using `getConfigurations().getByName(...)` when configuring source sets, which means that it was forcing the creation of `Configuration`s and so finalising the values of these attributes before `baseline-java-versions` had a chance to configure the `languageVersion` etc.

## After this PR
==COMMIT_MSG==
`baseline-exact-dependencies` is now far more lazy around `Configuration` creation in order to support Gradle 8.
==COMMIT_MSG==

### DIFF
--- a/gradle-baseline-java/build.gradle
+++ b/gradle-baseline-java/build.gradle
@@ -61,6 +61,7 @@ tasks.test.dependsOn tasks.publishToMavenLocal
 test {
     environment 'CIRCLE_ARTIFACTS', "${buildDir}/artifacts"
     environment 'CIRCLE_TEST_REPORTS', "${buildDir}/circle-reports"
+    systemProperty 'ignoreDeprecations', 'true'
 }
 
 gradlePlugin {
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineExactDependencies.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineExactDependencies.java
@@ -49,6 +49,7 @@ import org.gradle.api.artifacts.ResolvedArtifact;
 import org.gradle.api.artifacts.ResolvedDependency;
 import org.gradle.api.artifacts.component.ComponentIdentifier;
 import org.gradle.api.artifacts.component.ProjectComponentIdentifier;
+import org.gradle.api.attributes.Attribute;
 import org.gradle.api.attributes.LibraryElements;
 import org.gradle.api.attributes.Usage;
 import org.gradle.api.plugins.JavaPluginConvention;
@@ -115,20 +116,24 @@ public final class BaselineExactDependencies implements Plugin<Project> {
                         }
                     });
 
-                    // Without this, the 'checkUnusedDependencies correctly picks up project dependency on java-library'
-                    // test fails, by not causing gradle run the jar task, but resolving the path to the jar (rather
-                    // than to the classes directory), which then doesn't exist.
-                    // Specifically, we need to pick up the LIBRARY_ELEMENTS_ATTRIBUTE, which is being configured on
-                    // compileClasspath in JavaBasePlugin.defineConfigurationsForSourceSet, but we can't reference it
-                    // directly because that would require us to depend on Gradle 5.6.
-                    // Instead, we just copy the attributes from compileClasspath.
-                    conf.getAttributes()
-                            .attributeProvider(LibraryElements.LIBRARY_ELEMENTS_ATTRIBUTE, project.provider(() -> {
-                                return compileClasspath
-                                        .get()
-                                        .getAttributes()
-                                        .getAttribute(LibraryElements.LIBRARY_ELEMENTS_ATTRIBUTE);
-                            }));
+                    conf.getDependencies().addAllLater(project.provider(() -> {
+                        // Without this, the 'checkUnusedDependencies correctly picks up project dependency on
+                        // java-library'
+                        // test fails, by not causing gradle run the jar task, but resolving the path to the jar (rather
+                        // than to the classes directory), which then doesn't exist.
+                        // Specifically, we need to pick up the LIBRARY_ELEMENTS_ATTRIBUTE, which is being configured on
+                        // compileClasspath in JavaBasePlugin.defineConfigurationsForSourceSet, but we can't reference
+                        // it
+                        // directly because that would require us to depend on Gradle 5.6.
+                        // Instead, we just copy the attributes from compileClasspath.
+                        compileClasspath.get().getAttributes().keySet().forEach(attribute -> {
+                            Object value =
+                                    compileClasspath.get().getAttributes().getAttribute(attribute);
+                            conf.getAttributes().attribute((Attribute<Object>) attribute, value);
+                        });
+
+                        return Set.of();
+                    }));
 
                     conf.withDependencies(deps -> {
                         // Pick up GCV locks. We're making an internal assumption that this configuration exists,
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineExactDependencies.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineExactDependencies.java
@@ -237,7 +237,7 @@ public final class BaselineExactDependencies implements Plugin<Project> {
     }
 
     /**
-     * The SourceSet#getCompileConfigurationName() method got removed in Gradle 7. Because we want to stay
+     * The {@code SourceSet#getCompileConfigurationName()} method got removed in Gradle 7. Because we want to stay
      * compatible with Gradle 6 but can't compile this method, we reimplement it temporarily.
      * TODO(fwindheuser): Remove after dropping support for Gradle 6.
      */
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineExactDependencies.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineExactDependencies.java
@@ -49,7 +49,6 @@ import org.gradle.api.artifacts.ResolvedArtifact;
 import org.gradle.api.artifacts.ResolvedDependency;
 import org.gradle.api.artifacts.component.ComponentIdentifier;
 import org.gradle.api.artifacts.component.ProjectComponentIdentifier;
-import org.gradle.api.attributes.Attribute;
 import org.gradle.api.attributes.LibraryElements;
 import org.gradle.api.attributes.Usage;
 import org.gradle.api.plugins.JavaPluginConvention;
@@ -116,24 +115,20 @@ public final class BaselineExactDependencies implements Plugin<Project> {
                         }
                     });
 
-                    conf.getDependencies().addAllLater(project.provider(() -> {
-                        // Without this, the 'checkUnusedDependencies correctly picks up project dependency on
-                        // java-library'
-                        // test fails, by not causing gradle run the jar task, but resolving the path to the jar (rather
-                        // than to the classes directory), which then doesn't exist.
-                        // Specifically, we need to pick up the LIBRARY_ELEMENTS_ATTRIBUTE, which is being configured on
-                        // compileClasspath in JavaBasePlugin.defineConfigurationsForSourceSet, but we can't reference
-                        // it
-                        // directly because that would require us to depend on Gradle 5.6.
-                        // Instead, we just copy the attributes from compileClasspath.
-                        compileClasspath.get().getAttributes().keySet().forEach(attribute -> {
-                            Object value =
-                                    compileClasspath.get().getAttributes().getAttribute(attribute);
-                            conf.getAttributes().attribute((Attribute<Object>) attribute, value);
-                        });
-
-                        return Set.of();
-                    }));
+                    // Without this, the 'checkUnusedDependencies correctly picks up project dependency on java-library'
+                    // test fails, by not causing gradle run the jar task, but resolving the path to the jar (rather
+                    // than to the classes directory), which then doesn't exist.
+                    // Specifically, we need to pick up the LIBRARY_ELEMENTS_ATTRIBUTE, which is being configured on
+                    // compileClasspath in JavaBasePlugin.defineConfigurationsForSourceSet, but we can't reference it
+                    // directly because that would require us to depend on Gradle 5.6.
+                    // Instead, we just copy the attributes from compileClasspath.
+                    conf.getAttributes()
+                            .attributeProvider(LibraryElements.LIBRARY_ELEMENTS_ATTRIBUTE, project.provider(() -> {
+                                return compileClasspath
+                                        .get()
+                                        .getAttributes()
+                                        .getAttribute(LibraryElements.LIBRARY_ELEMENTS_ATTRIBUTE);
+                            }));
 
                     conf.withDependencies(deps -> {
                         // Pick up GCV locks. We're making an internal assumption that this configuration exists,
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineExactDependencies.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineExactDependencies.java
@@ -118,13 +118,11 @@ public final class BaselineExactDependencies implements Plugin<Project> {
 
                     project.afterEvaluate(_ignored -> {
                         // Without this, the 'checkUnusedDependencies correctly picks up project dependency on
-                        // java-library'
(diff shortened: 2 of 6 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineExactDependencies.java
```diff
@@ -119,80 +116,95 @@ private static void configureSourceSet(
...
-            });
-        });
+                        // Figure out what our compile dependencies are while ignoring dependencies we've inherited from
+                        // other source sets. For example, if we are `test`, some of our configurations extend from the
+                        // `main` source set:
+                        // testImplementation     extendsFrom(implementation)
+                        //  \-- testCompile       extendsFrom(compile)
+                        // We therefore want to look at only the dependencies _directly_ declared in the implementation
+                        // and compile configurations (belonging to our source set)
+                        Configuration implCopy = project.getConfigurations()
```
CRogers, line 140: I had originally tried to put this in a `beforeResolve`, where it makes most sense to live. But this causes GCV to have `ConcurrentModificationException`s, as it iterates over all the `Configuration`s and then resolves them while this section of code creates more `Configuration`s, meaning I had to keep it in `afterEvaluate`

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineExactDependencies.java
```diff
@@ -119,80 +116,95 @@ private static void configureSourceSet(
...
+                        // Specifically, we need to pick up the LIBRARY_ELEMENTS_ATTRIBUTE, which is being configured on
+                        // compileClasspath in JavaBasePlugin.defineConfigurationsForSourceSet, but we can't reference
+                        // it
+                        // directly because that would require us to depend on Gradle 5.6.
+                        // Instead, we just copy the attributes from compileClasspath.
+                        compileClasspath.get().getAttributes().keySet().forEach(attribute -> {
+                            Object value =
+                                    compileClasspath.get().getAttributes().getAttribute(attribute);
+                            conf.getAttributes().attribute((Attribute<Object>) attribute, value);
+                        });
```
CRogers, line 131: There's no way to lazily add a `Map` of attributes (aka lazily conditionally add an attribute). So it too must unfortunately live in an `afterEvaluate`.

gradle-baseline-java/src/test/groovy/com/palantir/baseline/GradleTestVersions.java
```diff
@@ -21,7 +21,7 @@
 
 public final class GradleTestVersions {
     public static final ImmutableList<String> VERSIONS =
-            ImmutableList.of(Baseline.MIN_GRADLE_VERSION.getVersion(), "7.1.1", "7.3");
+            ImmutableList.of(Baseline.MIN_GRADLE_VERSION.getVersion(), "7.6.2", "8.4");
```
felixdesouza, line 24: do you want this to be 8.4 or 8.3?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineExactDependencies.java
```diff
@@ -227,7 +239,7 @@ static String checkUnusedDependenciesNameForSourceSet(SourceSet sourceSet) {
     }
 
     /**
-     * The {@link SourceSet#getCompileConfigurationName()} method got removed in Gradle 7. Because we want to stay
+     * The SourceSet#getCompileConfigurationName() method got removed in Gradle 7. Because we want to stay
```
felixdesouza: nit:
```suggestion
     * The {@code SourceSet#getCompileConfigurationName()} method got removed in Gradle 7. Because we want to stay
```

gradle-baseline-java/build.gradle
```diff
@@ -61,6 +61,7 @@ tasks.test.dependsOn tasks.publishToMavenLocal
 test {
     environment 'CIRCLE_ARTIFACTS', "${buildDir}/artifacts"
     environment 'CIRCLE_TEST_REPORTS', "${buildDir}/circle-reports"
+    systemProperty 'ignoreDeprecations', 'true'
```
felixdesouza, line 64: how come?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineExactDependencies.java
```diff
@@ -119,80 +116,95 @@ private static void configureSourceSet(
...
+                    project.afterEvaluate(_ignored -> {
+                        // Without this, the 'checkUnusedDependencies correctly picks up project dependency on
+                        // java-library'
+                        // test fails, by not causing gradle run the jar task, but resolving the path to the jar (rather
+                        // than to the classes directory), which then doesn't exist.
+                        // Specifically, we need to pick up the LIBRARY_ELEMENTS_ATTRIBUTE, which is being configured on
+                        // compileClasspath in JavaBasePlugin.defineConfigurationsForSourceSet, but we can't reference
+                        // it
+                        // directly because that would require us to depend on Gradle 5.6.
+                        // Instead, we just copy the attributes from compileClasspath.
```
felixdesouza: nit: can you sort out comment wrapping

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineExactDependencies.java
```diff
@@ -119,80 +116,95 @@ private static void configureSourceSet(
...
-                });
+                    project.afterEvaluate(_ignored -> {
+                        // Without this, the 'checkUnusedDependencies correctly picks up project dependency on
+                        // java-library'
+                        // test fails, by not causing gradle run the jar task, but resolving the path to the jar (rather
+                        // than to the classes directory), which then doesn't exist.
+                        // Specifically, we need to pick up the LIBRARY_ELEMENTS_ATTRIBUTE, which is being configured on
+                        // compileClasspath in JavaBasePlugin.defineConfigurationsForSourceSet, but we can't reference
+                        // it
+                        // directly because that would require us to depend on Gradle 5.6.
```
felixdesouza: probably not for this PR, but surely we can raise the min gradle version? does the concern here still hold without it?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineExactDependencies.java
```diff
@@ -119,80 +116,95 @@ private static void configureSourceSet(
...
+                        //  \-- testCompile       extendsFrom(compile)
+                        // We therefore want to look at only the dependencies _directly_ declared in the implementation
+                        // and compile configurations (belonging to our source set)
+                        Configuration implCopy = project.getConfigurations()
+                                .getByName(sourceSet.getImplementationConfigurationName())
+                                .copy();
+
+                        // Without these, explicitCompile will successfully resolve 0 files and you'll waste 1 hour
+                        // trying to figure out why.
+                        project.getConfigurations().add(implCopy);
```
felixdesouza, line 146: curious whether this would have fixed your initial attempts for the flattener PR?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineExactDependencies.java
felixdesouza: I suppose it just comes down to directly referencing the attribute like we did in line 117? instead of copying all the attributes?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineExactDependencies.java
felixdesouza: I think we can get around this particular usage of `afterEvaluate` since we don't actually need to copy the map, just the attribute from the `compileClasspath` => we know the attribute we want and can do use the attribute provider method for a single attribute key.

Additionally, I *think* this code is only really effective for gradle versions < 5.6? => if our min version is high enough, this bit of code is redundant as it's already covered by https://github.com/palantir/gradle-baseline/blob/36e20deaa2c3bc97c92deaae62516e8a03c7a082/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineExactDependencies.java#L112-L116 ?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineExactDependencies.java
```diff
@@ -119,80 +116,95 @@ private static void configureSourceSet(
...
+                            // a bad time resolving this with GCV, if you have direct dependencies without corresponding
+                            // entries in
+                            // versions.props, but instead rely on getting a version for them from the lock file.
+                            compileCopy.setCanBeResolved(false);
+                            compileCopy.setCanBeConsumed(false);
+
+                            project.getConfigurations().add(compileCopy);
+
+                            conf.extendsFrom(compileCopy);
+                        });
```
felixdesouza: redundant? our min version is effectively 7.0 if you apply *all* the plugins, I think it's 6.7 if you don't apply javaversions

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineExactDependencies.java
```diff
@@ -119,80 +116,95 @@ private static void configureSourceSet(
...
-                        projectsEvaluated.get()
-                                || (project.getGradle().getStartParameter().isConfigureOnDemand()
-                                        && project.getState().getExecuted()),
-                        "Tried to resolve %s too early.",
-                        explicitCompile));
+                    // Since we are copying configurations before resolving 'explicitCompile', make double sure that
+                    // it's not
+                    // being resolved (or dependencies realized via `.getIncoming().getDependencies()`) too early.
+                    AtomicBoolean projectsEvaluated = new AtomicBoolean();
+                    project.getGradle().projectsEvaluated(g -> projectsEvaluated.set(true));
```
felixdesouza, line 186: an aside, do you want some sort of similar check in the flattener?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineExactDependencies.java
```diff
@@ -119,80 +116,95 @@ private static void configureSourceSet(
...
+                    });
+                });
 
         TaskProvider<CheckUnusedDependenciesTask> sourceSetUnusedDependencies = project.getTasks()
                 .register(
                         checkUnusedDependenciesNameForSourceSet(sourceSet), CheckUnusedDependenciesTask.class, task -> {
                             task.dependsOn(sourceSet.getClassesTaskName());
                             task.setSourceClasses(sourceSet.getOutput().getClassesDirs());
-                            task.dependenciesConfiguration(explicitCompile);
+                            task.getDependenciesConfigurations().add(explicitCompile);
```
felixdesouza, line 205: wacky idea, not actually sure, but can't we just use the `getHierarchy()` method of the configuration we're initially copying? and then take the first one? do we actually need/want a copy, or did we only take a copy because that was the only way to get the first level without the other dependencies from super configurations?

felixdesouza: because with that, you can just do a `conf.map(c -> Iterables.getFirst(c.getHierarchy()))`, it's also a `LinkedSet` populated in DFS manner so in theory it should have what we want assuming we don't want/need a separate copy

felixdesouza (commented):
not everything is actionable, or necessarily for this PR, but would be interesting to see, as it could simplify the class a lot

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineExactDependencies.java
CRogers: This was the first thing I tried and it didn't work. Sometimes the library elements attribute does not exist on source configuration. This then causes `attribute(Attribute, Provider<T>)` to throw [as per the javadoc](https://docs.gradle.org/7.6.2/javadoc/org/gradle/api/attributes/AttributeContainer.html#attributeProvider-org.gradle.api.attributes.Attribute-org.gradle.api.provider.Provider-):

> This method can NOT be used to discard the value of an property. Specifying a null provider will result in an IllegalArgumentException being thrown. When the provider has no value at finalization time, an IllegalStateException - regardless of whether or not a convention has been set.

gradle-baseline-java/src/test/groovy/com/palantir/baseline/GradleTestVersions.java
CRogers: Might as well use 8.4 as that is latest.

gradle-baseline-java/build.gradle
CRogers: we discussed this internally some time ago - I actually want to set it on everything:

> should we just ignoreDeprecations for every gradle project? Some reasons why:
> 1. Deprecations often come from other repos we end up running as part of tests. For example, in [internal-repo] there are some tests that run tasks from [other-internal-repo], that then fail with deprecations warnings due to [other-internal-repo]'s tasks. Since we can’t fix them in the [internal-repo], we end up adding `--warning-mode=none` to every test, losing deprecation warnings anyway. Or worse, people add `System.setProperty('ignoreDeprecations', 'true')` to individual tests as nebula suggests, making the build flake.
> 2. It’s impossible to see what caused deprecation warnings using nebula-test. The only way I found out to get good info is to add the buildscan plugin to the test.
> 3. We don’t tend to fix deprecations until a major version remove the feature anyway.
> 4. This leads to our gradle plugin repos being super out of date for their own gradle wrapper, meaning they get wrecked by stuff like not supporting jackson 2.15.2's java 20 code.

In this case, we get deprecation warnings about GCV not working with Gradle 9 when running in 8.3/8.4 - something we can't even fix in this repo.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineExactDependencies.java
CRogers: I was excited when I read this comment a few days ago and gave it a try, but it did absolutely nothing to help.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineExactDependencies.java
CRogers: Maybe, tbh I'd rather just leave it while we still have some repos on Gradle 6 (which I should burn down as part of this work).

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineExactDependencies.java
CRogers: Not really sure I follow here, the first time I saw this plugin was this PR. I'd like to focus on getting this to work at all with Gradle 8 first, if we want to improve it/change it in some way it seems less risky to spend time on that intentionally in another change.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineExactDependencies.java
CRogers: We could probably aim to do some cleanup of minimum gradle versions (including the horrifying `GradleWorkarounds` files that exist in many repos) once we move to Gradle 8.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineExactDependencies.java
CRogers: Honestly, it's [a bit redundant given GCV does the same thing](https://github.com/palantir/gradle-consistent-versions/blob/f407769c01dba863147e6108bd12e6990ab2ebed/src/main/java/com/palantir/gradle/versions/VersionsPropsPlugin.java#L220-L230)

felixdesouza:
👍

PR #2645, opened 2023-10-17
Lazy exact deps - take three

CRogers:
## Before this PR
The previous PR (#2644) was not actually lazy enough to fix the original problem (described in #2639). There are still errors of the form:

```
**A problem occurred evaluating root project 'log-receiver-root'.
> Failed to apply plugin class 'com.palantir.baseline.plugins.javaversions.BaselineJavaVersion'.
   > The value for property 'languageVersion' is final and cannot be changed any further.
```

It turns out these are from the eager copying of attributes from the `compileClasspath` configuration.

## After this PR
==COMMIT_MSG==
Ensure that the baseline-exact-dependencies plugin does not eagerly force the creation of the `compileClasspath` configuration, which causes "'languageVersion' is final and cannot be changed any further." errors in Gradle >=8.3.
==COMMIT_MSG==

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineExactDependencies.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineExactDependencies.java
@@ -55,7 +55,6 @@ import org.gradle.api.plugins.JavaPluginConvention;
 import org.gradle.api.tasks.SourceSet;
 import org.gradle.api.tasks.TaskProvider;
 import org.gradle.util.GUtil;
-import org.gradle.util.GradleVersion;
 
 /** Validates that java projects declare exactly the dependencies they rely on, no more and no less. */
 public final class BaselineExactDependencies implements Plugin<Project> {
@@ -109,12 +108,10 @@ public final class BaselineExactDependencies implements Plugin<Project> {
                         attributes.attribute(
                                 Usage.USAGE_ATTRIBUTE, project.getObjects().named(Usage.class, Usage.JAVA_API));
                         // Ensure we resolve the classes directory for local projects where possible, rather than the
-                        // 'jar' file. We can only do this on Gradle 5.6+, otherwise do nothing.
-                        if (GradleVersion.current().compareTo(GradleVersion.version("5.6")) >= 0) {
-                            attributes.attribute(
-                                    LibraryElements.LIBRARY_ELEMENTS_ATTRIBUTE,
-                                    project.getObjects().named(LibraryElements.class, LibraryElements.CLASSES));
-                        }
+                        // 'jar' file.
+                        attributes.attribute(
+                                LibraryElements.LIBRARY_ELEMENTS_ATTRIBUTE,
+                                project.getObjects().named(LibraryElements.class, LibraryElements.CLASSES));
                     });
 
                     conf.withDependencies(deps -> {
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineExactDependencies.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineExactDependencies.java
@@ -49,7 +49,6 @@ import org.gradle.api.artifacts.ResolvedArtifact;
 import org.gradle.api.artifacts.ResolvedDependency;
 import org.gradle.api.artifacts.component.ComponentIdentifier;
 import org.gradle.api.artifacts.component.ProjectComponentIdentifier;
-import org.gradle.api.attributes.Attribute;
 import org.gradle.api.attributes.LibraryElements;
 import org.gradle.api.attributes.Usage;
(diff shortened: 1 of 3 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineExactDependencies.java
```diff
@@ -120,19 +117,6 @@ private static void configureSourceSet(
                         }
                     });
 
-                    // Without this, the 'checkUnusedDependencies correctly picks up project dependency on java-library'
-                    // test fails, by not causing gradle run the jar task, but resolving the path to the jar (rather
```
CRogers, line 124: this test passes without it though????

gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineExactDependenciesTest.groovy
```diff
@@ -277,6 +278,26 @@ class BaselineExactDependenciesTest extends AbstractPluginTest {
         with(':checkUnusedConstraints', '--stacktrace', '--write-locks').withDebug(true).build()
     }
 
+    def 'in Gradle >=8.3 you can set the toolchain language version without it being finalised'() {
```
felixdesouza, line 281: this is a failing test in the end?

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineExactDependencies.java
```diff
@@ -120,19 +117,6 @@ private static void configureSourceSet(
...
-                    // than to the classes directory), which then doesn't exist.
-
-                    // Specifically, we need to pick up the LIBRARY_ELEMENTS_ATTRIBUTE, which is being configured on
-                    // compileClasspath in JavaBasePlugin.defineConfigurationsForSourceSet, but we can't reference it
-                    // directly because that would require us to depend on Gradle 5.6.
-                    // Instead, we just copy the attributes from compileClasspath.
-                    compileClasspath.get().getAttributes().keySet().forEach(attribute -> {
-                        Object value = compileClasspath.get().getAttributes().getAttribute(attribute);
-                        conf.getAttributes().attribute((Attribute<Object>) attribute, value);
-                    });
```
CRogers, line 134: From the comments alone I think this is safe to remove as _apparently_ we only want the `LIBRARY_ELEMENTS_ATTRIBUTE`, which we already set previously. The other attributes we may not actually need. If we really wanted to we could either call the internal [JvmPluginServices#configureAsCompileClasspath](https://github.com/gradle/gradle/blob/080ed571d872395e25b6e7ae3af449991016524a/subprojects/platform-jvm/src/main/java/org/gradle/api/plugins/jvm/internal/DefaultJvmPluginServices.java#L63) (dangerous imo) or copy what it's doing with the extra attributes. I am unsure if this is actually safe, but will test on a variety of internal repos manually to check. It works on some simple ones at least.

felixdesouza: I think this makes sense, otherwise, the stuff in line 116 is wrong, and not sufficient right? since this bit of code always run unconditionally.

An aside, do you want to remove the gradle 5.6 conditional check? in line 116-119?

gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineExactDependenciesTest.groovy
CRogers: Yeah, I had just stupidly not set the gradle version to 8.4, so it was running with 7.x

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineExactDependencies.java
CRogers: Yeah, I might as well remove that.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineExactDependencies.java
```diff
@@ -176,6 +157,9 @@ private static void configureSourceSet(
 
             explicitCompile.get().extendsFrom(implCopy);
 
+            Optional<Configuration> maybeCompile =
```
felixdesouza, line 160: I think as a follow up potentially, once we're out of the woods, we can delete this, I don't think it will ever run because that configuration doesn't exist anymore

felixdesouza (approved):

CRogers:
Ok, I tried this on 3 chonky internal projects in addition to the simple few I tried and the `checkUnusedDependencies` task still works (for the projects it has enabled). Pretty much no one seems to use `checkImplicitDependencies` lol. Write locks also worked for all of them.

PR #2663, opened 2023-11-05
Allow leading underscore in variable names

pkoenig10:
This checkstyle rules conflicts with the `StrictUnusedVariable` check and is difficult or impossible to work around. Devs have run into this conflict numerous times over the years.

For example, I recently had a legitimate case where I wanted to ignore a return value annotated with `@CheckReturnValue`. The [`CheckReturnValue` docs](https://errorprone.info/bugpattern/CheckReturnValue) recommend assigning the value to an unused variable instead of using `@SuppressWarnings` (which seems like good advice given that `@SuppressWarnings` applies to the entire method). However it's impossible to do this with the checkstyle rule and `StrictUnusedVariable` check as they exist today.

This feels like a fairly minor concession with some meaningful ergonomic improvements.


pkoenig10:
It looks like this was discussed previously in https://github.com/palantir/gradle-baseline/pull/1104 and rejected. Not sure if anything has changed since then.

I suppose the risk is fairly high. Even if there are some legitimate use cases to ignore `@CheckReturnValue`, ignoring the return value in illegitimate cases can be fairly disastrous.

PR #2664, opened 2023-11-05
Exclude JUnit5 extensions from StrictUnusedVariable

pkoenig10:
If you have a JUnit5 extension that needs to be explicitly constructed (ie. you can't use `@ExtendWith`), then this will trigger the `StrictUnusedVariable` check if that extension isn't used by test code.

This affects nearly every internal repo that uses our internal integration testing extension. These repos are forced to add `@SuppressWarnings("StrictUnusedVariable")` which is unfortunate.

This is the JUnit5 equivalent of the existing `org.junit.Rule` exemption.


carterkozak (approved):
good idea, thanks!

PR #2667, opened 2023-11-06
stream to iterator is unsafe

lsingh123:
Adds an error prone warning for calling `Stream.iterator`, which loads the entire stream into memory and may cause OOMs ([JDK issue](https://bugs.openjdk.org/browse/JDK-8202307)). 

## Possible downsides?
For small streams, calling iterator on stream may be reasonable. I think it's clearer to explicitly collect the stream because users may not realize that `iterator` loads the stream into memory, but open to ideas.

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StreamToIterator.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StreamToIterator.java
@@ -33,7 +33,10 @@ import java.util.stream.Stream;
         link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
         linkType = BugPattern.LinkType.CUSTOM,
         severity = SeverityLevel.WARNING,
-        summary = "Calling iterator on stream buffers the stream into memory.")
+        summary =
+                "Avoid calling iterator on stream, as it buffers the stream into memory and may cause unexpected OOMs."
+                        + "For small streams, consider collecting the stream first to make it clear to readers that"
+                        + " the stream will be loaded into memory.")
 public final class StreamToIterator extends BugChecker implements BugChecker.MethodInvocationTreeMatcher {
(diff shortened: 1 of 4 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StreamToIterator.java
```diff
@@ -0,0 +1,52 @@
...
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
+        severity = SeverityLevel.WARNING,
+        summary =
+                "Avoid calling iterator on stream, as it buffers the stream into memory and may cause unexpected OOMs."
+                        + "For small streams, consider collecting the stream first to make it clear to readers that"
+                        + " the stream will be loaded into memory.")
```
carterkozak, line 39: This is fair, and a subtle interaction that is probably not expected in most cases, but I'm a bit worried about how common it may be. The suggested alternative isn't necessarily better, but makes it more obvious reading the code that data may be buffered -- in many cases, the stream isn't fully buffered, so the suggestion would result in adding buffering (albeit at a known point, with fewer spooky interactions).

In general when we are worried about resource utilization (cpu and/or memory) we should stay clear of streams imo.

lsingh123: That's true - I was concerned that people probably use streams often for innocuous use-cases where buffering really won't make a difference. Our team has been bit by this twice, but maybe we really should just not be using streams at all.

lsingh123:
See Carter's comment above; while this behavior can be dangerous, it will probably produce many false positives and the appropriate behavior is actually to avoid using streams when performance is critical.

PR #2670, opened 2023-11-09
Allow leading _ in Java local var names

CVdV-au:
## Before this PR

errorprone: CheckReturnValue requires methods with a return value use them (and @CanIgnoreReturnValue isn't an option).
If you really don't want to use the result, then assign it to a variable: `var unused = ...`

But doing that then results in another errorprone: StrictUnusedVariable: The local variable 'unused' is never read. Intentional occurrences are acknowledged by renaming the unused variable with a leading underscore. '_unused'

Checkstyles localVariableName rejects that (however allows method args to have leading _).
## After this PR

==COMMIT_MSG==
Allow local vars to have a leading _ to align with errorProne and be consistent with method arg names.
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-2670.v2.yml
+++ b/changelog/@unreleased/pr-2670.v2.yml
@@ -1,0 +1,6 @@
+type: fix
+fix:
+  description: Allow local vars to have a leading _ to align with errorProne and be
+    consistent with method arg names.
+  links:
+  - https://github.com/palantir/gradle-baseline/pull/2670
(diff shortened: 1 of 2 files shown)


CVdV-au:
#2663 raised this recently, and was originally proposed back in #1104.

The argument against this seems a bit circular: https://github.com/palantir/gradle-baseline/pull/1104#discussion_r355551386 as @SuppressWarnings("CheckReturnValue") is less granular.

Issue #2668, opened 2023-11-07
[IllegalSafeLoggingArgument] Prevent subclasses from overriding methods to be less safe


ash211:
## What happened?

While working on another PR, I saw something that was roughly this situation:
```
interface MyInterface {
  @Safe
  String myMethod();
}

class MyClass implements MyInterface {
  @Unsafe
  String myMethod() {
    return "The secret is 42";
  }
}
```

(with `SafeLoggable` and `RemoteException`)

## What did you want to happen?

I would expect `IllegalSafeLoggingArgument` to throw an error, but it did not.

## Next Steps

Should verify what the current behavior is with a test in this repo.

carterkozak:
Yes, we should cause compilation to fail in that case, as well as the case where an unmarked class attempts to implement two interfaces with different safety values for a given method -- these examples violate the liskov substitution principle.

carterkozak:
https://github.com/palantir/gradle-baseline/pull/2699

PR #2671, opened 2023-11-09
Add test coverage for interface method vs class annotation conflict

ash211:
Addresses https://github.com/palantir/gradle-baseline/issues/2668

### DIFF
--- a/baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/IllegalSafeLoggingArgumentTest.java
+++ b/baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/IllegalSafeLoggingArgumentTest.java
@@ -862,6 +862,47 @@ class IllegalSafeLoggingArgumentTest {
     }
 
     @Test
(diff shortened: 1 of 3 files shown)


baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/IllegalSafeLoggingArgumentTest.java
```diff
@@ -861,6 +861,47 @@ public void testInvalidValue_AnnotatedSupertype() {
...
+                        "    @Safe",
+                        "    public String unsafeMethod() {",
+                        "      return \"unsafe\";",
+                        "    }",
+                        "  }",
+                        "}")
+                .doTest();
+    }
+
+    // TODO(#2668): the @Safe interface method should conflict with the @Unsafe class method
```
ash211, line 884: I wasn't sure if this was an omission in IllegalSafeLoggingArgument or in SafeLoggingPropagation, so added to both.

@carterkozak Am I right that this is a situation where we should be failing?

PR #2672, opened 2023-11-09
Lint records with array fields

raiju:
## Before this PR

Records with array fields end up with surprising equality behavior: comparisons are performed using `.equals`, which end up being value comparisons on most types, but are reference comparisons for arrays.

```
byte[] a = new byte[] {1, 2};
byte[] b = new byte[] {1, 2};

System.out.println(a.equals(b)); // -> false
System.out.println(Arrays.equals(a, b)); // -> true
```

Note that this is already warned for when done in code: https://errorprone.info/bugpattern/ArrayEquals

This is also handled explicitly by immutables: https://github.com/immutables/immutables/blob/7c78f66766bd96e9f37ed02b866c2a08871cd89a/mirror/src/org/immutables/mirror/processor/Mirrors.generator#L510

## After this PR

==COMMIT_MSG==
Warn when using records with array fields as their equality behavior is surprising
==COMMIT_MSG==

Question for reviewer: should we attempt to auto-gen equals/hashCode? I'd prefer not to, as these tend to be more stylistic/converting to immutables might be preferable in these cases.

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DangerousRecordArrayField.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DangerousRecordArrayField.java
@@ -16,14 +16,6 @@
 
 package com.palantir.baseline.errorprone;
 
-import static com.google.errorprone.matchers.Matchers.allOf;
-import static com.google.errorprone.matchers.Matchers.equalsMethodDeclaration;
-import static com.google.errorprone.matchers.Matchers.hashCodeMethodDeclaration;
-import static com.google.errorprone.matchers.Matchers.instanceEqualsInvocation;
-import static com.google.errorprone.matchers.Matchers.instanceHashCodeInvocation;
-import static com.google.errorprone.matchers.Matchers.not;
-import static com.google.errorprone.matchers.Matchers.singleStatementReturnMatcher;
-
 import com.google.auto.service.AutoService;
 import com.google.errorprone.BugPattern;
 import com.google.errorprone.BugPattern.SeverityLevel;
@@ -51,10 +43,12 @@ import com.sun.tools.javac.code.Symbol.ClassSymbol;
 public final class DangerousRecordArrayField extends BugChecker implements BugChecker.ClassTreeMatcher {
 
     private static final Matcher<VariableTree> IS_ARRAY_VARIABLE = Matchers.isArrayType();
-    private static final Matcher<MethodTree> NON_TRIVIAL_EQUALS =
-            allOf(equalsMethodDeclaration(), not(singleStatementReturnMatcher(instanceEqualsInvocation())));
-    private static final Matcher<MethodTree> NON_TRIVIAL_HASHCODE =
-            allOf(hashCodeMethodDeclaration(), not(singleStatementReturnMatcher(instanceHashCodeInvocation())));
+    private static final Matcher<MethodTree> NON_TRIVIAL_EQUALS = Matchers.allOf(
+            Matchers.equalsMethodDeclaration(),
+            Matchers.not(Matchers.singleStatementReturnMatcher(Matchers.instanceEqualsInvocation())));
+    private static final Matcher<MethodTree> NON_TRIVIAL_HASHCODE = Matchers.allOf(
+            Matchers.hashCodeMethodDeclaration(),
+            Matchers.not(Matchers.singleStatementReturnMatcher(Matchers.instanceHashCodeInvocation())));
 
     @Override
     public Description matchClass(ClassTree classTree, VisitorState state) {
@@ -62,15 +56,35 @@ public final class DangerousRecordArrayField extends BugChecker implements BugCh
         if (!ASTHelpers.isRecord(classSymbol)) {
             return Description.NO_MATCH;
         }
-        boolean hasArrayField = false;
-        boolean hasEquals = false;
-        boolean hasHashCode = false;
+        if (!hasArrayField(classTree, state)) {
+            return Description.NO_MATCH;
+        }
+        if (hasNonTrivialEqualsAndHashCode(classTree, state)) {
+            return Description.NO_MATCH;
+        }
+
+        return buildDescription(classTree).build();
+    }
+
+    private static boolean hasArrayField(ClassTree classTree, VisitorState state) {
         for (Tree member : classTree.getMembers()) {
             if (member instanceof VariableTree) {
                 VariableTree variableTree = (VariableTree) member;
 
-                hasArrayField = hasArrayField || IS_ARRAY_VARIABLE.matches(variableTree, state);
-            } else if (member instanceof MethodTree) {
+                if (IS_ARRAY_VARIABLE.matches(variableTree, state)) {
+                    return true;
+                }
+            }
+        }
+
+        return false;
+    }
+
+    private static boolean hasNonTrivialEqualsAndHashCode(ClassTree classTree, VisitorState state) {
+        boolean hasEquals = false;
+        boolean hasHashCode = false;
+        for (Tree member : classTree.getMembers()) {
+            if (member instanceof MethodTree) {
                 MethodTree methodTree = (MethodTree) member;
 
                 // We want to check if the equals & hashCode methods have actually been overridden (i.e. don't just
@@ -80,10 +94,6 @@ public final class DangerousRecordArrayField extends BugChecker implements BugCh
             }
         }
 
-        if (!hasArrayField || (hasEquals && hasHashCode)) {
-            return Description.NO_MATCH;
-        }
-
-        return buildDescription(classTree).build();
+        return hasEquals && hasHashCode;
     }
 }
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DangerousRecordArrayField.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DangerousRecordArrayField.java
@@ -37,8 +37,8 @@ import com.sun.tools.javac.code.Symbol.ClassSymbol;
 @AutoService(BugChecker.class)
 @BugPattern(
         summary = "Record type has an array field and hasn't overridden equals/hashcode. By default array equality"
-                + " performs reference equality only. Consider using a List, using Immutables, or overriding"
-                + " equals/hashCode.",
+                + " performs reference equality only. Consider using an immutable Collection for the field, using"
+                + " Immutables instead of the record, or overriding equals/hashCode in the record.",
         severity = SeverityLevel.WARNING)
 public final class DangerousRecordArrayField extends BugChecker implements BugChecker.ClassTreeMatcher {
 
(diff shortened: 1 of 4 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DangerousRecordArrayField.java
```diff
@@ -0,0 +1,99 @@
...
+        severity = SeverityLevel.WARNING)
+public final class DangerousRecordArrayField extends BugChecker implements BugChecker.ClassTreeMatcher {
+
+    private static final Matcher<VariableTree> IS_ARRAY_VARIABLE = Matchers.isArrayType();
+    private static final Matcher<MethodTree> NON_TRIVIAL_EQUALS = Matchers.allOf(
+            Matchers.equalsMethodDeclaration(),
+            Matchers.not(Matchers.singleStatementReturnMatcher(Matchers.instanceEqualsInvocation())));
+    private static final Matcher<MethodTree> NON_TRIVIAL_HASHCODE = Matchers.allOf(
+            Matchers.hashCodeMethodDeclaration(),
+            Matchers.not(Matchers.singleStatementReturnMatcher(Matchers.instanceHashCodeInvocation())));
```
carterkozak: I don't think we have test coverage for `Matchers.not(Matchers.singleStatementReturnMatcher(Matchers.instanceHashCodeInvocation()))`. Fwiw I think we may be better off searching for both equals and hashCode implementations, there are enough edge cases in how those may be implemented that existence alone should disarm this check.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DangerousRecordArrayField.java
```diff
@@ -0,0 +1,99 @@
...
+            return Description.NO_MATCH;
+        }
+        if (!hasArrayField(classTree, state)) {
+            return Description.NO_MATCH;
+        }
+        if (hasNonTrivialEqualsAndHashCode(classTree, state)) {
+            return Description.NO_MATCH;
+        }
+
+        return buildDescription(classTree).build();
```
carterkozak, line 62: Perhaps we should suggest replacing `T[]` with `ImmutableList<T>` (with an automated refactor?)

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DangerousRecordArrayField.java
```diff
@@ -0,0 +1,99 @@
...
+import com.sun.source.tree.Tree;
+import com.sun.source.tree.VariableTree;
+import com.sun.tools.javac.code.Symbol.ClassSymbol;
+
+/**
+ * Warns that users should not have a {@link java.util.regex.Pattern} as a key to a Set or Map.
+ */
+@AutoService(BugChecker.class)
+@BugPattern(
+        summary = "Record type has an array field and hasn't overridden equals/hashcode. By default array equality"
```
carterkozak, line 39: We could mention `toString` here as well

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DangerousRecordArrayField.java
raiju: The issue is that `record` implicitly generates an `equals` & `hashCode` method. We can't actually check for presence of the override in user-written code (at least, to my knowledge/from my poking while making this PR).

Because of this we also have test coverage btw, the reason `testArray_withEqualsNoHashcode` fails is this matcher (as `record` generated a trivial hashCode implementation).

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DangerousRecordArrayField.java
raiju: As in, encourage overriding toString? I'm not sure that's a great idea on the average `byte[]` containing record.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DangerousRecordArrayField.java
raiju: I think I disagree for two reasons:
 - Going from e.g. `byte[]` to `ImmutableList<Byte>` has _a lot_ of overheads (both memory & conversion). So I'm not a fan of automatically refactoring. These can end up being quite big (e.g. imagine if an file-serving service ever gets this as an automatic refactor).
 - Restricting to `ImmutableList` (as opposed to `List`), while ensuring the typesystem helps you, also creates a bunch of code noise & extra allocations for very little gain given our programming culture. I don't really ever see people mutating the type of value that get inserted into records.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DangerousRecordArrayField.java
carterkozak: Not necessarily, but it is worth pointing out that array.toString does not provide much information compared to a collection.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DangerousRecordArrayField.java
carterkozak: I imagine the ClassTree can be searched for method overrides, but the ClassSymbol would include the implicit implementations

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DangerousRecordArrayField.java
raiju: Note that `Arrays.asList(Byte[])` ~> `List<Byte>` & `Arrays.asList(byte[])` ~> `List<byte[]>`; to my knowledge there's no jdk-included zero-copy way to transform a byte[] to a List<Byte>.

If we assume guava we could recommend `Bytes.asList` (but that only solves the problem for byte[], not for other types; this also doesn't return an immutable list).

carterkozak: Idea isn't to automatically do the refactor, but to add a suggestion inline to nudge folks toward best practices. I've found it helpful when error-prone checks fail with both a human readable description as well as the text "did you mean `foo.bar(baz)`?"

> Going from e.g. `byte[]` to `ImmutableList<Byte>`

We can provide other specific recommendations for binary data like the conjure `Bytes` type.

> Restricting to ImmutableList (as opposed to List)

I'm not suggesting that we restrict to ImmutableList, rather that it makes sense as a default recommendation.

The idea here is to use the check failure opportunity to share as much context as we can to help folks make the best, most informed decisions.

> I don't really ever see people mutating the type of value that get inserted into records.

It's very hard to see, this sort of thing often becomes a bug when it occurs on the original list after the record has been created and passed elsewhere. I don't think our default stance should be to allow subtle mutability. I agree wholeheartedly that it's often necessary for performance, but I think that should be the exception rather than the rule because debugging issues resulting from unexpected mutations is a nightmare.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DangerousRecordArrayField.java
raiju: Where would you expect that? For example,
```
        System.out.println("For " + classSymbol.getSimpleName());
        System.out.println(" Enclosed elements: "
                + classSymbol.getEnclosedElements().stream()
                        .map(Symbol::flatName)
                        .collect(Collectors.toSet()));
```
results in:
```
For MyRecordWithEquals
 Enclosed elements: [payload, toString, hashCode, name, <init>, equals]
For MyRecordWithNone
 Enclosed elements: [payload, toString, hashCode, name, <init>, equals]
For MyRecordWithNone
 Enclosed elements: [payload, toString, hashCode, name, <init>, equals]
For MyRecordWithHashCode
 Enclosed elements: [payload, toString, hashCode, name, <init>, equals]
```

Note that all of them show up identically, despite all of them missing at least one override.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DangerousRecordArrayField.java
raiju: Note that conjure's `Bytes` type still does a full copy, which is not ideal when the goal was to return a tuple (a classic use-case of `record`).

raiju: I'm happy to see if I can auto-generate an equality/hashCode implementation.

carterkozak: > Note that conjure's Bytes type still does a full copy, which is not ideal when the goal was to return a tuple (a classic use-case of record).

In such cases the equals and hashCode methods are seldom used either, no?

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DangerousRecordArrayField.java
carterkozak: I think you want to check `classTree.getMembers()`, but I haven't had a chance to verify this

carterkozak: It's possible the symbol could also work if the `Flags.SYNTHETIC` modifier is set on the method.

raiju: Yes, this actually does differentiate between the two:
```
        System.out.println(" Members: "
                + classTree.getMembers().stream()
                        .filter(tree -> tree instanceof JCTree.JCMethodDecl)
                        .map(tree -> (JCTree.JCMethodDecl) tree)
                        .map(method -> method.name)
                        .collect(Collectors.toSet()));
                        ```

raiju: Interesting, I had already changed it to using that API, but I think as I went through multiple refactors going from a working to a clean solution I left in a check that was necessary when I was using the ClassSymbol related APIs.

This is fixed now.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DangerousRecordArrayField.java
raiju: Correct; we could also try to detect the usage of equals/hashCode on such records & complain there, but I figured we'd start somewhere.

Separately, in the project that made me realize the issue, the only place where equality was checked was in the unit tests.

carterkozak: Agree it's best to flag the definition rather than call sites, detecting collections which eventually rely on hashCode/equals is tricky business. Happy to merge this as is, I think automatically adding equals/hashCode to records would be great, especially any time we migrate existing code into one of our codebases, but I don't want to block on that. Opens the door to other potential checks to prevent bit-rot, for example ensuring all fields are taken into account in overridden equality and hashCode methods, so that new element additions don't violate the contract.

raiju: Yeah, 100% agree on the equality/hashCode validity question. Will try to think of ways of doing this, but this is a little far out of my comfort zone w/ error-prone.

carterkozak (approved):

PR #2677, opened 2023-11-18
Add `Visitor` to the list of `DO_NOT_IMPORT_INNER` for intellij formatting

MGHawes:
- Putting this up as a suggestion
- Personally I find it very annoying that Intellij auto-imports the inner visitor calls so that you only see:
```java
.map(value -> value.accept(new Visitor<>() {
  ....
```
- It makes it pretty hard to know what is being visited without mousing-over etc.
- This is just the intellij setting so it doesn't ban the usage, just makes it a conscious choice

==COMMIT_MSG==
Add `Visitor` to the list of `DO_NOT_IMPORT_INNER`
==COMMIT_MSG==

## Possible downsides?
Maybe people rely on this? I searched our internal repos and the usage of inline raw visitors is pretty uncommon in general

### DIFF
--- a/changelog/@unreleased/pr-2677.v2.yml
+++ b/changelog/@unreleased/pr-2677.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Add `Visitor` to the list of `DO_NOT_IMPORT_INNER`
+  links:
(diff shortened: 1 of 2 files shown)


carterkozak (approved):
thanks!

PR #2693, opened 2024-01-05
deprecated conjure check

esword:
## Before this PR
No way to enforce removal of references to deprecated-for-removal conjure calls

## After this PR

==COMMIT_MSG==
New errorprone check that will flag usage of deprecated-for-removal conjure endpoints.
==COMMIT_MSG==

### DIFF
--- a/README.md
+++ b/README.md
@@ -230,6 +230,7 @@ Safe Logging can be found at [github.com/palantir/safe-logging](https://github.c
 - `JooqBatchWithoutBindArgs`: jOOQ batch methods that execute without bind args can cause performance problems.
 - `InvocationTargetExceptionGetTargetException`: InvocationTargetException.getTargetException() predates the general-purpose exception chaining facility. The Throwable.getCause() method is now the preferred means of obtaining this information. [(source)](https://docs.oracle.com/en/java/javase/17/docs/api//java.base/java/lang/reflect/InvocationTargetException.html#getTargetException())
 - `PreferInputStreamTransferTo`: Prefer JDK `InputStream.transferTo(OutputStream)` over utility methods such as `com.google.common.io.ByteStreams.copy(InputStream, OutputStream)`, `org.apache.commons.io.IOUtils.copy(InputStream, OutputStream)`, `org.apache.commons.io.IOUtils.copyLong(InputStream, OutputStream)`.
+- `ConjureEndpointDeprecatedForRemoval`: Conjure endpoints marked with Deprecated and `forRemoval = true` should not be used as they will are scheduled to be removed in the next major version.
 
 ### Programmatic Application
 
--- a/README.md
+++ b/README.md
@@ -230,7 +230,7 @@ Safe Logging can be found at [github.com/palantir/safe-logging](https://github.c
 - `JooqBatchWithoutBindArgs`: jOOQ batch methods that execute without bind args can cause performance problems.
 - `InvocationTargetExceptionGetTargetException`: InvocationTargetException.getTargetException() predates the general-purpose exception chaining facility. The Throwable.getCause() method is now the preferred means of obtaining this information. [(source)](https://docs.oracle.com/en/java/javase/17/docs/api//java.base/java/lang/reflect/InvocationTargetException.html#getTargetException())
 - `PreferInputStreamTransferTo`: Prefer JDK `InputStream.transferTo(OutputStream)` over utility methods such as `com.google.common.io.ByteStreams.copy(InputStream, OutputStream)`, `org.apache.commons.io.IOUtils.copy(InputStream, OutputStream)`, `org.apache.commons.io.IOUtils.copyLong(InputStream, OutputStream)`.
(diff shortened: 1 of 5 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DeprecatedConjure.java
```diff
@@ -0,0 +1,70 @@
...
+                + " @SuppressWarnings(\"DeprecatedConjure\").")
+public final class DeprecatedConjure extends BugChecker
+        implements BugChecker.MethodInvocationTreeMatcher, BugChecker.MemberReferenceTreeMatcher {
+
+    private static final String CONJURE_CLIENT_ENDPOINT = "com.palantir.conjure.java.lib.internal.ClientEndpoint";
+
+    private static final Matcher<Tree> FULL_DEPRECATED_CONJURE_ENDPOINT_MATCHER =
+            Matchers.symbolMatcher((symbol, state) -> {
+                return ASTHelpers.hasAnnotation(symbol, CONJURE_CLIENT_ENDPOINT, state)
+                        && symbol.isDeprecatedForRemoval();
```
carterkozak: I think we may want to reverse the order of this check for efficiency -- `ASTHelpers.hasAnnotation(symbol, CONJURE_CLIENT_ENDPOINT, state)` is relatively expensive and uncommon, where `symbol.isDeprecatedForRemoval()` should be a relatively simple bitwise comparison [`(flags_field & DEPRECATED_REMOVAL) != 0`](https://github.com/openjdk/jdk/blob/faa9c6909dda635eb008b9dada6e06fca47c17d6/src/jdk.compiler/share/classes/com/sun/tools/javac/code/Symbol.java#L378-L380).

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DeprecatedConjure.java
```diff
@@ -0,0 +1,70 @@
...
+import com.sun.source.tree.MethodInvocationTree;
+import com.sun.source.tree.Tree;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        severity = BugPattern.SeverityLevel.ERROR,
+        summary = "You should not use Conjure endpoints that are marked for removal as that may block"
+                + " upgrades in the near future. You can explicitly disable this check on a case-by-case basis using"
+                + " @SuppressWarnings(\"DeprecatedConjure\").")
+public final class DeprecatedConjure extends BugChecker
```
carterkozak: Naming is hard. "Deprecated conjure" may seem like conjure is deprecated, and doesn't make it obvious that deprecated-for-removal is more severe than standard deprecation.

While a bit wordy, perhaps something along the lines of `ConjureEndpointDeprecatedForRemoval`?

carterkozak (commented):
Generally lgtm, great tests! Two very minor nits

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DeprecatedConjure.java
esword: Mostly agree, though still unresolved what we should do if client code is using a deprecated-for-removal object or field - makes having the term "EndPoint" in the name a little confusing.  Not a huge deal, or we may choose to make that a separate check if we do implement it.

carterkozak: >  if client code is using a deprecated-for-removal object or field

Only endpoints support this flag, objects and fields aren't deprecated-for-removal

README.md
```diff
@@ -230,6 +230,7 @@ Safe Logging can be found at [github.com/palantir/safe-logging](https://github.c
 - `JooqBatchWithoutBindArgs`: jOOQ batch methods that execute without bind args can cause performance problems.
 - `InvocationTargetExceptionGetTargetException`: InvocationTargetException.getTargetException() predates the general-purpose exception chaining facility. The Throwable.getCause() method is now the preferred means of obtaining this information. [(source)](https://docs.oracle.com/en/java/javase/17/docs/api//java.base/java/lang/reflect/InvocationTargetException.html#getTargetException())
 - `PreferInputStreamTransferTo`: Prefer JDK `InputStream.transferTo(OutputStream)` over utility methods such as `com.google.common.io.ByteStreams.copy(InputStream, OutputStream)`, `org.apache.commons.io.IOUtils.copy(InputStream, OutputStream)`, `org.apache.commons.io.IOUtils.copyLong(InputStream, OutputStream)`.
+- `ConjureEndpointDeprecatedForRemoval`: Conjure endpoints marked with Deprecated and `forRemoval = true` should not be used as they will are scheduled to be removed in the next major version.
```
carterkozak: Typo in `they will are scheduled`

Let's edit this down to something like:

> Conjure endpoints marked with Deprecated and `forRemoval = true` should not be used as they are scheduled to be removed.

At this layer we can make no guarantees about major versions, some products provide stronger guarantees than others.

esword: done

carterkozak (approved):

PR #2699, opened 2024-01-08
fix #2668: Methods must take into account supertypes

### LINKED ISSUE #2668: [IllegalSafeLoggingArgument] Prevent subclasses from overriding methods to be less safe

carterkozak:
See #2668
==COMMIT_MSG==
Methods must take into account supertypes
==COMMIT_MSG==


bjlaub (approved):
lgtm

PR #2703, opened 2024-01-19
Safety propagation takes into account known subtypes

carterkozak:
==COMMIT_MSG==
Safety propagation takes into account known subtypes
==COMMIT_MSG==

## Possible downsides?
Rollout is tricky

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/MoreASTHelpers.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/MoreASTHelpers.java
@@ -22,13 +22,19 @@ import com.google.errorprone.VisitorState;
 import com.google.errorprone.fixes.SuggestedFixes;
 import com.google.errorprone.util.ASTHelpers;
 import com.sun.source.tree.CatchTree;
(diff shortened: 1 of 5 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/MoreASTHelpers.java
```diff
@@ -123,5 +129,39 @@ public static Type getResultType(Tree tree) {
                 : ASTHelpers.getType(tree);
     }
 
+    public static boolean isSealed(@Nullable ClassSymbol classSymbol) {
+        if (classSymbol == null) {
+            return false;
+        }
+        long flags = classSymbol.flags();
+        return (flags & (1L << 62)) != 0;
```
bjlaub, line 137: are there JDK docs for this?

carterkozak: Yes, in `com.sun.tools.javac.code.Flags`, but only in newer releases, where we target jdk-11

bmarcaur (dismissed):

bmarcaur (approved):

PR #2705, opened 2024-01-20
Publish plugin marker artifact for baseline-error-prone plugin

pkoenig10:
This allows the plugin to be used with the [Gradle plugin DSL](https://docs.gradle.org/current/userguide/plugins.html#sec:plugins_block).


carterkozak (approved):
No harm in listing it, however some plugins don't work very well with the plugins dsl due to inter-plugin dependencies. The DSL creates an isolated classloader for each plugin, so it cannot interact directly with classes from another plugin, where the `classpath '<plugin artifact>'` model can. Ideally we could support the dsl model across the board, but it would take a fair bit of gradle plugin refactoring to get there... iirc most of the issues are related to coupling with sls-packaging.

PR #2708, opened 2024-01-30
Prevent Unnecessary Propagation When Using Array Containers

bmarcaur:
## Before this PR
The tests have extremely clear examples, but TLDR when the return type of a method was `UnsafeType[]` it would cause that method to propagate its safety information. This differs from returning `UnsafeType` and `List<UnsafeType>`.

## After this PR
`Array` containers are treated more similarly to `Collection` types.

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/safety/SafetyAnnotations.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/safety/SafetyAnnotations.java
@@ -33,6 +33,7 @@ import com.sun.tools.javac.code.Symbol.VarSymbol;
 import com.sun.tools.javac.code.SymbolMetadata;
 import com.sun.tools.javac.code.TargetType;
 import com.sun.tools.javac.code.Type;
+import com.sun.tools.javac.code.Type.ArrayType;
 import com.sun.tools.javac.code.Type.WildcardType;
 import com.sun.tools.javac.util.List;
 import java.util.HashSet;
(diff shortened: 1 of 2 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/safety/SafetyAnnotations.java
```diff
@@ -108,6 +107,20 @@ public static Safety getSafety(Tree tree, VisitorState state) {
         return Safety.mergeAssumingUnknownIsSame(symbolSafety, treeTypeSafety, symbolTypeSafety);
     }
 
+    private static Safety getTypeSafety(Type treeType, VisitorState state) {
+        if (treeType == null) {
+            return Safety.UNKNOWN;
+        }
+
+        // The type of an array is not stored as a tree, but rather an adjacent type
+        if (treeType instanceof Type.ArrayType) {
```
carterkozak, line 116: Can we handle this in the same place we take care of other wrapper types (e.g. Optional, Collection)?

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/SafeLoggingPropagationTest.java
```diff
@@ -1037,6 +1037,49 @@ void testIgnoresOutOfScopeReturns() {
...
+                        "Test.java",
+                        "import com.palantir.logsafe.*;",
+                        "public final class Test {",
+                        "  @Unsafe",
+                        "  private static class UnsafeType {}",
+                        "  public UnsafeType[] getType() { return new UnsafeType[]{ new UnsafeType() }; }",
+                        "}")
+                .expectUnchanged()
+                .doTest();
+    }
```
carterkozak, line 1066: Great tests!

bmarcaur: Thanks ❤️

bmarcaur:
Gonna re-push to a non-branch, https://github.com/palantir/gradle-baseline/pull/2709

PR #2709, opened 2024-01-30
Prevent Unnecessary Propagation When Using Array Containers

bmarcaur:
## Before this PR
The tests have extremely clear examples, but TLDR when the return type of a method was `UnsafeType[]` it would cause that method to propagate its safety information. This differs from returning `UnsafeType` and `List<UnsafeType>`.

## After this PR
`Array` containers are treated more similarly to `Collection` types.

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/safety/SafetyAnnotations.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/safety/SafetyAnnotations.java
@@ -205,6 +205,12 @@ public final class SafetyAnnotations {
             }
         }
         Safety typeArgumentCombination = SAFETY_IS_COMBINATION_OF_TYPE_ARGUMENTS.getSafety(type, state, dejaVu);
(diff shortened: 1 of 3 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/safety/SafetyAnnotations.java
```diff
@@ -205,6 +205,12 @@ private static Safety getSafetyInternal(Type type, VisitorState state, Set<Strin
             }
         }
         Safety typeArgumentCombination = SAFETY_IS_COMBINATION_OF_TYPE_ARGUMENTS.getSafety(type, state, dejaVu);
+        // Arrays are difficult to pipe through the above combiner without a large refactor, since the AST
+        // is not quite a tree it stores its type information adjacent to the node.
+        if (type instanceof Type.ArrayType) {
```
bmarcaur, line 210: IMPL based on feedback in the previous PR https://github.com/palantir/gradle-baseline/pull/2708#discussion_r1472018766

carterkozak (approved):
Fantastic, thank you!

PR #2711, opened 2024-02-01
Add deprecation warning to ./gradlew idea task

### LINKED ISSUE #2437: Add note to try using native idea gradle integration instead of the idea task
## What happened?

I've seen several users (myself included) ask about the `idea` task in internal chat channels and be told that the native IntelliJ integration with gradle is superior to using baseline's `idea` task.  Many developers have `gw openIdea` burned into their memory though, and it continues to mostly work fine, so people don't know to try the native integration.

As a way of spreading awareness, I suggest outputting a message to console when running the `idea` task that native gradle integration is preferred.  It might look something like this:

```
$ ./gradlew idea
<snip>
Note: Baseline now recommends IntelliJ's native gradle integration instead of the 'idea' task.  
      Use it by opening the folder directly with IntelliJ (run `idea .`) instead of `./gradlew openIdea`.
      See https://www.jetbrains.com/help/idea/gradle.html#gradle_import_project_start
$
```

I think this can be done around here: https://github.com/palantir/gradle-baseline/blob/ef871ad9d2a608a1790758f96bac9472836e907c/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy#L86

crogoz:
Fixes https://github.com/palantir/gradle-baseline/issues/2437

## After this PR

==COMMIT_MSG==
Add deprecation warning to ./gradlew idea task
==COMMIT_MSG==

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
@@ -47,7 +47,6 @@ import org.gradle.plugins.ide.idea.model.ModuleDependency
 // TODO(dfox): separate the xml manipulation (which really benefits from groovy syntax) from typed things
 //@CompileStatic
 class BaselineIdea extends AbstractBaselinePlugin {
-
     void apply(Project project) {
         this.project = project
 
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
@@ -24,9 +24,6 @@ import com.palantir.baseline.plugins.javaversions.ChosenJavaVersion
 import com.palantir.baseline.util.GitUtils
 import groovy.transform.CompileStatic
 import groovy.xml.XmlUtil
-import org.gradle.api.logging.Logger
-import org.gradle.api.logging.Logging
-
 import java.nio.file.Files
 import java.nio.file.Path
 import java.nio.file.Paths
@@ -50,8 +47,6 @@ import org.gradle.plugins.ide.idea.model.ModuleDependency
 // TODO(dfox): separate the xml manipulation (which really benefits from groovy syntax) from typed things
 //@CompileStatic
 class BaselineIdea extends AbstractBaselinePlugin {
-
-    private static final Logger log = Logging.getLogger(BaselineIdea.class);
 
     void apply(Project project) {
         this.project = project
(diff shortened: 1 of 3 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
```diff
@@ -80,7 +80,12 @@ class BaselineIdea extends AbstractBaselinePlugin {
             }
         }
 
-        project.getTasks().named("idea").configure(idea -> idea.doLast(cleanup))
+        project.getTasks().named("idea").configure(idea -> {
+            idea.doFirst(_t -> project.getLogger().warn("DEPRECATED: Using `./gradlew idea` is no longer recommended, some " +
+                    "functionality may not function as expected. Instead, we suggest opening the project directly in " +
+                    "IntelliJ or running `idea .` (if the JetBrains utility is installed) for a smoother experience."))
```
CRogers: I think we need to tell people exactly how to get out of their situation, which is to:

1. Close the intellij project
2. Run `./gradlew cleanIdea`
3. Run `rm -rf .idea || true`
4. Open project in IntelliJ or use `idea .` if installed by JetBrains Toolbox.

Maybe say they only need to do step 4 for new projects?

crogoz: updated in https://github.com/palantir/gradle-baseline/pull/2711/commits/2d7e707d279332ad6250f006859f9630337a2c8f

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineIdea.groovy
```diff
@@ -80,7 +80,19 @@ class BaselineIdea extends AbstractBaselinePlugin {
             }
         }
 
-        project.getTasks().named("idea").configure(idea -> idea.doLast(cleanup))
+        project.getTasks().named("idea").configure(idea -> {
+            idea.doFirst(_t -> project.getLogger().warn("""
+                DEPRECATED: Using `./gradlew idea` is no longer recommended, some functionality may not function as expected.
```
crogoz, line 85: We still have a lot of occurrences in our docs that use `./gradlew idea` (https://github.com/search?q=org%3Apalantir+%22.%2Fgradlew+idea%22&type=code). Should we update these before merging?

CRogers: We probably should, but that is a serious amount of manual work.

CRogers (dismissed):

CRogers:
c'mon changelog-bot

CRogers:
:+1:

PR #2717, opened 2024-02-07
Ignore `org.immutables:value::annotations` when adding Immutable incremental arg

pkoenig10:
https://github.com/palantir/conjure-java-runtime-api/pull/1095 broke consumers who did not themselves use the Immutables annotation processor.

```
warning: The following options were not recognized by any processor: '[immutables.gradle.incremental]'
1 error
1 warning
```

This is not a regression of https://github.com/palantir/gradle-baseline/pull/2465 - the test added in that PR still passes.

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineImmutables.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineImmutables.java
@@ -21,7 +21,9 @@ import java.util.Collections;
 import java.util.Objects;
 import org.gradle.api.Plugin;
 import org.gradle.api.Project;
-import org.gradle.api.artifacts.Dependency;
+import org.gradle.api.artifacts.component.ComponentIdentifier;
+import org.gradle.api.artifacts.component.ModuleComponentIdentifier;
+import org.gradle.api.artifacts.result.ResolvedComponentResult;
 import org.gradle.api.tasks.SourceSet;
 import org.gradle.api.tasks.SourceSetContainer;
 import org.gradle.api.tasks.compile.JavaCompile;
@@ -74,12 +76,22 @@ public final class BaselineImmutables implements Plugin<Project> {
         return project
                 .getConfigurations()
                 .getByName(sourceSet.getAnnotationProcessorConfigurationName())
-                .getAllDependencies()
+                .getIncoming()
+                .getResolutionResult()
+                .getAllComponents()
                 .stream()
                 .anyMatch(BaselineImmutables::isImmutablesValue);
     }
 
-    private static boolean isImmutablesValue(Dependency dependency) {
-        return Objects.equals(dependency.getGroup(), "org.immutables") && Objects.equals(dependency.getName(), "value");
+    private static boolean isImmutablesValue(ResolvedComponentResult component) {
+        ComponentIdentifier id = component.getId();
+
+        if (!(id instanceof ModuleComponentIdentifier)) {
+            return false;
+        }
+
+        ModuleComponentIdentifier moduleId = (ModuleComponentIdentifier) id;
+
+        return Objects.equals(moduleId.getGroup(), "org.immutables") && Objects.equals(moduleId.getModule(), "value");
     }
 }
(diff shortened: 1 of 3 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineImmutables.java
```diff
@@ -76,22 +74,12 @@ private static boolean hasImmutablesProcessor(Project project, SourceSet sourceS
...
-
-        if (!(id instanceof ModuleComponentIdentifier)) {
-            return false;
-        }
-
-        ModuleComponentIdentifier moduleId = (ModuleComponentIdentifier) id;
-
-        return Objects.equals(moduleId.getGroup(), "org.immutables") && Objects.equals(moduleId.getModule(), "value");
+    private static boolean isImmutablesValue(Dependency dependency) {
+        return Objects.equals(dependency.getGroup(), "org.immutables") && Objects.equals(dependency.getName(), "value");
```
CRogers: If people bring in the actual immutables processor jar (not `::annotations` classifier) transitively and it ends up being used we should still actually apply this? Couldn't we fix this by keeping the code looking at transitives but making sure one of the components has a group `org.immutables`, artifactId `value` _and_ no classifier? Might have to use the `artifacts` rather than `components` api from Gradle to do this from memory (so you can get the classifiers)?

pkoenig10: > Couldn't we fix this by keeping the code looking at transitives but making sure one of the components has a group org.immutables, artifactId value and no classifier?

This is what I tried to do at first, but I couldn't figure out how to get at the classifier here. Although it is just more complexity.

If https://github.com/palantir/conjure-java-runtime-api/pull/1099 doesn't fix the issue I'll play around with this a bit more.

Although I have to imagine that bringing in the immutables processor as a transitive dependencies and then using it is pretty unlikely. Honestly, if you're using an annotation processor and not declaring an explicit dependency, then you have no right to complain if your build is slow because we didn't enable incremental compilation for you.

CRogers: I _swear_ I remember seeing the horror of some annotation processor produce an immutables class that was then compiled by immutables... It was a long time ago, sense may have prevailed.

CRogers: > you have no right to complain if your build is slow because we didn't enable incremental compilation for you

I agree with the sentiment but the problem is it just silently becomes slower with no indication why... and devs often just suffer in silence. I guess this is telling me we should probably build something that makes a loud log line or fails the build when it detects an non-incremental annotation processor somehow (in an ideal world).

CRogers: I updated it with the version that checks the classifier using the old artifact APIs (Gradle still does not have a configuration cacheable api for maven classifiers 😢: https://github.com/gradle/gradle/issues/23702).

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineImmutables.java
```diff
@@ -76,22 +76,27 @@ private static boolean hasImmutablesProcessor(Project project, SourceSet sourceS
...
         if (!(id instanceof ModuleComponentIdentifier)) {
             return false;
         }
 
         ModuleComponentIdentifier moduleId = (ModuleComponentIdentifier) id;
 
-        return Objects.equals(moduleId.getGroup(), "org.immutables") && Objects.equals(moduleId.getModule(), "value");
+        // The actual annotation processor jar has no classifier, we must make sure not to match on the
+        // `annotations` jar which has the `annotations` classifier
+        boolean noClassifier = resolvedArtifact.getClassifier() == null;
```
CRogers, line 96: Should probably have a test for this...

pkoenig10 (commented):
👍 

Confirmed that this fixes the issue, thanks for the help @CRogers!

CRogers:
:+1:

schlosna (approved):

PR #2734, opened 2024-02-27
Remove old gradle-junit-reports code

CRogers:
## Before this PR
We have made a better version of this plugin that can be found at https://github.com/palantir/gradle-failure-reports/.

This was previously not even applied any more since #2355.

## After this PR
==COMMIT_MSG==
Remove gradle-junit-reports.
==COMMIT_MSG==

## Possible downsides?
Removing the code could be called a break - only two things actually add `gradle-junit-reports` as a direct dep internally, and both those are no longer used.


crogoz (dismissed):

CRogers:
:+1:

crogoz (approved):

PR #2736, opened 2024-02-29
turn off "Insert imports for inner classes"

iamdanfox:
## Before this PR

Our internal ski product database tooling does a lot of codegen and creates many inner classes called `Key`. Today, I find that I often want to reference a few of these classes (e.g. `MyTable.Key key = ...` and `OtherTable.Key key2 = ...`) but IntelliJ _immediately_ simplifies the first import into `Key key = ...` which then trips over the `BadImport` static analysis which complaints about commonly-used names being ambiguous.

I have started just unticking this box:

image

## After this PR
==COMMIT_MSG==
turn off "Insert imports for inner classes"
==COMMIT_MSG==

### DIFF
--- a/.baseline/idea/intellij-java-palantir-style.xml
+++ b/.baseline/idea/intellij-java-palantir-style.xml
@@ -30,7 +30,7 @@
             <package name="" static="false" withSubpackages="true" />
           </value>
         </option>
-        <option name="INSERT_INNER_CLASS_IMPORTS" value="true" />
+        <option name="INSERT_INNER_CLASS_IMPORTS" value="false" />
         <GroovyCodeStyleSettings>
           <option name="CLASS_COUNT_TO_USE_IMPORT_ON_DEMAND" value="999" />
           <option name="NAMES_COUNT_TO_USE_IMPORT_ON_DEMAND" value="999" />
--- a/changelog/@unreleased/pr-2736.v2.yml
+++ b/changelog/@unreleased/pr-2736.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: turn off "Insert imports for inner classes"
+  links:
+  - https://github.com/palantir/gradle-baseline/pull/2736


carterkozak:
fwiw I strongly prefer not to import inner classes as well, since the outer class name should provide a fair bit of context (more so than a package provides for a top-level class name). Other folks may have different opinions though, so I'll leave this up for another day or so in case others have feedback.

LouisCaubet:
Main reason we want this is because some other part of the tooling makes compilation fail when inner classes are imported that way:

Screenshot 2024-02-29 at 17 32 03

pkoenig10:
My default is also to prefer avoiding inner imports, but I think this is hugely context dependent.

For example, [AtlasDB](https://github.com/palantir/atlasdb)'s generated table classes contains all row, column, and column value for that table and the table name is repeated in each of the types. For example, here's a real set of imports from an internal product - and this is not uncommon:

```java
import com.palantir.mp.atlas.generated.ClientsTable;
import com.palantir.mp.atlas.generated.ClientsTable.AllowedClientsIdxTable;
import com.palantir.mp.atlas.generated.ClientsTable.AllowedClientsIdxTable.AllowedClientsIdxColumn;
import com.palantir.mp.atlas.generated.ClientsTable.AllowedClientsIdxTable.AllowedClientsIdxRow;
import com.palantir.mp.atlas.generated.ClientsTable.ClientsNamedColumn;
import com.palantir.mp.atlas.generated.ClientsTable.ClientsRow;
import com.palantir.mp.atlas.generated.ClientsTable.ClientsRowResult;
import com.palantir.mp.atlas.generated.ClientsTable.OrganizationIdToOrganizationClientIdIdxTable;
import com.palantir.mp.atlas.generated.ClientsTable.OrganizationIdToOrganizationClientIdIdxTable.OrganizationIdToOrganizationClientIdIdxColumn;
import com.palantir.mp.atlas.generated.ClientsTable.OrganizationIdToOrganizationClientIdIdxTable.OrganizationIdToOrganizationClientIdIdxRow;
import com.palantir.mp.atlas.generated.ClientsTable.OrganizationIdToPersonalClientIdIdxTable;
import com.palantir.mp.atlas.generated.ClientsTable.OrganizationIdToPersonalClientIdIdxTable.OrganizationIdToPersonalClientIdIdxColumn;
import com.palantir.mp.atlas.generated.ClientsTable.OrganizationIdToPersonalClientIdIdxTable.OrganizationIdToPersonalClientIdIdxRow;
import com.palantir.mp.atlas.generated.ClientsTable.OwnerUserIdToPersonalClientIdIdxTable;
import com.palantir.mp.atlas.generated.ClientsTable.OwnerUserIdToPersonalClientIdIdxTable.OwnerUserIdToPersonalClientIdIdxRow;
```

I'd much prefer to leave this option to users. IntelliJ makes it easy to configure this globally, so you only have to disable inner import or define a set of exclusions once. If our tooling overrides the global setting, then I have to manually re-enable this for ever repo I work on. In general, I'd much prefer we give people the flexibility to configure their development environments how they would like rather than imposing a particular workflow.

iamdanfox:
ok - shall I just delete the xml line entirely then?

PR #2742, opened 2024-03-14
override remover fixer

esword:
this is a sample for an errorprone fixer to break connections between a class that implements a conjure interface and methods that are about to be removed from the interface.  

More than likely this should not go into baseline.  Just doing it here for now as a reference

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/conjure/RemoveLater.java
+++ b/baseline-error-prone/src/main/java/com/palantir/conjure/RemoveLater.java
@@ -1,0 +1,28 @@
+/*
+ * (c) Copyright 2024 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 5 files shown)


baseline-error-prone/src/main/java/com/palantir/conjure/RemoveLater.java
```diff
@@ -0,0 +1,28 @@
...
+package com.palantir.conjure;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.CLASS)
+@Target(ElementType.METHOD)
+public @interface RemoveLater {
```
esword, line 26: is there a standard annotation we could use?

esword:
experimental.  moved to https://github.palantir.build/gotham/gotham-conjure-validator

PR #2745, opened 2024-03-15
Promote BaselineJavaVersionExtension.library in README

pkoenig10:
The vast majority of the time that repos explicitly configure Java versions in a subproject, they are doing so because they want a subproject to be considered a library. Using the explicit Java version configuration to accomplish this makes Java version upgrades harder to automate.

This PR updates the README to include documentation for `library()` and to promote it in favor of configuring the versions explicitly.


carterkozak (approved):

carterkozak:
Thanks!

PR #2746, opened 2024-03-16
Suppress more checks in Gradle plugin projects

pkoenig10:
Disable the `CatchBlockLogException` and `Slf4jConstantLogMessage` checks in Gradle plugins.

The Gradle [`Logger`](https://docs.gradle.org/current/javadoc/org/gradle/api/logging/Logger.html) interface extends the slf4j [`Logger`](https://www.slf4j.org/api/org/slf4j/Logger.html) interface which triggers these checks when certain log levels are used.


felixdesouza:
👍

PR #2748, opened 2024-03-18
Ignore intra-project dependencies in checkUnusedDependencies

pkoenig10:
## Before this PR

The `checkUnusedDependenciesTest` task does not work with [Gradle test fixtures](https://docs.gradle.org/current/userguide/java_testing.html#sec:java_test_fixtures) if the test code does not itself use the test fixture.

The `test` configurations [extend from](https://github.com/gradle/gradle/blob/85d30969f4672bb2739550b4de784910a6810b7a/subprojects/plugins/src/main/java/org/gradle/api/plugins/JavaPlugin.java#L416-L419) the `main` configuration, so this "dependency" is not visible to `checkUnusedDependenciesTest`. But the `test` configurations do not extend from the `testFixtures` configurations. Instead the dependency between the `test` and `testFixtures` is handled as a normal dependency, which is visible to `checkUnusedDependenciesTest`.

This PR is motivated by a code generation tool we are building where we want to generate test utilities into the `testFixture` source set. It will not be uncommon for the test classes in the project producing the generated code to not use these utilities and we want to avoid causing `checkUnusedDependenciesTest` failures.

## After this PR

Dependencies to artifacts produced by the same project are ignored by the `checkUnusedDependencies` task.

This seems like a simple way to accomplish this goal with relatively little downside. I don't think there are many other cases that would result in intra-project dependencies like this, and if there are we probably would want to ignore them also.

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineExactDependencies.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineExactDependencies.java
@@ -182,6 +182,10 @@ public final class BaselineExactDependencies implements Plugin<Project> {
                             task.setSourceClasses(sourceSet.getOutput().getClassesDirs());
                             task.getDependenciesConfigurations().add(explicitCompile);
 
+                            // ignore intra-project dependencies, which are typically added automatically for things
+                            // like test fixtures
+                            task.ignore(project.getGroup().toString(), project.getName());
+
                             // this is liberally applied to ease the Java8 -> 11 transition
                             task.ignore("javax.annotation", "javax.annotation-api");
 
@@ -278,7 +282,7 @@ public final class BaselineExactDependencies implements Plugin<Project> {
                     .append(projectComponentId.getProjectPath())
                     .append("')");
             if (withName) {
-                builder.append(" (").append(artifact.getId().getDisplayName()).append(")");
+                builder.append(" <-- ").append(artifact.getName());
             }
             return builder.toString();
         }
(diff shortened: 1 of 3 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineExactDependencies.java
```diff
@@ -278,7 +282,7 @@ private static String asDependencyString(ResolvedArtifact artifact, boolean with
                     .append(projectComponentId.getProjectPath())
                     .append("')");
             if (withName) {
-                builder.append(" (").append(artifact.getId().getDisplayName()).append(")");
+                builder.append(" <-- ").append(artifact.getName());
```
pkoenig10, line 285: Drive by to improve upon this output added in https://github.com/palantir/gradle-baseline/pull/948. The current output is verbose and has confused users.

**Before**

```
Execution failed for task ':foo:checkUnusedDependenciesMain'.
> Found 1 dependencies unused during compilation, please delete them from 'foo/build.gradle' or choose one of the suggested fixes:
        project(':bar') (main (project :bar))
```

**After**

```
Execution failed for task ':foo:checkUnusedDependenciesMain'.
> Found 1 dependencies unused during compilation, please delete them from 'foo/build.gradle' or choose one of the suggested fixes:
        project(':bar') <-- main
```

pkoenig10:
I have no idea why these tests are failing. They don't fail for me locally and them seemingly have nothing to do with the change in this PR.

carterkozak (approved):

PR #2749, opened 2024-03-18
Check runtime classpath contains compatible jars

CRogers:
## Before this PR
Internally, we had an incident where as part of our Java 17 library upgrade, a product that cannot yet take Java 17 (but will soon) had a Java 17 transitive dep get upgraded. This caused it to fail when deployed.

They had `runtime = 17`. Had they had `runtime = 11`, we should have hopefully caught this issue.

## After this PR
==COMMIT_MSG==
Check the runtime classpath contains jar that contain only class which are compatible with the intended runtime java language version.
==COMMIT_MSG==

We iterate through all the jars and error out if there are jars which have versions greater than the configured runtime can support.

Example redacted failure from internal product on the bad release:

```
The runtime classpath has the following jars which contain classes that have too high a java language version. We're expecting the bytecode major version to be no more than 55 for java language version 11. Examples classes in each jar:
  
  /Users/callumr/.gradle/caches/modules-2/files-2.1/com.palantir.libA/libA/2.31.0/202452ab4311b23029de9049dbf4caea317d8d5d/libA-2.31.0.jar: com/palantir/libA/EventDefinition.class has bytecode major version 61
  /Users/callumr/.gradle/caches/modules-2/files-2.1/com.palantir.libA/libA-config/2.31.0/6c1e409d8082140df0dcf5d981c1a31f017cce56/libA-config-2.31.0.jar: com/palantir/libA/config/EventConfig$Builder.class has bytecode major version 61
  /Users/callumr/.gradle/caches/modules-2/files-2.1/com.palantir.libA/libA-dialogue/2.31.0/c71a0368595a18aca427963dd2b4a734fce83ffa/libA-dialogue-2.31.0.jar: com/palantir/libA/api/DialogueEventConsumerEndpoints$1.class has bytecode major version 61
  /Users/callumr/.gradle/caches/modules-2/files-2.1/com.palantir.libA/libA-objects/2.31.0/c6232283d30f4c256ab711c3fad3da7dafa5e373/libA-objects-2.31.0.jar: com/palantir/libA/api/Event$Builder.class has bytecode major version 61
  /Users/callumr/.gradle/caches/modules-2/files-2.1/com.palantir.libB/libB/0.31.0/bd9b09e503f2d4234c5d8b6bd5ae185d5da3c0c7/libB-0.31.0.jar: com/palantir/libB/Thing.class has bytecode major version 61

```

## Possible downsides?
I think this may block some baseline upgrades, but the errors are probably worth looking into. We can use our internal error collection to find all these cases.


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/javaversions/BaselineJavaVersion.java
```diff
@@ -86,7 +86,17 @@ public void execute(CheckJavaVersionsTask task) {
                             task.getRuntimeVersion().set(extension.runtime());
                         }
                     });
-            project.getTasks().named("check").configure(check -> check.dependsOn(checkJavaVersions));
+
+            TaskProvider<CheckClasspathCompatible> checkRuntimeClasspathCompatible = project.getTasks()
+                    .register("checkRuntimeClasspathCompatible", CheckClasspathCompatible.class, task -> {
```
CRogers, line 91: Was originally going to do this for compileClasspath too, but I think I'd rather try and just roll out runtime first.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/javaversions/CheckClasspathCompatible.java
```diff
@@ -0,0 +1,118 @@
...
+import java.util.stream.Stream;
+import org.gradle.api.DefaultTask;
+import org.gradle.api.file.ConfigurableFileCollection;
+import org.gradle.api.provider.Property;
+import org.gradle.api.tasks.Classpath;
+import org.gradle.api.tasks.Console;
+import org.gradle.api.tasks.Input;
+import org.gradle.api.tasks.TaskAction;
+
+public abstract class CheckClasspathCompatible extends DefaultTask {
```
felixdesouza, line 35: auto parallelizable is unused here?

felixdesouza:
is there no way to use `org.gradle.jvm.version` and set that as a variant attribute? it seems like it'd do what you want (that said, I haven't actually dabbled with this myself so YMMV)

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/javaversions/CheckClasspathCompatible.java
CRogers: Couldn't get it to work in this repo 😢. Couldn't work out why. This should still have all the benefits as I have written in "good" way (I think?).

CRogers:
You'd think that's what happens automatically when you just build your project. But it evidently doesn't always work for transitives who are then bumped up via a constraint (via GCV).

felixdesouza:
oof, that kinda sucks

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/javaversions/CheckClasspathCompatible.java
felixdesouza: weird! I think you still have the deps on it though? should be good to go without it

felixdesouza:
👍

felixdesouza:
👍

PR #2752, opened 2024-03-19
Handle relocated multi-release class files in checkRuntimeClasspathCompatible

pkoenig10:
See https://github.com/gradle/gradle/issues/24515

The `checkRuntimeClasspathCompatible` check added in https://github.com/palantir/gradle-baseline/pull/2749 produces false positives when using `gradleApi()`.

```
The runtime classpath has the following jars which contain classes that have too high a java language version. We're expecting the bytecode major version to be no more than 61 for java language version 17. Examples classes in each jar:

/home/circleci/.gradle/caches/8.5/generated-gradle-jars/gradle-api-8.5.jar: org/gradle/internal/impldep/META-INF/versions/19/com/fasterxml/jackson/core/io/doubleparser/FastDoubleSwar.class has bytecode major version 63

* Causal chain is:
	org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':my-gradle-plugin:checkRuntimeClasspathCompatible'.
	java.lang.RuntimeException: The runtime classpath has the following jars which contain classes that have too high a java language version. We're expecting the bytecode major version to be no more than 61 for java language version 17. Examples classes in each jar:

/home/circleci/.gradle/caches/8.5/generated-gradle-jars/gradle-api-8.5.jar: org/gradle/internal/impldep/META-INF/versions/19/com/fasterxml/jackson/core/io/doubleparser/FastDoubleSwar.class has bytecode major version 63
```

### DIFF
--- a/changelog/@unreleased/pr-2752.v2.yml
+++ b/changelog/@unreleased/pr-2752.v2.yml
@@ -1,0 +1,5 @@
+type: fix
+fix:
+  description: Handle relocated multi-release class files in checkRuntimeClasspathCompatible
+  links:
+  - https://github.com/palantir/gradle-baseline/pull/2752
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/javaversions/CheckClasspathCompatible.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/javaversions/CheckClasspathCompatible.java
@@ -76,7 +76,7 @@ public abstract class CheckClasspathCompatible extends DefaultTask {
                     .flatMap(entry -> {
                         // We don't care about higher versions of classes in multi-release jars as JVMs will only
                         // load classes from here that match or are higher than their current version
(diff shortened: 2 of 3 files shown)


pkoenig10:
Since this seems like a Gradle problem, I'm happy to do something a bit more targeted here. Although I wouldn't be too surprised if there are other libraries that do the same thing.

CRogers:
Thanks for the fix - one of my tasks for today is to check using the papaya data that this hasn't blocked too many baseline upgrades. Just gonna add a test for this case.

CRogers:
Thanks for the fix - one of my tasks for today is to check using the excavator data that this hasn't blocked too many baseline upgrades. Just gonna add a test for this case.

CRogers:
Funnily enough, the jrengelman shadow plugin does the right thing here and moves multi-release jar classes to `META-INF/versions/x/relocated/package/`

CRogers:
:+1:

carterkozak (approved):

CRogers:
Looked at the internal data, seems this was the only error across all our repos for this new task.

PR #2772, opened 2024-04-02
Check eclipse collection api factory

Shibi-bala:
## Before this PR

We've seen this deadlock issue on a service where this package was the root cause. Adding an errorprone rule for this.

Full description of issue is here: https://github.com/palantir/atlasdb/pull/7073

## After this PR

==COMMIT_MSG==
==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/EclipseCollectionsApiFactoryUsage.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/EclipseCollectionsApiFactoryUsage.java
@@ -1,0 +1,58 @@
+/*
+ * (c) Copyright 2024 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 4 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/EclipseCollectionsApiFactoryUsage.java
```diff
@@ -0,0 +1,58 @@
...
+import com.sun.source.tree.ImportTree;
+import com.sun.tools.javac.code.Type;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        summary = "Import of type uses non-canonical name",
+        severity = BugPattern.SeverityLevel.ERROR,
+        documentSuppression = false)
+public final class EclipseCollectionsApiFactoryUsage extends BugChecker implements BugChecker.ImportTreeMatcher {
+    private static final String BAD_ECLIPSE_COLLECTIONS_USAGE_SUBSTRING = "org.eclipse.collections.api.factory.";
```
Shibi-bala, line 36: I can't find an errorprone rule that deals with libraries we don't like. I suppose because that is pretty opinionated. 

Anyway, decided against making this check too generic (supporting other bad imports and fixes) since there would be a lot of abstraction written for one case (so far) to be able to support various ways of fixing issues.

PR #2776, opened 2024-04-08
Allow safe implementations of unsafe classes

pkoenig10:
## Before this PR

The `IllegalSafeLoggingArgument` check allows an `@Unsafe` subtype of `@Safe` type, which is can be dangerous.

```java
// This is allowed

@Safe
interface Foo {}

@Unsafe
class Bar implements Foo {}

// Oops, we will now safe log an unsafe value
Foo getFoo() {
    return new Bar();
}
```

Conversely, the `IllegalSafeLoggingArgument` check does not allow a `@Safe` subtype of an `@Unsafe` type, even though this should be safe.

```java
// This is not allowed

@Unsafe
interface Foo {}

@Safe
class Bar implements Foo {}
```

This just seems like a bug in the `IllegalSafeLoggingArgument` check.

## After this PR

Inverts the ancestor check in `IllegalSafeLoggingArgument` to:
- Disallow an `@Unsafe` subtype of `@Safe` type 
- Allow a `@Safe` subtype of `@Unsafe` type

This appears to be a subset of the functionality in https://github.com/palantir/gradle-baseline/pull/2699.

## Possible downsides?

Is there something I'm not understanding about this check?

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/IllegalSafeLoggingArgument.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/IllegalSafeLoggingArgument.java
@@ -349,7 +349,7 @@ public final class IllegalSafeLoggingArgument extends BugChecker
             return Description.NO_MATCH;
         }
         Safety ancestorSafety = SafetyAnnotations.getTypeSafetyFromAncestors(tree, state);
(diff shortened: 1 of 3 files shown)


carterkozak:
Both are dangerous — I have a pair of PRs up which resolve this

PR #2783, opened 2024-04-22
Gradle JDK automanagement integration for toolchains 

crogoz:
## After this PR

Part of the Gradle JDK automanagement workflow see: https://github.com/palantir/gradle-jdks/issues/333
Required by: https://github.com/palantir/gradle-jdks/pull/341

Adding a `javaVersions` property `setupJdkToolchains` (defaulting to enabled = old workflow). When disabled (via gradle-jdks), the `BaselineJavaVersion` plugin will set the custom `javaCompile/javaLauncher/javaDocTool` based on:  https://docs.gradle.org/current/userguide/toolchains.html#toolchains_for_tasks

==COMMIT_MSG==
Gradle JDK automanagement integration for toolchains 
==COMMIT_MSG==


gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineJavaVersionIntegrationTest.groovy
```diff
@@ -341,6 +343,51 @@ class BaselineJavaVersionIntegrationTest extends IntegrationSpec {
...
     }
 
+    def '#gradleVersionNumber: jdkToolchainsAutoManagement only sets the language versions'() {
+        when:
+        // language=gradle
+        buildFile << '''
+        javaVersions {
+            libraryTarget = 11
+            runtime = 21
+            jdkToolchainsAutoManagement = true
```
CRogers: This generally looks good, but I thinks this is a bit confusing to me, as you are enabling something which actually ends up disabling code in this plugin (which also happens to be about managing jdkToolchains). I'd probably call this something like `setupJdkToolchains = false` instead.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/javaversions/BaselineJavaVersionsExtension.java
```diff
@@ -126,4 +129,19 @@ public final void jdks(LazyJdks lazyJdks) {
...
     }
+
+    /**
+     * Overrides toolchain configuration using the new gradle-jdks workflow.
+     * If the property is set to true, then the plugin will not attempt to configure toolchains. The toolchains were
+     * already configured using the new gradle-jdks workflow by setting the following properties:
+     * "org.gradle.java.installations.paths", "org.gradle.java.installations.auto-download" and
+     * "org.gradle.java.installations.auto-detect".
+     */
+    public final Property<Boolean> jdkToolchainsAutoManagement() {
```
CRogers: I'd ignore the existing style here and prefix this with `get`. Gradle special cases this and you do not need to write the `set` method below to get this style of gradle script code:

```gradle
jdkToolchainsAutoManagement = false
```

Fwiw, I actually despise prefixing "normal" java with `get` or `set` - it's usually unnecessary and makes me remember to the "java bean ask-don't-tell" years - but with Gradle we need to add them for extensions/tasks/artifact transforms/generally all Gradle managed types.

gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineJavaVersionIntegrationTest.groovy
```diff
@@ -343,14 +343,14 @@ class BaselineJavaVersionIntegrationTest extends IntegrationSpec {
...
-    def '#gradleVersionNumber: jdkToolchainsAutoManagement only sets the language versions'() {
+    def '#gradleVersionNumber: when setupJdkToolchains=false no toolchains are configured by gradle-baseline'() {
         when:
         // language=gradle
         buildFile << '''
         javaVersions {
             libraryTarget = 11
             runtime = 21
-            jdkToolchainsAutoManagement = true
+            setupJdkToolchains = true
```
CRogers, line 355: shouldn't this be false?

crogoz: good spot, yes, it should. I realized that these test actually use the auto-detected jdks instead of using the `jdk-latest` plugin, which make my test successful regardless of whether the flag is set or not. Fixing it now.

gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineJavaVersionIntegrationTest.groovy
```diff
@@ -341,6 +344,86 @@ class BaselineJavaVersionIntegrationTest extends IntegrationSpec {
...
+                languageVersion = JavaLanguageVersion.of(21)
+                vendor = JvmVendorSpec.ADOPTIUM
+            }
+        }
+        '''.stripIndent(true)
+        file('src/main/java/Main.java') << java11CompatibleCode
+        File compiledClass = new File(projectDir, "build/classes/java/main/Main.class")
+
+        then:
+        ExecutionResult result = runTasksSuccessfully('compileJava', 'run')
```
CRogers, line 417: I normally put this in `when:` and all the setup stuff before `when:`, just make this a bit more readable/immediately see what this is running.

CRogers (dismissed):

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/javaversions/BaselineJavaVersion.java
```diff
@@ -101,127 +101,150 @@ public void execute(CheckJavaVersionsTask task) {
...
+            JavaToolchains baselineConfiguredJavaToolchains,
+            BaselineJavaVersionsExtension rootExtension,
+            JavaToolchainService javaToolchainService) {
+
+        project.getTasks().withType(JavaCompile.class).configureEach(javaCompileTask -> {
+            setJavaCompiler(
+                    javaCompileTask, rootExtension, baselineConfiguredJavaToolchains, javaToolchainService, target);
+            javaCompileTask.getOptions().getCompilerArgumentProviders().add(new EnablePreviewArgumentProvider(target));
+            // Set sourceCompatibility to opt out of '-release', allowing opens/exports to be used.
+            javaCompileTask.doFirst(task -> ((JavaCompile) task)
```
CRogers: TaskActions must be written using `new Action` rather than lambas, otherwise Gradle will make this task never up-to-date. It can't get a serialised name for a lambda, but can for a task. There is an errorprone for this, but errorprone does not run in here as for some reason it's using a groovy source.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/javaversions/BaselineJavaVersion.java
```diff
@@ -101,127 +101,150 @@ public void execute(CheckJavaVersionsTask task) {
...
-                            options.setSource(target.get().javaLanguageVersion().toString());
-                        }
-                    }
-                });
-            }
+        project.getTasks().withType(Javadoc.class).configureEach(javadocTask -> {
+            setJavaDocTool(javadocTask, rootExtension, baselineConfiguredJavaToolchains, javaToolchainService, target);
+
+            // javadocTask doesn't allow us to add a CommandLineArgumentProvider, so we do it just in time
+            javadocTask.doFirst(task -> {
```
CRogers: Ditto.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/javaversions/BaselineJavaVersion.java
```diff
@@ -101,127 +101,150 @@ public void execute(CheckJavaVersionsTask task) {
...
+                    // yes, javadoc truly takes a single-dash where everyone else takes a double dash
+                    options.addBooleanOption("-enable-preview", true);
+                    options.setSource(target.get().javaLanguageVersion().toString());
+                }
+            });
         });
 
         // checkstyle.getJavaLauncher() was added in Gradle 7.5
         if (GradleVersion.current().compareTo(GradleVersion.version("7.5")) >= 0) {
-            project.getTasks().withType(Checkstyle.class).configureEach(new Action<Checkstyle>() {
```
CRogers, line 148: configureEach/stuff that isn't a task action or a CommandLineArgumentsProvider are fine to have as a lambda though

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/javaversions/BaselineJavaVersion.java
```diff
@@ -101,127 +101,150 @@ public void execute(CheckJavaVersionsTask task) {
...
+                    .getJavaLauncher()
+                    .set(getJavaLauncher(
+                            rootExtension, baselineConfiguredJavaToolchains, javaToolchainService, target));
+            groovyCompileTask
+                    .getOptions()
+                    .getCompilerArgumentProviders()
+                    .add(new EnablePreviewArgumentProvider(target));
+
+            // Set sourceCompatibility to opt out of '-release', allowing opens/exports to be used.
+            groovyCompileTask.doFirst(task -> ((GroovyCompile) task)
```
CRogers: Another task action lambda

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/javaversions/BaselineJavaVersion.java
crogoz: oh, that is such a hidden behaviour :(  Will change all of these back to new Action()

crogoz: so then we need to use Action here as well: https://github.com/palantir/gradle-baseline/pull/2783/files#diff-bd31167bf1fd132bca9a139faa8fc09fb838bc5067c668f99b116be494a0e8b6R104 right ? as this is a taskAction

CRogers: Yep, anything given to doFirst/doLast.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/javaversions/BaselineJavaVersion.java
```diff
@@ -88,11 +99,18 @@ public void execute(CheckJavaVersionsTask task) {
...
             TaskProvider<CheckClasspathCompatible> checkRuntimeClasspathCompatible = project.getTasks()
-                    .register("checkRuntimeClasspathCompatible", CheckClasspathCompatible.class, task -> {
-                        task.getClasspathName().set("runtime");
-                        task.getJavaVersion().set(extension.runtime());
-                        task.getClasspath().setFrom(project.getConfigurations().getByName("runtimeClasspath"));
-                    });
+                    .register(
+                            "checkRuntimeClasspathCompatible",
+                            CheckClasspathCompatible.class,
+                            new Action<CheckClasspathCompatible>() {
```
CRogers: Register is fine for lambas, it's only task actions/CommandLineArgumentProviders that need the lambda

CRogers (approved):

PR #2790, opened 2024-04-30
Document how to enable checkUnusedDependencies

ash211:
## Before this PR
I mistakenly thought that com.palantir.baseline-exact-dependencies being applied (by default) meant that that `check` would run `checkUnusedDependencies`.  Actually it does not, and repos wanting this behavior need to opt into it.

## After this PR
==COMMIT_MSG==
Document how to enable checkUnusedDependencies
==COMMIT_MSG==

This PR documents how repos can do that, so it's an easy copy+paste if desired.

## Possible downsides?
None known

### DIFF
--- a/README.md
+++ b/README.md
@@ -386,6 +386,16 @@ checkImplicitDependencies {
 }
 ```
 
+These tasks are created by default, but are not configured to run as part of `check`.  To do that, add to the root `/build.gradle`:
+
+```gradle
+allprojects {
+    apply plugin: 'java-library'
+
+    tasks.check.dependsOn checkUnusedDependencies, checkImplicitDependencies
+}
+```
+
 ## com.palantir.baseline-encoding
 
 This plugin sets the encoding for JavaCompile tasks to `UTF-8`.


README.md
```diff
@@ -386,6 +386,16 @@ checkImplicitDependencies {
 }
 ```
 
+These tasks are created by default, but are not configured to run as part of `check`.  To do that, add to the root `/build.gradle`:
+
+```gradle
+allprojects {
+    apply plugin: 'java-library'
```
ash211, line 393: Without this line, my internal test repo fails with:

```
-----------
* Where:
Build file '/Volumes/git/my-project/build.gradle' line: 77

* What went wrong:
A problem occurred evaluating root project 'my-project'.
> Could not get unknown property 'checkUnusedDependencies' for root project 'my-project' of type org.gradle.api.Project.

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
```

Maybe there's an alternate pattern that we should recommend instead, of only applying to subprojects that are Java type?

pkoenig10: I probably wouldn't be prescriptive in this documentation about putting this in `allprojects` - that's something that consumers can decide for themselves.

I'd just write something like:

```diff
+ tasks.check.dependsOn tasks.checkUnusedDependencies
```

PR #2791, opened 2024-05-01
Reduce memory usage of `checkClasspathCompatible`

CRogers:
## Before this PR
A project internally was running out of heap in CI. We did a JFR and determined that `JarFile` and the sheer number of `InflaterInputStream`s it creates was using up gigabytes of space. `DataInputStream` was also allocating a lot (it allocates 240+ bytes [the moment you create it](https://github.com/openjdk/jdk17/blob/4afbcaf55383ec2f5da53282a1547bac3d099e9d/src/java.base/share/classes/java/io/DataInputStream.java#L57-L61)).

## After this PR
Use `JarInputStream` instead, which does not appear to have the same issue.

==COMMIT_MSG==
Reduce memory usage of `checkClasspathCompatible`
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-2791.v2.yml
+++ b/changelog/@unreleased/pr-2791.v2.yml
@@ -1,0 +1,5 @@
+type: fix
+fix:
+  description: Reduce memory usage of `checkClasspathCompatible`
+  links:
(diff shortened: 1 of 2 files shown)


felixdesouza (approved):
Thanks!

PR #2802, opened 2024-05-17
Add gradle daemon target to BaselineJavaVersionsExtension

crogoz:
## After this PR

Required for the Gradle Automanagement work.  All the java versions requested by the `BaselineJavaVersionsExtension` and the `BaselineJavaVersionExtension` will be materialized as files inside the `gradle/jdks/*` directory. The gradle daemon version will be materialized in the file`gradle/gradle-daemon-jdk-version`. WIP PR here: https://github.com/palantir/gradle-jdks/compare/cr/patching2...cr/patchin2-generate-gradle?expand=1

==COMMIT_MSG==
Add gradle daemon target to BaselineJavaVersionsExtension
==COMMIT_MSG==


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/javaversions/BaselineJavaVersionsExtension.java
```diff
@@ -77,6 +82,14 @@ public final void setLibraryTarget(String value) {
         libraryTarget.set(version.javaLanguageVersion());
     }
 
+    public final Property<JavaLanguageVersion> getDaemonTarget() {
```
CRogers, line 85: I don't think this should live here - imo it should live in gradle-jdks in it's own extension (`jdks`) as it's not needed in baseline at all. It should also be possible to use gradle-jdks without baseline-java-versions - it's not at the moment, but with the conclusion of the new work we should be able to remove the dependency.

crogoz: I was thinking that for consistency it would be easier to have all the java versions in the same extension

CRogers: I can see how that would be nice but I think it still logically makes sense as part of gradle-jdks. We've done similar things in the past (palantir-java-format config in the spotless part of baseline) and ended up having to unwind it later, which was kinda painful.

PR #2804, opened 2024-05-22
Adding TestFactory as a method that ignores DesignForExntesion along other test methods

htaan6:
## Before this PR

Have to manually suppress DesignForExtension every time.

## After this PR

==COMMIT_MSG==
Adding [TestFactory](https://junit.org/junit5/docs/5.4.1/api/org/junit/jupiter/api/TestFactory.html) (for DynamicTests) as part of the methods that ignore this annotation.

==COMMIT_MSG==

### DIFF
--- a/.baseline/checkstyle/checkstyle.xml
+++ b/.baseline/checkstyle/checkstyle.xml
@@ -406,7 +406,7 @@
             <property name="switchBlockAsSingleDecisionPoint" value="true"/>
         </module>
         <module name="DesignForExtension"> <!-- Java Coding Guidelines: Design for extension -->
-            <property name="ignoredAnnotations" value="ParameterizedTest, Test, Before, BeforeEach, After, AfterEach, BeforeClass, BeforeAll, AfterClass, AfterAll"/>
+            <property name="ignoredAnnotations" value="ParameterizedTest, Test, Before, BeforeEach, After, AfterEach, BeforeClass, BeforeAll, AfterClass, AfterAll, TestFactory"/>
         </module>
         <module name="JavadocMethod"> <!-- Java Style Guide: Where Javadoc is used -->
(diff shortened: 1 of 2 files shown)


fawind (dismissed):

fawind (approved):

.baseline/checkstyle/checkstyle.xml
```diff
@@ -406,7 +406,7 @@
             <property name="switchBlockAsSingleDecisionPoint" value="true"/>
         </module>
         <module name="DesignForExtension"> <!-- Java Coding Guidelines: Design for extension -->
-            <property name="ignoredAnnotations" value="ParameterizedTest, Test, Before, BeforeEach, After, AfterEach, BeforeClass, BeforeAll, AfterClass, AfterAll"/>
+            <property name="ignoredAnnotations" value="ParameterizedTest, Test, Before, BeforeEach, After, AfterEach, BeforeClass, BeforeAll, AfterClass, AfterAll, TestFactory"/>
```
pkoenig10, line 409: FYI you can't modify this file. It's managed by baseline and will just be reverted when baseline upgrades.

https://github.com/palantir/gradle-baseline/pull/2807

See [the README](https://github.com/palantir/gradle-baseline#compalantirbaseline-checkstyle) for how to add custom suppressions.

fawind: Hmpf, good flag, sorry should have noticed this!

Think it's still reasonable to add this to the template checkstyle.xml if multiple people run into this? But would need to modify the template here:
https://github.com/palantir/gradle-baseline/blob/ff8ac41def94d235721c39639ce0c2ceaf1ccd4f/gradle-baseline-java-config/resources/checkstyle/checkstyle.xml#L427

PR #2806, opened 2024-05-29
Remove ClassInitializationDeadlock

pkoenig10:
This is in preparation to upgrade ErrorProne. We can't upgrade until we remove this check and upgrade Baseline in this repo because the `ClassInitializationDeadlock` check conflicts with the identically named check contributed upstream in https://github.com/google/error-prone/issues/2062.

Trying to compile with Error Prone 2.27.0 produces:

```
java.lang.IllegalArgumentException: Cannot combine scanner suppliers with different implementations of 'ClassInitializationDeadlock': com.palantir.baseline.errorprone.ClassInitializationDeadlock, com.google.errorprone.bugpatterns.ClassInitializationDeadlock
```

In a follow up PR I will upgrade Baseline and Error Prone so we once again have a `ClassInitializationDeadlock` check.


fawind (approved):
Nice that this got accepted back upstream 🎉

PR #2808, opened 2024-05-30
Upgrade Error Prone

pkoenig10:
See https://github.com/palantir/gradle-baseline/pull/2806


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/CatchSpecificity.java
```diff
@@ -213,7 +213,7 @@ public Void visitInstanceOf(InstanceOfTree instanceOfNode, Void ignored) {
 
                                 @Override
                                 public Void visitParenthesized(ParenthesizedTree node, Void ignored) {
-                                    return node.getExpression().accept(this, ignored);
+                                    return node.getExpression().accept(this, null);
```
pkoenig10, line 216: See https://errorprone.info/bugpattern/VoidUsed

fawind (approved):

PR #2813, opened 2024-06-05
cache key for JarClassHasher includes artifact classifier

bjlaub:
## Before this PR
Previously, we used the `ModuleVersionIdentifier` associated with a `ResolvedArtifact` as the key for this cache, which contains sha256 hashes of each class file within a jar. Unfortunately this is not sufficient for distinguishing artifacts with the same maven coordinate, but different classifiers, which may contain completely different content within the jars.

This leads to problems when two artifacts from the same maven coordinate but different classifiers are on the classpath, wherein `checkClassUniqueness` potentially reports incorrect results as it may rely on hashes cached from a different artifact altogether (but for the same `ModuleVersionIdentifier`).

As a practical example (and where this issue was discovered), the coordinate `com.google.cloud.bigdataoss:gcs-connector:hadoop3-2.2.19` contains both a "normal" jar artifact, and one with the `shaded` classifier. The latter includes classes that may conflict with other dependencies on the classpath. If `JarClassHasher` hashes and subsequently caches the classfiles from the "normal" jar first, when we encounter the "shaded" jar we will erroneously used the cached values from the "normal" jar for the shaded one when considering classpath conflicts.

## After this PR
This change includes the artifact classifier in the cache key, to distinguish between artifacts with the same `ModuleVersionIdentifier` but different classifiers, which are actually different artifacts.

==COMMIT_MSG==
cache key for JarClassHasher includes artifact classifier
==COMMIT_MSG==

## Possible downsides?
This will possibly cause `checkClassUniqueness` to fail on consumers that may have committed lockfiles under the old behavior, where we potentially missed possible classpath conflicts due to the incorrect caching behavior.

### DIFF
--- a/changelog/@unreleased/pr-2813.v2.yml
+++ b/changelog/@unreleased/pr-2813.v2.yml
@@ -1,0 +1,5 @@
+type: fix
+fix:
+  description: cache key for JarClassHasher includes artifact classifier
+  links:
+  - https://github.com/palantir/gradle-baseline/pull/2813
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/services/ClassUniquenessArtifactIdentifier.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/services/ClassUniquenessArtifactIdentifier.java
@@ -1,0 +1,29 @@
+/*
+ * (c) Copyright 2024 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.baseline.services;
+
+import java.util.Optional;
+import org.gradle.api.artifacts.ModuleVersionIdentifier;
+import org.immutables.value.Value;
+
+@Value.Immutable
+@Value.Style(optionalAcceptNullable = true)
+public interface ClassUniquenessArtifactIdentifier {
+    ModuleVersionIdentifier moduleVersionIdentifier();
+
+    Optional<String> classifier();
+}
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/services/JarClassHasher.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/services/JarClassHasher.java
@@ -42,14 +42,8 @@ import org.slf4j.Logger;
 
 public abstract class JarClassHasher implements BuildService<BuildServiceParameters.None>, AutoCloseable {
 
-    @Value.Immutable
-    interface CacheKey {
-        ModuleVersionIdentifier moduleVersionIdentifier();
-
-        String classifier();
-    }
-
-    private final Cache<CacheKey, Result> cache = Caffeine.newBuilder().build();
+    private final Cache<ClassUniquenessArtifactIdentifier, Result> cache =
+            Caffeine.newBuilder().build();
 
     public static final class Result {
         private final ImmutableSetMultimap<String, HashCode> hashesByClassName;
@@ -68,7 +62,7 @@ public abstract class JarClassHasher implements BuildService<BuildServiceParamet
     }
 
     public final Result hashClasses(ResolvedArtifact resolvedArtifact, Logger logger) {
-        CacheKey key = ImmutableCacheKey.builder()
+        ClassUniquenessArtifactIdentifier key = ImmutableClassUniquenessArtifactIdentifier.builder()
                 .moduleVersionIdentifier(resolvedArtifact.getModuleVersion().getId())
                 .classifier(Strings.nullToEmpty(resolvedArtifact.getClassifier()))
                 .build();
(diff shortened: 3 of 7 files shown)


felixdesouza:
I think this makes sense, but I haven't really looked inside this class at all. 

I think a lot of the complexity of 1-1 mapping of artifacts goes away here with artifact transforms. Which as an added bonus, can be made incremental. In any case, this is a strict improvement.

Each artifact (which corresponds to `ResolvedArtifact`) will end up being transformed into its classes, and then you can transform them again to produce a file with hashes in. It will all be at the unit of a `ResolvedArtifact`. That would be the safest way to do it. 

I'm not sure how we get around the corrupted cache issue however as it's not something we can easily fix without automatically adding entries to the lock file (which is undesirable).

In a similar vein, do we not want it to also take into consideration `ResolvedArtifact#extension` ?

bjlaub:
FLUP: we should include the classifier in the conflicting maven coordinate names within the lockfile.

bjlaub:
> In a similar vein, do we not want it to also take into consideration `ResolvedArtifact#extension` ?

maybe, but i have no evidence at the moment that we'd have e.g. two artifacts with the same `ModuleVersionIdentifier` but different extensions, _that both contain class files_, that would cause a problem here. I'm inclined to maybe address that in a separate PR if/when it comes up.

bjlaub:
I've pushed a change to include the classifier in the lock state. Here's what that looks like:
```
[com.google.cloud.bigdataoss:gcs-connector (classifier=shaded), com.google.cloud.bigdataoss:gcsio]
  - com.google.cloud.hadoop.gcsio.authorization.AuthorizationHandler
[com.google.cloud.bigdataoss:gcs-connector (classifier=shaded), com.google.cloud.bigdataoss:util-hadoop]
  - com.google.cloud.hadoop.util.AccessTokenProvider
  - com.google.cloud.hadoop.util.AccessTokenProvider$AccessToken
  - com.google.cloud.hadoop.util.AccessTokenProvider$AccessTokenType
[com.google.cloud.bigdataoss:gcs-connector (classifier=shaded), com.google.cloud.bigdataoss:util]
  - com.google.cloud.hadoop.util.AccessBoundary
  - com.google.cloud.hadoop.util.AccessBoundary$Action
  - com.google.cloud.hadoop.util.AutoValue_AccessBoundary
[com.google.cloud.bigdataoss:gcs-connector, com.google.cloud.bigdataoss:gcs-connector (classifier=shaded)]
  - com.google.cloud.hadoop.fs.gcs.AutoValue_SyncableOutputStreamOptions
  - com.google.cloud.hadoop.fs.gcs.AutoValue_SyncableOutputStreamOptions$1
  - com.google.cloud.hadoop.fs.gcs.AutoValue_SyncableOutputStreamOptions$Builder
  - com.google.cloud.hadoop.fs.gcs.CoopLockFsck
  - com.google.cloud.hadoop.fs.gcs.CoopLockFsckRunner
  - com.google.cloud.hadoop.fs.gcs.CoopLockFsckRunner$1
  - com.google.cloud.hadoop.fs.gcs.CoopLockFsckRunner$2
  - com.google.cloud.hadoop.fs.gcs.FileSystemDescriptor
  - com.google.cloud.hadoop.fs.gcs.FsBenchmark
  - com.google.cloud.hadoop.fs.gcs.GoogleHadoopFS
  - com.google.cloud.hadoop.fs.gcs.GoogleHadoopFSInputStream
  - com.google.cloud.hadoop.fs.gcs.GoogleHadoopFileSystem
  - com.google.cloud.hadoop.fs.gcs.GoogleHadoopFileSystemBase
  - com.google.cloud.hadoop.fs.gcs.GoogleHadoopFileSystemBase$1
  - com.google.cloud.hadoop.fs.gcs.GoogleHadoopFileSystemBase$GcsFileChecksum
  - com.google.cloud.hadoop.fs.gcs.GoogleHadoopFileSystemBase$GcsFileChecksumType
  - com.google.cloud.hadoop.fs.gcs.GoogleHadoopFileSystemBase$GlobAlgorithm
  - com.google.cloud.hadoop.fs.gcs.GoogleHadoopFileSystemBase$OutputStreamType
  - com.google.cloud.hadoop.fs.gcs.GoogleHadoopFileSystemConfiguration
  - com.google.cloud.hadoop.fs.gcs.GoogleHadoopOutputStream
  - com.google.cloud.hadoop.fs.gcs.GoogleHadoopSyncableOutputStream
  - com.google.cloud.hadoop.fs.gcs.HadoopConfigurationProperty
  - com.google.cloud.hadoop.fs.gcs.HadoopCredentialConfiguration
  - com.google.cloud.hadoop.fs.gcs.InMemoryGlobberFileSystem
  - com.google.cloud.hadoop.fs.gcs.SyncableOutputStreamOptions
  - com.google.cloud.hadoop.fs.gcs.SyncableOutputStreamOptions$Builder
  - com.google.cloud.hadoop.fs.gcs.auth.AbstractDelegationTokenBinding
  - com.google.cloud.hadoop.fs.gcs.auth.AbstractDelegationTokenBinding$TokenSecretManager
  - com.google.cloud.hadoop.fs.gcs.auth.DelegationTokenIOException
  - com.google.cloud.hadoop.fs.gcs.auth.GcsDelegationTokens
  - com.google.cloud.hadoop.fs.gcs.auth.GcsDtFetcher
```

(the above was taken from a project that included both the "normal" and `shaded` classifier artifacts for `gcs-connector`, which actually was previously not being caught by this task before the most recent change).

bjlaub:
I'm confident that this is an important fix for correctness reasons (we are not accurately capturing class name conflicts in some situations currently); but we might want to pause rolling this out until we can take stock of how it will affect some internal projects which I know currently have broken lockfiles, and figure out a strategy for managing that first.

CRogers:
This looks fine but I'd add a test that checks this behaviour so we don't regress in the future.

bjlaub:
> This looks fine but I'd add a test that checks this behaviour so we don't regress in the future.

@CRogers added a test here: https://github.com/palantir/gradle-baseline/pull/2813/commits/849b75a929813f7b7c1f97e906d57bedd60fd721

CRogers:
:+1:

PR #2814, opened 2024-06-05
Fix #2812

### LINKED ISSUE #2812: bug in StringBuilderConstantParameters fixer
## What happened?

Correctly spots a `StringBuilderConstantParameters` here, but suggests an incorrect fix:

```java
new StringBuilder(foo)
  .append("_")
  .append(bar)
  .toString()
  .toLowerCase();
```

Suggestion:

````java
foo + "_" + bar.toLowerCase();
````

````java
(foo + "_" + bar).toLowerCase();
````

### PR DESCRIPTION
**Pull request by nakulj**
## Before this PR
Tests miss the case when a method is called on the built string, see https://github.com/palantir/gradle-baseline/issues/2812

## After this PR
Parens are now surround the append result, ensuring that the post-build method call is applied to the entire built string, not just the last element.

## Possible downsides?
Adds parens even when unnecessary. These can be caught and fixed by [UnnecessaryParentheses](https://errorprone.info/bugpattern/UnnecessaryParentheses), but ideally they'd only be added if necessary.

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StringBuilderConstantParameters.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StringBuilderConstantParameters.java
@@ -95,6 +95,8 @@ public final class StringBuilderConstantParameters extends BugChecker
         return buildDescription(tree)
                 .setMessage(MESSAGE)
                 .addFix(SuggestedFix.builder()
+                        .prefixWith(tree, "(")
+                        .postfixWith(tree, ")")
                         .replace(
                                 tree,
                                 Streams.concat(
--- a/baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/StringBuilderConstantParametersTests.java
+++ b/baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/StringBuilderConstantParametersTests.java
@@ -312,6 +312,30 @@ public class StringBuilderConstantParametersTests {
     }
 
     @Test
+    public void suggestedFixHandlesMethodCalledOnBuilt() {
+        RefactoringValidator.of(StringBuilderConstantParameters.class, getClass())
+                .addInputLines(
+                        "Test.java",
+                        "class Test {",
+                        "   String f() {",
+                        "       return new StringBuilder()",
+                        "           .append(\"foo\")",
+                        "           .append(\"bar\")",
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by palantirtech** [comment]
Thanks for your interest in [palantir/gradle-baseline](https://github.com/palantir/gradle-baseline), @nakulj! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

File: baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StringBuilderConstantParameters.java [hunk 1]
```diff
@@ -95,6 +95,8 @@ public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState
         return buildDescription(tree)
                 .setMessage(MESSAGE)
                 .addFix(SuggestedFix.builder()
+                        .prefixWith(tree, "(")
+                        .postfixWith(tree, ")")
```
pkoenig10, line 99: I don't think it's desirable to unconditionally wrap the suggested fix with parenthesis. This will result in unidiomatic code in the common case.

pkoenig10 (changes requested):

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StringBuilderConstantParameters.java
nakulj: Do you have any suggestions on how to check if the parens are necessary?

pkoenig10: Sorry, I don't have any suggestions. I suspect this is fairly non-trivial.

I'm not sure it's worthwhile to fix this issue - I imagine it would significantly complicated the logic here.

nakulj: Yeah, it seems pretty hard.
That said it seems to me that unidiomatic code is preferable to incorrect code- which is what the fixer currently suggests, especially since [UnnecessaryParentheses](https://errorprone.info/bugpattern/UnnecessaryParentheses) will automatically catch and fix the unidiomatic generated code.

PR #2815, opened 2024-06-06
Upgrade Error Prone

pkoenig10:
In an internal repo, we're seeing `compileJava` tasks hang indefinitely. Looking at the JFRs we see:
- A ton of `TreePath.Result` errors
- A ton of `TreePath` allocations inside `SafeLoggingPropagation`

I suspect the Error Prone upgrade in https://github.com/palantir/gradle-baseline/pull/2808 might be the cause. Before I spend time investigating further. I'd like to upgrade Error Prone to the latest and see if that helps. It appears there were some [fixes for crashes in the latest release](https://github.com/google/error-prone/releases/tag/v2.28.0).

Screenshot 2024-06-06 at 10 56 42 AM

Screenshot 2024-06-06 at 10 52 26 AM

Screenshot 2024-06-06 at 10 55 43 AM


pkoenig10:
After further investigation, the real fix is in https://github.com/palantir/gradle-baseline/pull/2816.

Still can't hurt to upgrade Error Prone though.

carterkozak (approved):

PR #2816, opened 2024-06-07
Return ConditionalTransferResult for conditional nodes in SafetyPropagationTransfer

pkoenig10:
# Problem statement

In an internal repo we experienced a massive slowdown in compile times after adding a single if block with some instanceof pattern-matching to a function with some existing similarly shaped if blocks. The code looked something like:

```java
if (input.foo() instanceof RedFoo foo
    && foo.bar() instanceof BlueBar bar
    && bar.baz() instanceof GreenBaz baz) {
    return baz;
}
```

After adding this code our full compile times when from about 2 minutes to 12 minutes due to excessive creation of `TreeResult` values and GC thrashing. Taking a JFR it was clear that `SafetyPropagationTransfer` was the culprit.

There's a fairly extensive internal thread we're I've recorded my progress while debugging this, so I'm not going to repeat everything here. I do have a minimal repro with the logging that I used to help highlight the problem. This commit also includes a visualization of the CFG of the repro, which I found useful to get a sense for what was happening.

**The repro is in this commit:** https://github.com/palantir/gradle-baseline/commit/4b0286989f894a08765c349fc8c9adec8557954a

# Changes in this PR

As best I can tell, the issue is that we are not properly returning `ConditionalTransferResult` for conditional CFG nodes. This causes problems because when performing an analysis, we may have a `TransferResult` with separate `then` and `else` stores, where the `then` store contains the computed safety result for a pattern-matched variable (the relevant checkerframework code is [here](https://github.com/eisop/checker-framework/blob/c08538fbc8e167a931cbe405614b52663a8e67be/dataflow/src/main/java/org/checkerframework/dataflow/analysis/ForwardAnalysisImpl.java#L469-L554)). But we end up dropping this computed result due to the `leastUpperBound` behavior of `TransferInputResult.getRegularStore()`, which intersects the `then` and `else` stores.

This PR changes the behavior of `SafetyPropagationTransfer` to return `ConditionalTransferResult` for conditional CFG nodes.

Like most of the code in `SafetyPropagationTransfer`, I mimicked the behavior of `AbstractNullnessPropagationTransfer` which does something very similar. I'm not sure why we ever diverged from the `AbstractNullnessPropagationTransfer`  behavior.

After making this change, the test case from repro compiles quickly and without ever triggering an analysis in `getCapturedLocalVariableSafety`.

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/safety/SafetyPropagationTransfer.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/safety/SafetyPropagationTransfer.java
@@ -74,6 +74,7 @@ import javax.annotation.Nullable;
 import javax.lang.model.element.VariableElement;
 import javax.lang.model.type.TypeMirror;
 import org.checkerframework.errorprone.dataflow.analysis.Analysis;
+import org.checkerframework.errorprone.dataflow.analysis.ConditionalTransferResult;
 import org.checkerframework.errorprone.dataflow.analysis.ForwardAnalysisImpl;
 import org.checkerframework.errorprone.dataflow.analysis.ForwardTransferFunction;
 import org.checkerframework.errorprone.dataflow.analysis.RegularTransferResult;
@@ -447,6 +448,7 @@ public final class SafetyPropagationTransfer implements ForwardTransferFunction<
         }
     }
 
+    @CheckReturnValue
     private static TransferResult<Safety, AccessPathStore<Safety>> noStoreChanges(
             Safety value, TransferInput<?, AccessPathStore<Safety>> input) {
         return new RegularTransferResult<>(value, input.getRegularStore());
@@ -457,6 +459,12 @@ public final class SafetyPropagationTransfer implements ForwardTransferFunction<
             Safety value, TransferInput<?, AccessPathStore<Safety>> input, ReadableUpdates updates) {
         ResultingStore newStore = updateStore(input.getRegularStore(), updates);
         return new RegularTransferResult<>(value, newStore.store, newStore.storeChanged);
+    }
+
+    @CheckReturnValue
+    private static TransferResult<Safety, AccessPathStore<Safety>> conditionalResult(
+            Safety value, AccessPathStore<Safety> thenStore, AccessPathStore<Safety> elseStore, boolean storeChanged) {
+        return new ConditionalTransferResult<>(value, thenStore, elseStore, storeChanged);
     }
 
     @CheckReturnValue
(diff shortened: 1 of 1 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/safety/SafetyPropagationTransfer.java
```diff
@@ -746,41 +754,50 @@ public TransferResult<Safety, AccessPathStore<Safety>> visitGreaterThan(
...
+        // '!a' in source is safe, regardless of 'a'.
+        return conditionalResult(
+                Safety.SAFE,
+                /* thenStore= */ input.getElseStore(),
+                /* elseStore= */ input.getThenStore(),
+                /*
+                 * Weird case: We swap the contents of the THEN and ELSE stores without otherwise modifying
+                 * them. Presumably that can still count as a change?
+                 */
+                !input.getThenStore().equals(input.getElseStore()));
```
carterkozak, line 800: I don't think this counts as a change, iirc the changed flag means we've updated the safety of a symbol that has already been tracked. For example

```java
Object object = null; // `object` is currently safe because null is safe.
object = getUnsafeValue(); // This changes the safety of object from safe to unsafe, so storeChanged=true is required
```

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/safety/SafetyPropagationTransfer.java
```diff
@@ -1124,17 +1141,20 @@ private Safety getTypeConversionSafety(
...
-            return updateRegularStore(Safety.SAFE, input, updates);
+            thenUpdates.set(bindingVariables.get(0), safety);
         }
-        // Otherwise types themselves are generally safe, boolean results of type checks are always safe.
-        return noStoreChanges(Safety.SAFE, input);
+
+        ResultingStore thenStore = updateStore(input.getThenStore(), thenUpdates);
+
+        return new ConditionalTransferResult<>(
+                Safety.SAFE, thenStore.store, input.getElseStore(), thenStore.storeChanged);
```
carterkozak, line 1157: 👍

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/safety/SafetyPropagationTransfer.java
pkoenig10: This was just copied from `AbstractNullnessPropagationTransfer`. Do you still want me to change this? I don't know what the correct behavior is here.

https://github.com/google/error-prone/blob/5f4024afe7ff778bfee414671054fd90173a3a4d/check_api/src/main/java/com/google/errorprone/dataflow/nullnesspropagation/AbstractNullnessPropagationTransfer.java#L337-L347

carterkozak: Hmm, they probably know what they're doing. Ideally we'd be able to cover this with a test, but let's not block on that

carterkozak (approved):

PR #2820, opened 2024-06-14
baseline-java-versions: Support Gradle 8.6+ by not inspecting publications

CRogers:
## Before this PR
baseline-java-versions does not work in Gradle 8.6+. This error is produced:

```
   > Circular evaluation detected: task ':log-receiver-api:log-receiver-api-dialogue:compileJava' property 'javaCompiler'
      -> map(org.gradle.jvm.toolchain.JavaCompiler flatmap(map(property(com.palantir.baseline.plugins.javaversions.ChosenJavaVersion, map(com.palantir.baseline.plugins.javaversions.ChosenJavaVersion provider(?) check-type())))) check-type())
      -> flatmap(map(property(com.palantir.baseline.plugins.javaversions.ChosenJavaVersion, map(com.palantir.baseline.plugins.javaversions.ChosenJavaVersion provider(?) check-type()))))
      -> map(property(com.palantir.baseline.plugins.javaversions.ChosenJavaVersion, map(com.palantir.baseline.plugins.javaversions.ChosenJavaVersion provider(?) check-type())))
      -> property(com.palantir.baseline.plugins.javaversions.ChosenJavaVersion, map(com.palantir.baseline.plugins.javaversions.ChosenJavaVersion provider(?) check-type()))
      -> map(com.palantir.baseline.plugins.javaversions.ChosenJavaVersion provider(?) check-type())
      -> provider(?)
      -> set(interface org.gradle.api.publish.maven.MavenArtifact, org.gradle.api.internal.provider.AbstractCollectionProperty$CollectingSupplier@48eab27f)
      -> map(property 'component')
      -> provider(?)
      -> task ':log-receiver-api:log-receiver-api-dialogue:compileJava' property 'javaCompiler'
```

On further inspection, it appears that accessing information about the `MavenArtifact`s for jars in the maven publishing extension requires forcing the `javaCompiler` property the compile task in Gradle 8.6+. Since we use info from `MavenArtifact` to configure the `javaCompiler` property (checking to see whether there are any jar artifacts), this makes an unresolvable circular evaluation.

## After this PR
==COMMIT_MSG==
Prevent "Circular evaluation detected" error when using baseline-java-versions on Gradle 8.6+.
==COMMIT_MSG==

We avoid this by just not looking the published artifacts. Instead, we expand the list of known library and known distribution plugins and only check with them.

## Possible downsides?
* This may work less generally for non-palantir consumers of this plugin who do not use our distribution plugins. This plugin has always been focused on making it easy for Palantir consumers though.
* The behaviour is now different. If someone was using non-standard plugins but depended on this plugin checking for jar non-artifacts to determine library status, it would now downgrade the java version they used to the libraryTarget, possibly causing errors.

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/javaversions/BaselineJavaVersions.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/javaversions/BaselineJavaVersions.java
@@ -16,14 +16,19 @@
 
 package com.palantir.baseline.plugins.javaversions;
 
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import java.util.Objects;
 import org.gradle.api.GradleException;
+import org.gradle.api.Named;
 import org.gradle.api.Plugin;
 import org.gradle.api.Project;
 import org.gradle.api.provider.Property;
 import org.gradle.api.provider.Provider;
+import org.gradle.api.publish.Publication;
 import org.gradle.api.publish.PublishingExtension;
+import org.gradle.api.publish.ivy.IvyPublication;
+import org.gradle.api.publish.maven.MavenPublication;
 import org.gradle.util.GradleVersion;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -73,6 +78,14 @@ public final class BaselineJavaVersions implements Plugin<Project> {
     }
 
     private static boolean isLibrary(Project project, BaselineJavaVersionExtension projectVersions) {
+        if (GradleVersion.current().compareTo(GradleVersion.version("8.6")) >= 0) {
+            return newIsLibrary(project, projectVersions);
+        } else {
+            return oldIsLibrary(project, projectVersions);
+        }
+    }
+
+    private static boolean newIsLibrary(Project project, BaselineJavaVersionExtension projectVersions) {
         Property<Boolean> libraryOverride = projectVersions.overrideLibraryAutoDetection();
         if (libraryOverride.isPresent()) {
             log.debug(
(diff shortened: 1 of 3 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/javaversions/BaselineJavaVersions.java
```diff
@@ -39,10 +34,14 @@ public final class BaselineJavaVersions implements Plugin<Project> {
     public static final String EXTENSION_NAME = "javaVersions";
 
     public static final GradleVersion MIN_GRADLE_VERSION = GradleVersion.version("7.0");
-    // 'nebula.maven-publish' and 'com.palantir.shadow-jar' create publications lazily which cause inconsistencies
-    // based on ordering.
+
     private static final ImmutableSet<String> LIBRARY_PLUGINS =
-            ImmutableSet.of("nebula.maven-publish", "com.palantir.shadow-jar", "com.palantir.external-publish-jar");
+            ImmutableSet.of("com.palantir.external-publish-jar", "com.palantir.publish-jar");
```
carterkozak, line 44: I suspect we'll need to add some other plugins here eventually (possibly in a pluggable way :-s) but no need to implement that in the initial version.

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/javaversions/BaselineJavaVersions.java
```diff
@@ -98,38 +109,9 @@ private static boolean isLibrary(Project project, BaselineJavaVersionExtension p
...
-        if (publication instanceof IvyPublication) {
-            IvyPublication ivyPublication = (IvyPublication) publication;
-            return ivyPublication.getArtifacts().stream().anyMatch(artifact -> "jar".equals(artifact.getExtension()));
-        }
-        // Default to true for unknown publication types to avoid setting higher jvm targets than necessary
-        log.warn(
-                "Unknown publication '{}' of type '{}'. Assuming project {} is a library",
-                publication,
-                publication.getClass().getName(),
-                project.getName());
```
carterkozak: There's an argument to be made for hiding this logic behind a gradle 8.6 version comparison to begin with, so we can experiment without impacting all gradle-baseline users in case something goes wrong. Not something that I'd want to keep long-term, but may make initial rollout a bit easier

carterkozak (commented):
Looks good to me, my commends are just a braindump :-)

gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/javaversions/BaselineJavaVersions.java
CRogers: Nice idea, I think I'll do this. We can then just check all the Gradle upgrade excavators made after a certain date are merging.

CRogers: Although I think the behaviour will still be different unless I switch all the old logic and all the new logic on the gradle version (and not have the new plugins)

CRogers: Kept the old behaviour for now, need a +1.

felixdesouza:
👍

PR #2822, opened 2024-06-17
Remove old pre Gradle 8.6 baseline-java-versions behaviour

CRogers:
## Before this PR
In #2820, we changed the behaviour of baseline-java-versions, but only when Gradle >=8.6 was being used. Now we have upgraded a big chunk of repos to Gradle 8.7 and no issues have appeared.

## After this PR
==COMMIT_MSG==
Remove pre-Gradle 8.6 baseline-java-versions behaviour.
==COMMIT_MSG==

I'll probably wait a few more days just to check no one reports issues before merging. Wanted to make the PR ahead of time so we don't forget to delete the old behaviour.

### DIFF
--- a/changelog/@unreleased/pr-2822.v2.yml
+++ b/changelog/@unreleased/pr-2822.v2.yml
@@ -1,0 +1,5 @@
+type: fix
+fix:
+  description: Remove pre-Gradle 8.6 baseline-java-versions behaviour.
+  links:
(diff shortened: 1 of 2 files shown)


felixdesouza:
👍

carterkozak (approved):

CRogers:
No one has complained so far!

PR #2825, opened 2024-06-21
String.valueOf and String.copyValueOf should propagate log safety annotations

lisasgoh:
## After this PR

==COMMIT_MSG==
String.valueOf and String.copyValueOf should propagate log safety annotations
==COMMIT_MSG==


pkoenig10 (approved):
Thanks!

carterkozak (approved):
Thanks!

Issue #2828, opened 2024-07-11
Old version of spotless java creating compilation issues


daberkow:
## What happened?

With https://github.com/palantir/gradle-baseline/blob/5e773859eee2463ca8b262f8730efff13cdb32db/versions.props#L49 we have spotless java kept at 6.6.0, https://github.com/diffplug/spotless/issues/834#issuecomment-1827659711 says we need to bump to at least 6.7.1 to avoid a bug I am seeing where the following stack trace occurs (I copied the stack trace from the link before).

```
Step 'removeUnusedImports' found problem in 'functional-tests/src/main/java/com/comp/atlas/oyo/upgrade/configuration/TesterConfiguration.java':

null

java.lang.reflect.InvocationTargetException
at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
```

## What did you want to happen?

I see a note that "Newer spotless versions have issues resolving dependencies at configuration time", but latest is 6.24.0, and we are on a build from 2022.

PR #2829, opened 2024-07-11
Bump spotless and versionlock

daberkow:
## Before this PR
Issues arose running spotless checks, as shown in #2828 

## After this PR
==COMMIT_MSG==
Update spotless java to 6.25.0 from 6.6.0
==COMMIT_MSG==

## Possible downsides?
It was noted that updates to spotless "Newer spotless versions have issues resolving dependencies at configuration time", but the build seems to be working.


daberkow:
This has been tested locally against a repo showing the issue, and bumping did correct the issue. I realized 6.25.0 is out, and will bump again.

daberkow:
I am doing some checks, 6.25.0 is passing in circle but failing on my local system. 6.24.0 is passing in both. Trying to figure out why.

versions.props
```diff
@@ -45,8 +45,7 @@ org.mockito:* = 5.12.0
 # Don't upgrade, we will remove this in a future release.
 # Users should depend on this plugin directly such that it will get updated by excavator.
 com.palantir.javaformat:gradle-palantir-java-format = 1.1.0
-# Newer spotless versions have issues resolving dependencies at configuration time
-com.diffplug.spotless:spotless-plugin-gradle = 6.6.0
```
pkoenig10, line 49: The issues that prompted this pin in https://github.com/palantir/gradle-baseline/pull/2312 is still open:

https://github.com/diffplug/spotless/issues/1380

Do we have reason to believe this issue is fixed?

daberkow: In testing this and coil mentioned in the issue. Coil still has it, this repo seems to currently not have it, and people seem to think its related to indirect dependencies being loaded. Gradle does correct the issue if it comes up, it just causes a perf hit. Without bumping spotless, repos using OpenJFx and some other libs can compile at all. If we do not update this lib, repos with that issue will be forced to disable baseline-spotless entirely.

daberkow: The bug reports say that gradle has a catch for this and the worst case scenario is: if there are indirect dependencies and it has to work through them, hitting this regression, then it is a small perf hit, but doesnt fail builds. Where as without this we cant compile UI apps at all. :/

daberkow: Is there a path to get spotless updated here?

daberkow:
https://github.com/palantir/gradle-baseline/pull/2313
https://github.com/diffplug/spotless/issues/1380 
With gradle-consistent-versions this will cause a full build failure. 

Pointed out, for anyone who has issues, you can add the newer spotless version to your classpath to override what is given here, and work from there.

PR #2830, opened 2024-07-24
ResourceIdentifierGetEqualsUsage

schlosna:
## After this PR

==COMMIT_MSG==
ResourceIdentifierGetEqualsUsage rewrites code using `ResourceIdentifier#get*().equals(Object)` to `ResourceIdentifier#has*(String)` to avoid allocations. See https://github.com/palantir/resource-identifier/pull/423
==COMMIT_MSG==


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ResourceIdentifierGetEqualsUsage.java
```diff
@@ -0,0 +1,95 @@
...
+                String replacement = SOURCE_GET_EQUALS_PATTERN.matcher(source).replaceAll("has$1(");
+                return buildDescription(tree)
+                        .addFix(SuggestedFix.builder()
+                                .replace(tree, replacement)
+                                .build())
+                        .build();
+            }
+
+            String replacement = SOURCE_EQUALS_GET_PATTERN.matcher(source).replaceAll("$2.has$3($1");
+            SuggestedFix fix = SuggestedFix.builder().replace(tree, replacement).build();
```
schlosna: I'm not a huge fan of the regex based replacement here, but it works for the cases that matter. I suspect there's a better AST matching mechanism to use if one is suggested.

schlosna (commented):
finally had a few minutes to get back to this and ran locally against a few repos with a decent number of automated improvements, so my sense is this is worth moving forward.

baseline-error-prone/build.gradle
```diff
@@ -13,26 +13,28 @@ dependencies {
     implementation 'io.github.eisop:dataflow-errorprone'
 
     testImplementation gradleApi()
-    testImplementation 'com.palantir.tokens:auth-tokens'
     testImplementation 'com.fasterxml.jackson.core:jackson-annotations'
     testImplementation 'com.google.errorprone:error_prone_test_helpers'
+    testImplementation 'com.palantir.conjure.java.api:test-utils'
+    testImplementation 'com.palantir.conjure.java:conjure-lib'
+    testImplementation 'com.palantir.ri:resource-identifier'
```
schlosna, line 20: this is the only real change in test deps, but OCD got the best of me and I had to sort these

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ResourceIdentifierGetEqualsUsage.java
pkoenig10: Here's how I'd write this to avoid the regex.

https://github.com/palantir/gradle-baseline/commit/613549413121458c38630461ba9cf41a1f01c2fd

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/ResourceIdentifierGetEqualsUsageTest.java
```diff
@@ -0,0 +1,193 @@
...
+                .addOutputLines(
+                        "Test.java",
+                        "import com.palantir.ri.ResourceIdentifier;",
+                        "public class Test {",
+                        "  private static final String SERVICE = \"FOO\";",
+                        "  boolean f(ResourceIdentifier rid) {",
+                        "    return rid.hasService(SERVICE);",
+                        "  }",
+                        "}")
+                .doTest();
```
pkoenig10: nit: If this case is important enough to test for `getService`, then we should test it for every method.

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/ResourceIdentifierGetEqualsUsageTest.java
```diff
@@ -0,0 +1,193 @@
...
+ * limitations under the License.
+ */
+
+package com.palantir.baseline.errorprone;
+
+import org.junit.jupiter.api.Test;
+
+final class ResourceIdentifierGetEqualsUsageTest {
+    @Test
+    void testHasInstance() {
```
pkoenig10: nit: Order these tests service, instance, type, locator to match the RID component order

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/ResourceIdentifierGetEqualsUsageTest.java
```diff
@@ -0,0 +1,193 @@
...
+                        "public class Test {",
+                        "  boolean f(ResourceIdentifier rid) {",
+                        "    return rid.hasInstance(\"test\");",
+                        "  }",
+                        "}")
+                .doTest();
+    }
+
+    @Test
+    void testHasInstanceFlipped() {
```
pkoenig10: nit: This isn't anything specific about instance

```suggestion
    void testInverse() {
```

pkoenig10 (commented):
LGTM, just a couple nits

pkoenig10 (dismissed):

pkoenig10 (approved):

PR #2833, opened 2024-08-08
Optional.toString() should throw error

Priyadarshanvijay:
## Before this PR

Any usages of Optional.toString() are allowed and cause hard to debug runtime bugs where the actual serialized string is either of:
```text
Optional[ACTUAL_VALUE]
Optional.empty
```
depending on the value contained by the optional.

## After this PR

Usages of Optional.toString() will be discouraged and users will be prompted to use Optional.get().toString() instead.

==COMMIT_MSG==
Optional.toString() should throw error
==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/OptionalToString.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/OptionalToString.java
@@ -32,9 +32,8 @@ import java.util.Optional;
 @BugPattern(
         link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
         linkType = BugPattern.LinkType.CUSTOM,
-        severity = SeverityLevel.ERROR,
-        summary = "Optional.toString() does not stringifies the value contained by the Optional"
-                + " object. Did you mean Optional.get().toString() instead?")
+        severity = SeverityLevel.WARNING,
+        summary = "Optional.toString() does not stringifies the value contained by the Optional object.")
 public final class OptionalToString extends BugChecker implements BugChecker.MethodInvocationTreeMatcher {
 
     private static final Matcher<ExpressionTree> OPTIONAL_TO_STRING_METHOD = MethodMatchers.instanceMethod()
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/OptionalToString.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/OptionalToString.java
@@ -1,0 +1,53 @@
+/*
+ * (c) Copyright 2024 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.baseline.errorprone;
+
+import com.google.auto.service.AutoService;
+import com.google.errorprone.BugPattern;
+import com.google.errorprone.BugPattern.SeverityLevel;
+import com.google.errorprone.VisitorState;
+import com.google.errorprone.bugpatterns.BugChecker;
+import com.google.errorprone.matchers.Description;
+import com.google.errorprone.matchers.Matcher;
+import com.google.errorprone.matchers.method.MethodMatchers;
+import com.sun.source.tree.ExpressionTree;
+import com.sun.source.tree.MethodInvocationTree;
+import java.util.Optional;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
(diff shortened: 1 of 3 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/OptionalToString.java
```diff
@@ -0,0 +1,53 @@
...
+import com.sun.source.tree.MethodInvocationTree;
+import java.util.Optional;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
+        severity = SeverityLevel.ERROR,
+        summary = "Optional.toString() does not stringifies the value contained by the Optional"
+                + " object. Did you mean Optional.get().toString() instead?")
```
pkoenig10: I don't think we want to suggest `Optional.get().toString()`. This is not always appropriate, as the optional may be empty.

Even if we do want to suggest a fix, the right way to do this is with Error Prone's suggested fixes, rather than in the description, so that the fixes can be applied automatically.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/OptionalToString.java
```diff
@@ -0,0 +1,53 @@
...
+import com.google.errorprone.matchers.method.MethodMatchers;
+import com.sun.source.tree.ExpressionTree;
+import com.sun.source.tree.MethodInvocationTree;
+import java.util.Optional;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
+        severity = SeverityLevel.ERROR,
```
pkoenig10:
```suggestion
        severity = SeverityLevel.WARNING,
```

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/OptionalToString.java
```diff
@@ -0,0 +1,53 @@
...
+        summary = "Optional.toString() does not stringifies the value contained by the Optional"
+                + " object. Did you mean Optional.get().toString() instead?")
+public final class OptionalToString extends BugChecker implements BugChecker.MethodInvocationTreeMatcher {
```
pkoenig10, line 37: If we want this, we need to devise a roll out plan, since there may be a lot of existing code that does this and we don't want Baseline upgrades to be blocked.

Typically we do this by automating the fix, but that seems more difficult here since the correct fix is context dependnent.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/OptionalToString.java
Priyadarshanvijay: Agreed, but not sure how we can use Error Prone's suggested fixes here since like you highlighted, the actual fix depends on the context

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/OptionalToString.java
Priyadarshanvijay: Would changing SeverityLevel to WARNING instead of error unblock the upgrades? Or should we suggest people to add `SuppressWarnings` if this is the behavior they want?
I think if there are many places in existing codebases where people are using this then this probably points to some runtime error in their code and shouldn't it be better to fix it manually?

carterkozak:
> cause hard to debug runtime bugs

Can you provide some examples? Most of the uses of `Optional#toString` I'm aware of are within `toString` implementations where I'd argue they're entirely reasonable. Runtime behavior generally shouldn't depend on calling `Object.toString`.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/OptionalToString.java
```diff
@@ -0,0 +1,52 @@
...
+import com.sun.source.tree.ExpressionTree;
+import com.sun.source.tree.MethodInvocationTree;
+import java.util.Optional;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
+        severity = SeverityLevel.WARNING,
+        summary = "Optional.toString() does not stringifies the value contained by the Optional object.")
```
carterkozak, line 36: `Optional#toString` does stringify the contained value, and decorates the result with `Optional[<delegate string value>]`  to make it clear that the value is an optional.

Priyadarshanvijay:
> > cause hard to debug runtime bugs
> 
> Can you provide some examples? Most of the uses of `Optional#toString` I'm aware of are within `toString` implementations where I'd argue they're entirely reasonable. Runtime behavior generally shouldn't depend on calling `Object.toString`.

This typically happens when returning string values from functions and asserting over them/storing them in a datastore. We caught this issue in a few of our ETE tests while making changes and it was difficult to debug why things were failing. Once we found the issue, we thought it would be better if our IDE/gradle checks could warn us about this since we almost always want the stringified value of the object wrapped in Optional rather than the decorated string.

If it's not a common occurance, can we add this as an Experimental check which needs to be turned on manually? (Not sure if this is even possible)

pkoenig10:
> returning string values from functions

Can you clarify what you mean by this? Are you describing a Java method that returns `String`? Or something else?

If you are describing a Java method that returns `String`, what does that have to do with `Optional.toString()`?

> storing them in a datastore

> we almost always want the stringified value of the object

Why are we storing stringified values in a database? `toString` is generally a one way conversion that has no backwards compatibility guarantees. We should almost always be persisting data in a structured format (ex. JSON, etc.).

PR #2834, opened 2024-08-08
Refactor: move to abstract classes for baseline-java-versions extensions

CRogers:
## Before this PR
The `baseline-java-versions` extensions use the old "manually making properties" style of Gradle.

## After this PR
==COMMIT_MSG==
Move to abstract classes for baseline-java-versions extensions
==COMMIT_MSG==

...which is more modern Gradle, before I add to these.

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/javaversions/BaselineJavaVersionsExtension.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/javaversions/BaselineJavaVersionsExtension.java
@@ -105,6 +105,12 @@ public abstract class BaselineJavaVersionsExtension implements BaselineJavaVersi
         getDistributionTarget().set(ChosenJavaVersion.fromString(value));
     }
 
+    /** @deprecated Use {@link #getRuntime()} instead. */
+    @Deprecated
+    public final Property<ChosenJavaVersion> runtime() {
+        return getRuntime();
+    }
+
     @Override
     public final void setRuntime(int value) {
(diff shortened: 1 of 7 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/javaversions/BaselineJavaVersionsExtension.java
```diff
@@ -74,42 +85,41 @@ public final void setLibraryTarget(String value) {
...
 
     @Override
     public final void setDistributionTarget(int value) {
-        distributionTarget.set(ChosenJavaVersion.of(value));
+        getDistributionTarget().set(ChosenJavaVersion.of(value));
     }
 
     /** Accepts inputs such as '17_PREVIEW'. */
     @Override
     public final void setDistributionTarget(String value) {
```
CRogers, line 104: Because these setter overrides exist, Gradle won't let us create an abstract `getDistributionTarget()` property, which is extremely frustrating. We need these to remain as we need to be able to let people set a plain string or integer. Unsure whether this is intended behaviour by Gradle or not.

CRogers: I don't particularly want to call the actual property something else; I guess we just abandon this idea.

CRogers: Gradle bug report: https://github.com/gradle/gradle/issues/26691#issuecomment-2277814640

PR #2835, opened 2024-08-13
Provide more specific error message for `StrictUsedVariable` check based on unused variable type

### LINKED ISSUE #2361: `StrictUnusedVariable` check should suggest removing variables assigned in constructor but never read

IlanaRadinsky:
Fixes https://github.com/palantir/gradle-baseline/issues/2361

## Before this PR
Turns out this was just a confusing error message. Previously, a code snippet like
```
public class Test {
    private final String myField;

    public Test(String myField) {
        this.myField = myField;
    }
}
```
would result in an error message
```
error: [StrictUnusedVariable] The field 'field' is never read. Intentional occurrences are acknowledged by renaming the unused variable with a leading underscore. '_field', for example.
  private String field;
                 ^
  Did you mean to remove this line?
```

If you applied the fix from the error message, checks would fail because the field is read but has 'StrictUnusedVariable' suppressed because of its name (prefixed with an underscore).

But really, the correct fix is suggested below the code snippet/link - to remove the field (`Did you mean to remove this line?`).

## After this PR

==COMMIT_MSG==
Provide more specific error message for `StrictUsedVariable` check based on unused variable type
==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StrictUnusedVariable.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StrictUnusedVariable.java
@@ -260,14 +260,13 @@ public final class StrictUnusedVariable extends BugChecker implements BugChecker
             } else {
                 fixes = buildUnusedVarFixes(symbol, allUsageSites, state);
             }
+            String assignmentDescriptor = unused instanceof VariableTree ? "" : "assignment to this ";
+            String descriptor = assignmentDescriptor + describeVariable(symbol);
             state.reportMatch(buildDescription(unused)
                     .setMessage(String.format(
-                            "%s %s '%s' is never read. Intentional occurrences are acknowledged by renaming "
-                                    + "the unused variable with a leading underscore. '_%s', for example.",
-                            unused instanceof VariableTree ? "The" : "The assignment to this",
-                            describeVariable(symbol),
-                            symbol.name,
-                            symbol.name))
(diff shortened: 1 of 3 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StrictUnusedVariable.java
```diff
@@ -253,28 +251,14 @@ public Description matchCompilationUnit(CompilationUnitTree tree, VisitorState s
...
                 if (index == -1) {
-                    fixes = buildUnusedLambdaParameterFix(symbol, entry.getValue(), state);
+                    reportUnusedLambdaParameter(symbol, unused, specs, state);
                 } else {
-                    fixes = buildUnusedParameterFixes(symbol, methodSymbol, allUsageSites, state);
+                    reportUnusedParameter(symbol, methodSymbol, unused, allUsageSites, state);
                 }
             } else {
-                fixes = buildUnusedVarFixes(symbol, allUsageSites, state);
+                reportUnusedAssignment(symbol, unused, specs, allUsageSites, state);
```
IlanaRadinsky, line 259: This only reports a match if the var is a local var, so no need to do it for all vars (including parameters) as we were before.

PR #2836, opened 2024-08-14
Test for implicit toString() safety

ash211:
I would expect all these tests to pass, but the latter 2 do not.

### DIFF
--- a/baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/IllegalSafeLoggingArgumentTest.java
+++ b/baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/IllegalSafeLoggingArgumentTest.java
@@ -717,6 +717,120 @@ class IllegalSafeLoggingArgumentTest {
     }
 
     @Test
+    public void testStringFormatSafety() {
+        helper().addSourceLines(
+                        "Test.java",
+                        "import com.palantir.logsafe.*;",
+                        "class Test {",
+                        "  void f(",
+                        "      @Safe String safeParam,",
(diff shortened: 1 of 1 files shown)


baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/IllegalSafeLoggingArgumentTest.java
```diff
@@ -716,6 +716,120 @@ public void testStringConcatAssignmentSafety() {
...
+                        "    public String toString() {",
+                        "      return \"safe\";",
+                        "    }",
+                        "  }",
+                        "  class UnsafeToStringClass {",
+                        "    @Unsafe",
+                        "    @Override",
+                        "    public String toString() {",
+                        "      return \"unsafe\";",
+                        "    }",
```
ash211, line 761: The conjure generator for unsafe aliases generates code like this. Example type:

```
      ObjectTypeId:
        alias: string
        safety: unsafe
```

carterkozak: I don't think this is true -- the alias generator aplies an `@Unsafe` annotation to both toString and the class. Similarly, the SafeLoggingPropagation error-prone check requires us to propagate safety information from `toString()` to the top level class.

ash211: Ah true.  Please disregard -- I think the error I'm hunting for is elsehwere

PR #2840, opened 2024-08-19
Upgrade Baseline and Error Prone

pkoenig10:
See https://github.com/palantir/gradle-baseline/pull/2839


.baseline/copyright/999_palantir.txt
```diff
@@ -0,0 +1 @@
+(c) Copyright ${today.year} Palantir Technologies Inc. All rights reserved.
```
carterkozak: This should not be added

pkoenig10: Hmm this was generated by `baselineUpdateConfig`, will figure out why.

pkoenig10: This is also generated if you run `baselineUpdateConfig` on develop. I can remove it here, but I suspect that an excavator will automatically generate it in the future.

carterkozak: The excavator deletes new copyright files added by the task, when running the update task we need to do that step manually

carterkozak (approved):

PR #2843, opened 2024-08-20
Simplify error-prone refactoring code

CRogers:
## Before this PR
Now we've upgraded to recent errorprone, there are some workarounds we had for bugs that are no longer necessary. Now that https://github.com/google/error-prone/issues/3908 is fixed (and [released in 2.26.0](https://github.com/google/error-prone/commit/9da2d5580e3939f97ef2e91278b330a56b5ed1fe)) we can avoid explicitly disabling checks when patching.

## After this PR
==COMMIT_MSG==
Simplify error-prone refactoring code
==COMMIT_MSG==

### DIFF
--- a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
+++ b/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
@@ -16,21 +16,17 @@
 
 package com.palantir.baseline.plugins;
 
-import com.google.common.base.Joiner;
 import com.google.common.base.Splitter;
 import com.google.common.base.Strings;
-import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.MoreCollectors;
 import com.palantir.baseline.extensions.BaselineErrorProneExtension;
 import java.util.List;
-import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
-import net.ltgt.gradle.errorprone.CheckSeverity;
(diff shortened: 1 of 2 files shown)


gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java
```diff
@@ -172,42 +168,23 @@ private static void configureErrorProneOptions(
...
-                    Optional<List<String>> specificChecks = getSpecificErrorProneChecks(project);
-                    if (specificChecks.isPresent()) {
-                        List<String> errorProneChecks = specificChecks.get();
-                        // Work around https://github.com/google/error-prone/issues/3908 by explicitly enabling any
-                        // check we want to use patch checks for (ensuring it is not disabled); if this is fixed, the
-                        // -Xep:*:ERROR arguments could be removed
-                        return Iterables.concat(
-                                errorProneChecks.stream()
-                                        .map(checkName -> "-Xep:" + checkName + ":ERROR")
-                                        .collect(Collectors.toList()),
```
CRogers, line 199: I was expecting this change to [fail this test](https://github.com/palantir/gradle-baseline/blob/f9539d53d1a1bfd945e9fc9f1db8b499e73cdbf1/gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineErrorProneIntegrationTest.groovy#L208-L238), but it is actually working 🤔

CRogers: I think the tests in `BaselineErrorProneIntegrationTest` are very broken, they still pass if you stop disabling the checks.

carterkozak: I recall some configuration ordering issues where checks may be configured before the gradle-baseline defaults are applied and override error-prone settings

PR #2851, opened 2024-08-28
Safety propagation avoids recursive analysis for pattern match variables

### LINKED ISSUE #2850: Slow compilation for large pattern matching switch
## What happened?

Similar to #2328. This code snippet takes a very long time to compile (and so does the code snippet in [the previous issue](https://github.com/palantir/gradle-baseline/issues/2328#issuecomment-1194171048)):
```
String id(Parent parent) {
    return switch (parent) {
        case Child1 child -> child.id();
        case Child2 child -> child.id();
        case Child3 child -> child.id();
        case Child4 child -> child.id();
        case Child5 child -> child.id();
        case Child6 child -> child.id();
        case Child7 child -> child.id();
        case Child8 child -> child.id();
    };
}
```
The 7-branch case only takes ~3s for me, but the 8-branch case takes ~20s, the 9-branch case takes ~3m, growing exponentially.

This should take very little time.

## Other notes

Just like #2328, this looks to also be due to the `SafeLoggingPropagation` check: each pattern matching variable `child` is considered a captured reference and recursively triggers an analysis [here](https://github.com/palantir/gradle-baseline/blob/bcd9d2ef8b93e873dfdb38bc9f01add92a9b259b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/safety/SafetyPropagationTransfer.java#L872C26-L872C56).

kevinychen:
Fixes #2850 

```
  return switch (parent) {
      case Child1 child -> child.id();
      case Child2 child -> child.id();
  };
```
In a pattern matching switch, we never seem to add the pattern-matched variables (`child`) to the `AccessPathStore`. For example, for the pattern-matching instance of, we update the store [here](https://github.com/palantir/gradle-baseline/blob/bcd9d2ef8b93e873dfdb38bc9f01add92a9b259b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/safety/SafetyPropagationTransfer.java#L1159). (It might be possible to update `visitSwitchExpressionNode` in a similar way, but I think that requires Java 17+ compiler APIs.)

This means that when we visit `child.id()`, we don't have a safety value for `child` and we enter [this code path](https://github.com/palantir/gradle-baseline/blob/bcd9d2ef8b93e873dfdb38bc9f01add92a9b259b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/safety/SafetyPropagationTransfer.java#L872C26-L872C56),  triggering a recursive analysis.

## After this PR

In this PR, every time we need to lazily compute a safety value in `visitLocalVariable`, we also add it to the store so downstream nodes don't need to recompute it. The first time we visit `child`, it's in the pattern binding (`case Child1 child`) and we compute the safety quickly [here](https://github.com/palantir/gradle-baseline/blob/bcd9d2ef8b93e873dfdb38bc9f01add92a9b259b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/safety/SafetyPropagationTransfer.java#L869). The subsequent times we visit `child` (`child.id()`), we now get the safety from the store instead of triggering a recursive analysis.

==COMMIT_MSG==
Fix slow compilation for large pattern matching switch
==COMMIT_MSG==

## Possible downsides?

It's probably weird to add a safety value to the store in `visitLocalVariable`, since semantically it should be added higher up (e.g. for instance of, it's correctly added in `visitInstanceOf`. But I think the entire `if (safety == null) { ... }` block in `visitLocalVariable` is intended to handle these weird cases where we didn't properly add a safety value to the store, so this change at least feels consistent.

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/safety/SafetyPropagationTransfer.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/safety/SafetyPropagationTransfer.java
@@ -871,6 +871,9 @@ public final class SafetyPropagationTransfer implements ForwardTransferFunction<
                 // No safety information found, likely a captured reference used within a lambda or anonymous class.
                 safety = getCapturedLocalVariableSafety(node);
             }
+            ReadableUpdates updates = new ReadableUpdates();
+            updates.set(node, safety);
+            return updateRegularStore(safety, input, updates);
         }
         return noStoreChanges(safety, input);
     }
--- a/baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/IllegalSafeLoggingArgumentTest.java
+++ b/baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/IllegalSafeLoggingArgumentTest.java
@@ -1067,6 +1067,39 @@ class IllegalSafeLoggingArgumentTest {
     }
 
     @Test
+    @Timeout(2)
+    public void testSwitchExpressionPatternMatching() {
+        helper().addSourceLines(
+                        "Test.java",
+                        "import com.palantir.logsafe.*;",
+                        "class Test {",
+                        "  sealed interface Parent {}",
+                        "  record Child1() implements Parent {}",
+                        "  record Child2() implements Parent {}",
+                        "  record Child3() implements Parent {}",
+                        "  record Child4() implements Parent {}",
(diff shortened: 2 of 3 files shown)


carterkozak (dismissed):
thanks for the fix! There's clearly some additional work we need to do in order to propagate safety from the switch case rules back to the result, but that can be handled separately.

carterkozak (approved):

PR #2878, opened 2024-09-18
disable PreferSafeLogger and StrictUnusedVariable for intellij plugin development

FinlayRJW:
## Before this PR
Inellij plugins had to manually disable error prone options

## After this PR
Now error prone options are auto disabled for intellij plugins

### DIFF
--- a/changelog/@unreleased/pr-2878.v2.yml
+++ b/changelog/@unreleased/pr-2878.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: disable PreferSafeLogger and StrictUnusedVariable  for intellij
+  links:
+  - https://github.com/palantir/gradle-baseline/pull/2878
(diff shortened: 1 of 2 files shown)


CRogers (approved):

pkoenig10:
Hey, I don't think this is a good idea for the reasons mentioned in https://github.com/palantir/gradle-baseline/pull/2010#issuecomment-1012606175.

If my compilation is going to fail in CI, I want to to fail locally. Otherwise developers just waste cycles when they push changes and then compilation fails in CI. And when I need to fix this I want to be able to verify locally that I've addressed all the failures, otherwise I just have to repeat this cycle again.

I think we can debate what the correct default is, but I feel pretty strongly that we must at least give users the option to enable these if they want.

pkoenig10:
Ignore me, I misunderstood this PR. This applies to IntelliJ plugin development, not compilation in IntelliJ.

PR #2883, opened 2024-09-19
Fix `checkUnusedDependencies` for multi-source projects

crogoz:
## Before this PR

`checkUnusedDependencies` fails when there is a dependency on a project that has multi source types (eg. java & scala) but it only uses a source type (eg. scala)

## After this PR
Fixes `checkUnusedDependencies` for multi-source projects.

==COMMIT_MSG==
Fix `checkUnusedDependencies` for multi-source projects
==COMMIT_MSG==


gradle-baseline-java/src/main/java/com/palantir/baseline/tasks/CheckUnusedDependenciesTask.java
```diff
@@ -77,13 +77,17 @@ public final void checkUnusedDependencies() {
...
-        Set<ResolvedArtifact> necessaryArtifacts = Streams.stream(
+        Set<String> necessaryArtifactsDeclaration = Streams.stream(
                         sourceClasses.get().iterator())
                 .flatMap(BaselineExactDependencies::referencedClasses)
                 .flatMap(BaselineExactDependencies.INDEXES::classToArtifacts)
+                .map(BaselineExactDependencies::asString)
                 .collect(Collectors.toSet());
 
-        Set<ResolvedArtifact> possiblyUnused = Sets.difference(declaredArtifacts, necessaryArtifacts);
+        Set<ResolvedArtifact> possiblyUnused = StreamEx.of(declaredArtifacts)
```
crogoz: Doing a set.difference is not correct here. 
A resolvedArtifact is an object that contains the sourcesPaths as well eg. `classes/java/main` or `classes/scala/main`. A dependency (eg. `project('my-subproject')`) can contain multiple sourcePaths (in the case of a 'scala' libraries that have both java & scala sources) and only a sourceType (eg. scala) might be consumed. In that case when doing the set.difference, we would end up with a possiblyUnusedArtifact =`ResolvedArtifact(project('my-project'), sourcePath=classes/java/main)`

As a solution, comparing if the actual dependency string is used should be enough.

esword: why the use of StreamEx vs just doing filter?  I'm leery of bringing in a whole other dependency for one usage.

crogoz: removed it in https://github.com/palantir/gradle-baseline/pull/2883/commits/be84e807b64ea7fc7eaabd4780968644a2844c28. 

I thought I could apply Felix's comment from a different PR here as well: `https://github.com/palantir/gradle-jdks/pull/403#discussion_r1747113978`

esword (approved):

PR #2891, opened 2024-09-26
Correctly propagate safety information for Multimap collectors

pkoenig10:
The test cases added in this PR currently produce false positives.


carterkozak (dismissed):
Thanks!

carterkozak (approved):

PR #2896, opened 2024-09-30
Add `org.mockito.ArgumentMatchers.*` to `AvoidStaticImport` exclusions

MGHawes:
## Before this PR
- We did not permit `any()` or `eq(...)` static imports

## After this PR
- Add `org.mockito.ArgumentMatchers.*` to `AvoidStaticImport` exclusions
- Is in-line with the other suppressions here for Mockito and AssertJ

==COMMIT_MSG==
Add `org.mockito.ArgumentMatchers.*` to `AvoidStaticImport` exclusions
==COMMIT_MSG==

### DIFF
--- a/.baseline/checkstyle/checkstyle.xml
+++ b/.baseline/checkstyle/checkstyle.xml
@@ -77,7 +77,7 @@
         </module>
         <module name="AvoidStarImport"/> <!-- Java Style Guide: No wildcard imports -->
         <module name="AvoidStaticImport"> <!-- Java Style Guide: No static imports -->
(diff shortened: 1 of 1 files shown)


pkoenig10:
We already suppress `AvoidStaticImport` in test files. So this PR shouldn't be necessary.

https://github.com/palantir/gradle-baseline/blob/28f8b665dd7234dcd161efb8114632c8b8b2fe3b/gradle-baseline-java-config/resources/checkstyle/checkstyle-suppressions.xml#L13

PR #2912, opened 2024-10-10
forbid full atlasdb range scans

bavardage:
## After this PR

==COMMIT_MSG==
Forbid doing full atlas range scans (without thinking about it!)
==COMMIT_MSG==


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DangerousRangeScanUsage.java
```diff
@@ -0,0 +1,51 @@
...
+    private static final long serialVersionUID = 1L;
+
+    private static final Matcher<ExpressionTree> STRING_INTERN_METHOD_MATCHER = MethodMatchers.instanceMethod()
+            .onExactClass("com.palantir.atlasdb.keyvalue.api.RangeRequest")
+            .named("all")
+            .withNoParameters();
+
+    @Override
+    public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
+        if (STRING_INTERN_METHOD_MATCHER.matches(tree, state)) {
```
schlosna:
```suggestion
    private static final Matcher<ExpressionTree> ALL_RANGE_METHOD_MATCHER = MethodMatchers.instanceMethod()
            .onExactClass("com.palantir.atlasdb.keyvalue.api.RangeRequest")
            .named("all")
            .withNoParameters();

    @Override
    public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
        if (ALL_RANGE_METHOD_MATCHER.matches(tree, state)) {
```

schlosna (commented):
Overall supportive. Can we add quick unit test?

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DangerousRangeScanUsage.java
schlosna: We may also want to warn on invocations of `getAllRowsUnordered()`

bavardage: good shout on `getAllRowsUnordered` - will push that as a separate PR

bavardage:
added unit test + other fixes (e.g. the variable name you called out)!

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DangerousRangeScanUsage.java
```diff
@@ -0,0 +1,51 @@
...
+        linkType = BugPattern.LinkType.CUSTOM,
+        severity = SeverityLevel.WARNING,
+        summary = "Should not use full atlasdb range scans")
+public final class DangerousRangeScanUsage extends BugChecker implements BugChecker.MethodInvocationTreeMatcher {
+    private static final long serialVersionUID = 1L;
+
+    private static final Matcher<ExpressionTree> ALL_RANGE_METHOD_MATCHER = MethodMatchers.staticMethod()
+            .onClass("com.palantir.atlasdb.keyvalue.api.RangeRequest")
+            .named("all")
+            .withNoParameters();
```
carterkozak, line 42: Since we own the source for this type, perhaps we should deprecate the method and add something like `@DoNotCall("Do not use full atlasdb range scans")` instead of adding a new custom error-prone rule. What do you think?

felixdesouza: there's also `@RestrictedApi` if you need to allowlist usages within Atlas instead of suppressing within atlas

esword: @felixdesouza - what is this magic you speak of (RestrictedApi)? So wish I'd known about that last winter with the massive deprecations we did.

felixdesouza: haha yeah. It's this: https://errorprone.info/bugpattern/RestrictedApi

felixdesouza: [javadoc](https://errorprone.info/api/latest/com/google/errorprone/annotations/RestrictedApi.html) is better explainer

bavardage:
Cool - will use that instead!

PR #2921, opened 2024-10-18
Upgrade errorprone/checkerframework to latest releases supporting JDK-11

carterkozak:
==COMMIT_MSG==
Upgrade errorprone/checkerframework to latest releases supporting JDK-11
==COMMIT_MSG==


carterkozak:
Looks like the excavator is already handling this for the most part

PR #2944, opened 2024-11-07
Use suppressible-error-prone

CRogers:
## Before this PR
[suppressible-error-prone](https://github.com/palantir/suppressible-error-prone) now exists, which lets us suppress error prones easily and moves a lot of the heavy lifting code out of this repo in that other repo.

## After this PR
==COMMIT_MSG==
Use suppressible-error-prone so it's easy to suppress error prones that are currently failing
==COMMIT_MSG==

Much more info at https://github.com/palantir/suppressible-error-prone/pull/6.

This basically turns this repo in just "configuration" for suppressible-error-prone.

Requires https://github.com/palantir/suppressible-error-prone/pull/17


gradle-baseline-java/build.gradle
```diff
@@ -65,6 +67,7 @@ test {
     environment 'CIRCLE_TEST_REPORTS', "${buildDir}/circle-reports"
 
     systemProperty 'ignoreDeprecations', 'true'
+    environment 'ORG_GRADLE_PROJECT_baselineErrorProneVersion', project.version
```
felixdesouza, line 73: what is this?

gradle-baseline-java/src/main/java/com/palantir/baseline/extensions/BaselineErrorProneExtension.java
```diff
@@ -97,14 +94,19 @@ public class BaselineErrorProneExtension {
...
+    @Deprecated
+    public final SetProperty<String> getPatchChecks() {
+        return suppressibleErrorProneExtension.getPatchChecks();
     }
 
-    public final ListProperty<String> getPatchChecks() {
-        return patchChecks;
+    public BaselineErrorProneExtension(SuppressibleErrorProneExtension suppressibleErrorProneExtension) {
+        this.suppressibleErrorProneExtension = suppressibleErrorProneExtension;
+        suppressibleErrorProneExtension.getPatchChecks().addAll(DEFAULT_PATCH_CHECKS);
```
felixdesouza, line 111: perhaps it already exists, but should we be worried that it's easy to wipe out the defaults just by doing `patchChecks = [ ... ] ` ?

CRogers: I think this is a valid thing to do, although probably not the right thing to do. I think we should just not advertise this ability at all.

gradle-baseline-java/build.gradle
CRogers: Add comment

felixdesouza:
👍  💃

PR #2946, opened 2024-11-08
Add StreamFlatMapOptional error-prone check

schlosna:
## Before this PR

JFR profiles indicated some uses of `Stream<Optional<T>>.flatMap(Optional::stream)` were causing unnecessary allocations & GC pressure. 

## After this PR

==COMMIT_MSG==
If one has a `Stream<Optional<T>> stream` of size `N` and does `stream.flatMap(Optional::stream)`, you’ll end up allocating `N` extra streams — one for each `Optional` input element. When `N` is large, those allocations can cause extra GC cycles and pauses if allocation rate is high enough leading to issues with latency, throughput, and allocation sensitive code paths.

`Stream.filter(Optional::isPresent).map(Optional::get)` is more efficient than `Stream.flatMap(Optional::stream)` as it does not allocate a new `Stream` for every element in the stream.
==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StreamFlatMapOptional.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StreamFlatMapOptional.java
@@ -1,0 +1,79 @@
+/*
+ * (c) Copyright 2024 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.baseline.errorprone;
+
+import com.google.auto.service.AutoService;
+import com.google.errorprone.BugPattern;
+import com.google.errorprone.VisitorState;
+import com.google.errorprone.bugpatterns.BugChecker;
+import com.google.errorprone.fixes.SuggestedFix;
+import com.google.errorprone.matchers.Description;
+import com.google.errorprone.matchers.Matcher;
+import com.google.errorprone.matchers.method.MethodMatchers;
+import com.google.errorprone.util.ASTHelpers;
+import com.sun.source.tree.ExpressionTree;
+import com.sun.source.tree.MethodInvocationTree;
+import java.util.List;
+import java.util.Optional;
+import java.util.function.Function;
+import java.util.stream.Stream;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
+        severity = BugPattern.SeverityLevel.WARNING,
+        summary = "`Stream.filter(Optional::isPresent).map(Optional::get)` is more efficient than "
+                + "`Stream.flatMap(Optional::stream)`")
+public final class StreamFlatMapOptional extends BugChecker implements BugChecker.MethodInvocationTreeMatcher {
+    private static final long serialVersionUID = 1L;
+
+    private static final Matcher<ExpressionTree> STREAM_FLAT_MAP = MethodMatchers.instanceMethod()
+            .onDescendantOf(Stream.class.getName())
+            .namedAnyOf("flatMap")
+            .withParameters(Function.class.getName());
+
+    private static final Matcher<ExpressionTree> OPTIONAL_STREAM = MethodMatchers.instanceMethod()
+            .onDescendantOf(Optional.class.getName())
+            .named("stream")
+            .withNoParameters();
+
+    @Override
+    public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
+        if (STREAM_FLAT_MAP.matches(tree, state)) {
+            ExpressionTree stream = ASTHelpers.getReceiver(tree);
+            if (stream != null) {
+                ExpressionTree streamReceiver = ASTHelpers.getReceiver(stream);
+                if (streamReceiver != null) {
+                    List<? extends ExpressionTree> arguments = tree.getArguments();
+                    if (arguments != null && arguments.size() == 1) {
+                        if (OPTIONAL_STREAM.matches(arguments.get(0), state)) {
+                            String replacement = state.getSourceForNode(ASTHelpers.getReceiver(tree.getMethodSelect()))
+                                    + ".filter(Optional::isPresent).map(Optional::get)";
+                            SuggestedFix fix = SuggestedFix.builder()
+                                    .addImport("java.util.Optional")
+                                    .replace(tree, replacement)
+                                    .build();
+                            return buildDescription(tree).addFix(fix).build();
+                        }
+                    }
+                }
+            }
+        }
+        return Description.NO_MATCH;
+    }
+}
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StreamFlatMapOptional.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StreamFlatMapOptional.java
@@ -57,19 +57,16 @@ public final class StreamFlatMapOptional extends BugChecker implements BugChecke
         if (STREAM_FLAT_MAP.matches(tree, state)) {
             ExpressionTree stream = ASTHelpers.getReceiver(tree);
             if (stream != null) {
-                ExpressionTree streamReceiver = ASTHelpers.getReceiver(stream);
-                if (streamReceiver != null) {
-                    List<? extends ExpressionTree> arguments = tree.getArguments();
-                    if (arguments != null && arguments.size() == 1) {
-                        if (OPTIONAL_STREAM.matches(arguments.get(0), state)) {
-                            String replacement = state.getSourceForNode(ASTHelpers.getReceiver(tree.getMethodSelect()))
-                                    + ".filter(Optional::isPresent).map(Optional::get)";
-                            SuggestedFix fix = SuggestedFix.builder()
-                                    .addImport("java.util.Optional")
-                                    .replace(tree, replacement)
-                                    .build();
-                            return buildDescription(tree).addFix(fix).build();
-                        }
+                List<? extends ExpressionTree> arguments = tree.getArguments();
+                if (arguments != null && arguments.size() == 1) {
+                    if (OPTIONAL_STREAM.matches(arguments.get(0), state)) {
+                        String replacement = state.getSourceForNode(ASTHelpers.getReceiver(tree.getMethodSelect()))
+                                + ".filter(Optional::isPresent).map(Optional::get)";
+                        SuggestedFix fix = SuggestedFix.builder()
+                                .addImport("java.util.Optional")
+                                .replace(tree, replacement)
+                                .build();
+                        return buildDescription(tree).addFix(fix).build();
                     }
                 }
             }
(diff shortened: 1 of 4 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StreamFlatMapOptional.java
```diff
@@ -0,0 +1,76 @@
...
+        if (STREAM_FLAT_MAP.matches(tree, state)) {
+            ExpressionTree stream = ASTHelpers.getReceiver(tree);
+            if (stream != null) {
+                List<? extends ExpressionTree> arguments = tree.getArguments();
+                if (arguments != null && arguments.size() == 1) {
+                    if (OPTIONAL_STREAM.matches(arguments.get(0), state)) {
+                        String replacement = state.getSourceForNode(ASTHelpers.getReceiver(tree.getMethodSelect()))
+                                + ".filter(Optional::isPresent).map(Optional::get)";
+                        SuggestedFix fix = SuggestedFix.builder()
+                                .addImport("java.util.Optional")
```
carterkozak: It can be a bit safer to use `qualifyType` for these, in case something like the old guava Optional type is also imported in the same class.

e.g. https://github.com/palantir/gradle-baseline/blob/4ae1e054ed729efd7814dd888e7486cad3ac2aa8/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferCollectionTransform.java#L64

This replaces `addImport` and potentially uses a fully qualified reference if another type with the same simple name is already imported.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StreamFlatMapOptional.java
```diff
@@ -0,0 +1,76 @@
...
+                                .replace(tree, replacement)
```
carterkozak: In case there are multiple `Stream.flatMap(Optional::stream)` occurrences in a single call chain, it may be helpful to make a more specific replacement, allowing a single round of applying the fixes to produce a passing build.

Instead of replacing the full `tree`, we can do something along these lines:

```java
.replace(
    state.getEndPosition(ASTHelpers.getReceiver(tree.getMethodSelect())),
    state.getEndPosition(tree),
    ".filter(Optional::isPresent).map(Optional::get)")
```

I imagine a test could produce this case using something along these lines:
```java
Stream<String> f(Stream<Optional<Optional<String>>> in) {
    return in.flatMap(Optional::stream).flatMap(Optional::stream);
}
```

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StreamFlatMapOptional.java
```diff
@@ -0,0 +1,76 @@
...
+            .withNoParameters();
+
+    @Override
+    public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
+        if (STREAM_FLAT_MAP.matches(tree, state)) {
+            ExpressionTree stream = ASTHelpers.getReceiver(tree);
+            if (stream != null) {
+                List<? extends ExpressionTree> arguments = tree.getArguments();
+                if (arguments != null && arguments.size() == 1) {
+                    if (OPTIONAL_STREAM.matches(arguments.get(0), state)) {
```
carterkozak: error-prone provides a utility function which may be helpful here. Sometimes I opt not to use it in order to get access to parameters required to build the suggested fix, but I figured I'd call it out in case you find it helpful:

```java
private static final Matcher<ExpressionTree> 
STREAM_FLATMAP_OPTIONAL_STREAM = Matchers.methodInvocation(
    STREAM_FLAT_MAP,
    // Any of the three MatchTypes are reasonable in this case, given a single arg
    MatchType.LAST,
    OPTIONAL_STREAM);
```

carterkozak (commented):
This looks fantastic, thanks for putting it together!

I added a few comments which aren't blockers, goal is to give you as much context as possible since you mentioned that you'd like to write more of this style of check :-)

schlosna:
Thanks @carterkozak , that looks much better. Added a few more test cases as well.

carterkozak (approved):
Thanks!

pkoenig10:
Hey I don't think this is a good as a blanket rule. It makes the code harder to read, less resilient to refactors, and encourages potentially error-prone patterns.

When devs see things like `.map(Optional::get)` they are encouraged to do that elsewhere and can easily forget the present check. The vast majority of code that does `.flatMap(Optional::stream)` is not performance sensitive enough for this to matter. While I agree this is useful to avoid allocations, I don't think the cost is unambiguously worth the benefit.

For example, this check changed roughly 65 instances of this in our internal authentication service in https://g.p.b/foundry/multipass/pull/21715.

@carterkozak @schlosna can we consider reverting this?

schlosna:
@pkoenig10 are there any places you have found that are incorrect? I just looked through that PR and all of the replacements look correct to me, and given the service this seems like a net improvement to avoid unnecessary GC overhead. 

`Optional` is effectively boxing nullable references, which I know we like express in APIs, but streaming through these is an internal implementation detail that we should make efficient by default. I think you would agree that avoiding boxed types in performance sensitive places is beneficial given heavy use in MP. Most uses that this check fixes are on hot paths that are at least `O(N)` if not `O(N * M)` or worse, and are places we should not be eating additional allocation & CPU cycles on Optional. I can share a few internal JFR showing some services with 3-5% of total allocations due to `Stream.flatMap(Optional::stream)`. I want to avoid future regressions, thus this automated fix.

I'm not convinced folks will incorrectly use `Optional#get()` more because of this check. IntelliJ will appropriately flag uses of `Optional#get()` that are not guarded by `Optional#isPresent()` or `Optional#isEmpty()`. Ideally folks would use those in combination with `Optional#orElseThrow()` and we would have error-prone checks enforcing at compile time, or in less performance sensitive places `Optional#map()`.


See also "Effective Java" 3rd edition, Item 55: Return optionals judiciously and Stuart Marks' https://stuartmarks.wordpress.com/wp-content/uploads/2017/10/optionalmotherofallbikesheds-devoxxus2017.pdf for external bike shed opinions.

pkoenig10:
> @pkoenig10 are there any places you have found that are incorrect?

I'm not suggesting that the automated refactors are incorrect - I agree that they are correct.

I can see the arguments both ways here. I certainly agree we should be doing this in performance sensitive places. But these allocations are inconsequential for 99% of the code we write.

> I'm not convinced folks will incorrectly use `Optional#get()` more because of this check. IntelliJ will appropriately flag uses of `Optional#get()` that are not guarded by `Optional#isPresent()` or `Optional#isEmpty()`. 

I see devs makes these kinds of mistakes all the time - pattern-matching from existing code without thinking more deeply about what they are actually writing. IntelliJ lints and code reviews are tools to mitigate this, but they aren't foolproof in the way that compile-time checks (ie. typechecking) or tests are.

I won't belabor the point. I understand the arguments here but just don't personally agree that it is desirable to declare that all uses of `.flatMap(Optional::stream)` should be replaced by `.filter(Optional::isPresent).map(Optional:get)`.

carterkozak:
I think the key piece is that sometimes this optimization is incredibly meaningful, but sometimes code doesn’t need to be optimal.

so, do we force folks to make this distinction based on their knowledge of the scope they’re working within, or do we eliminate the option so that folks don’t have to think about it. If we give them the option, are we confident they’ll get it right most of the time? Likely not, since copy/paste is common and tends to occur without much thought about the perf requirements of the original vs new code (and often times code which wasn’t meant to be hot becomes hot in a refactor or new call path).

I think the additional verbosity is a reasonable trade off, and likely to lead to fewer problems than the alternative, but it’s not a hill that I’d die on!

pkoenig10:
Makes sense, I don't disagree with reasons here, thanks for engaging!

schlosna:
https://github.com/palantir/gradle-baseline/pull/2946 is a draft PR to catch cases where one attempts to `Stream#map(Optional::get)` without `Stream#filter(Optional::isPresent)`.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StreamFlatMapOptional.java
```diff
@@ -0,0 +1,79 @@
...
+            OPTIONAL_STREAM);
+
+    @Override
+    public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
+        if (STREAM_FLATMAP_OPTIONAL_STREAM.matches(tree, state)) {
+            ExpressionTree receiver = ASTHelpers.getReceiver(tree.getMethodSelect());
+            if (receiver != null) {
+                SuggestedFix.Builder fix = SuggestedFix.builder();
+                String optionalType = SuggestedFixes.qualifyType(state, fix, Optional.class.getCanonicalName());
+                String replacement = ".filter(" + optionalType + "::isPresent).map(" + optionalType + "::get)";
```
pkoenig10, line 70: Should we use `orElseThrow` instead of `get`? The Java docs state that `orElseThrow` should be preferred and I've come to agree with this recommendation.

https://docs.oracle.com/javase/10/docs/api/java/util/Optional.html#get()

pkoenig10, line 70: Should we use `orElseThrow` instead of `get`? The Java docs state that `orElseThrow` should be preferred and I've come to agree with this recommendation.

https://docs.oracle.com/javase/10/docs/api/java/util/Optional.html#get()

Screenshot 2024-11-11 at 10 39 39 PM

schlosna: Good call https://github.com/palantir/gradle-baseline/pull/2950

ash211:
Current JDKs make us choose between what's "readable":

```
Stream<Optional<MyObject>> mystream;

mystream
  .flatMap(Optional::stream)
 ```

and what's efficient:

```
Stream<Optional<MyObject>> mystream;

mystream
  .filter(Optional::isPresent)
  .map(Optional::get)
```

Is there a possible API addition to the JDK that would be both readable and efficient?

**Comment by wi11dey** [comment]
noting that for projects that use StreamEx, `.mapPartial(Function.identity())` should have the same perf characteristics of `.filter(Optional::isPresent).map(Optional::get)` based on the implementation https://github.com/amaembo/streamex/blob/master/src/main/java/one/util/streamex/AbstractStreamEx.java#L768-L771, and is arguably as readable/idiomatic as `.flatMap(Optional::stream)`

**Comment by schlosna** [comment]
> Current JDKs make us choose between what's "readable":
> 
> ```
> Stream<Optional<MyObject>> mystream;
> 
> mystream
>   .flatMap(Optional::stream)
> ```
> 
> and what's efficient:
> 
> ```
> Stream<Optional<MyObject>> mystream;
> 
> mystream
>   .filter(Optional::isPresent)
>   .map(Optional::get)
> ```
> 
> Is there a possible API addition to the JDK that would be both readable and efficient?

Using [Stream `.mapMulti(Optional::ifPresent)` would be the readable & efficient JDK 16+ version](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/stream/Stream.html#mapMulti(java.util.function.BiConsumer)). See https://github.com/Nikolas-Charalambidis/java-16-mapmulti-benchmark (h/t @westinrm ).

I have not had cycles to try automating the refactor from `.flatMap(Optional::stream)` to `.<T>mapMulti(Optional::ifPresent)` as the generic type inference is a bit beyond my error-prone kung-fu and limited hacking time.

**Comment by pkoenig10** [comment]
Given https://github.com/palantir/gradle-baseline/pull/2898, I think it's safe to make that change.

We probably would also want to replace `.filter(Optional::isPresent).map(Optional::get)` with this.

> I have not had cycles to try automating the refactor from .flatMap(Optional::stream) to .<T>mapMulti(Optional::ifPresent) as the generic type inference is a bit beyond my error-prone kung-fu and limited hacking time.

One option is to cheat a bit and perform a regex match on the source code.

**Comment by schlosna** [comment]
Started a draft in https://github.com/palantir/gradle-baseline/pull/2996 but need to tweak the fix

**Comment by fawind** [comment]
~~I'm trying the `mapMulti` refactoring but running into compile errors when using the method reference on more complex stream pipelines:~~

```java
.mapMulti(Optional::ifPresent)
.collect(ImmutableList.toImmutableList()));
```

~~This fails compile with:~~

```
Required type: List<MyClass>
Provided: ImmutableList<Object>
no instance(s) of type variable(s) exist so that Object conforms to <MyClass
```

~~I can get it to compile by typing out the full signature though:~~

```java
.mapMulti((Optional<MyClass> maybeValue, Consumer<MyClass> consumer) ->
        maybeValue.ifPresent(consumer))
.collect(ImmutableList.toImmutableList()));
```

~~This is actually much more verbose then the previously used `.filter(#isPresent).map(#get)`. Is there a trick to make mapMulti work more nicely here?~~


**Edit:** Never mind, was just a bit lost. You can provide a type hint using `.<MyClass>mapMulti(Optional::ifPresent)` which solves this nicely! Errorprone's suggested refactoring is also handling this correctly already!

PR #2948, opened 2024-11-09
Replace Collectors.toUnmodifiableList() with Stream#toList()

schlosna:
## Before this PR

Uses of `stream.collect(Collectors.toUnmodifiableList())` could be replaced with `stream.toList()` added in JDK 16 and provides a more efficient lower allocation implementation that returns an immutable `List` that does not contain `null` elements.

Similar to https://github.com/palantir/gradle-baseline/pull/2946

## After this PR

==COMMIT_MSG==
[JDK 16 (in JDK-8256441)](https://bugs.openjdk.org/browse/JDK-8256441) added
[`Stream#toList()`](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/stream/Stream.html#toList()) provides an optimized version that uses JDK internals to avoid some of the additional array copies that make stream collection expensive.
==COMMIT_MSG==

## Possible downsides?

This error-prone check does not modify usages of `Collectors.toList()` as that returns a modifiable `List` that may contain `null` elements. It also does not modify usages of `ImmutableList.toImmutableList()` as some use sites leverage Guava `ImmutableList` types to avoid copies via `ImmutableList.copyOf(someImmutableList)`.

### DIFF
--- a/baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/StreamToListTest.java
+++ b/baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/StreamToListTest.java
@@ -1,0 +1,63 @@
+/*
+ * (c) Copyright 2024 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.baseline.errorprone;
+
+import org.junit.jupiter.api.Test;
+
+class StreamToListTest {
+    @Test
+    public void test() {
+        fix().addInputLines(
+                        "Test.java",
+                        "import java.util.List;",
+                        "import java.util.stream.Collectors;",
+                        "import java.util.stream.Stream;",
+                        "public class Test {",
+                        "  List<String> f0(Stream<String> in) {",
+                        "    return in.toList();",
+                        "  }",
+                        // Collectors.toList() supports nulls & is mutable while Stream#toList() does not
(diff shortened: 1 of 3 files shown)


baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/StreamToListTest.java
```diff
@@ -0,0 +1,63 @@
...
+        fix().addInputLines(
+                        "Test.java",
+                        "import java.util.List;",
+                        "import java.util.stream.Collectors;",
+                        "import java.util.stream.Stream;",
+                        "public class Test {",
+                        "  List<String> f0(Stream<String> in) {",
+                        "    return in.toList();",
+                        "  }",
+                        // Collectors.toList() supports nulls & is mutable while Stream#toList() does not
```
carterkozak, line 33: I think `Stream#toList()` is meant to allow nulls (based on a test this seems to be the case, and the javadoc suggests implementations should be equivalent to `Collections.unmodifiableList(new ArrayList<>(Arrays.asList(this.toArray())))`.

Unlike `Collectors.toList()`, `Collectors.toUnmodifiableList()` doesn't allow nulls, so this should be a safe replacement, but I think this comment is a bit off.

Confusingly `List.of` produces an immutable list which doesn't allow nulls, so the newer apis are all over the place wrt nullability...

pkoenig10:
It's not always possible to replace `toUnmodifiableList()` with `toList()` when using subclasses. The type signatures of these methods are:

```java
// Stream.collect 
<R, A> R collect(Collector<? super T, A, R> collector) 

// Collectors.toUnmodifiableList 
Collector<T, ?, List<T>> toUnmodifiableList()

// Stream.toList 
List<T> toList()
```

So if I have a `Stream<Dog>` that I want to turn into a `List<Animal>`, I can't use `toList()`.

This is a not uncommon pattern that I've hit a couple times. For example:

```java
List<Foo> foos = arguments.stream()
    .map(args -> new FooImpl(args))
    // This returns List<FooImpl>, not List<Foo>
    .toList();
```

You can work around this by explicitly declaring the `map` type parameter, but it is a but cumbersome. I don't think it would be unacceptable to say that we require this, but I just want to point this out.

```java
List<Foo> foos = arguments.stream()
    .<Foo>map(args -> new FooImpl(args))
    .toList();
```

carterkozak:
I think we can compare the generic component of the `Stream` receiver to the generic component of the `toUnmodifiableList()` method invocation, and only propose replacement when they match.

(unverified)

PR #2949, opened 2024-11-12
StreamOptionalGetWithoutFilter

schlosna:
## Before this PR

Discussion of https://github.com/palantir/gradle-baseline/pull/2946 identified potential value in identifying places where `Stream`s of `Optional`s should get if the optional is present before mapping via `Optional::get`.

## After this PR

==COMMIT_MSG==
Add `StreamOptionalGetWithoutFilter` to identify places where `Stream<Optional<?>>` should call `filter(Optional::isPresent)` before `map(Optional::get)`.

Similar to [`OptionalNotPresent`](https://errorprone.info/bugpattern/OptionalNotPresent) but for streams.
==COMMIT_MSG==


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StreamOptionalGetWithoutFilter.java
```diff
@@ -0,0 +1,91 @@
...
+        linkType = BugPattern.LinkType.CUSTOM,
+        severity = SeverityLevel.WARNING,
+        summary = "Stream<Optional<?>> should call filter(Optional::isPresent) before map(Optional::get)",
+        explanation = "Calling map(Optional::get) on a Stream<Optional<?>> without first calling "
+                + "filter(Optional::isPresent) can cause NoSuchElementException.")
+public final class StreamOptionalGetWithoutFilter extends BugChecker implements BugChecker.MethodInvocationTreeMatcher {
+
+    private static final Matcher<ExpressionTree> OPTIONAL_GET_METHOD = Matchers.staticMethod()
+            .onClass(Optional.class.getName())
+            .named("get")
```
pkoenig10, line 49: Probably also want to include `orElseThrow` here.

pkoenig10 (commented):
How confident are we that there are no legitimate uses to call `.map(Optional::get)` without filtering? I can't think of an specific uses, but I wouldn't be surprised if there were some.

carterkozak:
I'm not sure it's worth maintaining a check for this, unless we see failures begin to crop up in practice. I can imagine false positives causing more total friction than a couple failures getting through. It's certainly possible that my estimation is off, though.

schlosna:
Superseded by https://github.com/palantir/gradle-baseline/pull/2996

PR #2950, opened 2024-11-12
Prefer Optional::orElseThrow over Optional::get

schlosna:
## Before this PR

https://github.com/palantir/gradle-baseline/pull/2946#discussion_r1837439292

## After this PR

==COMMIT_MSG==
Prefer Optional::orElseThrow over Optional::get
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-2950.v2.yml
+++ b/changelog/@unreleased/pr-2950.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Prefer Optional::orElseThrow over Optional::get
+  links:
+  - https://github.com/palantir/gradle-baseline/pull/2950
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StreamFlatMapOptional.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StreamFlatMapOptional.java
@@ -39,7 +39,7 @@ import java.util.stream.Stream;
         link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
         linkType = BugPattern.LinkType.CUSTOM,
         severity = BugPattern.SeverityLevel.WARNING,
-        summary = "`Stream.filter(Optional::isPresent).map(Optional::get)` is more efficient than "
+        summary = "`Stream.filter(Optional::isPresent).map(Optional::orElseThrow)` is more efficient than "
                 + "`Stream.flatMap(Optional::stream)`")
 public final class StreamFlatMapOptional extends BugChecker implements BugChecker.MethodInvocationTreeMatcher {
     private static final long serialVersionUID = 1L;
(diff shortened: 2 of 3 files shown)


changelog/@unreleased/pr-2950.v2.yml
```diff
@@ -0,0 +1,5 @@
+type: improvement
+improvement:
+  description: Prefer Optional::orElseThrow over Optional::get
```
schlosna, line 3: Open question: Do we also want a more generic `PreferOptionalOrElseThrow` check to rewrite `Optional::get` & `opt.get()` to `Optional::orElseThrow` & `opt.orElseThrow()`

pkoenig10: I'd be in favor of this. Especially to fix consumers that have already applied `StreamFlatMapOptional`.

pkoenig10:
I'm in favor of this, but would like @carterkozak to weigh in.

carterkozak:
I suppose I'd mentally modeled `Optional#get` as "I've verified that the optional is present, and would like to extract the value", and `Optional#orElseThrow` as "Extract the value, but throwing is reasonable if the optional is empty", and sort of discounted using `orElseThrow` without args in favor of providing a more specific SafeLoggable exception. I'd never put much thought into this distinction, and I'm not sure it makes sense to have two ways of doing basically the same thing. It's worth verifying if `Optional` support in `NullAway` guards `get()` and `orElseThrow()` in the same way, or only checks `get()` but considers `orElseThrow()` as a guard.

pkoenig10:
I've come to prefer `orElseThrow()` primarily for readability. With `get()`, it can be non-obvious that a particular piece of code can throw. This is especially valuable if:
- The code that checks if the value is present is, for whatever reason, further away from the code that extracts the value.
- The `get()` is called directly rather than with a method reference, so it's less obvious that the type of the value is an `Optional`. Particularly since we have a number of other types with `get()` methods, like Conjure aliases.

pkoenig10:
`orElseThrow()` was introduced in [JDK-8140281](https://bugs.openjdk.org/browse/JDK-8140281) for exactly this reason. The author ultimately plans to deprecate `get()` in [JDK-8160606](https://bugs.openjdk.org/browse/JDK-8160606) - although I suspect that will never happen.

carterkozak:
Given [JDK-8160606](https://bugs.openjdk.org/browse/JDK-8160606), I agree. I suspect you're right that `get` may not actually be removed, but it seems reasonable to align with the JDKs design goals.

Unfortunately it seems that `NullAway:CheckOptionalEmptiness` only protects `Optional.get()`, not `Optional.orElseThrow()`. Only a few projects use this option, since we don't enable it by default anymore after some projects hit bugs: https://github.com/palantir/gradle-baseline/pull/2396
Probably worth looking into re-enabling, though. We've upgraded NullAway several times since that issue.

schlosna:
Closing in favor of draft https://github.com/palantir/gradle-baseline/pull/2996

PR #2954, opened 2024-11-14
Bump suppressible-error-prone to 1.5.0 to get Java 11 target for libs used in compilation

CRogers:
## Before this PR
More information in https://github.com/palantir/suppressible-error-prone/pull/22

## After this PR
==COMMIT_MSG==
Bump suppressible-error-prone to 1.5.0 to get Java 11 target for libs used in compilation
==COMMIT_MSG==


felixdesouza:
👍

PR #2962, opened 2024-11-22
promote PreferCommonAnnotations to WARNING

rhuffy:
## Before this PR

Check introduced in #2505. SeverityLevel.SUGGESTION makes it easy to miss.

## After this PR

==COMMIT_MSG==
promote PreferCommonAnnotations to WARNING
==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferCommonAnnotations.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferCommonAnnotations.java
@@ -39,7 +39,7 @@ import java.util.Objects;
 @AutoService(BugChecker.class)
 @BugPattern(
         summary = "Prefer the common version of annotations over other copies.",
-        severity = SeverityLevel.SUGGESTION)
+        severity = SeverityLevel.WARNING)
 public final class PreferCommonAnnotations extends BugChecker implements ImportTreeMatcher {
 
     /** ClassName -> preferred import. */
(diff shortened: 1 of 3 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferCommonAnnotations.java
```diff
@@ -39,7 +39,7 @@
 @AutoService(BugChecker.class)
 @BugPattern(
         summary = "Prefer the common version of annotations over other copies.",
-        severity = SeverityLevel.SUGGESTION)
+        severity = SeverityLevel.WARNING)
 public final class PreferCommonAnnotations extends BugChecker implements ImportTreeMatcher {
```
CRogers: imports can't be suppressed using `@SuppressedWarnings`, so this cannot be auto-suppressed. If it had an autofix that worked every time, it would be fine to make this error level. But I don't think this works every time as repos may not have guava as an import (that said, we could make this a ["ConditionalPatchCheck"](https://github.com/palantir/suppressible-error-prone#:~:text=//%20You%20can%20also%20conditionally%20opt%20into%20patch%20checks%20if%20a%20module%20is%20in%20the%20transitive%0A%20%20%20%20%20%20%20%20//%20runtimeClasspath%20of%20the%20source%20set%3A) on the presence of guava, which should make it safe).

I don't actually know how suppressible-error-prone behaves with errors on imports - maybe try running this on an internal repo using maven local and see what happens.

PR #2966, opened 2024-11-26
baseline-error-prone targets jdk-17+

carterkozak:
==COMMIT_MSG==
baseline-error-prone targets jdk-17+
==COMMIT_MSG==

Risk is that we lose errorprone support for projects targeting jdk11, where we use a jdk11 for compilation.
In those cases, failure modes range from silently ignoring error-prone to build failures. Given that we do not support jdk11 anymore, this is a reasonable risk.

### DIFF
--- a/baseline-error-prone/build.gradle
+++ b/baseline-error-prone/build.gradle
@@ -2,7 +2,7 @@ apply plugin: 'java-library'
 apply plugin: 'com.palantir.external-publish-jar'
 
 javaVersion {
-    target = 11
+    target = 17
     runtime = 21
 }
 
--- a/changelog/@unreleased/pr-2966.v2.yml
+++ b/changelog/@unreleased/pr-2966.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: baseline-error-prone targets jdk-17+
+  links:
+  - https://github.com/palantir/gradle-baseline/pull/2966


pkoenig10:
@carterkozak what would it take to merge this? This is [blocking](https://app.circleci.com/pipelines/github/palantir/gradle-baseline/7466/workflows/583381ab-deb2-4717-9077-e3371bc054ee/jobs/40817) the Error Prone upgrade in https://github.com/palantir/gradle-baseline/pull/2982, which is blocking some upgrades due to the typical:
- Error Prone break their API (https://github.com/google/error-prone/commit/3b16d6147af7b8ac0a682893daddf9b0055e5f89)
- Guava upgrades to this version of Error Prone (https://github.com/google/guava/commit/cf508b773c389b7ab9b3d8ecf991ef2e4b681717)
- Another Error Prone checks library upgrades to this version of Guava
- Upgrading these custom Error Prone check libraries in consumers fails with `java.lang.NoSuchMethodError: 'boolean com.google.errorprone.util.ASTHelpers.isRuleKind(com.sun.source.tree.CaseTree)'`

As of 2.33.0, Error Prone requires compiling using Java 17.

https://github.com/google/error-prone/releases/tag/v2.33.0

It seems like `com.palantir.baseline-java-versions` doesn't support using a newer compiler to target older versions, which is what Error Prone requires.

pkoenig10:
Should we consider not auto-upgrading Guava in Error Prone check repos? This is [not the first time](https://github.com/palantir/gradle-baseline/pull/2839) this specific failure mode has happened.

pkoenig10 (dismissed):

pkoenig10 (approved):

PR #2984, opened 2024-12-18
failing testcases for safe-logging check when using Runnables/Suppliers/etc

Blizzara:
## Before this PR

I tried to see if I could fix this myself, but I lack the knowledge of this meta-java stuff. I did see in debugger that at [this line](https://github.com/palantir/gradle-baseline/blob/2270796a5160f4a1b61d8b3690a323fdb3cfe8ec/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/IllegalSafeLoggingArgument.java#L168), the `parameterSafety` is correctly `SAFE`, and the `argument` is an `@Unsafe Object value`, but then the SafetyAnalysis returns as `UNKNOWN` leading to the check below passing. But not sure if this is even the correct place to catch it, or if it should rather be caught somewhere else.

Also my `//` asserts in the test cases might be wrong, but if you remove them, the tests will pass with no errors - which they shouldn't, I think.

## After this PR

==COMMIT_MSG==
==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/IllegalSafeLoggingArgumentTest.java
+++ b/baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/IllegalSafeLoggingArgumentTest.java
@@ -2121,6 +2121,51 @@ class IllegalSafeLoggingArgumentTest {
                 .doTest();
     }
 
+    @Test
+    public void testInterfaceDefaultMethodReturningRunnable() {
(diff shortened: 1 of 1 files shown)


carterkozak:
Thanks for this report, I've got a proposed fix up here, including one of your tests: #2988
I've verified all three do in fact pass, but the issue is the same in all of them, so it doesn't matter if the lambda is returned/assigned/ignored.

PR #2986, opened 2025-01-07
baseline-error-prone takes advantage of jdk17 compiler APIs

carterkozak:
This removes some reflection in favor of direct method invocations now that we target JDK-17 bytecode.
Reflection is expensive, and harder to read.

==COMMIT_MSG==
baseline-error-prone takes advantage of jdk17 compiler APIs
==COMMIT_MSG==


mpritham (approved):
lgtm

PR #2988, opened 2025-01-08
safe-logging analysis correctly handles out-of-scope parameter references

carterkozak:
Reported in #2984

==COMMIT_MSG==
safe-logging analysis correctly handles out-of-scope parameter references
==COMMIT_MSG==

Note that this has also been updated to improve the `isPatternBinding` check using java 17 target compatibility now that we've upgraded.
This should dramatically improve performance in areas which reference parameters in lambdas as well, as we can avoid an additional set of compilations :-)

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/safety/SafetyPropagationTransfer.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/safety/SafetyPropagationTransfer.java
@@ -870,14 +870,16 @@ public final class SafetyPropagationTransfer implements ForwardTransferFunction<
             // 4. Pattern matching (input instanceof String str)
 
             // Cast a wide net for all throwables (covers catch statements)
+            Safety treeSafety = SafetyAnnotations.getSafety(node.getTree(), state);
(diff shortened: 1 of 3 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/safety/SafetyPropagationTransfer.java
```diff
@@ -884,24 +886,9 @@ public TransferResult<Safety, AccessPathStore<Safety>> visitLocalVariable(
...
-            return false;
-        }
-        Tree enclosing = parentPath.getLeaf();
-        if (enclosing == null) {
-            return false;
-        }
-        // JCBindingPattern is newer than our compilation target, so we match strings to avoid
-        // complex build-system configurations.
-        String kindString = enclosing.getKind().name();
-        return "BINDING_PATTERN".equals(kindString);
```
carterkozak, line 904: not sure why I originally checked for an enclosing binding pattern instead of the binding variable type before, most likely searching in the dark based on older jdk-11 sources

ash211:
Looks like this would catch some log safety issues in my repo.  Looking forward to having this additional protection land!

mpritham (approved):
Thanks!

PR #2990, opened 2025-01-09
Test case: missed BUG for captured lambda parameter type's safety annotation

ash211:
I would expect the test to pass as written. I can get it to pass by swapping to the commented-out line, which has `new UnsafeClass()` directly instead of as a parameter.


ash211:
Looks like this would be solved by active PR https://github.com/palantir/gradle-baseline/pull/2988

PR #2996, opened 2025-01-15
`Stream.mapMulti(Optional::ifPresent)` is more efficient than `Stream.flatMap(Optional::stream)`

schlosna:
## Before this PR

https://github.com/palantir/gradle-baseline/pull/2946 added a check to convert `Stream.flatMap(Optional::stream)` to more efficient `Stream.filter(Optional::isPresent).map(Optional::get)`. There was discussion in https://github.com/palantir/gradle-baseline/pull/2950 whether the autofix should prefer `Optional::orElseThrow` over `Optional::get` and using [`Stream.mapMulti(Optional::ifPresent)` as a more concise and efficient alternative](https://github.com/palantir/gradle-baseline/pull/2946#issuecomment-2593654541).

## After this PR

==COMMIT_MSG==

Converts `Stream.flatMap(Optional::stream)`, `Stream.filter(Optional::isPresent).map(Optional::get)`, and `filter(Optional::isPresent).map(Optional::orElseThrow)` to `Stream.mapMulti(Optional::ifPresent)`. 

==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StreamFlatMapOptional.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StreamFlatMapOptional.java
@@ -35,6 +35,7 @@ import com.sun.source.tree.MethodInvocationTree;
 import com.sun.tools.javac.code.Type;
 import java.io.Serial;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.Optional;
 import java.util.function.Function;
 import java.util.function.Predicate;
@@ -124,17 +125,11 @@ public final class StreamFlatMapOptional extends BugChecker implements MethodInv
 
         SuggestedFix.Builder fix = SuggestedFix.builder();
 
-        long receiverCount = args(ASTHelpers.getReceiverType(receiver)).count();
-        Stream<Type> args = args(ASTHelpers.getReceiverType(expressionTree));
-        long argCount = args.count();
-        boolean shouldQualifyType =
-                args(elementType).findAny().isPresent() || receiverCount > 1 || (receiverCount == 0 && argCount > 0);
-
         String qualifiedType = "";
-        if (shouldQualifyType) {
+        if (needsQualification(receiver, expressionTree, elementType)) {
             qualifiedType = qualifyType(state, fix, receiverType.getTypeArguments());
             if (qualifiedType.isEmpty()) {
-                qualifiedType = SuggestedFixes.qualifyType(state, fix, elementType);
+                qualifiedType = qualifyType(state, fix, Collections.singleton(elementType));
             }
         }
 
@@ -146,8 +141,19 @@ public final class StreamFlatMapOptional extends BugChecker implements MethodInv
                 .build();
     }
 
+    private static boolean needsQualification(
+            ExpressionTree receiver, ExpressionTree expressionTree, Type elementType) {
+        long receiverCount = args(ASTHelpers.getReceiverType(receiver)).count();
+        long treeArgCount = args(ASTHelpers.getReceiverType(expressionTree)).count();
+        return args(elementType).findAny().isPresent() || receiverCount > 1 || (receiverCount == 0 && treeArgCount > 0);
+    }
+
     private static String qualifyType(
             VisitorState state, SuggestedFix.Builder fix, Collection<Type> receiverTypeArguments) {
+        if (receiverTypeArguments.isEmpty()
+                || args(receiverTypeArguments).findAny().isEmpty()) {
+            return "";
+        }
         return args(receiverTypeArguments)
                 .map(type -> SuggestedFixes.qualifyType(state, fix, type))
                 .collect(Collectors.joining(", ", "<", ">"));
@@ -157,7 +163,7 @@ public final class StreamFlatMapOptional extends BugChecker implements MethodInv
         return args(type.getTypeArguments());
     }
 
-    private static Stream<Type> args(Collection<Type> types) {
-        return types.stream().flatMap(type -> type.getTypeArguments().stream());
+    static Stream<Type> args(Collection<Type> types) {
+        return types.stream().flatMap(t -> t.getTypeArguments().stream());
     }
 }
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StreamFlatMapOptional.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StreamFlatMapOptional.java
@@ -41,7 +41,6 @@ import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
-import org.jspecify.annotations.Nullable;
 
 @AutoService(BugChecker.class)
 @BugPattern(
@@ -92,27 +91,22 @@ public final class StreamFlatMapOptional extends BugChecker implements MethodInv
     @Override
     public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
         if (STREAM_FLATMAP_OPTIONAL_STREAM.matches(tree, state)) {
-            ExpressionTree receiver = ASTHelpers.getReceiver(tree);
-            return fix(tree, state, receiver, receiver);
+            return fix(tree, state, tree);
         }
 
         if (STREAM_MAP_GET.matches(tree, state)) {
             ExpressionTree mapTree = ASTHelpers.getReceiver(tree);
             if (mapTree != null && STREAM_FILTER_IS_PRESENT.matches(mapTree, state)) {
-                ExpressionTree filterTree = ASTHelpers.getReceiver(mapTree);
-                return fix(tree, state, tree.getMethodSelect(), filterTree);
+                return fix(tree, state, mapTree);
             }
         }
 
         return Description.NO_MATCH;
     }
 
-    private Description fix(
-            MethodInvocationTree tree,
-            VisitorState state,
-            @Nullable ExpressionTree receiver,
-            @Nullable ExpressionTree expressionTree) {
-        if (receiver == null || expressionTree == null) {
+    private Description fix(MethodInvocationTree tree, VisitorState state, ExpressionTree expressionTree) {
+        ExpressionTree receiver = ASTHelpers.getReceiver(expressionTree);
+        if (receiver == null) {
             return Description.NO_MATCH;
         }
 
@@ -121,16 +115,10 @@ public final class StreamFlatMapOptional extends BugChecker implements MethodInv
             return Description.NO_MATCH;
         }
 
-        Type receiverType = ASTHelpers.getType(receiver);
-        if (receiverType == null) {
-            return Description.NO_MATCH;
-        }
-
         // try to elide the type arguments for mapMulti if it compiles, otherwise fallback on including qualified types
-        SuggestedFix fix = Optional.of(getSuggestedFix(SuggestedFix.builder(), tree, state, expressionTree, ""))
+        SuggestedFix fix = Optional.of(getSuggestedFix(SuggestedFix.builder(), tree, state, receiver, ""))
                 .filter(f -> SuggestedFixes.compilesWithFix(f, state))
-                .orElseGet(
-                        () -> getQualifiedSuggestedFix(tree, state, expressionTree, receiverType.getTypeArguments()));
+                .orElseGet(() -> getQualifiedSuggestedFix(tree, state, receiver, List.of(elementType)));
         return buildDescription(tree).addFix(fix).build();
     }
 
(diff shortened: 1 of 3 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StreamFlatMapOptional.java
```diff
@@ -67,7 +68,13 @@ public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState
...
                 SuggestedFix.Builder fix = SuggestedFix.builder();
                 String optionalType = SuggestedFixes.qualifyType(state, fix, Optional.class.getCanonicalName());
-                String replacement = ".filter(" + optionalType + "::isPresent).map(" + optionalType + "::get)";
+                Type receiverType = ASTHelpers.getType(receiver);
+                com.sun.tools.javac.util.List<Type> receiverTypeArguments = receiverType.getTypeArguments();
+                com.sun.tools.javac.util.List<Type> optionalTypeArgs = receiverTypeArguments.isEmpty()
+                        ? com.sun.tools.javac.util.List.nil()
+                        : receiverTypeArguments.get(0).getTypeArguments();
+                String type = typeArgument(optionalTypeArgs, state, fix);
+                String replacement = "." + type + "mapMulti(" + optionalType + "::ifPresent)";
```
carterkozak: I think we may be able to simplify a bit using `ASTHelpers.getType(tree)`, which gives us `Stream<T>` as opposed to unwrapping `Stream<Optional<T>>`

schlosna: thanks, updated to use ASTHelpers as well as handling `Stream.filter(Optional::isPresent).map(Optional::get)` and `filter(Optional::isPresent).map(Optional::orElseThrow)`.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StreamFlatMapOptional.java
```diff
@@ -30,50 +32,132 @@
...
-                        .build();
+            ExpressionTree methodSelect = tree.getMethodSelect();
+            ExpressionTree receiver = ASTHelpers.getReceiver(methodSelect);
+            return fix(tree, state, receiver, receiver);
+        }
+
+        if (STREAM_MAP_GET.matches(tree, state)) {
+            ExpressionTree mapTree = ASTHelpers.getReceiver(tree.getMethodSelect());
+            if (mapTree != null && STREAM_FILTER_IS_PRESENT.matches(mapTree, state)) {
+                ExpressionTree filterTree = ASTHelpers.getReceiver(mapTree);
```
carterkozak: I don't think this should ever be null, but we may want to add a null check out of an abundance of caution

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StreamFlatMapOptional.java
```diff
@@ -30,50 +32,132 @@
...
-                return buildDescription(tree)
-                        .addFix(fix.replace(state.getEndPosition(receiver), state.getEndPosition(tree), replacement)
-                                .build())
-                        .build();
+            ExpressionTree methodSelect = tree.getMethodSelect();
+            ExpressionTree receiver = ASTHelpers.getReceiver(methodSelect);
+            return fix(tree, state, receiver, receiver);
+        }
+
+        if (STREAM_MAP_GET.matches(tree, state)) {
```
carterkozak, line 97: Unclear if this is the best fit for this check, vs a separate error-prone check (if it's uncommon outside of the previous automatic fix, we could even run it via excavator without enforcing every compilation).

Certainly easiest to roll out here, and given the current state of the world it probably makes the most sense

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StreamFlatMapOptional.java
```diff
@@ -30,50 +32,132 @@
...
+            return Description.NO_MATCH;
+        }
+
+        SuggestedFix.Builder fix = SuggestedFix.builder();
+
+        long receiverCount = args(ASTHelpers.getReceiverType(receiver)).count();
+        Stream<Type> args = args(ASTHelpers.getReceiverType(expressionTree));
+        long argCount = args.count();
+        boolean shouldQualifyType =
+                args(elementType).findAny().isPresent() || receiverCount > 1 || (receiverCount == 0 && argCount > 0);
```
carterkozak: Worth running this on a couple larger internal projects to verify the results compile successfully. If there are edge cases we can't easily isolate, I think it should be safe to over-qualify.

schlosna: Yeah, this logic isn't quite right yet. I've tested out on a few local repos and there are some places were I'm not generating qualified types that need them.

I'm trying to avoid adding unnecessary qualifying type args, but there are at least a few cases I haven't quite nailed down that need qualification (e.g. some stream chains extracting layers of optionals).

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StreamFlatMapOptional.java
```diff
@@ -30,50 +32,147 @@
...
 
 @AutoService(BugChecker.class)
 @BugPattern(
         link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
         linkType = BugPattern.LinkType.CUSTOM,
-        severity = BugPattern.SeverityLevel.WARNING,
-        summary = "`Stream.filter(Optional::isPresent).map(Optional::get)` is more efficient than "
-                + "`Stream.flatMap(Optional::stream)`")
-public final class StreamFlatMapOptional extends BugChecker implements BugChecker.MethodInvocationTreeMatcher {
+        severity = SeverityLevel.SUGGESTION,
```
schlosna:
```suggestion
        severity = SeverityLevel.WARNING,
```

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StreamFlatMapOptional.java
schlosna: yeah, this should be much less common after automated roll out, though folks may still write `.filter(Optional::isPresent).map(Optional::get)` so this would keep that cleaned up.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StreamFlatMapOptional.java
schlosna: added check into `fix` method as `receiver` and `expressionTree` are nullable

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StreamFlatMapOptional.java
carterkozak: > I'm trying to avoid adding unnecessary qualifying type args, but there are at least a few cases I haven't quite nailed down that need qualification (e.g. some stream chains extracting layers of optionals).

Thinking about this a bit more, I think it's actually a good fit for the incredibly expensive `SuggestedFixes.compilesWithFix` utility, because in the failing case, we can qualify the method reference, and needn't pay the cost again (unlike `LambdaMethodReference` which used `compilesWithFix` in the matching component, meaning it must run the test in every subsequent compilation).

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StreamFlatMapOptional.java
```diff
@@ -30,50 +32,146 @@
...
+        }
+
+        return Stream.<Supplier<SuggestedFix>>of(
+                        () -> getSuggestedFix(SuggestedFix.builder(), tree, state, expressionTree, ""),
+                        () -> getQualifiedSuggestedFix(tree, state, expressionTree, receiverType.getTypeArguments()),
+                        () -> getQualifiedSuggestedFix(tree, state, expressionTree, List.of(elementType)))
+                .<SuggestedFix>mapMulti((f, consumer) -> {
+                    SuggestedFix fix = f.get();
+                    if (SuggestedFixes.compilesWithFix(fix, state)) {
+                        consumer.accept(fix);
```
carterkozak: I think we should guarantee that we propose _something_ after checking `compilesWithFix`, this could make future compilations expensive if compilation doesn't succeed on any of the options

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StreamFlatMapOptional.java
```diff
@@ -30,50 +32,146 @@
...
+
+        Type receiverType = ASTHelpers.getType(receiver);
+        if (receiverType == null) {
+            return Description.NO_MATCH;
+        }
+
+        return Stream.<Supplier<SuggestedFix>>of(
+                        () -> getSuggestedFix(SuggestedFix.builder(), tree, state, expressionTree, ""),
+                        () -> getQualifiedSuggestedFix(tree, state, expressionTree, receiverType.getTypeArguments()),
+                        () -> getQualifiedSuggestedFix(tree, state, expressionTree, List.of(elementType)))
```
carterkozak: I think these produce the same result in the tests, as long as one of the lines is present, everything seems to pass. Are there any known cases where we need both?

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StreamFlatMapOptional.java
```diff
@@ -30,50 +32,146 @@
...
-            if (receiver != null) {
-                SuggestedFix.Builder fix = SuggestedFix.builder();
-                String optionalType = SuggestedFixes.qualifyType(state, fix, Optional.class.getCanonicalName());
-                String replacement = ".filter(" + optionalType + "::isPresent).map(" + optionalType + "::get)";
-                return buildDescription(tree)
-                        .addFix(fix.replace(state.getEndPosition(receiver), state.getEndPosition(tree), replacement)
-                                .build())
-                        .build();
+            ExpressionTree methodSelect = tree.getMethodSelect();
+            ExpressionTree receiver = ASTHelpers.getReceiver(methodSelect);
```
carterkozak: `ASTHelpers.getReceiver` follows `MethodInvocationTree.getMethodSelect()`, we can pass `tree` directly for the same result
```suggestion
            ExpressionTree receiver = ASTHelpers.getReceiver(tree);
```

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StreamFlatMapOptional.java
```diff
@@ -30,50 +32,146 @@
...
+            return fix(tree, state, receiver, receiver);
+        }
+
+        if (STREAM_MAP_GET.matches(tree, state)) {
+            ExpressionTree mapTree = ASTHelpers.getReceiver(tree.getMethodSelect());
```
carterkozak:
```suggestion
            ExpressionTree mapTree = ASTHelpers.getReceiver(tree);
```

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StreamFlatMapOptional.java
schlosna: I think we can remove the latter

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StreamFlatMapOptional.java
```diff
@@ -128,19 +126,12 @@ private Description fix(
...
-                    }
-                })
-                .findFirst()
-                .map(fix -> buildDescription(tree).addFix(fix).build())
-                .orElse(Description.NO_MATCH);
+        // try to elide the type arguments for mapMulti if it compiles, otherwise fallback on including qualified types
+        SuggestedFix fix = Optional.of(getSuggestedFix(SuggestedFix.builder(), tree, state, expressionTree, ""))
+                .filter(f -> SuggestedFixes.compilesWithFix(f, state))
+                .orElseGet(
+                        () -> getQualifiedSuggestedFix(tree, state, expressionTree, receiverType.getTypeArguments()));
```
carterkozak: If it's safe to replace `receiverType.getTypeArguments()` with `List.of(elementType)`, we can remove the `receiver` argument to this function entirely. I'd imagine that should be safe, since we're pulling the resulting stream type of the original function (e.g. what's returned by `.map(Optional::get)`) as opposed to the intermediate/input state of the stream such a function is invoked upon.

schlosna: thanks, that updated and simplified things quite a bit

carterkozak (approved):
Looks great, thanks!

I've added do-not-merge in order to roll out the refactor via an excavator ahead of merging this into gradle-baseline, that way we decouple most of the code churn from the baseline upgrade. I've tagged you on the excavator which has largely identical error-prone code -- let me know if that rollout strategy seems reasonable

carterkozak:
@schlosna Feel free to remove the do-not-merge label whenever you like, the lions share of these have successfully merged already

Issue #2997, opened 2025-01-16
Improve slow error prones: LambdaMethodReference and UnnecessarilyQualified


ash211:
I ran with `-PerrorProneTimings` from https://github.com/palantir/suppressible-error-prone/pull/38 on an internal repo and had a gradle module with these numbers:

```
Total errorprone time: 71.293000 seconds 
 
ErrorProneName: <time taken> (percentage of all errorprone time) 
 
LambdaMethodReference                            : 29.449000 seconds (41.31%) 
UnnecessarilyQualified                           : 12.412000 seconds (17.41%) 
IllegalSafeLoggingArgument                       : 5.425000 seconds  (7.61%) 
NullAway                                         : 4.582000 seconds  (6.43%) 
SafeLoggingPropagation                           : 1.283000 seconds  (1.80%) 
ParameterName                                    : 1.026000 seconds  (1.44%)
```

I'm surprised that LambdaMethodReference and UnnecessarilyQualified are so expensive. Especially compared to SafeLoggingPropagation, which seems like it'd be much more expensive to trace log safety through a codebase.

I've disabled these two slow error prones on a large internal repo and saw non-incremental compile speedup from 10m27s to 7m9s.

To re-enable these at some point in the future, is there anything we can do to improve performance and reduce runtime of these checks?

pkoenig10:
I think `UnnecessarilyQualified` is so expensive because it matches on every `MemberSelectTree` and iterates through the imports. Member select trees are incredibly common. I imagine that we can somehow cache/precompute the import information per compilation tree in a way that only requires a constant time lookup.

carterkozak:
I think there's some very low hanging fruit around `UnnecessarilyQualified`, I can take a quick crack at it shortly

carterkozak:
See #2998

carterkozak:
We should probably not run `LambdaMethodReference` by default -- I'd implemented it initially in order to clean up lambdas left by other refactors (e.g. `OptionalOrElseGetMethodInvocation`, see the example from the original PR: #763).

It's particularly expensive because it's difficult to determine whether generic type inference will work correctly, so we use the `SuggestedFixes.compilesWithFix` utility to attempt compilation with proposed changes. Compilation is expensive! Over time, the build will succeed with more and more of these sub-compilation-failures for lambdas that cannot be replaced, multiplying out the compilation times in some areas.

[...]

PR #2998, opened 2025-01-16
Dramatically simplify the UnnecessarilyQualified check

carterkozak:
## Before this PR

See #2997

Previously we did a rather invasive search on every MemberSelectTree.

## After this PR

Now, we first check that the symbol is a type, and the select is operating upon a package, ruling out unnecessary treepath traversals and import scans.

==COMMIT_MSG==
Dramatically simplify the UnnecessarilyQualified error-prone check
==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/UnnecessarilyQualifiedTest.java
+++ b/baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/UnnecessarilyQualifiedTest.java
@@ -50,6 +50,31 @@ class UnnecessarilyQualifiedTest {
                 .doTest();
     }
 
+    @Test
+    void testEnclosedClassNotFlaggedWhenImportIsPresent() {
+        fix().addInputLines(
+                        "Test.java",
+                        "import java.util.Map;",
+                        "import java.util.Map.Entry;",
+                        "import java.util.Set;",
+                        "class Test {",
+                        "  java.util.Map.Entry<?, ?> get() {",
+                        "    return null;",
+                        "  }",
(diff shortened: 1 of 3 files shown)


baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/UnnecessarilyQualifiedTest.java
```diff
@@ -50,6 +50,31 @@ void testUnnecessarilyQualifiedType() {
...
+                        "    return null;",
+                        "  }",
+                        "}")
+                .addOutputLines(
+                        "Test.java",
+                        "import java.util.Map;",
+                        "import java.util.Map.Entry;",
+                        "import java.util.Set;",
+                        "class Test {",
+                        "  Map.Entry<?, ?> get() {",
```
carterkozak, line 71: Some may prefer for this to end up as `Entry` rather than `Map.Entry`, but I'd argue that either one is better than `java.util.Map.Entry` (and I prefer to list enclosing classes rather than importing them, but that's a stylistic choice).

pkoenig10 (approved):

ash211:
Thank you for this change -- it reduced `UnnecessarilyQualified` time by over 100x on an internal repo!

Before:
```
UnnecessarilyQualified                           : 12.412000 seconds (17.41%) 
```

After:
```
UnnecessarilyQualified                           : 108 millis       (0.74%)
```

An unscientific test -- there other changes in the repo than just this change. But it is vastly better now, and I'm going to turn this back on.

PR #2999, opened 2025-01-17
Disable expensive LambdaMethodReference by default

carterkozak:
See #2997

LambdaMethodReference is incredibly expensive, even compared to log safety flow analysis! It was originally added to clean up the results of other excavators, not to prevent runtime failures. We will apply the cleanup asynchronously via an excavator, rather than on a per-compilation basis, delayed fixups are a reasonable tradeoff here to dramatically improve compilation performance.

While we could move the code entirely into excavator instead of maintaining it here, however that has some risk for folks who explicitly configure/disable the `LambdaMethodReference` check (specifically, they would fail with:
`LambdaMethodReference not a valid checker name`)

==COMMIT_MSG==
Disable expensive LambdaMethodReference by default
==COMMIT_MSG==


mpritham (approved):
Thanks!

PR #3024, opened 2025-01-28
Fix StreamFlatMapOptional wildcard types

schlosna:
## Before this PR

The auto-fixes for #2996 would not generate proper capturing wildcard type arguments.

## After this PR

==COMMIT_MSG==
Properly handle generic wildcards in StreamFlatMapOptional generated fixes.
==COMMIT_MSG==


carterkozak (approved):
thanks!

PR #3041, opened 2025-02-18
Handle safety analysis for `this` and `super` references

carterkozak:
## Before this PR
Previously we failed to handle safety metadata for `this` and `super` references similarly to other variable references. See the new tests for examples. Most notably, this meant one could implement a `SafeLoggable` exception thusly without a warning:
```java
@Override
public String getLogMessage() {
    return getMessage(); // oh no! This is dangerous!
}
```
The danger is subtler than one might realize -- if such an exception is created with a known, safe message, it may be fine, however if it's constructed with a ctor that delegates to `public Throwable(Throwable cause)`, then `cause.toString()` will be used as the exception message, which is almost certainly unsafe, even (or especially!) if the cause implements `SafeLoggable` due to the following:
```java
public Throwable(Throwable cause) {
    fillInStackTrace();
    detailMessage = (cause==null ? null : cause.toString());
    this.cause = cause;
}
```

## After this PR
==COMMIT_MSG==
Handle safety analysis for `this` and `super` references
==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/safety/SafetyPropagationTransfer.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/safety/SafetyPropagationTransfer.java
@@ -49,7 +49,9 @@ import com.sun.tools.javac.code.Symbol;
 import com.sun.tools.javac.code.Symbol.ClassSymbol;
 import com.sun.tools.javac.code.Symbol.MethodSymbol;
 import com.sun.tools.javac.code.Symbol.RecordComponent;
+import com.sun.tools.javac.code.Symbol.TypeSymbol;
 import com.sun.tools.javac.code.Symbol.VarSymbol;
+import com.sun.tools.javac.code.Type;
 import com.sun.tools.javac.code.Type.ClassType;
 import com.sun.tools.javac.processing.JavacProcessingEnvironment;
 import java.io.Closeable;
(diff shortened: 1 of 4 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/safety/SafetyPropagationTransfer.java
```diff
@@ -1058,25 +1060,40 @@ public TransferResult<Safety, AccessPathStore<Safety>> visitArrayAccess(
...
+        return visitThisOrSuper(node, input);
     }
 
     @Override
     public TransferResult<Safety, AccessPathStore<Safety>> visitSuper(
             SuperNode node, TransferInput<Safety, AccessPathStore<Safety>> input) {
+        return visitThisOrSuper(node, input);
+    }
+
+    public TransferResult<Safety, AccessPathStore<Safety>> visitThisOrSuper(
```
bjlaub: does this need to be public?

carterkozak: nope, good catch!

carterkozak: Updated

bjlaub (approved):

PR #3048, opened 2025-02-21
[Safe-logging] Verify safety of returns in lambda expressions

aldexis:
## Before this PR
We wouldn't check correspondence between a functional interface's return type's declared safety and a lambda's actual return value, when that value's safety can be computed

## After this PR
==COMMIT_MSG==
[Safe-logging] Verify safety of returns in lambda expressions
==COMMIT_MSG==

~Note: This does not yet handle statement-base lambdas (need to do some more code-digging for this)~

I don't know much about Java compilation and static analysis primitives, so please tell me if there are better or simpler ways to do these things.

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/IllegalSafeLoggingArgument.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/IllegalSafeLoggingArgument.java
@@ -17,6 +17,7 @@
 package com.palantir.baseline.errorprone;
 
 import com.google.auto.service.AutoService;
+import com.google.common.collect.Iterables;
 import com.google.errorprone.BugPattern;
 import com.google.errorprone.VisitorState;
 import com.google.errorprone.bugpatterns.BugChecker;
@@ -33,6 +34,7 @@ import com.sun.source.tree.AssignmentTree;
 import com.sun.source.tree.ClassTree;
 import com.sun.source.tree.CompoundAssignmentTree;
 import com.sun.source.tree.ExpressionTree;
+import com.sun.source.tree.LambdaExpressionTree;
 import com.sun.source.tree.MethodInvocationTree;
 import com.sun.source.tree.MethodTree;
 import com.sun.source.tree.NewClassTree;
@@ -53,6 +55,8 @@ import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
+import javax.lang.model.element.Modifier;
 
 /**
  * Ensures that safe-logging annotated elements are handled correctly by annotated method parameters.
@@ -78,7 +82,8 @@ public final class IllegalSafeLoggingArgument extends BugChecker
                 BugChecker.MethodTreeMatcher,
                 BugChecker.VariableTreeMatcher,
                 BugChecker.NewClassTreeMatcher,
-                BugChecker.ClassTreeMatcher {
+                BugChecker.ClassTreeMatcher,
+                BugChecker.LambdaExpressionTreeMatcher {
 
     private static final String UNSAFE_ARG = "com.palantir.logsafe.UnsafeArg";
     private static final Matcher<ExpressionTree> SAFE_ARG_OF_METHOD_MATCHER = MethodMatchers.staticMethod()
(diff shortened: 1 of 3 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/IllegalSafeLoggingArgument.java
```diff
@@ -427,4 +432,86 @@ public Description matchClass(ClassTree tree, VisitorState state) {
...
+        }
+        MethodSymbol method = Iterables.getOnlyElement(abstractMethods);
+        Type returnType = method.getReturnType();
+
+        com.sun.tools.javac.util.List<Type> classTypeArguments =
+                variableType.tsym.asType().getTypeArguments();
+
+        for (int i = 0; i < classTypeArguments.size(); i++) {
+            // TODO: check whether this is the correct way to compare the type arguments
+            if (ASTHelpers.isSameType(classTypeArguments.get(i), returnType, state)) {
```
aldexis: Not too sure about this check - I used .equals when using the expression evaluation in IntelliJ, but compilation doesn't seem happy with it, and I found this method which seemed to do what I want

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/IllegalSafeLoggingLambdaTest.java
```diff
@@ -86,6 +86,90 @@ void testNestedAnonymousInLambdaUnsafeExternalData() {
...
+    //                        "import com.palantir.logsafe.*;",
+    //                        "import java.util.function.*;",
+    //                        "class Test {",
+    //                        "  static void f(@Unsafe Object value) {",
+    //                        "    // BUG: Diagnostic contains: Dangerous argument value: arg is 'UNSAFE'",
+    //                        "    Supplier<@Safe Object> supplier = () -> {return value;};",
+    //                        "  }",
+    //                        "}")
+    //                .doTest();
+    //    }
```
aldexis: Will either remove or make it support statement lambdas, if I can figure it out and it's not too much extra code (otherwise I'll move to a later PR)

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/IllegalSafeLoggingLambdaTest.java
```diff
@@ -86,6 +86,90 @@ void testNestedAnonymousInLambdaUnsafeExternalData() {
...
+
+    @Test
+    @SuppressWarnings("MisformattedTestData")
```
aldexis: Need to figure out how to make the block strings pass `MisformattedTestData`

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/IllegalSafeLoggingArgument.java
```diff
@@ -427,4 +432,86 @@ public Description matchClass(ClassTree tree, VisitorState state) {
...
+        while (path != null && !(path.getLeaf() instanceof VariableTree)) {
+            path = path.getParentPath();
+        }
+        if (path == null) {
+            // The lambda isn't directly assigned to a variable, so we don't know how to analyze it (yet)
+            // TODO(aldexis): handle other ways that a lambda can be used, e.g. returned from a method
+            return Description.NO_MATCH;
+        }
+        VariableTree variable = (VariableTree) path.getLeaf();
+        Type variableType = ASTHelpers.getType(variable);
```
carterkozak: I think we can simplify the implementation and solve your TODO thusly (worth verifying with some additional test cases where a lambda is returned, or provided as an argument into a method invocation)
```suggestion
        Type variableType = ASTHelpers.getType(tree);
```

aldexis: Indeed, this worked much better! I've used this and added the relevant test cases (lamdba as return type, and lambda as argument in a method invocation)

Thanks for pointing me to this!

I was pattern-matching from the `matchReturn` method. Is the reason this doesn't work there because we want not only the return type of the method, but also the potential annotation on the method itself, which we can't derive from `ASTHelper.getType(tree)`? (I'll check with the debugger, since I'm curious, but asking in case you know/remember off the top of your head)

EDIT: Interestingly (but also rather logically), `ASTHelpers.getType(tree.getExpression())` on a return statement would work if the expression is e.g. a lambda, whose type is inferred from the method return type, but not if we can infer the type in some more specific way (e.g. `return new SafeObject()` would have the expression have the type `SafeObject` regardless of the method return type, which makes sense to me, though it means that we can only rely on it this directly in the case of lambdas, which don't have an intrinsic type and need to be inferred from the surrounding regardless)

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/IllegalSafeLoggingArgument.java
```diff
@@ -427,4 +432,76 @@ public Description matchClass(ClassTree tree, VisitorState state) {
...
+
+        return buildDescription(tree)
+                .setMessage(String.format(
+                        "Dangerous return type in lambda function: expected return type is '%s' "
+                                + "but the lambda is returning '%s'.",
+                        requiredReturnSafety, resultSafety))
+                .build();
+    }
+
+    private Type getFunctionalInterfaceReturnType(Type variableType, VisitorState state) {
```
aldexis: Going to replace this entire method with `ASTHelpers.targetType(state.withPath(new TreePath(state.getPath(), tree)));` which delegates to `state.getTypes().findDescriptorType(getType(lambdaExpressionTree)).getReturnType();` and is a much simpler version of this

aldexis:
I've now added support for statement lambdas as well, and various test cases. This should be ready for review.

Please let me know if you would prefer me to split the statement lambdas to their own PR, considering they share little code in the end

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/IllegalSafeLoggingArgument.java
```diff
@@ -427,4 +462,43 @@ public Description matchClass(ClassTree tree, VisitorState state) {
...
+
+        return buildDescription(tree)
+                .setMessage(String.format(
+                        "Dangerous return value: result is '%s' but the lambda expects return '%s'.",
+                        resultSafety, requiredReturnSafety))
+                .build();
+    }
+
+    private Safety getLambdaRequiredReturnSafety(LambdaExpressionTree tree, VisitorState state) {
+        TargetType returnType = ASTHelpers.targetType(state.withPath(new TreePath(state.getPath(), tree)));
```
carterkozak, line 494: Ah, very nice! This ends up calling `state.getTypes().findDescriptorType(ASTHelpers.getType(lambdaExpressionTree)).getReturnType()`, where the `findDescriptorType` provides the functional interface method contract

carterkozak: Does this catch issues with switch expressions as well?

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/IllegalSafeLoggingLambdaTest.java
```diff
@@ -86,6 +86,195 @@ void testNestedAnonymousInLambdaUnsafeExternalData() {
...
+                        import java.util.function.*;
+
+                        class Test {
+                          static Supplier<@Safe Object> f(@Unsafe Object value) {
+                            // BUG: Diagnostic contains: Dangerous return value:
+                            // result is 'UNSAFE' but the lambda expects return 'SAFE'.
+                            return () -> value;
+                          }
+                        }
+                        """)
```
carterkozak, line 252: Mind adding another test case where we define a custom interface along the liens of `interface MaybeSupplier<T> { Optional<T> get(); }`? I think it will work correctly, but it catches an issue from a previous commit where type-parameters were matched with the return type, rather than using the resolved info

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/IllegalSafeLoggingArgument.java
aldexis: Ah I haven't checked. Do you have an example in mind of such an issue? I'm not entirely sure about what you're referring to

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/IllegalSafeLoggingLambdaTest.java
aldexis: Something like this? https://github.com/palantir/gradle-baseline/pull/3048/commits/c97ee70949b869358c40ff7aa0e939af796aa48b

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/IllegalSafeLoggingArgument.java
```diff
@@ -260,11 +263,24 @@ public Description matchReturn(ReturnTree tree, VisitorState state) {
         while (path != null && path.getLeaf() instanceof StatementTree) {
             path = path.getParentPath();
         }
-        if (path == null || !(path.getLeaf() instanceof MethodTree)) {
+        if (path == null || !(path.getLeaf() instanceof MethodTree || path.getLeaf() instanceof LambdaExpressionTree)) {
             return Description.NO_MATCH;
         }
-        MethodTree method = (MethodTree) path.getLeaf();
+
+        if (path.getLeaf() instanceof MethodTree method) {
```
carterkozak: style nit: we can combine these two sets of conditionals together to get rid of the `throw new IllegalStateException` since it should be impossible.

```java
if (path != null) {
    if (path.getLeaf() instanceof MethodTree method) {
        // etc
    } else if (path.getLeaf() instanceof LambdaExpressionTree lambda) {
        // etc
    }
}
return Description.NO_MATCH;

File: baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/IllegalSafeLoggingLambdaTest.java (hunk 7 above)
Comment (carterkozak) [reply baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/IllegalSafeLoggingLambdaTest.java:252]: Perfect, thanks!

File: baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/IllegalSafeLoggingArgument.java (hunk 8 above)
Comment (aldexis) [reply baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/IllegalSafeLoggingArgument.java]: Oh yeah, that's much nicer

File: baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/IllegalSafeLoggingArgument.java (hunk 6 above)
Comment (carterkozak) [reply baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/IllegalSafeLoggingArgument.java:494]: Something along these lines:
```java
void f(@Unsafe Object unsafeValue) {
  @Safe Object result = switch (myEnum) {
      case ONE -> unsafeValue;
      case TWO -> "safe";
  }
}
```

aldexis: Thanks! Unfortunately, it seems like the following test is still failing. I'll add to a separate PR listing edge cases I've found and look deeper into it
```java
  import com.palantir.logsafe.*;
  import java.util.function.*;

  enum MyEnum {
    ONE,
    TWO
  }

  class Test {
    void f(MyEnum myEnum, @Unsafe Object unsafeValue) {
      @Safe
      Object result =
          switch (myEnum) {
              // BUG: Diagnostic contains: Dangerous
            case ONE -> unsafeValue;
            case TWO -> "safe";
          };
    }
  }
```

EDIT: pushed it in https://github.com/palantir/gradle-baseline/pull/3051

carterkozak (approved):
Great work, thanks!

PR #3050, opened 2025-02-24
[Safe-logging] Infer lambda parameter safety from surrounding context

aldexis:
## Before this PR
Using a lambda's parameter as safe when we know it is unsafe isn't caught.

Unfortunately, the lambda parameter types passed in `SafetyPropagationTransfer#initialStore` from `ForwardAnalysisImpl#getParameters` don't seem to contain the type annotations. However, they are present in the `underlyingAst` argument which is also given, so we can instead recompute the argument safety from the expected lambda type.

## After this PR
Builds on top of https://github.com/palantir/gradle-baseline/pull/3048, which handles checking the returned values in a lambda

==COMMIT_MSG==
[Safe-logging] Infer lambda parameter safety from surrounding context
==COMMIT_MSG==


carterkozak (approved):
Great work, especially the test coverage!

PR #3052, opened 2025-02-24
[Safe-logging] Verify member reference type parameters when cast to functional interface

aldexis:
## Before this PR
Passing a member reference as a lambda will currently not catch cases where the type it gets passed to has different requirements around safety of the return or parameter types.

## After this PR
==COMMIT_MSG==
[Safe-logging] Verify member reference type parameters when cast to functional interface
==COMMIT_MSG==

This is done by comparing the safety between the method reference and the functional interface it gets cast into, and verifying that:
- The interface's return type safety is _less safe_ than the method reference
  - This means you can do e.g. `Supplier<@Unsafe String> f = this::fun` when `fun` returns `@Safe`, but not `@Supplier<@Safe String> f = this::fun` when `fun` returns `@Unsafe`, as in the latter case, you would then start using the unsafe result from `fun` as safe when calling `f`
- The interface's argument's type safety is _more safe_ than the method reference
  - This means you can do e.g. `Consumer<@Safe String> f = this::fun` when `fun` takes an `@Unsafe`, but not `Consumer<@Unsafe String> f = this::fun` when `fun` takes a `@Safe`, as otherwise calling `f.accept(s)` would let you pass an unsafe string to `fun`

To summarize:
**Allowed**
```java
import com.palantir.logsafe.*;
import java.util.function.*;

class Test {
  Supplier<@Unsafe String> func = this::fun;

  @Safe
  String fun() {
    return "safe";
  }
}
```

```java
import com.palantir.logsafe.*;
import java.util.function.*;

class Test {
  Consumer<@Safe String> func = this::fun;

  void fun(@Unsafe Object ob) {}
}
```

**Forbidden**
```java
import com.palantir.logsafe.*;
import java.util.function.*;

class Test {
  Supplier<@Safe String> func = this::fun;

  @Unsafe
  String fun() {
    return "unsafe";
  }
}
```

```java
import com.palantir.logsafe.*;
import java.util.function.*;

class Test {
  Consumer<@Unsafe String> func = this::fun;

  void fun(@Safe Object ob) {}
}
```

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/IllegalSafeLoggingArgument.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/IllegalSafeLoggingArgument.java
@@ -502,52 +502,85 @@ public final class IllegalSafeLoggingArgument extends BugChecker
 
     @Override
     public Description matchMemberReference(MemberReferenceTree tree, VisitorState state) {
-        // This is the type the reference gets "cast" into, whether through assignment, return type, argument, etc
-        Type castType = state.getTypes().findDescriptorType(ASTHelpers.getType(tree));
-        // The reference will be used as the expected type. This means:
-        //   * the safety of the return type of the expected must be "lower" than the actual safety of the reference
-        //   * the safety of the arguments of the expected must be "higher" than the actual safety of the reference
-        // i.e. "Supplier<@Safe String> f = this::method" shouldn't be allowed, when method returns unsafe
-        // Similarly, "@Consumer<@Unsafe String> f = this::method" shouldn't be allowed, when method expects to take a
-        //   safe argument (since it might log it)
+        Type type = ASTHelpers.getType(tree);
+        if (type == null) {
+            return Description.NO_MATCH;
+        }
 
-        Safety castTypeReturnSafety = SafetyAnnotations.getSafety(castType.getReturnType(), state);
+        // This is the type the reference gets "cast" into, whether through assignment, return type, argument, etc
+        Type castType = state.getTypes().findDescriptorType(type);
+        if (castType == null) {
+            return Description.NO_MATCH;
+        }
 
         MethodSymbol methodSymbol = ASTHelpers.getSymbol(tree);
+
+        // These methods rely on state.reportMatch to report all failures in one pass if multiple are found
+        handleMemberReferenceReturnType(tree, castType, methodSymbol, state);
+        handleMemberReferenceParameterTypes(tree, castType, methodSymbol, state);
+
+        return Description.NO_MATCH;
+    }
+
+    /**
+     * Verifies the return type of a method reference, when cast to a different type.
+     *
+     * This means that we expect the cast type's return type safety and will use it as such.
+     * If e.g. the cast type says it will return SAFE, but the reference returns UNSAFE, that's a problem.
+     * On the other hand, if the cast returns UNSAFE and the reference returns SAFE, that's fine.
+     */
+    private void handleMemberReferenceReturnType(
+            MemberReferenceTree tree, Type castType, MethodSymbol methodSymbol, VisitorState state) {
+        Safety castTypeReturnSafety = SafetyAnnotations.getSafety(castType.getReturnType(), state);
+
+        if (castTypeReturnSafety.allowsAll()) {
+            // Fast path - the type we're casting to allows all values (unknown or do-not-log), so we don't need to
+            //   check further
+            return;
+        }
+
         Safety referenceReturnSafety = Safety.mergeAssumingUnknownIsSame(
                 // This gets the combined safety annotations of the method and any of its supers
                 SafetyAnnotations.getSafety(methodSymbol, state),
                 SafetyAnnotations.getSafety(methodSymbol.getReturnType(), state));
 
-        // The reference will get used as the cast type
-        // This means that we expect the cast type's return type safety and will use it as such
-        // If e.g. the cast type says it will return SAFE, but the reference returns UNSAFE, that's a problem
-        // On the other hand, if the cast returns UNSAFE and the reference returns SAFE, that's fine
         if (!castTypeReturnSafety.allowsValueWith(referenceReturnSafety)) {
-            return buildDescription(tree)
+            state.reportMatch(buildDescription(tree)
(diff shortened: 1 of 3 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/IllegalSafeLoggingArgument.java
```diff
@@ -497,4 +499,67 @@ private Safety getLambdaRequiredReturnSafety(LambdaExpressionTree tree, VisitorS
...
+        // This is the type the reference gets "cast" into, whether through assignment, return type, argument, etc
+        Type castType = state.getTypes().findDescriptorType(ASTHelpers.getType(tree));
+        // The reference will be used as the expected type. This means:
+        //   * the safety of the return type of the expected must be "lower" than the actual safety of the reference
+        //   * the safety of the arguments of the expected must be "higher" than the actual safety of the reference
+        // i.e. "Supplier<@Safe String> f = this::method" shouldn't be allowed, when method returns unsafe
+        // Similarly, "@Consumer<@Unsafe String> f = this::method" shouldn't be allowed, when method expects to take a
+        //   safe argument (since it might log it)
+
+        Safety castTypeReturnSafety = SafetyAnnotations.getSafety(castType.getReturnType(), state);
```
carterkozak, line 534: should we check `allowsAll()` here to avoid checking the method symbol/returnType below in the `UNKNOWN` and `DO_NOT_LOG` cases?

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/IllegalSafeLoggingArgument.java
```diff
@@ -497,4 +499,67 @@ private Safety getLambdaRequiredReturnSafety(LambdaExpressionTree tree, VisitorS
...
+                // This gets the combined safety annotations of the method and any of its supers
+                SafetyAnnotations.getSafety(methodSymbol, state),
+                SafetyAnnotations.getSafety(methodSymbol.getReturnType(), state));
+
+        // The reference will get used as the cast type
+        // This means that we expect the cast type's return type safety and will use it as such
+        // If e.g. the cast type says it will return SAFE, but the reference returns UNSAFE, that's a problem
+        // On the other hand, if the cast returns UNSAFE and the reference returns SAFE, that's fine
+        if (!castTypeReturnSafety.allowsValueWith(referenceReturnSafety)) {
+            return buildDescription(tree)
```
carterkozak: If we use `state.reportMatch(`, we can report problems with the return type as well as parameters in the same evaluation. It's not really the standard way we like to use error-prone, but I do find it helpful to describe all the issues at once for the safelogging check specifically

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/IllegalSafeLoggingArgument.java
```diff
@@ -497,4 +499,67 @@ private Safety getLambdaRequiredReturnSafety(LambdaExpressionTree tree, VisitorS
...
+        // If the cast type says it accepts an UNSAFE parameter, but the reference needs a SAFE one,
+        //   then this should break
+        // If the cast type accepts SAFE, and the reference needs UNSAFE, that's fine, since we're less permissive
+        for (int i = 0; i < methodSymbol.getParameters().size(); i++) {
+            Type expectedParameterType = castType.getParameterTypes().get(i);
+            Safety expectedParameterSafety = SafetyAnnotations.getSafety(expectedParameterType, state);
+
+            VarSymbol parameter = methodSymbol.getParameters().get(i);
+            Type referenceParameterType = parameter.type;
+            Safety referenceParameterSafety = SafetyAnnotations.getSafety(referenceParameterType, state);
```
carterkozak, line 575: Can we check this first, and potentially bypass extracting safety of `expectedParameterSafety` in the fast path?

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/IllegalSafeLoggingArgument.java
```diff
@@ -497,4 +499,67 @@ private Safety getLambdaRequiredReturnSafety(LambdaExpressionTree tree, VisitorS
         }
         return SafetyAnnotations.getSafety(returnType.type(), state);
     }
+
+    @Override
+    public Description matchMemberReference(MemberReferenceTree tree, VisitorState state) {
+        // This is the type the reference gets "cast" into, whether through assignment, return type, argument, etc
+        Type castType = state.getTypes().findDescriptorType(ASTHelpers.getType(tree));
```
carterkozak: Should we check if `ASTHelpers.getType(tree)` returns null or `findDescriptorType` returns null and exit early?

aldexis: Good call!

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/IllegalSafeLoggingArgument.java
aldexis: Yes! I didn't want to return immediately, to avoid cases where this would hide the parameter check, but what I'll do is combine this with your other comment about using stateMatch for the return type check (which I missed would hide the argument check) and just extract them to separate methods for easier reading

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/IllegalSafeLoggingArgument.java
aldexis: Great idea indeed!

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/IllegalSafeLoggingArgument.java
aldexis: Yes definitely!

carterkozak (approved):
Fantastic work, I especially appreciate the inline comments!

carterkozak:
Feel free to add "merge-when-ready" if you're ready to release this

PR #3053, opened 2025-02-24
Use JUnit Platform by default

pkoenig10:
The JUnit 5.12.0 upgrade is causing test failures in all of our repos. See some prior discussion in https://github.com/junit-team/junit5/issues/4335.

```
Caused by: org.junit.platform.commons.JUnitException: TestEngine with ID 'junit-jupiter' failed to discover tests
        ...
Caused by: org.junit.platform.commons.JUnitException: OutputDirectoryProvider not available; probably due to unaligned versions of the junit-platform-engine and junit-platform-launcher jars on the classpath/module path.
        ...
```

The [Gradle documentation](https://docs.gradle.org/8.12/userguide/upgrading_version_8.html#test_framework_implementation_dependencies) states that, in the absence of test suites, the necessary test framework dependencies (ex. `junit-platform-launcher`) must be explicitly declared. I think we've just gotten lucky that nothing in the JUnit implementation has actually required this dependency until now.

The best solution here is to migrate `BaselineTesting` to configure tests using the newer [`jvm-test-suites`](https://docs.gradle.org/8.12/userguide/jvm_test_suite_plugin.html) plugin, which is applied automatically by the [`java`](https://github.com/palantir/gradle-baseline/pull/3053) plugin and already used to configure the default `test` test suite. The test suites plugin will automatically add the necessary test framework dependencies, and consumers can use their existing version management tool (ex. [gradle-consistent-versions](https://github.com/palantir/gradle-consistent-versions)) to ensure consistent versions of the JUnit dependencies.

Consumers will just need to add a `org.junit.platform:*` constraints in their `versions.props`. We can easily automate this (this is already done).

### DIFF
--- a/gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineTesting.java
+++ b/gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineTesting.java
@@ -69,6 +69,9 @@ public final class BaselineTesting implements Plugin<Project> {
                 task.dependsOn(checkJUnitDependencies);
             });
 
+            // For backwards compatibility reasons, Gradle uses the legacy JUnit4 test toolchain for the default
+            // test suite and the JUnit Platform test toolchain for everything else.
+            // We want to use junit jupiter by default for the default test suite.
             testingExtension
                     .getSuites()
                     .named(
@@ -76,11 +79,16 @@ public final class BaselineTesting implements Plugin<Project> {
                             JvmTestSuite.class,
                             JvmTestSuite::useJUnitJupiter);
 
+            // Gradle <8 does not automatically add junit toolchain deps. When the test-sets plugin is used instead
+            // of jvm-test-suites, these deps are not added either. We make sure that we lazily add these deps to
+            // the correct configurations if required. Note: we can't lazily configure test tasks or test suite then
+            // add the dependencies in that configuration action, as the Configurations may be resolved before the
+            // tasks/suites are realised, hence we have to work from the Configurations directly.
+            // See https://github.com/gradle/gradle/pull/26369
             project.getConfigurations().configureEach(configuration -> {
                 configuration.getDependencies().addAllLater(project.provider(() -> {
                     return testSourceSetsWhereJunitToolchainDepsHaveNotBeenAutomaticallyAdded(project)
                             .flatMap(sourceSet -> {
-                                // See https://github.com/gradle/gradle/pull/26369
                                 if (configuration.getName().equals(sourceSet.getImplementationConfigurationName())) {
                                     return Stream.of(project.getDependencies().create(JUNIT_JUPITER));
                                 }
@@ -109,28 +117,6 @@ public final class BaselineTesting implements Plugin<Project> {
         });
     }
 
-    private static Stream<SourceSet> testSourceSetsWhereJunitToolchainDepsHaveNotBeenAutomaticallyAdded(
-            Project project) {
-
-        SourceSetContainer sourceSets = project.getExtensions().getByType(SourceSetContainer.class);
-        TestingExtension testingExtension = project.getExtensions().getByType(TestingExtension.class);
-
-        // Gradle <8 does not automatically add junit toolchain deps - see https://github.com/gradle/gradle/pull/21919
-        boolean gradleVersionLessThan8 = GradleVersion.current().compareTo(GRADLE_8) < 0;
-
-        return project.getTasks().withType(Test.class).getNames().stream().flatMap(testTaskName -> {
-            if (gradleVersionLessThan8 || testTaskNotCreatedByJvmTestSuites(testingExtension, testTaskName)) {
-                return Stream.of(sourceSets.getByName(testTaskName));
-            }
-
-            return Stream.empty();
-        });
-    }
-
-    private static boolean testTaskNotCreatedByJvmTestSuites(TestingExtension testingExtension, String testTaskName) {
-        return !testingExtension.getSuites().getNames().contains(testTaskName);
-    }
-
     private void configureTestTask(Test task) {
         task.systemProperty("junit.platform.output.capture.stdout", "true");
         task.systemProperty("junit.platform.output.capture.stderr", "true");
@@ -153,5 +139,27 @@ public final class BaselineTesting implements Plugin<Project> {
                     .getEvents()
                     .addAll(Set.of(TestLogEvent.STARTED, TestLogEvent.PASSED, TestLogEvent.SKIPPED));
         }
+    }
+
+    private static Stream<SourceSet> testSourceSetsWhereJunitToolchainDepsHaveNotBeenAutomaticallyAdded(
+            Project project) {
+
+        SourceSetContainer sourceSets = project.getExtensions().getByType(SourceSetContainer.class);
+        TestingExtension testingExtension = project.getExtensions().getByType(TestingExtension.class);
+
+        // Gradle <8 does not automatically add junit toolchain deps - see https://github.com/gradle/gradle/pull/21919
+        boolean gradleVersionLessThan8 = GradleVersion.current().compareTo(GRADLE_8) < 0;
+
+        return project.getTasks().withType(Test.class).getNames().stream().flatMap(testTaskName -> {
+            if (gradleVersionLessThan8 || testTaskNotCreatedByJvmTestSuites(testingExtension, testTaskName)) {
+                return Stream.of(sourceSets.getByName(testTaskName));
+            }
+
+            return Stream.empty();
+        });
+    }
+
+    private static boolean testTaskNotCreatedByJvmTestSuites(TestingExtension testingExtension, String testTaskName) {
+        return !testingExtension.getSuites().getNames().contains(testTaskName);
     }
 }
--- a/gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineTesting.java
+++ b/gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineTesting.java
@@ -19,13 +19,13 @@ package com.palantir.baseline.plugins;
 import com.palantir.baseline.tasks.CheckJUnitDependencies;
 import java.util.Objects;
 import java.util.Set;
+import java.util.stream.Stream;
 import org.gradle.api.Plugin;
 import org.gradle.api.Project;
-import org.gradle.api.plugins.JavaPluginExtension;
 import org.gradle.api.plugins.JvmTestSuitePlugin;
 import org.gradle.api.plugins.jvm.JvmTestSuite;
 import org.gradle.api.tasks.SourceSet;
-import org.gradle.api.tasks.TaskCollection;
+import org.gradle.api.tasks.SourceSetContainer;
 import org.gradle.api.tasks.TaskProvider;
 import org.gradle.api.tasks.testing.Test;
 import org.gradle.api.tasks.testing.logging.TestLogEvent;
@@ -60,7 +60,6 @@ public final class BaselineTesting implements Plugin<Project> {
         });
 
         project.getPluginManager().withPlugin("java", unusedPlugin -> {
-            JavaPluginExtension javaPluginExtension = project.getExtensions().getByType(JavaPluginExtension.class);
             TestingExtension testingExtension = project.getExtensions().getByType(TestingExtension.class);
 
             TaskProvider<CheckJUnitDependencies> checkJUnitDependencies =
@@ -72,44 +71,64 @@ public final class BaselineTesting implements Plugin<Project> {
 
             testingExtension
                     .getSuites()
-                    .named(JvmTestSuitePlugin.DEFAULT_TEST_SUITE_NAME, JvmTestSuite.class, testSuite -> {
-                        testSuite.useJUnitJupiter();
-                    });
+                    .named(
+                            JvmTestSuitePlugin.DEFAULT_TEST_SUITE_NAME,
+                            JvmTestSuite.class,
+                            JvmTestSuite::useJUnitJupiter);
 
-            javaPluginExtension.getSourceSets().configureEach(sourceSet -> {
-                TaskCollection<Test> testTasks = project.getTasks()
-                        .withType(Test.class)
-                        .matching(task -> task.getName().equals(sourceSet.getName()));
+            project.getConfigurations().configureEach(configuration -> {
+                configuration.getDependencies().addAllLater(project.provider(() -> {
+                    return testSourceSetsWhereJunitToolchainDepsHaveNotBeenAutomaticallyAdded(project)
+                            .flatMap(sourceSet -> {
+                                // See https://github.com/gradle/gradle/pull/26369
+                                if (configuration.getName().equals(sourceSet.getImplementationConfigurationName())) {
+                                    return Stream.of(project.getDependencies().create(JUNIT_JUPITER));
+                                }
 
-                // We must eagerly create the test task in order to add the junit-platform-launcher dependency.
-                // Otherwise the dependencies will already have been resolved by the time we create the test task.
-                testTasks.all(task -> {
-                    // See https://github.com/gradle/gradle/pull/21919
-                    if (GradleVersion.current().compareTo(GRADLE_8) < 0) {
-                        addJUnitJupiterTestToolchainDependencies(project, sourceSet);
-                    }
+                                if (configuration.getName().equals(sourceSet.getRuntimeOnlyConfigurationName())) {
+                                    return Stream.of(project.getDependencies().create(JUNIT_PLATFORM_LAUNCHER));
+                                }
 
-                    // For test tasks not created using test suites, we must explicitly the test to use JUnit Platform
-                    // and add the junit-platform-launcher dependency.
-                    if (testingExtension.getSuites().findByName(sourceSet.getName()) == null) {
-                        task.useJUnitPlatform();
-                        addJUnitJupiterTestToolchainDependencies(project, sourceSet);
-                    }
-                });
+                                return Stream.empty();
+                            })
+                            .toList();
+                }));
+            });
 
-                testTasks.configureEach(task -> {
-                    configureTestTask(task);
+            project.getTasks().withType(Test.class).configureEach(task -> {
+                configureTestTask(task);
 
-                    task.dependsOn(checkJUnitDependencies);
-                });
+                task.dependsOn(checkJUnitDependencies);
+
+                // For test tasks not created using test suites (ie using the old unbroken dome test-suites plugin),
+                // we must explicitly use the JUnit Platform
+                if (testTaskNotCreatedByJvmTestSuites(testingExtension, task.getName())) {
+                    task.useJUnitPlatform();
+                }
             });
         });
     }
 
-    private void addJUnitJupiterTestToolchainDependencies(Project project, SourceSet sourceSet) {
-        // See https://github.com/gradle/gradle/pull/26369
-        project.getDependencies().add(sourceSet.getImplementationConfigurationName(), JUNIT_JUPITER);
-        project.getDependencies().add(sourceSet.getRuntimeOnlyConfigurationName(), JUNIT_PLATFORM_LAUNCHER);
+    private static Stream<SourceSet> testSourceSetsWhereJunitToolchainDepsHaveNotBeenAutomaticallyAdded(
+            Project project) {
+
+        SourceSetContainer sourceSets = project.getExtensions().getByType(SourceSetContainer.class);
+        TestingExtension testingExtension = project.getExtensions().getByType(TestingExtension.class);
+
+        // Gradle <8 does not automatically add junit toolchain deps - see https://github.com/gradle/gradle/pull/21919
+        boolean gradleVersionLessThan8 = GradleVersion.current().compareTo(GRADLE_8) < 0;
+
+        return project.getTasks().withType(Test.class).getNames().stream().flatMap(testTaskName -> {
+            if (gradleVersionLessThan8 || testTaskNotCreatedByJvmTestSuites(testingExtension, testTaskName)) {
+                return Stream.of(sourceSets.getByName(testTaskName));
+            }
+
+            return Stream.empty();
+        });
+    }
+
+    private static boolean testTaskNotCreatedByJvmTestSuites(TestingExtension testingExtension, String testTaskName) {
+        return !testingExtension.getSuites().getNames().contains(testTaskName);
     }
 
     private void configureTestTask(Test task) {
--- a/gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineTesting.java
+++ b/gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineTesting.java
@@ -100,8 +100,6 @@ public final class BaselineTesting implements Plugin<Project> {
             project.getTasks().withType(Test.class).configureEach(task -> {
                 configureTestTask(task);
 
-                task.dependsOn(checkJUnitDependencies);
-
                 // For test tasks not created using test suites (ie using the old unbroken dome test-suites plugin),
                 // we must explicitly use the JUnit Platform
                 if (testTaskNotCreatedByJvmTestSuites(testingExtension, task.getName())) {
--- a/gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineTesting.java
+++ b/gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineTesting.java
@@ -16,34 +16,29 @@
 
 package com.palantir.baseline.plugins;
 
-import com.google.common.base.Predicate;
-import com.google.common.collect.ImmutableSet;
 import com.palantir.baseline.tasks.CheckJUnitDependencies;
-import com.palantir.baseline.util.VersionUtils;
-import java.lang.reflect.Method;
 import java.util.Objects;
-import java.util.Optional;
 import java.util.Set;
 import org.gradle.api.Plugin;
 import org.gradle.api.Project;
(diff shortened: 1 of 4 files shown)


gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineTesting.java
```diff
@@ -64,88 +56,52 @@ public void apply(Project project) {
...
-
-                        ifHasResolvedCompileDependenciesMatching(
-                                project,
-                                sourceSet,
-                                BaselineTesting::requiresJunitPlatform,
-                                () -> fixSourceSet(project, sourceSet));
+            testingExtension
+                    .getSuites()
+                    .named(JvmTestSuitePlugin.DEFAULT_TEST_SUITE_NAME, JvmTestSuite.class, testSuite -> {
+                        testSuite.useJUnitJupiter();
```
pkoenig10: For backwards compatibility reasons, Gradle uses the legacy JUnit4 test toolchain for the default `test` test suite and the JUnit Platform test toolchain for everything else.

https://github.com/gradle/gradle/blob/4bb4af6421896c3ef094349be8e33098d343db2b/platforms/jvm/plugins-jvm-test-suite/src/main/java/org/gradle/api/plugins/jvm/internal/DefaultJvmTestSuite.java#L91-L100

Here we configure the default `test` test suite to also use the JUnit Platform test toolchain.

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineTesting.java
```diff
@@ -64,88 +56,52 @@ public void apply(Project project) {
...
+                        .matching(task -> task.getName().equals(sourceSet.getName()));
+
+                // We must eagerly create the test task in order to add the junit-platform-launcher dependency.
+                // Otherwise the dependencies will already have been resolved by the time we create the test task.
+                testTasks.all(task -> {
+                    // See https://github.com/gradle/gradle/pull/21919
+                    if (GradleVersion.current().compareTo(GRADLE_8) < 0) {
+                        project.getDependencies()
+                                .add(sourceSet.getRuntimeOnlyConfigurationName(), JUNIT_PLATFORM_LAUNCHER);
+                    }
```
pkoenig10: Gradle didn't start automatically adding this dependency until Gradle 8, so for earlier versions of Gradle we replicate that behavior here.

https://github.com/gradle/gradle/pull/21919

Once we drop support for Gradle 7, this should be removed.

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineTesting.java
```diff
@@ -64,88 +56,52 @@ public void apply(Project project) {
...
-                            .anyMatch(spec);
-
-                    if (anyMatch) {
-                        runnable.run();
+                    // For test tasks not creating using test suites, we must explicitly the test to use JUnit Platform
+                    // and add the junit-platform-launcher dependency.
+                    if (testingExtension.getSuites().findByName(sourceSet.getName()) == null) {
+                        task.useJUnitPlatform();
+                        project.getDependencies()
+                                .add(sourceSet.getRuntimeOnlyConfigurationName(), JUNIT_PLATFORM_LAUNCHER);
```
pkoenig10: This only exists to support the [testsets plugin](https://github.com/unbroken-dome/gradle-testsets-plugin). 

Once we migrate users from the testsets plugin to test suites, this should be removed.

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineTesting.java
```diff
@@ -64,88 +56,52 @@ public void apply(Project project) {
...
-        }
-
-        // unbroken dome does this
-        Task task2 = proj.getTasks().findByName(ss.getName());
-        if (task2 instanceof Test) {
-            return Optional.of((Test) task2);
-        }
-        return Optional.empty();
-    }
+            javaPluginExtension.getSourceSets().configureEach(sourceSet -> {
```
pkoenig10: It is a bit weird that we iterate over source sets to find tests. Once we no longer support the testsets plugin (and thus, all tests are declared using test suites), we should change this to iterating over the test suites instead. Same goes for the `CheckJUnitDependencies` task.

gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineTestingIntegrationTest.groovy
```diff
@@ -68,62 +81,98 @@ class BaselineTestingIntegrationTest extends IntegrationSpec {
...
+        class JqwikTest { 
             @Property
             void test(@ForAll byte value) {}
         }
         '''.stripIndent(true)
 
-    def '#gradleVersionNumber: capable of running both junit4 and junit5 tests'() {
+    def '#gradleVersionNumber: runs JUnit4 tests'() {
         when:
         gradleVersion = gradleVersionNumber
```
pkoenig10, line 86: Only one of these tests was testing on both Gradle 7 and 8. I've updated these tests so we're testing both Gradle versions on all of the tests (except a couple at the bottom where the Gradle version is not relevant).

gradle-baseline-java/src/main/java/com/palantir/baseline/tasks/CheckJUnitDependencies.java
```diff
@@ -49,178 +42,122 @@ public CheckJUnitDependencies() {
...
-        return getProject().provider(() -> getProbablyTestSourceSets()
-                .map(SourceSet::getRuntimeClasspathConfigurationName)
-                .map(getProject().getConfigurations()::getByName)
-                .collect(toList()));
-    }
+                    getLogger().info("Analyzing source set {} with task {}", ss.getName(), task.getName());
 
-    private Stream<SourceSet> getProbablyTestSourceSets() {
-        return getProject().getExtensions().getByType(JavaPluginExtension.class).getSourceSets().stream()
-                .filter(ss -> !ss.getName().equals(SourceSet.MAIN_SOURCE_SET_NAME));
```
pkoenig10, line 75: I've removed this filtering because:
- It's inconsistent with the `BaselineTesting` plugin, where we don't filter the source sets.
- It's almost certainly unnecessary. If it's a `Test` task, we probably want to perform this validation.

gradle-baseline-java/src/main/java/com/palantir/baseline/tasks/CheckJUnitDependencies.java
```diff
@@ -49,178 +42,122 @@ public CheckJUnitDependencies() {
...
+            if (usesJqwik && !hasJqwikEngine) {
+                throw new IllegalStateException("Some tests use Jqwik, but the '"
+                        + task.getName()
+                        + "' task is not using the Jqwik engine."
+                        + " To ensure your Jqwik tests are run, add the following dependency:\n\n"
+                        + "    "
+                        + ss.getRuntimeOnlyConfigurationName()
+                        + " 'net.jqwik:jqwik-engine'\n\n");
+            }
+        } else {
```
pkoenig10, line 150: In practice, we should never reach this unless someone has explicitly configured a test suite to use something other than JUnit Platform. I don't think we care to support any other test frameworks (all the things we we care about integrate with JUnit Platform using a custom engine), so I'm tempted to just remove this.

pkoenig10:
There may be a small number of repos that require manual changes after this upgrade. Those are repos with projects that exclusively use JUnit4 and are using the legacy JUnit4 test runner (which is the default) and not the JUnit Platform test runner.

Given that we performed a pretty extensive JUnit5 migration, the number of internal repos that meet this criteria is quite small (internal code search seems to confirm this, although I can't be certain from textual search alone). It is likely limited to the handful of repos that explicitly disabled the JUnit5 migration.

These repos don't have to migrate any of their tests to use JUnit5 APIs - the just need to migrate to use the JUnit Platform test runner (running their JUnit4 tests using the JUnit Vintage engine).

Alternatively, these repos can explicitly set the test runner to the JUnit4 test runner. I've verified that this works as expected, removing the JUnit Platorms dependencies rather than adding to them.

```gradle
testing {
    suites {
        test {
            useJUnit()
        }
    }
}
```

gradle-baseline-java/src/main/java/com/palantir/baseline/tasks/CheckJUnitDependencies.java
```diff
@@ -49,178 +42,122 @@ public CheckJUnitDependencies() {
...
-        boolean spock1Dependency = hasDep(deps, CheckJUnitDependencies::isSpock1);
-        boolean spock2OrGreaterDependency = hasDep(deps, CheckJUnitDependencies::isSpock2OrGreater);
-        String testRuntimeOnly = ss.getRuntimeOnlyConfigurationName();
-        boolean junitPlatformEnabled = BaselineTesting.useJUnitPlatformEnabled(task);
-
-        // If some testing library happens to provide the junit-jupiter-api, then users might start using the
-        // org.junit.jupiter.api.Test annotation, but as JUnit4 knows nothing about these, they'll silently not run
-        // unless the user has wired up the dependency correctly.
-        if (sourceSetMentionsJUnit5Api(ss)) {
-            String runtime = ss.getRuntimeClasspathConfigurationName();
```
pkoenig10, line 96: This produces incorrect suggestions, dependencies should be declared using the runtime only configuration, not the runtime classpath configuration.

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineTesting.java
```diff
@@ -64,88 +57,56 @@ public void apply(Project project) {
...
-                "Detected 'org:junit.jupiter:junit-jupiter', enabling useJUnitPlatform() on {}",
-                maybeTestTask.get().getName());
-        enableJunit5ForTestTask(maybeTestTask.get());
-    }
-
-    public static Optional<Test> getTestTaskForSourceSet(Project proj, SourceSet ss) {
-        String testTaskName = ss.getTaskName(null, "test");
+            project.getTasks().named(LifecycleBasePlugin.CHECK_TASK_NAME).configure(task -> {
+                task.dependsOn(checkJUnitDependencies);
+            });
```
pkoenig10, line 70: I was surprised that `checkJUnitDependencies` didn't depends on the `check` task - I think we probably want this.

gradle-baseline-java/src/main/java/com/palantir/baseline/tasks/CheckJUnitDependencies.java
```diff
@@ -49,178 +44,134 @@ public CheckJUnitDependencies() {
...
     private void validateSourceSet(SourceSet ss, Test task) {
         Set<ResolvedComponentResult> deps = getProject()
                 .getConfigurations()
                 .getByName(ss.getRuntimeClasspathConfigurationName())
                 .getIncoming()
                 .getResolutionResult()
                 .getAllComponents();
-        boolean junitJupiterIsPresent = hasDep(deps, CheckJUnitDependencies::isJunitJupiter);
-        boolean vintageEngineExists = hasDep(deps, CheckJUnitDependencies::isVintageEngine);
-        boolean spock1Dependency = hasDep(deps, CheckJUnitDependencies::isSpock1);
```
pkoenig10, line 87: Internal code search shows that there are no remaining users of Spock 1, so I've removed this.

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineTesting.java
```diff
@@ -64,88 +59,60 @@ public void apply(Project project) {
...
-        return Optional.empty();
-    }
+            javaPluginExtension.getSourceSets().configureEach(sourceSet -> {
+                TaskCollection<Test> testTasks = project.getTasks()
+                        .withType(Test.class)
+                        .matching(task -> task.getName().equals(sourceSet.getName()));
+
+                // We must eagerly create the test task in order to add the junit-platform-launcher dependency.
+                // Otherwise the dependencies will already have been resolved by the time we create the test task.
+                testTasks.all(task -> {
```
CRogers: It feels like this might be happening because the configuration to add the dependency happens on the dependencyset rather than test task or on source sets, so lazily configuring the test task/source set will not add the dep at the right time. I wonder if we can configure the dependencies of each Configuration instead - this is pretty much as lazy as possible, only being generated when the Configuration is resolved. Something like:

```java
project.getConfigurations().configureEach(configuration -> {
    configuration.getDependencies().addAllLater(project.provider(() -> {
        for (Test test : project.getTasks().withType(Test.class)) {
            SourceSet sourceSet = sourceSets.getByName(test.getName());
            if (testingExtension.getSuites().findByName(sourceSet.getName()) != null) {
                return Set.of();
            }

            if (configuration.getName().equals(sourceSet.getImplementationConfigurationName())) {
                return Set.of(project.getDependencies().create(JUNIT_JUPITER));
            }

            if (configuration.getName().equals(sourceSet.getRuntimeOnlyConfigurationName())) {
                return Set.of(project.getDependencies().create(JUNIT_PLATFORM_LAUNCHER));
            }
        }

        return Set.of();
    }));
});
```

If `addAllLater` doesn't work (maybe someone iterates over the dependencyset) maybe `beforeResolve` will do the trick.

CRogers: I guess we need to do the gradle 8 checking too

`task.useJUnitPlatform();` below can happen in a `configureEach` now, no need for `.all`.

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineTesting.java
```diff
@@ -64,88 +59,60 @@ public void apply(Project project) {
...
-                    boolean anyMatch = deps.getResolutionResult().getAllComponents().stream()
-                            .map(ResolvedComponentResult::getId)
-                            .filter(ModuleComponentIdentifier.class::isInstance)
-                            .map(ModuleComponentIdentifier.class::cast)
-                            .anyMatch(spec);
-
-                    if (anyMatch) {
-                        runnable.run();
+                    // For test tasks not created using test suites, we must explicitly the test to use JUnit Platform
+                    // and add the junit-platform-launcher dependency.
```
CRogers: Is this effectively just going to be people still using `test-sets`?

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineTesting.java
```diff
@@ -64,88 +59,60 @@ public void apply(Project project) {
...
-        Task task2 = proj.getTasks().findByName(ss.getName());
-        if (task2 instanceof Test) {
-            return Optional.of((Test) task2);
-        }
-        return Optional.empty();
-    }
+            javaPluginExtension.getSourceSets().configureEach(sourceSet -> {
+                TaskCollection<Test> testTasks = project.getTasks()
+                        .withType(Test.class)
+                        .matching(task -> task.getName().equals(sourceSet.getName()));
```
CRogers: Fun fact, I recently discovered that `.matching` realises all the tasks immediately, not just `.all`:
* [gradle-guide](https://github.com/palantir/gradle-guide/blob/8ca7aafffc5481d76fd3e2f7dde07ecc6e90d026/guide/avoiding-unnecessary-configuration.md#iterating-over-tasks)
* [gradle docs](https://docs.gradle.org/8.13/userguide/task_configuration_avoidance.html#:~:text=matching()%20requires%20all%20tasks%20to%20be%20created)

gradle-baseline-java/src/main/java/com/palantir/baseline/tasks/CheckJUnitDependencies.java
```diff
@@ -49,178 +44,141 @@ public CheckJUnitDependencies() {
...
-            Optional<Test> maybeTestTask = BaselineTesting.getTestTaskForSourceSet(getProject(), ss);
-            if (!maybeTestTask.isPresent()) {
-                // source set doesn't have a test task, e.g. 'schema'
-                return;
-            }
-            Test task = maybeTestTask.get();
+        getProject()
+                .getExtensions()
+                .getByType(JavaPluginExtension.class)
+                .getSourceSets()
```
CRogers, line 69: Calling `.getProject()`/accessing mutable domain state inside task actions is banned for configuration-cache and will throw in some future version of Gradle. Task actions should only access their input/output properties.

It feels weird, but the recommended way to do this is to have the input to the task be just the final result of the check, then the task action does basically nothing other than throw or not. So in this case we could have a `@Optional @Input Property<String> getErrorMessage` here, then just throw the error message if it exists.

Since inputs are only evaluated right before the task runs, it essentially happens at the same time as the task action. But it is configuration cacheable/supports modern gradle.

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineTesting.java
pkoenig10: That would be my expectation, but who knows.

gradle-baseline-java/src/main/java/com/palantir/baseline/tasks/CheckJUnitDependencies.java
pkoenig10: Can we address this in a follow-up? This task was already calling `getProject()` and I'd prefer not to make this PR any larger.

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineTesting.java
CRogers: I guess there is some risk here that the task will start failing on people's repos, but that's probably good info to burn down (as they probably aren't running tests)

gradle-baseline-java/src/main/java/com/palantir/baseline/tasks/CheckJUnitDependencies.java
CRogers: I just did this as it's a simple change.

gradle-baseline-java/src/main/java/com/palantir/baseline/tasks/CheckJUnitDependencies.java
```diff
@@ -16,211 +16,194 @@
...
 
-public class CheckJUnitDependencies extends DefaultTask {
+public abstract class CheckJUnitDependencies extends DefaultTask {
+    @Input
+    public abstract ListProperty<String> getErrorMessages();
 
     public CheckJUnitDependencies() {
         setGroup("Verification");
         setDescription("Ensures the correct JUnit4/5 dependencies are present, otherwise tests may silently not run");
-        getOutputs().upToDateWhen(_task -> true);
```
CRogers, line 47: Tasks with no outputs are never up-to-date

pkoenig10: Err sorry, I did not intend to delete this.

CRogers:
:+1:

pkoenig10:
👍

pkoenig10:
Thanks for fixing this up!

FinlayRJW:
👍

pkoenig10:
Ran into one small papercut in a repo that does not have any tests.

Because it does not have any tests, it does not currently have a `org.junit.platform` dependency, so our automation did not add a `org.junit.platform:` constraint to this repo.

Gradle declares a [versioned dependency](https://github.com/gradle/gradle/blob/881e969e9f5a529a9d82d934a76f159ac5331915/platforms/jvm/plugins-jvm-test-suite/src/main/java/org/gradle/api/testing/toolchains/internal/JUnitJupiterTestToolchain.java#L37) for `junit-jupiter` [but not for](https://github.com/gradle/gradle/blob/881e969e9f5a529a9d82d934a76f159ac5331915/platforms/jvm/plugins-jvm-test-suite/src/main/java/org/gradle/api/testing/toolchains/internal/JUnitPlatformTestToolchain.java#L51-L53) `junit-platform-launcher`. Because of this dependency resolution fails:

```
$ ./gradlew --no-scan example:dependencies --configuration testRuntimeClasspath

> Task :example:dependencies

------------------------------------------------------------
Project ':example'
------------------------------------------------------------

testRuntimeClasspath - Runtime classpath of source set 'test'.
+--- root project :
|    \--- root project : (*)
+--- root project : (*)
+--- org.junit.jupiter:junit-jupiter:5.8.2
|    +--- org.junit.jupiter:junit-jupiter-api:5.8.2
|    |    +--- org.opentest4j:opentest4j:1.2.0
|    |    +--- org.junit.platform:junit-platform-commons:1.8.2
|    |    |    \--- org.apiguardian:apiguardian-api:1.1.2
|    |    \--- org.apiguardian:apiguardian-api:1.1.2
|    +--- org.junit.jupiter:junit-jupiter-params:5.8.2
|    |    +--- org.junit.jupiter:junit-jupiter-api:5.8.2 (*)
|    |    \--- org.apiguardian:apiguardian-api:1.1.2
|    \--- org.junit.jupiter:junit-jupiter-engine:5.8.2
|         +--- org.junit.platform:junit-platform-engine:1.8.2
|         |    +--- org.opentest4j:opentest4j:1.2.0
|         |    +--- org.junit.platform:junit-platform-commons:1.8.2 (*)
|         |    \--- org.apiguardian:apiguardian-api:1.1.2
|         +--- org.junit.jupiter:junit-jupiter-api:5.8.2 (*)
|         \--- org.apiguardian:apiguardian-api:1.1.2
\--- org.junit.platform:junit-platform-launcher FAILED
```

---

The test suites `useJUnitJupiter()` API allows you to set the version of `junit-jupiter` dependency but AFAICT there is not way to set the version of the `junit-platform-launcher` by setting the `JUnitPlatformToolchainParameters.getPlatformVersion()` property.

I'm not sure how much effort we should put in to address this - I'm not sure how common it is to have a repo with no tests. 

The ideal solution is to ensure that every repo has a `org.junit.platform:*` constraint.

Alternatively we could unconditionally add a versioned `junit-platform-launcher` here, but I'd prefer not to take this responsibility from Gradle.

I'm considering putting up a Gradle PR proposing that `JUnitPlatformTestToolchain` fallback to a default versioned dependency rather than a versionless dependency.

pkoenig10:
Hmm I think there's something else going on here. This works with a super minimal repro.

Maybe GCV is somehow causing issues?

```groovy
plugins {
    id 'java'
}

repositories {
    mavenCentral()
}

testing {
    suites {
        test {
            useJUnitJupiter()
        }
    }
}
```

```
$ ./gradlew --no-scan dependencies --configuration testRuntimeClasspath

> Task :dependencies

------------------------------------------------------------
Root project 'example'
------------------------------------------------------------

testRuntimeClasspath - Runtime classpath of source set 'test'.
+--- org.junit.jupiter:junit-jupiter:5.8.2
|    +--- org.junit:junit-bom:5.8.2
|    |    +--- org.junit.jupiter:junit-jupiter:5.8.2 (c)
|    |    +--- org.junit.jupiter:junit-jupiter-api:5.8.2 (c)
|    |    +--- org.junit.jupiter:junit-jupiter-engine:5.8.2 (c)
|    |    +--- org.junit.jupiter:junit-jupiter-params:5.8.2 (c)
|    |    +--- org.junit.platform:junit-platform-launcher:1.8.2 (c)
|    |    +--- org.junit.platform:junit-platform-commons:1.8.2 (c)
|    |    \--- org.junit.platform:junit-platform-engine:1.8.2 (c)
|    +--- org.junit.jupiter:junit-jupiter-api:5.8.2
|    |    +--- org.junit:junit-bom:5.8.2 (*)
|    |    +--- org.opentest4j:opentest4j:1.2.0
|    |    \--- org.junit.platform:junit-platform-commons:1.8.2
|    |         \--- org.junit:junit-bom:5.8.2 (*)
|    +--- org.junit.jupiter:junit-jupiter-params:5.8.2
|    |    +--- org.junit:junit-bom:5.8.2 (*)
|    |    \--- org.junit.jupiter:junit-jupiter-api:5.8.2 (*)
|    \--- org.junit.jupiter:junit-jupiter-engine:5.8.2
|         +--- org.junit:junit-bom:5.8.2 (*)
|         +--- org.junit.platform:junit-platform-engine:1.8.2
|         |    +--- org.junit:junit-bom:5.8.2 (*)
|         |    +--- org.opentest4j:opentest4j:1.2.0
|         |    \--- org.junit.platform:junit-platform-commons:1.8.2 (*)
|         \--- org.junit.jupiter:junit-jupiter-api:5.8.2 (*)
\--- org.junit.platform:junit-platform-launcher -> 1.8.2
     +--- org.junit:junit-bom:5.8.2 (*)
     \--- org.junit.platform:junit-platform-engine:1.8.2 (*)
```

pkoenig10:
Oh this is because of our internal `disable-module-metadata` plugin.

https://github.com/junit-team/junit5/blob/c4faa6c26918fbbb8e3dbbde0e8d5077c8873d7c/junit-jupiter/junit-jupiter.gradle.kts#L8

Will continue the discussion internally.

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineTesting.java
pkoenig10: For posterity, this happened in at least one internal repo. They had an `integrationTest` source set with integration tests that were not being run because they did not have the necessary `junit-vintage-engine`.

The previous implementation is broken because the `integrationTest` source set is created before the `integrationTest` task. So `getTestTaskForSourceSet` doesn't find any task and the dependency is never created.

pkoenig10: This is actually also a problem for the built-in `test` source set. The source set is created before the `test` task, so we never set up any dependency.

PR #3066, opened 2025-03-14
baseline-java-versions: Regard projects as applicable for `libraryTarget` iff they have a jar publish plugin

CRogers:
## Before this PR
Currently we are _very_ conservative with what we choose to be "published library" vs "distribution only" code in baseline-java-versions. If there is any hint of publishing, even if there are no publications, we regard the project as a published library.

Internally, lots of repos have Gradle projects that are internal only (no jar is published) but happen to publish some non-jar resource (eg `conjure` or `audit-schema`). baseline-java-versions currently regards these as `libraryTarget` despite being internal only. People have started to use the hidden `override...` property to fix it, which is less than ideal: it should just do the right thing.

## After this PR
==COMMIT_MSG==
baseline-java-versions is now much less conservative about what it regards for `libraryTarget` instead of `distributionTarget`. `libraryTarget` will be used if and only if a jar is published using one of the Palantir jar publish plugins.
==COMMIT_MSG==

## Possible downsides?
* What about Gradle plugins etc bundled into maven-bundles? Is this going to now make them `distributionTarget`? Similar question for other jar based stuff packaged into maven bundles and run in prod (flink? magritte?)

---

I think baseline-java-versions needs a complete overhaul. This PR is a patch more than anything and the plugin still has some deep problems:

* It's very much in a "working by accident" state, as now Gradle requires the JDK levels very early on in the build, which just happens to be after the `publish-jar` plugin is applied. If it were just a bit earlier, it would not work.
* The providers to work out the target/runtime versions are called many times throughout the build and produce different values depending on the time they are called. Finalising them on read breaks everything at it finalises a calculation that's too early.
* It's very hard to work out _why_ a project is a certain runtime/target without running with extra logging.
* Some situations it's just wrong (eg libraryTarget test code depends on testing library project that is now marked as distributionTarget). Ideally it needs to take into account all the dependents for each project.

All of these could probably be solved by having lock files in each project. Then we can work out library vs distribution target later in the build ie at task input property calculation time rather than super early on in configuration time - then we could actually look at publications and inter-project dependencies etc safely

### DIFF
--- a/gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/javaversions/BaselineJavaVersions.java
+++ b/gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/javaversions/BaselineJavaVersions.java
@@ -69,7 +69,7 @@ public final class BaselineJavaVersions implements Plugin<Project> {
     private static boolean isLibrary(Project project, BaselineJavaVersionExtension projectVersions) {
         Property<Boolean> libraryOverride = projectVersions.overrideLibraryAutoDetection();
         if (libraryOverride.isPresent()) {
-            log.warn(
+            log.info(
                     "{} is considered a library because it has been overridden with library = true",
                     project.getDisplayName());
             return libraryOverride.get();
@@ -77,7 +77,7 @@ public final class BaselineJavaVersions implements Plugin<Project> {
 
         for (String plugin : LIBRARY_PLUGINS) {
             if (project.getPluginManager().hasPlugin(plugin)) {
-                log.warn(
+                log.info(
                         "{} is considered a library because the '{}' plugin is applied",
                         project.getDisplayName(),
                         plugin);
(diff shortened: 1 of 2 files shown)


gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/javaversions/BaselineJavaVersions.java
```diff
@@ -75,43 +69,38 @@ public void apply(Project project) {
...
+        // class files, allowing everyone to support the newer compilation level *before* libraries they depend on
+        // start requiring the newer compilation level. We want to upgrade distributionTarget first, wait for every
+        // repo to support the new compilation level, then allow libraries to be published with the new level.
+        //
+        // So the assumptions here are that (at least for Palantir code):
+        //   1. Only individual published jars matter for spreading higher compilation level requirements between repos
+        //      when we upgrade Java versions:
+        //        a) Jars inside dists/containers are not relevant as they will normally have JDK packaged with them.
+        //        b) Shaded/fat jars will still need to be published to have any kind of effect.
+        //   2. The *only* way to publish an individual jar is to use one of the above library plugins.
```
pkoenig10, line 99: Do we care about non-Palantir publishing plugins? I'm guessing not (especially since there's not a Gradle "JAR publishing" plugin, there's just the more general `maven-publish`), but just wanted to confirm we're okay with this tradeoff.

CRogers: I guess we might want the gradle plugin publish plugin too. I can't think of anything else (that we use at Palantir) that should be publishing individual jars.

Eventually, I want this is move this logic back to checking if there's any publication that has a jar artifact, but we can't do this without the rewrite listed in the description (due to [this](https://github.com/palantir/gradle-baseline/pull/2820))

CRogers: I guess eventually using the jar publication again would result in jars included in maven-bundles being regarded as libraries again. In that case, we probably wouldn't want to make the change in this PR. I'm still very unclear on whether a maven bundle should be libraryTarget or distributionTarget tbh, I _think_ it either doesn't matter (unless used by another module) or depends on the usage of the maven bundle.

PR #3068, opened 2025-03-14
Disallow duplicate files in archive tasks

pkoenig10:
Gradle 7.0 changed the default duplicates strategy from `WARN` to fail if there are duplicates and an explicit duplicates strategy has not been set.

By explicitly setting a duplicates strategy here, we are opting in to less strict behavior than the Gradle defaults. We definitely do not want to allow duplicate files in archives in order to ensure reproducibility.

If existing repos produce duplicate files and cannot easily fix it, they can explicitly set the duplicate strategy to `WARN` or lower.

### DIFF
--- a/gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineReproducibility.java
+++ b/gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineReproducibility.java
@@ -20,7 +20,6 @@ import com.palantir.baseline.tasks.CheckExplicitSourceCompatibilityTask;
 import org.gradle.api.Plugin;
 import org.gradle.api.Project;
 import org.gradle.api.Task;
-import org.gradle.api.file.DuplicatesStrategy;
 import org.gradle.api.plugins.JavaBasePlugin;
 import org.gradle.api.tasks.TaskProvider;
 import org.gradle.api.tasks.bundling.AbstractArchiveTask;
@@ -34,7 +33,6 @@ public final class BaselineReproducibility implements Plugin<Project> {
         project.getTasks().withType(AbstractArchiveTask.class).configureEach(t -> {
             t.setPreserveFileTimestamps(false);
             t.setReproducibleFileOrder(true);
-            t.setDuplicatesStrategy(DuplicatesStrategy.WARN);
         });
 
         project.getPluginManager().withPlugin("nebula.info", plugin -> {


changelog/@unreleased/pr-3068.v2.yml
```diff
@@ -0,0 +1,7 @@
+type: improvement
+improvement:
+  description: Baseline no longer explicitly sets the archive task duplicates strategy
+    to `WARN`. Archive tasks will now use the Gradle default behavior which will fail
+    if there are duplicates files and a duplicates strategy was not explicitly set.
```
CRogers: Isn't this just going to prevent baseline upgrading on a whole load of repos?

pkoenig10: I would hope not. If it does, we almost certainly want to know about it because it means those builds aren't reproducible.

gradle-baseline-java/src/main/resources/META-INF/gradle-plugins/com.palantir.baseline-checkstyle.properties
```diff
@@ -1 +0,0 @@
-implementation-class=com.palantir.baseline.plugins.BaselineCheckstyle
```
pkoenig10: These files are generated by the `gradlePlugin` config. It seems like we only declare about half of the plugins there. I've left the properties files for the ones that we don't declare there.

Ultimately we should declare them there, but I think that's outside the scope of this PR.

CRogers: Should probably not do this in the same PR.

pkoenig10: I have to at least remove these ones to stop using `include` in `gradle-baseline-java`.

changelog/@unreleased/pr-3068.v2.yml
CRogers: I think this is a pretty common issue - I've definitely seen the warning a lot. Even if it's bad, we can't block baseline upgrades for this kind of fix if the issue is widespread. It's a shame we can't easily find which repos this is happening in, then we could decided whether to do feature flag or just go for it.

Fwiw, sls-packging sets up it's own behaviour for `distTar`s:
* https://github.com/palantir/sls-packaging/blob/548f03e7cb2918f21af6e0d3a6c16402d553c1ac/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/service/DistTarTask.java#L109
* https://github.com/palantir/sls-packaging/blob/0345a84afa5bad80c01bb24555a8925d3609a61b/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/asset/AssetDistributionPlugin.java#L95

gradle-baseline-java/src/main/resources/META-INF/gradle-plugins/com.palantir.baseline-checkstyle.properties
CRogers: Ah, seems like the breaks from this are going to be pretty widespread then

CRogers: Oh wait, you mean the `'include'` change in baseline-only code

changelog/@unreleased/pr-3068.v2.yml
CRogers: I guess we could do an RC then a "background excavator" to find out how many failures there are before merging it in

gradle-baseline-java/src/main/resources/META-INF/gradle-plugins/com.palantir.baseline-checkstyle.properties
pkoenig10: Yes. I suppose I should probably pull that out into a separate PR.

pkoenig10: https://github.com/palantir/gradle-baseline/pull/3069

changelog/@unreleased/pr-3068.v2.yml
pkoenig10: Ack, I assumed this was not widespread.

PR #3078, opened 2025-03-20
Clarify library detection logic

pkoenig10:
I think it's easier to understand this logic as "if has publishing, do something" rather than "if does not have publishing, do something".

### DIFF
--- a/gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/javaversions/BaselineJavaVersions.java
+++ b/gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/javaversions/BaselineJavaVersions.java
@@ -113,24 +113,21 @@ public final class BaselineJavaVersions implements Plugin<Project> {
         }
 
         PublishingExtension publishing = project.getExtensions().findByType(PublishingExtension.class);
-        if (publishing == null) {
-            return new IsLibraryWithReason(false, "doesn't have any publishing extensions defined");
+        if (publishing != null) {
+            return new IsLibraryWithReason(
+                    true,
+                    String.format(
+                            "has publishing extensions with publications %s",
+                            publishing.getPublications().getNames()));
         }
 
-        // Better to be conservative with the java version rather than release something that is too high to be used.
         return new IsLibraryWithReason(
-                true,
+                false,
                 String.join(
                         "\n",
-                        "didn't match any other conditions that would indicate it was a distribution:",
-                        "  * It did not have a distribution plugin: " + DISTRIBUTION_PLUGINS,
-                        "  * It had a publishing extension, indicating *something* that isn't a known "
-                                + "distribution type is being published. The publications in this extensions are "
-                                + publishing.getPublications().getNames(),
-                        String.format(
-                                "Despite not having any library publish plugins (%s), this is conservatively "
-                                        + "regarded as a library for safety.",
-                                LIBRARY_PLUGINS)));
+                        "didn't match any other conditions that would indicate it was a library:",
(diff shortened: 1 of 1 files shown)


gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/javaversions/BaselineJavaVersions.java
```diff
@@ -111,24 +111,21 @@ private static IsLibraryWithReason isLibrary(Project project, BaselineJavaVersio
...
 
         PublishingExtension publishing = project.getExtensions().findByType(PublishingExtension.class);
-        if (publishing == null) {
-            return new IsLibraryWithReason(false, "doesn't have any publishing extensions defined");
+        if (publishing != null) {
+            return new IsLibraryWithReason(
+                    true,
+                    String.format(
+                            "has publishing extensions with publications %s",
+                            publishing.getPublications().getNames()));
```
aldexis, line 121: Based on the previous code, this seems to be an edge case for when _something_ is publishing, even though it's not using one of our library publishing plugins, so I'd keep some of the mentions regarding this
```suggestion
        // If the project seems to publish something without our publishing plugins,
        //   be conservative and consider this a library, to avoid releasing a library with too high a java version
        if (publishing != null) {
            return new IsLibraryWithReason(
                    true,
                    String.format(
                            "has publishing extensions with publications %s.\n" +
                            "Despite not having any library publish plugins (%s), this is conservatively regarded as a library for safety.",
                            publishing.getPublications().getNames(),
                            LIBRARY_PLUGINS));
```

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/javaversions/BaselineJavaVersions.java
```diff
@@ -111,24 +111,21 @@ private static IsLibraryWithReason isLibrary(Project project, BaselineJavaVersio
...
-                        "  * It did not have a distribution plugin: " + DISTRIBUTION_PLUGINS,
-                        "  * It had a publishing extension, indicating *something* that isn't a known "
-                                + "distribution type is being published. The publications in this extensions are "
-                                + publishing.getPublications().getNames(),
-                        String.format(
-                                "Despite not having any library publish plugins (%s), this is conservatively "
-                                        + "regarded as a library for safety.",
-                                LIBRARY_PLUGINS)));
+                        "didn't match any other conditions that would indicate it was a library:",
+                        "  * It did not have a distribution plugin: " + LIBRARY_PLUGINS,
```
aldexis:
```suggestion
                        "  * It did not have a library plugin: " + LIBRARY_PLUGINS,
```

aldexis (commented):
Lgtm, but leaving the opportunity for @felixdesouza, @CRogers or @FinlayRJW to review and +1

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/javaversions/BaselineJavaVersions.java
pkoenig10: I'm not sure this is particularly relevant. It seems simpler to just say "anything with a publishing extension is considered a library".

aldexis: I defer to @felixdesouza on what he'd prefer here

pkoenig10:
@felixdesouza can you take a look at this?

pkoenig10:
@aldexis @felixdesouza I'd still like to merge this, do have the time to take a look?

esword (approved):

PR #3083, opened 2025-03-29
Delete unused images/checkstyle-circle-failure.png

ash211:
## Before this PR
We have a 7yo image in this directory that afaict on [github search](https://github.com/search?q=repo%3Apalantir%2Fgradle-baseline+checkstyle-circle-failure.png&type=code) and internal sourcegraph search, is not being used anywhere in this repo.

## After this PR
==COMMIT_MSG==
Delete unused images/checkstyle-circle-failure.png
==COMMIT_MSG==

## Possible downsides?
- maybe it's used from somewhere outside the repo?


bmarcaur (approved):
Seems like it was simply orphaned when some docs were removed. Trivial enough to restore if we find out later.

PR #3090, opened 2025-04-08
Fix gradle 7 compat for checkClassUniquenessTask

esword:
## Before this PR
When run under gradle 7, the checkClassUniquenessTask gets this error:
LambdaConversionException: Invalid receiver type interface org.gradle.api.artifacts.Configuration; not a subtype of implementation type interface org.gradle.api.Named 

The problem comes in the multiple references to Configuration::getName in CheckClassUniquenessLockTask. The Configuration interface directly had a getName() method in gradle 7. This was removed in gradle 8 and instead Configuration extends Named.  While the method signature is the same, code compiled against gradle 8 will have an incorrect reference to it in the class file when run under gradle 7.

When compiling against the Gradle 7.6.4 API, the method that ends up in the class file is Configuration.getName.
When compiling against the Gradle 8.12.1 API, the method that ends up in the class file is Named.getName.
When you try to run the latter with Gradle 7.6.4, it fails because we attempt to invoke Named.getName on a Configuration object, but Configuration does not implement Named.

## After this PR
This method works around that by having the code check for what interfaces are implemented at runtime.  Could also use reflection to just find the getName method but this is a little more explicit as to why.
==COMMIT_MSG==
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-3090.v2.yml
+++ b/changelog/@unreleased/pr-3090.v2.yml
@@ -1,0 +1,5 @@
+type: fix
+fix:
+  description: Fix gradle 7 compat for checkClassUniquenessTask
+  links:
+  - https://github.com/palantir/gradle-baseline/pull/3090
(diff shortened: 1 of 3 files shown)


felixdesouza (dismissed):

felixdesouza:
👍

PR #3092, opened 2025-04-11
Move Slf4jLogsafeArgs to ERROR level

aldexis:
## Before this PR
This check was presumably introduced as WARNING to allow its rollout without actually breaking builds. However, this also means that it will be ignored in a lot of cases.

With https://github.com/palantir/suppressible-error-prone now being a thing, and the new capability to remove the for-rollout suppressions to automatically open PRs to flag these, this means we can safely move this to ERROR level. Cases that will now be flagged as errors that weren't previously can be easily suppressed by suppressible-error-prone, and flagged in separate PRs for fixing.

## After this PR
==COMMIT_MSG==
Move Slf4jLogsafeArgs to ERROR level
==COMMIT_MSG==


bjlaub:
👍

PR #3093, opened 2025-04-14
Replace getAllModuleArtifacts with getModuleArtifacts

pkoenig10:
`getAllModuleArtifacts` does not correctly handle circular dependencies. See https://github.com/gradle/gradle/issues/22850.

We should not try to "resolve" dependencies ourselves using `getAllModuleArtifacts`. Instead we should always let Gradle resolve configurations and then use the artifacts from the resolved configuration.

### DIFF
--- a/gradle-baseline-java/src/main/java/com/palantir/baseline/tasks/CheckImplicitDependenciesTask.java
+++ b/gradle-baseline-java/src/main/java/com/palantir/baseline/tasks/CheckImplicitDependenciesTask.java
@@ -29,7 +29,7 @@ import java.util.stream.Collectors;
 import org.gradle.api.DefaultTask;
 import org.gradle.api.artifacts.Configuration;
 import org.gradle.api.artifacts.ResolvedArtifact;
-import org.gradle.api.artifacts.ResolvedDependency;
+import org.gradle.api.artifacts.ResolvedConfiguration;
 import org.gradle.api.artifacts.component.ProjectComponentIdentifier;
 import org.gradle.api.file.FileCollection;
 import org.gradle.api.provider.ListProperty;
@@ -63,17 +63,20 @@ public class CheckImplicitDependenciesTask extends DefaultTask {
 
     @TaskAction
     public final void checkImplicitDependencies() {
-        Set<ResolvedDependency> declaredDependencies = dependenciesConfigurations.get().stream()
+        Set<ResolvedConfiguration> resolvedConfigurations = dependenciesConfigurations.get().stream()
                 .map(Configuration::getResolvedConfiguration)
-                .flatMap(resolved -> resolved.getFirstLevelModuleDependencies().stream())
                 .collect(Collectors.toSet());
-        BaselineExactDependencies.INDEXES.populateIndexes(declaredDependencies);
+        BaselineExactDependencies.INDEXES.populateIndexes(resolvedConfigurations);
+
+        Set<ResolvedArtifact> declaredArtifacts = resolvedConfigurations.stream()
+                .flatMap(resolved -> resolved.getFirstLevelModuleDependencies().stream())
+                .flatMap(dependency -> dependency.getModuleArtifacts().stream())
+                .filter(dependency ->
+                        BaselineExactDependencies.VALID_ARTIFACT_EXTENSIONS.contains(dependency.getExtension()))
+                .collect(Collectors.toSet());
 
         Set<List<ResolvedArtifact>> necessaryArtifacts = referencedClasses().stream()
                 .map(c -> BaselineExactDependencies.INDEXES.classToArtifacts(c).collect(Collectors.toList()))
-                .collect(Collectors.toSet());
-        Set<ResolvedArtifact> declaredArtifacts = declaredDependencies.stream()
-                .flatMap(dependency -> dependency.getModuleArtifacts().stream())
                 .collect(Collectors.toSet());
 
         List<ResolvedArtifact> usedButUndeclared = necessaryArtifacts.stream()
--- a/gradle-baseline-java/src/main/java/com/palantir/baseline/tasks/CheckUnusedDependenciesTask.java
+++ b/gradle-baseline-java/src/main/java/com/palantir/baseline/tasks/CheckUnusedDependenciesTask.java
@@ -30,7 +30,7 @@ import java.util.stream.Collectors;
 import org.gradle.api.DefaultTask;
 import org.gradle.api.artifacts.Configuration;
 import org.gradle.api.artifacts.ResolvedArtifact;
-import org.gradle.api.artifacts.ResolvedDependency;
+import org.gradle.api.artifacts.ResolvedConfiguration;
 import org.gradle.api.file.FileCollection;
 import org.gradle.api.provider.ListProperty;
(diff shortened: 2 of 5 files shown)


gradle-baseline-java/src/main/java/com/palantir/baseline/tasks/CheckImplicitDependenciesTask.java
```diff
@@ -63,18 +63,21 @@ public CheckImplicitDependenciesTask() {
...
+        Set<ResolvedConfiguration> resolvedConfigurations = dependenciesConfigurations.get().stream()
                 .map(Configuration::getResolvedConfiguration)
+                .collect(Collectors.toSet());
+        BaselineExactDependencies.INDEXES.populateIndexes(resolvedConfigurations);
+
+        Set<ResolvedArtifact> declaredArtifacts = resolvedConfigurations.stream()
                 .flatMap(resolved -> resolved.getFirstLevelModuleDependencies().stream())
+                .flatMap(dependency -> dependency.getModuleArtifacts().stream())
+                .filter(dependency ->
+                        BaselineExactDependencies.VALID_ARTIFACT_EXTENSIONS.contains(dependency.getExtension()))
```
pkoenig10, line 75: We did this filtering in `CheckUnusedDependenciesTask` but not here. I assume this was just an oversight.

gradle-baseline-java/src/test/groovy/com/palantir/baseline/AbstractPluginTest.groovy
```diff
@@ -46,6 +46,8 @@ class AbstractPluginTest extends Specification {
             .withProjectDir(projectDir)
             .withArguments(tasks)
             .withPluginClasspath()
+            .withDebug(true)
+            .forwardOutput()
```
pkoenig10, line 50: These are extremely helpful when debugging tests:
- `withDebug` allows breakpoints to work
- `forwardOutput` shows the build output in the test output

gradle-baseline-java/src/main/java/com/palantir/baseline/tasks/CheckUnusedDependenciesTask.java
```diff
@@ -83,55 +83,29 @@ public final void checkUnusedDependencies() {
...
-
-                if (!didYouMean.isEmpty()) {
-                    builder.append("\t\tDid you mean:\n");
-                    didYouMean.stream()
-                            .map(BaselineExactDependencies::asDependencyStringWithoutName)
-                            .sorted()
-                            .forEach(dependencyString -> builder.append("\t\t\t")
-                                    .append(dependencyString)
-                                    .append("\n"));
-                }
```
pkoenig10: Providing these suggestions requires:
- Mapping artifacts to dependencies (the `artifactsFromDependency` call here). This is no longer possible now that `Indexes` takes in configurations rather than dependencies.
- Computing the transitive dependencies of a dependency, rather than a configuration (the `getAllModuleArtifacts` call here). This is something we want to avoid.

I don't see a way to support this. In practice, I don't think these suggestions are that valuable. Either:
- You have `checkImplicitDependencies` enabled and thus you already have explicit dependencies for the suggestions, or
- You don't have `checkImplicitDependencies` enabled, but it's likely that you have at least one other dependency bringing in the transitive dependencies listed here

Given this, I opted to just remove these suggestions.

gradle-baseline-java/src/main/java/com/palantir/baseline/tasks/CheckUnusedDependenciesTask.java
```diff
@@ -83,55 +83,29 @@ public final void checkUnusedDependencies() {
...
-                        .collect(Collectors.toSet());
-
-                if (!didYouMean.isEmpty()) {
-                    builder.append("\t\tDid you mean:\n");
-                    didYouMean.stream()
-                            .map(BaselineExactDependencies::asDependencyStringWithoutName)
-                            .sorted()
-                            .forEach(dependencyString -> builder.append("\t\t\t")
-                                    .append(dependencyString)
-                                    .append("\n"));
```
CRogers: Doesn't this remove the helpful suggestions in the error messages?

pkoenig10: Yes, see https://github.com/palantir/gradle-baseline/pull/3093#discussion_r2042494732

gradle-baseline-java/src/main/java/com/palantir/baseline/tasks/CheckUnusedDependenciesTask.java
pkoenig10: I feel like it's better to leave this as a job for `compileJava`. `checkUnusedDependencies` failures are not a super common part of dev workflows, so I don't think the lack of suggestions will be much of a hindrance - you just remove the unused dependency and then address any subsequent compilation failures.

CRogers: I gave it a go to try and bring back the error message (partially as a test for our new AI tools) and it wasn't too bad - might as well keep it?

CRogers:
:+1:

felixdesouza:
👍

gradle-baseline-java/src/main/java/com/palantir/baseline/tasks/CheckUnusedDependenciesTask.java
pkoenig10: I don't think that works. Your commit did not revert the test change. When I revert the test change, it fails because the output does not include the suggestion.

```

* What went wrong:
Execution failed for task ':checkUnusedDependenciesMain'.
> Found 1 dependencies unused during compilation, please delete them from 'build.gradle' or choose one of the suggested fixes:
  	project(':sub-project-with-deps') <-- main
```

gradle-baseline-java/src/main/java/com/palantir/baseline/tasks/CheckUnusedDependenciesTask.java
```diff
@@ -83,40 +83,38 @@ public final void checkUnusedDependencies() {
...
-                ResolvedDependency dependency =
-                        BaselineExactDependencies.INDEXES.artifactsFromDependency(resolvedArtifact);
-                Set<ResolvedArtifact> didYouMean = dependency.getAllModuleArtifacts().stream()
-                        .filter(artifact ->
-                                BaselineExactDependencies.VALID_ARTIFACT_EXTENSIONS.contains(artifact.getExtension()))
+                // Suggest fixes by looking at all artifacts with the same module ID,
+                // filtering the ones we have declarations on, and mapping the remaining ones
+                // back to the jars they came from.
+                Set<ResolvedArtifact> didYouMean = BaselineExactDependencies.INDEXES
+                        .findArtifactsWithSameModuleId(resolvedArtifact)
```
pkoenig10, line 117: This doesn't seem equivalent to the previous implementation at all.

The intention of the existing code was to find implicit dependencies of the unused dependencies. I don't see how finding artifacts with the same module ID accomplishes that.

PR #3096, opened 2025-04-16
Convert flatMap(Streams::stream) to Stream.mapMulti(Optional::isPresent)

exclowd:
## Before this PR

https://github.com/palantir/gradle-baseline/pull/2996 introduced check to
convert `Stream.flatMap(Optional::stream)` to more efficient `Stream.mapMulit(Optional::isPresent)`
This PR just extends that to cover https://guava.dev/releases/23.0/api/docs/com/google/common/collect/Streams.html#stream-com.google.common.base.Optional-
Since that has the same implementation as `Optional::stream`, we should be
convert this as well.

## After this PR

==COMMIT_MSG==
Extend check to convert flatMap(Streams::stream) to Stream.mapMulti(Optional::isPresent)
==COMMIT_MSG==


baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/StreamFlatmapOptionalTest.java
```diff
@@ -147,6 +147,119 @@ public void test() {
                 .doTest();
     }
 
+    @Test
+    public void testGuavaStreams() {
```
pkoenig10: These should just be additional cases added to `test`.

I don't think we need to duplicate everything. Many of these cases are testing particular structures that we don't need to re-test for `Streams::stream`.

exclowd: Ack, updated! just testing the base case and the `Streams::stream(Collection)` case

pkoenig10 (approved):

PR #3101, opened 2025-05-05
Fix compatibility with error-prone 2.38.0

jmongeluzzi:
There were several breaking changes between error-prone 2.31.0 and 2.38.0. This change should fix compatibility without affecting functionality.

Because this repo applies the gradle-baseline plugin to itself, error-prone fails on this change because of the same issue it is intended to fix. To work around this, I have disabled the affected error-prone checks temporarily and will re-enable them in a subsequent PR that bumps gradle-baseline to the latest version.

Also, a change in error-prone (https://github.com/google/error-prone/pull/4699) means that patches are no longer applied to disabled checks, so I have commented out the test for this behavior. Restoring this behavior could be possible in the future via argument manipulation in https://github.com/palantir/suppressible-error-prone, but that is beyond the scope of this change.

==COMMIT_MSG==
Fix compatibility with error-prone 2.38.0.
==COMMIT_MSG==


build.gradle
```diff
@@ -50,6 +50,10 @@ allprojects {
         tasks.withType(JavaCompile) {
             options.compilerArgs += ['-Werror', '-Xlint:deprecation']
             options.errorprone {
+                // Temporarily disable errorprone, because it cannot pass due to the issue that is being fixed
+                // in #3101. Remove this line after a new version has been released.
+                getDisableAllChecks().set(true)
```
pkoenig10: Is it possible to make this a bit more targeted? Or are all checks broken?

[Example where we've done this previously](https://github.com/palantir/gradle-baseline/commit/f9539d53d1a1bfd945e9fc9f1db8b499e73cdbf1#diff-49a96e7eea8a94af862798a45174e6ac43eb4f8b4bd40759b5da63ba31ec3ef7R55-R56)

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ExecutorSubmitRunnableFutureIgnored.java
```diff
@@ -45,6 +47,10 @@ public final class ExecutorSubmitRunnableFutureIgnored extends AbstractReturnVal
             .named("submit")
             .withParameters(Runnable.class.getName());
 
+    public ExecutorSubmitRunnableFutureIgnored() {
+        super(ConstantExpressions.fromFlags(ErrorProneFlags.empty()));
+    }
```
pkoenig10, line 53: We should inject these parameters rather than using empty flags. We want to preserve the provided flags.

jmongeluzzi: Can you clarify which parameters you mean? What I've done should preserve the existing behavior. Previously this implicitly used the no-arg constructor of AbstractReturnValueIgnored which passed in the same empty set of flags as I am here. That no-arg constructor has been removed.

https://github.com/google/error-prone/blob/ab522c7dcac5e83b84828d5670595e5582d71fb3/core/src/main/java/com/google/errorprone/bugpatterns/AbstractReturnValueIgnored.java#L141

pkoenig10: > Can you clarify which parameters you mean?

`ConstantExpressions`. Looks at the other Error Prone checks that extend `AbstractReturnValue`.

The no-args constructor needs to be kept for service loading.

This code should look like:

```
public ExecutorSubmitRunnableFutureIgnored() {
    super(ConstantExpressions.fromFlags(ErrorProneFlags.empty()));
}

@Inject
ExecutorSubmitRunnableFutureIgnored(ConstantExpressions constantExpressions) {
    super(constantExpressions);
}
```

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/CardinalityEqualsZero.java
```diff
@@ -83,27 +83,21 @@ public Description matchBinary(BinaryTree tree, VisitorState state) {
...
-            case MEMBER_SELECT:
-                return ((MemberSelectTree) tree).getIdentifier().contentEquals("this");
-            default:
-                return false;
-        }
+        return switch (tree.getKind()) {
+            case IDENTIFIER -> ((IdentifierTree) tree).getName().contentEquals("this");
+            case MEMBER_SELECT -> ((MemberSelectTree) tree).getIdentifier().contentEquals("this");
+            default -> false;
+        };
```
pkoenig10, line 90: Can we please save unrelated refactors for a separate PR? It makes it hard to review this PR.

jmongeluzzi: This was required by error prone checks on the repo. Would you prefer I suppress in this PR?

pkoenig10: Which check?

pkoenig10: Oh this is a new one introduced in Error Prone.

https://errorprone.info/bugpattern/StatementSwitchToExpressionSwitch

CRogers: We should work out if this (and any other introduced check) is something we want to disable, or are comfortable will a load of automated suppressions for them. Perhaps it has autofixes?

pkoenig10: > Perhaps it has autofixes?

This check [only reports a match if it has an auto-fix](https://github.com/google/error-prone/blob/65630f20ff9c56ed1e0d7919155d54e368d8575b/core/src/main/java/com/google/errorprone/bugpatterns/StatementSwitchToExpressionSwitch.java#L215-L217). So I think it should be fine to roll this out.

pkoenig10: [StringConcatToTextBlock](https://errorprone.info/bugpattern/StringConcatToTextBlock) also feels a bit 🌶️, but [there are automated fixes for that too](https://github.com/google/error-prone/blob/65630f20ff9c56ed1e0d7919155d54e368d8575b/core/src/main/java/com/google/errorprone/bugpatterns/StringConcatToTextBlock.java#L188-L195). And given that text blocks seem objectively better than string concatenation, I'm inclined to leave that one also.

gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineErrorProneIntegrationTest.groovy
```diff
@@ -203,37 +203,40 @@ class BaselineErrorProneIntegrationTest extends AbstractPluginTest {
...
-                if (a.equals(b)) {
-                  System.out.println("arrays are equal!");
-                  Optional.of("hello").orElseGet(() -> System.getProperty("world"));
-                }
-            }
-        }
-        '''.stripIndent()
-    }
+// As of https://github.com/google/error-prone/pull/4699, error prone no longer applies patches to disabled checks.
+// Disabling this test since it is no longer valid and will require significant changes to restore the prior behavior.
```
pkoenig10, line 207: Can you elaborate on this? Why do we need to disable this test?

jmongeluzzi: Quoting from https://github.com/google/error-prone/pull/4699#discussion_r1899667974

> ❗ That said, my changes impact semantics ❗. Before the changes in this PR, Error Prone would ignore the -Xep:AssignmentUpdater:OFF flag and apply any suggestions by AssignmentUpdater. After my changes, AssignmentUpdater is ignored altogether.

Disabling a check results in it being set to OFF, so suggestions do not get applied.

pkoenig10: This feels like something we need to address before merging this PR. A lot of our internal automation is built on Error Prone and applying suggested fixes.

jmongeluzzi:
Closing this as the scope of what is needed to merge this has expanded to include changes to other gradle plugins (see 
 https://github.com/palantir/gradle-baseline/pull/3101#discussion_r2085707039). The workaround to exclude the affected checks is sufficient for my purposes.

PR #3117, opened 2025-06-02
Add errorprone to detect xml mapper without input / output factories

FinlayRJW:
## Before this PR
An xml mapper without input / output factories can cause bugs where shaded versions of woodstox are found etc
We had no error-prones to catch this common pitfall

## After this PR

We now have an error-prone to detect and fix bad usages of xml mapper. It catches and fixes the following:

### 1. Unsafe Default Constructor

```java
XmlMapper mapper = new XmlMapper();
```
-> 
```java
XmlMapper mapper = new XmlMapper(new WstxInputFactory(), new WstxOutputFactory());
```

### 2. Unsafe Factory Constructor

```java
XmlFactory factory = new XmlFactory();
XmlMapper mapper = new XmlMapper(factory);
```
-> 
```java
XmlMapper mapper = new XmlMapper(new XmlFactory(new WstxInputFactory(), new WstxOutputFactory()));
```

### 3. Unsafe Builder Pattern

```java
XmlMapper mapper = XmlMapper.builder().build();
```
-> 
```java
XmlMapper mapper = XmlMapper.builder(new XmlFactory(new WstxInputFactory(), new WstxOutputFactory())).build();
```

### 4. Unsafe Inline Factory Constructor

```java
XmlMapper mapper = new XmlMapper(new XmlFactory());
```
-> 
```java
XmlMapper mapper = new XmlMapper(new XmlFactory(new WstxInputFactory(), new WstxOutputFactory()));
```

### 5. Unsafe Builder with Inline Factory

```java
XmlMapper mapper = XmlMapper.builder(new XmlFactory()).build();
```
-> 
```java
XmlMapper mapper = XmlMapper.builder(new XmlFactory(new WstxInputFactory(), new WstxOutputFactory())).build();
```

==COMMIT_MSG==
Add errorprone to detect xml mapper without input / output factories
==COMMIT_MSG==

## Possible downsides?
Currently if there is a factory without input / output we delete it, inline it and add the factory, this could be bad if the factory is used else where


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/UnsafeXmlMapperConstruction.java
```diff
@@ -0,0 +1,285 @@
...
+            NewClassTree tree, List<? extends ExpressionTree> args, VisitorState state) {
+        if (args.size() != 1) {
+            return Description.NO_MATCH;
+        }
+        ExpressionTree arg = args.get(0);
+        if (isUnsafeXmlFactoryCtor(arg, state)) {
+            return buildAndDescribeFix(
+                    tree, "new XmlMapper(" + SAFE_FACTORY_CTOR + ")", ImportAction.REMOVE, Optional.empty(), state);
+        }
+        if (isIdentifierWithUnsafeXmlFactoryCtor(arg, state)) {
```
FinlayRJW, line 143: There is a case for when we have a state such as:

```java
XmlFactory factory = new XmlFactory(...);
XmlMapper mapper = new XmlMapper(factory);
```

that we should not inline it as we currently are but should instead replace the existing `XmlFactory` in place

There is also a risk that replace it is dangerous and we should only provide a fix in cases where it has no current inputs?

PR #3119, opened 2025-06-03
Make `gradle-baseline` configuration cache compatible 

FinlayRJW:
## Before this PR
Baseline failed on configuration cache for the following:

```
field `compilerArgumentProviders` of `org.gradle.api.tasks.compile.CompileOptions_Decorated`
    - bean of type `com.palantir.baseline.plugins.BaselineImmutables$ImmutablesCommandLineArgumentProvider`
        - field `resolvedArtifacts` of `com.palantir.baseline.plugins.BaselineImmutables$ImmutablesCommandLineArgumentProvider`
            - [error] cannot serialize object of type `org.gradle.api.internal.artifacts.result.DefaultResolvedArtifactResult`, a subtype of `org.gradle.api.artifacts.result.ArtifactResult`, as these are not supported with the configuration cache.
    - bean of type `com.palantir.baseline.plugins.BaselineModuleJvmArgs$1`
        - field `val$project` of `com.palantir.baseline.plugins.BaselineModuleJvmArgs$1`
            - [error] cannot serialize object of type `org.gradle.api.internal.project.DefaultProject`, a subtype of `org.gradle.api.Project`, as these are not supported with the configuration cache.
        - field `val$sourceSet` of `com.palantir.baseline.plugins.BaselineModuleJvmArgs$1`
            - [error] cannot serialize object of type `org.gradle.api.internal.tasks.DefaultSourceSet`, a subtype of `org.gradle.api.tasks.SourceSet`, as these are not supported with the configuration cache.
field `forkOptions` of `org.gradle.api.tasks.compile.CompileOptions_Decorated`
    - bean of type `org.gradle.api.tasks.compile.ForkOptions_Decorated`
        - field `jvmArgumentProviders` of `org.gradle.api.tasks.compile.ForkOptions_Decorated`
            - bean of type `com.palantir.baseline.plugins.BaselineImmutables$ImmutablesCommandLineArgumentProvider`
                - field `resolvedArtifacts` of `com.palantir.baseline.plugins.BaselineImmutables$ImmutablesCommandLineArgumentProvider`
                    - [error] cannot serialize object of type `org.gradle.api.internal.artifacts.result.DefaultResolvedArtifactResult`, a subtype of `org.gradle.api.artifacts.result.ArtifactResult`, as these are not supported with the configuration cache.

```

The build was also failing due to an issue with spotless not being configuration cache compatible - this was resolved with a version bump - issue:

```
    - [error] external process started `/usr/bin/git --version`
    - [error] external process started `/usr/bin/git config --system --show-origin --list -z`
    - [error] external process started `bash --login -c which git`
    - [error] external process started `xcode-select -p`
```

## After this PR

Now configuration cache is working and a test class `ConfigurationCacheTest` has been added to prevent regression.

We should have maintained lazyness but moved to configuration cacheable

==COMMIT_MSG==
Make baseline configuration cacheable 
==COMMIT_MSG==


gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
```diff
@@ -91,23 +98,29 @@ public void apply(Project project) {
...
                                 public Iterable<String> asArguments() {
                                     // The '--release' flag is set when BaselineJavaVersion is not used.
-                                    if (!project.getPlugins().hasPlugin(BaselineJavaVersion.class)) {
-                                        project.getLogger()
-                                                .debug(
-                                                        "BaselineModuleJvmArgs not applying args to compilation task "
-                                                                + "{} on {} due to lack of BaselineJavaVersion",
-                                                        javaCompile.getName(),
-                                                        project);
+                                    Project taskProject = javaCompile.getProject();
```
felixdesouza: how does this not give you an error for configuration cache?

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
```diff
@@ -410,4 +373,41 @@ enum OpensMode {
         COMPILATION,
         RUNTIME;
     }
+
+    public static final class ModuleJvmArgsArgumentProvider implements CommandLineArgumentProvider, Serializable {
```
felixdesouza: let's make this private

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
```diff
@@ -410,4 +373,41 @@ enum OpensMode {
...
+                    exports.get().stream(), classpathInfo.stream().flatMap(info -> info.exports().stream()));
+            Stream<String> allOpens =
+                    Stream.concat(opens.get().stream(), classpathInfo.stream().flatMap(info -> info.opens().stream()));
+
+            Stream<String> exportsArgs = allExports.distinct().sorted().flatMap(BaselineModuleJvmArgs::addExportArg);
+            Stream<String> opensArgs = allOpens.distinct().sorted().flatMap(BaselineModuleJvmArgs::addOpensArg);
+
+            ImmutableList<String> arguments =
+                    Stream.concat(exportsArgs, opensArgs).collect(ImmutableList.toImmutableList());
+            log.debug("BaselineModuleJvmArgs executing task on project {} with exports: {}", projectPath, arguments);
```
felixdesouza: you shouldn't use loggers from other classes i.e. in this case you're using the logger from `BaselineModuleJvmArgs`, make a new logger for `ModuleJbmArgsArgumentProvider` so you can distinguish where lines are coming from

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
```diff
@@ -410,4 +373,41 @@ enum OpensMode {
...
+                String projectPath) {
+            this.exports = exports;
+            this.opens = opens;
+            this.classpath = classpath;
+            this.projectPath = projectPath;
+        }
+
+        @Override
+        public Iterable<String> asArguments() {
+            ImmutableList<JarManifestModuleInfo> classpathInfo = collectClasspathInfo(null, classpath);
```
felixdesouza: a typical assumption is that most methods do not take in null parameters unless marked with `@Nullable`. So usually do that or mark it with `Optional`. I think in this case, `Project` is not actually used by anything and all of its overloads, so you can remove the parameter. You can probably make this plugin class abstract, and inject `ConfigurationContainer`, so then you really don't need any use for `project`. It just makes it far clearer to see what's going on especially when you're trying to get rid of references/capture to it.

gradle-baseline-java/src/test/groovy/com/palantir/baseline/ConfigurationCacheTest.groovy
```diff
@@ -0,0 +1,45 @@
...
+ * limitations under the License.
+ */
+
+package com.palantir.baseline
+
+import nebula.test.IntegrationTestKitSpec
+
+class ConfigurationCacheTest extends IntegrationTestKitSpec {
+
+    def setup() {
```
felixdesouza, line 23: have you tried just disabling this test to see whether this is what is pushing it over the edge?

changelog/@unreleased/pr-3119.v2.yml
```diff
@@ -0,0 +1,5 @@
+type: improvement
+improvement:
+  description: Make baseline configuration cacheable
```
felixdesouza:
```suggestion
  description: Make `gradle-baseline` configuration-cache compatible
```

This phrasing is somewhat more accurate and is similar to the nomenclature other plugins appear to be using.

felixdesouza (dismissed):

gradle-baseline-java/src/test/groovy/com/palantir/baseline/ConfigurationCacheTest.groovy
FinlayRJW: I don't think it can be this test as was failing from the first commit which doesn't have it

felixdesouza: oh that's so interesting

FinlayRJW: It looks like its hanging on this test:

```groovy
    def '#gradleVersionNumber: java 17 preview javadoc works'() {
        when:
        buildFile << '''
        javaVersions {
            libraryTarget = 11
            distributionTarget = '17_PREVIEW'
        }
        '''.stripIndent(true)
        mainJava << java17PreviewCode

        then:
        runTasksSuccessfully('javadoc', '-i')

        where:
        gradleVersionNumber << GRADLE_TEST_VERSIONS
    }
```

at least thats where it broke when I ran with ssh

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
FinlayRJW: good spot! hadn't noticed I did this

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
FinlayRJW: nice, have switch to injecting

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
FinlayRJW: huh - I don't know! will take a look

FinlayRJW: not sure why this was not blowing up as is banned but have refactored to use `ModuleJvmArgsArgumentProvider` as well

gradle-baseline-java/src/test/groovy/com/palantir/baseline/ConfigurationCacheTest.groovy
FinlayRJW: its magically working today - love that!

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
FinlayRJW: I think maybe this was just not being hit by the `classes` path?

FinlayRJW:
PJF is does not work with the bumped version of spotless - we must first rollout https://github.com/palantir/palantir-java-format/pull/1313 before we can merge this

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
```diff
@@ -183,52 +173,28 @@ public void execute(Task task) {
...
-                }
+            project.getTasks().withType(JavaExec.class).configureEach(javaExec -> {
+                javaExec.getJvmArgumentProviders()
+                        .add(new ModuleJvmArgsArgumentProvider(
+                                extension.exports(),
+                                extension.opens(),
+                                javaExec.getClasspath(),
+                                project.getPath(),
+                                OpensMode.RUNTIME,
+                                true));
```
felixdesouza: why is this true and not based on the plugin existence? like in the previous calls?

FinlayRJW: This is only used here (in complication mode):

```java
            if (mode == OpensMode.COMPILATION && !baselineJavaVersionEnabled) {
                log.debug(
                        "BaselineModuleJvmArgs not applying args to compilation task on {} due to lack of "
                                + "BaselineJavaVersion",
                        projectPath);
                return ImmutableList.of();
            }
```

and here we are always calling `OpensMode.RUNTIME` so we may as well pass in true - I think this would be clearer as an optional so will update

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
CRogers: It might be because the test uses `--dry-run`. I don't think Gradle will detect the usages unless the code is actually run (tasks are actually run? Can't see what this code was)

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineImmutables.java
```diff
@@ -43,14 +44,24 @@ public void apply(Project project) {
                 project.getTasks()
                         .named(sourceSet.getCompileJavaTaskName(), JavaCompile.class)
                         .configure(javaCompileTask -> {
+                            Provider<Set<ResolvedArtifactResult>> resolvedArtifactsProvider =
```
CRogers, line 50: Might be worth adding a comment here about doing this for configuration cache.

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
FinlayRJW: okay I've made this a bit nice by adding two overloads one for complication and one for runtime args so should be easier to understand now

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
FinlayRJW: your right! removing dry run reveals a bunch more issues!

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
```diff
@@ -406,8 +326,84 @@ static Builder builder() {
         class Builder extends ImmutableJarManifestModuleInfo.Builder {}
     }
 
-    enum OpensMode {
-        COMPILATION,
-        RUNTIME;
+    private static final class ModuleJvmArgsArgumentProvider implements CommandLineArgumentProvider, Serializable {
```
CRogers: I think here, you could make this an `abstract` class and then replace `Provider<Set<String>> exports` with `SetProperty<String> getExports()`. Since the java compile tasks lists these as `@Nested`:

```java
@Nested
public List<CommandLineArgumentProvider> getCompilerArgumentProviders() {
    return this.compilerArgumentProviders;
}
```

it means that we can also annotate thense all as `@Input`s and if there were task dependencies (say from the exports/opens - unlikely though) they would get linked up automatically.

Also makes it Serializable automatically.

Also reduces some boilerplate and allows injecting things.

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
FinlayRJW: cool if I remove --dry-run and switch to classes everything works with our changes

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
```diff
@@ -80,37 +89,21 @@ public void apply(Project project) {
...
-                                                    arguments);
-                                    return arguments;
-                                }
-                            });
+                            .add(new ModuleJvmArgsArgumentProvider(
+                                    extension.exports(),
+                                    extension.opens(),
+                                    getConfigurations().getByName(annotationProcessorConfigurationName.get()),
+                                    javaCompile.getProject().getPath(),
+                                    javaCompile.getProject().getPlugins().hasPlugin(BaselineJavaVersion.class)));
```
CRogers: The semantics of this have changed. Before, this was evaluating `project.getPlugins().hasPlugin(BaselineJavaVersion.class)` when task inputs were calculated, now `hasPlugin` is running during the configuration phase. Instead, I think we should make `Property<Boolean>` on the `ModuleJvmArgsArgumentProvider`, set it lazily to be this `hasPlugin`, then it will happen as late as possible.

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
```diff
@@ -406,8 +326,84 @@ static Builder builder() {
...
+            this.exports = exports;
+            this.opens = opens;
+            this.classpath = classpath;
+            this.projectPath = projectPath;
+            this.baselineJavaVersionEnabled = Optional.empty();
+        }
+
+        @Override
+        public Iterable<String> asArguments() {
+            if (baselineJavaVersionEnabled.isPresent() && !baselineJavaVersionEnabled.get()) {
```
CRogers: This could be `baselineJavaVersionEnabled.orElse(false)`. Same for if we're moving to a property, it would be `getBaselineJavaVersionEnabled().orElse(false)`.

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
FinlayRJW: okay I think I have done this correctly, made it abstract and used `@Inputs`

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
FinlayRJW: good spot have changed to `project.provider(() -> project.getPlugins().hasPlugin(BaselineJavaVersion.class)`

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineImmutables.java
FinlayRJW: added 

```
// Previously the anonymous classes where capturing project thus breaking the configuration
// cache, now we only supply a provider of ResolvedArtifactResults which is configuration
// cache compatible
```

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
```diff
@@ -406,8 +337,79 @@ static Builder builder() {
...
+        @InputFiles
+        @Classpath
+        public abstract ConfigurableFileCollection getClasspath();
+
+        @Input
+        public abstract Property<String> getProjectPath();
+
+        @Input
+        @org.gradle.api.tasks.Optional
+        public abstract Property<Boolean> getBaselineJavaVersionEnabled();
```
felixdesouza: these annotations have no effect, it's not a task. Everything should work just fine without them.

FinlayRJW: are you sure? without them I get

```
  - Type 'org.gradle.api.tasks.compile.JavaCompile' property 'options.compilerArgumentProviders.$1.baselineJavaVersionEnabled' is missing an input or output annotation.
    
    Reason: A property without annotation isn't considered during up-to-date checking.
    
    Possible solutions:
      1. Add an input or output annotation.
      2. Mark it as @Internal.
```

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
```diff
@@ -80,37 +94,24 @@ public void apply(Project project) {
...
+                    ModuleJvmArgsArgumentProvider provider =
+                            project.getObjects().newInstance(ModuleJvmArgsArgumentProvider.class);
+
+                    provider.getExports().set(extension.exports());
+                    provider.getOpens().set(extension.opens());
+                    provider.getClasspath()
+                            .from(getConfigurations().getByName(annotationProcessorConfigurationName.get()));
+                    provider.getProjectPath().set(project.getPath());
+                    provider.getBaselineJavaVersionEnabled()
+                            .set(project.provider(() -> project.getPlugins().hasPlugin(BaselineJavaVersion.class)));
```
felixdesouza: can't you use your helper here

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
FinlayRJW: I think they are needed - https://github.com/palantir/gradle-baseline/pull/3119#discussion_r2128685154

felixdesouza: sorry my bad, I missed the `@Nested` thing on the property that consumes it.

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
```diff
@@ -406,8 +334,79 @@ static Builder builder() {
...
+        private static final Logger log = Logging.getLogger(ModuleJvmArgsArgumentProvider.class);
+
+        @Input
+        public abstract SetProperty<String> getExports();
+
+        @Input
+        public abstract SetProperty<String> getOpens();
+
+        @InputFiles
+        @Classpath
```
CRogers: I think you only need `@Classpath` here.

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
```diff
@@ -80,37 +94,21 @@ public void apply(Project project) {
...
-                                                    javaCompile.getName(),
-                                                    project,
-                                                    arguments);
-                                    return arguments;
-                                }
-                            });
+                    ModuleJvmArgsArgumentProvider provider = newModuleJvmArgsArgumentProvider(
+                            project,
+                            extension,
+                            getConfigurations().getByName(annotationProcessorConfigurationName.get()));
```
CRogers: Ideally don't call `getByName` at configuration time, as it realises the Configuration (they can now be lazy). Instead called `named` and pass a provider, which you can then give to `newModuleJvmArgsArgumentProvider` (which might want to take an `Object` like `.from` does as other places give it a raw `FileCollection`).

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
```diff
@@ -80,37 +94,21 @@ public void apply(Project project) {
             // https://github.com/gradle/gradle/issues/18824
             // However, we set sourceCompatibility in BaselineJavaVersion to opt out of the '--release' flag.
             project.getExtensions().getByType(SourceSetContainer.class).configureEach(sourceSet -> {
+                Provider<String> annotationProcessorConfigurationName =
+                        project.provider(sourceSet::getAnnotationProcessorConfigurationName);
```
CRogers: This provider seems needless, you can just call `sourceSet.getAnnotationProcessorConfigurationName()` below inside the `javaCompileProvider.configure`

FinlayRJW: yep this is was a left over from how I was doing it originally updated

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
```diff
@@ -406,8 +330,78 @@ static Builder builder() {
         class Builder extends ImmutableJarManifestModuleInfo.Builder {}
     }
 
-    enum OpensMode {
-        COMPILATION,
-        RUNTIME;
+    public abstract static class ModuleJvmArgsArgumentProvider implements CommandLineArgumentProvider {
+        private static final Logger log = Logging.getLogger(ModuleJvmArgsArgumentProvider.class);
+
+        @Input
```
CRogers: On second thoughts... I think we should mark these all as `@Internal` because we don't really care what these values are for task up-to-dateness, the only truly important input to the task is the actual command line args that get calculated. If the classpath changes slightly but the arg we produce does not really change - then we shouldn't need to rerun the compilation task again, as the args to it will be the same.

CRogers (dismissed):

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineReleaseCompatibility.java
```diff
@@ -53,17 +53,23 @@ public void apply(Project project) {
...
         private final JavaCompile javaCompile;
 
+        private final boolean hasBaselineJavaVersionPlugin;
+        private final String projectString;
+
         private ReleaseFlagProvider(JavaCompile javaCompile) {
             this.javaCompile = javaCompile;
+            Project project = javaCompile.getProject();
+            this.hasBaselineJavaVersionPlugin = project.getPlugins().hasPlugin(BaselineJavaVersion.class);
+            this.projectString = project.toString();
```
felixdesouza: it's a bit weird to use the `toString` of the project, it was weird in the diff before that as well, but a bit more implicit. Try using the project path and call it as such

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineReleaseCompatibility.java
```diff
@@ -53,17 +53,23 @@ public void apply(Project project) {
     private static final class ReleaseFlagProvider implements CommandLineArgumentProvider {
         private final JavaCompile javaCompile;
 
+        private final boolean hasBaselineJavaVersionPlugin;
+        private final String projectString;
+
         private ReleaseFlagProvider(JavaCompile javaCompile) {
             this.javaCompile = javaCompile;
+            Project project = javaCompile.getProject();
+            this.hasBaselineJavaVersionPlugin = project.getPlugins().hasPlugin(BaselineJavaVersion.class);
```
felixdesouza: I don't believe this is safe to do. When you create this object - which is during configuration - the java version plugin could potentially be applied *after* this plugin. Meaning the `hasBaselineJavaVersionPlugin` is actually wrong, and behaviour is subtly changed. It still needs to happen as late as possible.

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
```diff
@@ -200,8 +200,16 @@ public void execute(Task task) {
             project.getTasks().withType(Jar.class).configureEach(new Action<Jar>() {
                 @Override
                 public void execute(Jar jar) {
-                    Provider<String> jarNameProvider = project.provider(jar::getName);
-                    Provider<String> projectPathProvider = project.provider(jar.getProject()::getPath);
+                    String jarName = jar.getName();
+                    String projectPath = jar.getProject().getPath();
+
+                    Provider<Set<String>> exportsProvider = extension.exports();
```
felixdesouza: I'd be more curious about why this is failing, extension is a relatively simple type, we shouldn't have to pull everything out of extensions no? it'll just make plugins significantly more verbose. So I would investigate what the reason is there. If there's a `Project` reference, we should look at what we need the project for and see what can be done to make it better.

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineReleaseCompatibility.java
FinlayRJW: switched to using a provider which should still be configuration cache safe and be lazy

``` java
this.hasBaselineJavaVersionPlugin =
                    project.provider(() -> project.getPlugins().hasPlugin(BaselineJavaVersion.class));
```

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
FinlayRJW: this was not needed - I was being a bit keen on pulling bits out! have reverted

changelog/@unreleased/pr-3119.v2.yml
```diff
@@ -0,0 +1,5 @@
+type: improvement
+improvement:
+  description: Make baseline configuration cacheable for compilation
```
felixdesouza:
```suggestion
  description: Make `gradle-baseline` configuration cache compatible for the `classes` task.
```

felixdesouza:
👍

FinlayRJW:
👍

CRogers (approved):

PR #3121, opened 2025-06-10
Upgrade Error Prone to 2.38.0

pkoenig10:
Another attempt at https://github.com/palantir/gradle-baseline/pull/3101.


gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineErrorProneIntegrationTest.groovy
```diff
@@ -203,38 +203,6 @@ class BaselineErrorProneIntegrationTest extends AbstractPluginTest {
         '''.stripIndent()
     }
 
-    def 'compileJava applies patches when errorProneApply contains specific checks including disabled'() {
```
pkoenig10, line 206: This is an Error Prone behavior change. See https://github.com/google/error-prone/pull/4699#discussion_r1899667974.

I'm not sure there is much we can do about this. I'm not sure if or how heavily we are relying on this behavior in our internal automations. It appears that we deliberately wanted this behavior in https://github.com/palantir/gradle-baseline/pull/1109, but then a lot of this logic was removed in https://github.com/palantir/gradle-baseline/pull/2944.

CRogers (approved):
Thanks!

PR #3124, opened 2025-06-12
Disable some more built in Google errorprones

CRogers:
## Before this PR
When we upgraded errorprone in https://github.com/palantir/gradle-baseline/pull/3121, it brought in a load of new checks. Some of these are not valuable or relevant with how we write code at Palantir.

We disabled some in https://github.com/palantir/gradle-baseline/pull/3123 already, but there are more.

## After this PR
==COMMIT_MSG==
Disable some new low-value Google provided errorprone checks:
* `CatchingUnchecked`
* `DifferentNameButSame`
* `PreferredInterfaceType`
* `ThrowSpecificExceptions`
* `UnnecessaryFinal`
* `UnnecessaryFullyQualified`
* `UnnecessarilyVisible`
==COMMIT_MSG==

## Possible downsides?
This is kinda a stop-gap measure until we fix suppressible-error-prone suppressing warnings.

PR #3126, opened 2025-06-13
Disable Google-authored errorprone checks that have been spuriously re-enabled by a bug in error-prone 2.37.0

CRogers:
## Before this PR
After upgrading to error-prone 2.38.0 in #3121, we started seeing a crazy number of "new" error-prone checks starting to be suppressed across our codebases:

Screenshot 2025-06-13 at 16 37 09

We attempted to disable a few in #3123 and #3124, but there are still loads lying around.

I spent today investigating deeper and found the real cause is a bug in error-prone, from this change: https://github.com/google/error-prone/pull/4699.

The issue is that on [this line of error-prone code](https://github.com/google/error-prone/blob/028246ba884a77849e6d6240073c2bb3af4c0862/check_api/src/main/java/com/google/errorprone/ErrorProneAnalyzer.java#L91), it checks to see whether a check has been disabled in the user provided options, but the default disabled checks are not disabled via user given options.

There's actually already an issue for this: https://github.com/google/error-prone/issues/4943

## After this PR
==COMMIT_MSG==
Disable Google-authored errorprone checks that have been spuriously re-enabled by a bug in error-prone 2.27.0
==COMMIT_MSG==

I will follow up with an errorprone PR, and pray they merge it.

### DIFF
--- a/gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineErrorProne.java
+++ b/gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineErrorProne.java
@@ -128,6 +128,10 @@ public final class BaselineErrorProne implements Plugin<Project> {
         // to be enabled by default. So we're disabling them here until error-prone is fixed, then these can all
         // be removed.
         // See https://github.com/google/error-prone/issues/4943
+        // Some of the checks in the original file:
+        // https://github.com/google/error-prone/blob/v2.38.0/core/src/main/java/
+        // com/google/errorprone/scanner/BuiltInCheckerSuppliers.java#L1168-L1307
+        // no longer exist and have been removed, so it's not the same.
         errorProneOptions.disable(
                 "AddNullMarkedToPackageInfo",
                 "AndroidJdkLibsChecker",
@@ -158,7 +162,6 @@ public final class BaselineErrorProne implements Plugin<Project> {
                 "DefaultLocale",
                 "DepAnn",
                 "DifferentNameButSame",
-                "EmptyIfStatement",
                 "EqualsBrokenForNull",
                 "EqualsMissingNullable",
                 "ExpectedExceptionChecker",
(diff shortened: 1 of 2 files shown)


gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineErrorProne.java
```diff
@@ -152,6 +124,152 @@ private static void configureErrorProneOptions(ErrorProneOptions errorProneOptio
                 // seamless rollout.
                 "SuperCallToObjectMethod");
 
+        // These checks are default disabled in error-prone. However, in error-prone 2.37.0 a bug has caused them
+        // to be enabled by default. So we're disabling them here until error-prone is fixed, then these can all
+        // be removed.
+        // See https://github.com/google/error-prone/issues/4943
+        errorProneOptions.disable(
```
pkoenig10, line 135: Could we reference `BuiltInCheckerSuppliers.DISABLED_CHECKS` here instead of copying this list?

pkoenig10 (dismissed):

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineErrorProne.java
CRogers: No, this a gradle plugin, that class is executed in the compiler context so we can't access it. Not without putting errorprone on the Gradle buildscript classpath, which I'd rather not do (and might not work as it expects to have access to types from the compiler).

bmarcaur (approved):

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineErrorProne.java
pkoenig10: Ah right 👍

Stephan202:
Hey @CRogers! Sorry to see the hassle this caused. PTAL at google/error-prone#5102 and let me know whether that works for you guys.

PR #3128, opened 2025-06-16
Allow violations of TimeInStaticInitializer rule in test code

invliD:
## Before this PR
The [TimeInStaticInitializer](https://errorprone.info/bugpattern/TimeInStaticInitializer) errorprone rule is enabled in both production and test code.

## After this PR
==COMMIT_MSG==
Allow violations of TimeInStaticInitializer rule in test code
==COMMIT_MSG==

While this rule is valuable for production code, initializing objects statically based on time is not unreasonable since test JVMs are short-lived anyway.

## Possible downsides?
🤷🏻‍♂️


pkoenig10:
I think this is still valuable. It's better for tests to be deterministic rather than use the current time.

pkoenig10:
Regardless, `TimeInStaticInitializer` already ignores test code.

https://github.com/google/error-prone/blame/d2540bf4c0acb59d56a35855608b396a04a5668a/core/src/main/java/com/google/errorprone/bugpatterns/time/TimeInStaticInitializer.java#L76-L78

invliD:
I do see these being flagged in test code internally. Closing anyway.

PR #3131, opened 2025-06-20
Improve logging and restore correct `--add-exports` handling in `BaselineModuleJvmArgs`

FinlayRJW:
## Before this PR
In https://github.com/palantir/gradle-baseline/pull/3119 we fixed configuration cache for baseline but this has introduced and error where exports where not being correctly added and the logging was incorrect:

```
2025-06-20T09:09:39.332-0400 [DEBUG] [com.palantir.baseline.plugins.BaselineModuleJvmArgs$ModuleJvmArgsArgumentProvider] BaselineModuleJvmArgs executing task on project property 'projectPath' with exports: []
```

This appears to have been caused by eagerly calling `test.getClasspath()` which caused the `classpath` to be resolved at configuration time, before all dependencies (including those added later or via plugins) were present. As a result, any module `--add-exports` or `--add-opens` entries contributed by jars added after the initial configuration phase were missed. This led to missing JVM arguments for runtime tasks such as Test and JavaExec.

## After this PR
This PR defers evaluation of the `classpath` by wiring the provider with a Callable/Provider that resolves the `classpath` at execution time, ensuring that all jars (including those added late) are correctly included. This restores the expected behavior: all relevant `--add-exports` and `--add-opens` flags are present for runtime tasks, regardless of when dependencies are added. Logging is fixed to show the actual task and project names rather than property objects

This results in a restored fucntionallity:

```
2025-06-20T15:01:18.095+0100 [DEBUG] [com.palantir.baseline.plugins.BaselineModuleJvmArgs$ModuleJvmArgsArgumentProvider] BaselineModuleJvmArgs executing <redacted> on project <redacted> with exports: [--add-exports, java.base/jdk.internal.platform=ALL-UNNAMED, --add-exports, java.base/sun.net=ALL-UNNAMED, --add-exports, java.management/sun.management=ALL-UNNAMED, --add-exports, jdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED, --add-exports, jdk.compiler/com.sun.tools.javac.code=ALL-UNNAMED, --add-exports, jdk.compiler/com.sun.tools.javac.file=ALL-UNNAMED, --add-exports, jdk.compiler/com.sun.tools.javac.main=ALL-UNNAMED, --add-exports, jdk.compiler/com.sun.tools.javac.parser=ALL-UNNAMED, --add-exports, jdk.compiler/com.sun.tools.javac.tree=ALL-UNNAMED, --add-exports, jdk.compiler/com.sun.tools.javac.util=ALL-UNNAMED, --add-exports, jdk.jfr/jdk.jfr.internal=ALL-UNNAMED, --add-opens, java.base/java.lang.invoke=ALL-UNNAMED, --add-opens, java.base/java.net=ALL-UNNAMED]
```

==COMMIT_MSG==
Fixes incorrect logging and restores correct handling of `--add-exports` and `--add-opens` in `BaselineModuleJvmArgs`

This change defers evaluation of the `classpath` to execution time, ensuring all relevant jars are included and their manifest entries are respected.
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-3131.v2.yml
+++ b/changelog/@unreleased/pr-3131.v2.yml
@@ -1,8 +1,8 @@
 type: fix
 fix:
   description: |-
-    Fixes incorrect logging and restores correct handling of `--add-exports` and `--add-opens` in BaselineModuleJvmArgs
+    Fixes incorrect logging and restores correct handling of `--add-exports` and `--add-opens` in `BaselineModuleJvmArgs`
 
-    This change defers evaluation of the classpath to execution time, ensuring all relevant jars are included and their manifest entries are respected.
+    This change defers evaluation of the `classpath` to execution time, ensuring all relevant jars are included and their manifest entries are respected.
   links:
   - https://github.com/palantir/gradle-baseline/pull/3131
--- a/changelog/@unreleased/pr-3131.v2.yml
+++ b/changelog/@unreleased/pr-3131.v2.yml
@@ -1,0 +1,8 @@
+type: fix
+fix:
+  description: |-
+    Fixes incorrect logging and restores correct handling of `--add-exports` and `--add-opens` in BaselineModuleJvmArgs
+
+    This change defers evaluation of the classpath to execution time, ensuring all relevant jars are included and their manifest entries are respected.
+  links:
+  - https://github.com/palantir/gradle-baseline/pull/3131
--- a/gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
+++ b/gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
@@ -99,7 +99,8 @@ public abstract class BaselineModuleJvmArgs implements Plugin<Project> {
                     ModuleJvmArgsArgumentProvider provider = newModuleJvmArgsArgumentProvider(
                             project,
                             extension,
-                            project.getConfigurations().named(sourceSet.getAnnotationProcessorConfigurationName()),
+                            project.files(project.getConfigurations()
+                                    .named(sourceSet.getAnnotationProcessorConfigurationName())),
                             javaCompile.getName());
                     provider.getBaselineJavaVersionEnabled()
                             .set(project.provider(() -> project.getPlugins().hasPlugin(BaselineJavaVersion.class)));
@@ -175,14 +176,20 @@ public abstract class BaselineModuleJvmArgs implements Plugin<Project> {
 
             project.getTasks().withType(Test.class).configureEach(test -> {
                 ModuleJvmArgsArgumentProvider provider = newModuleJvmArgsArgumentProvider(
-                        project, extension, (Callable<FileCollection>) test::getClasspath, test.getName());
+                        project,
+                        extension,
+                        project.files((Callable<FileCollection>) test::getClasspath),
+                        test.getName());
                 test.getJvmArgumentProviders().add(provider);
                 setTaskInputsFromExtension(test, extension);
             });
 
             project.getTasks().withType(JavaExec.class).configureEach(javaExec -> {
                 ModuleJvmArgsArgumentProvider provider = newModuleJvmArgsArgumentProvider(
-                        project, extension, (Callable<FileCollection>) javaExec::getClasspath, javaExec.getName());
+                        project,
+                        extension,
+                        project.files((Callable<FileCollection>) javaExec::getClasspath),
+                        javaExec.getName());
                 javaExec.getJvmArgumentProviders().add(provider);
                 setTaskInputsFromExtension(javaExec, extension);
             });
@@ -234,7 +241,10 @@ public abstract class BaselineModuleJvmArgs implements Plugin<Project> {
     }
 
     private static ModuleJvmArgsArgumentProvider newModuleJvmArgsArgumentProvider(
-            Project project, BaselineModuleJvmArgsExtension extension, Object classpath, String taskName) {
+            Project project,
+            BaselineModuleJvmArgsExtension extension,
+            ConfigurableFileCollection classpath,
+            String taskName) {
         ModuleJvmArgsArgumentProvider provider = project.getObjects().newInstance(ModuleJvmArgsArgumentProvider.class);
         provider.getExports().set(extension.exports());
         provider.getOpens().set(extension.opens());
@@ -380,9 +390,7 @@ public abstract class BaselineModuleJvmArgs implements Plugin<Project> {
                     isCompilation ? compilationArgs(allExports, allOpens) : runtimeArgs(allExports, allOpens);
 
             log.debug(
-                    isCompilation
-                            ? "BaselineModuleJvmArgs compiling {} on project {} with exports: {}"
-                            : "BaselineModuleJvmArgs executing {} on project {} with exports: {}",
+                    "BaselineModuleJvmArgs executing {} on project {} with exports: {}",
                     getTaskName().getOrNull(),
                     getProjectPath().getOrNull(),
                     args);
--- a/gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
+++ b/gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
@@ -28,6 +28,7 @@ import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.Callable;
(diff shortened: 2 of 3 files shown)


gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
```diff
@@ -376,9 +381,10 @@ public final Iterable<String> asArguments() {
 
             log.debug(
                     isCompilation
-                            ? "BaselineModuleJvmArgs compiling on {} with exports: {}"
-                            : "BaselineModuleJvmArgs executing task on project {} with exports: {}",
-                    getProjectPath(),
+                            ? "BaselineModuleJvmArgs compiling {} on project {} with exports: {}"
+                            : "BaselineModuleJvmArgs executing {} on project {} with exports: {}",
+                    getTaskName().getOrNull(),
+                    getProjectPath().getOrNull(),
```
pkoenig10: Do we really need this `isCompilation` conditional? It seems a bit confusing to log something like "compiling 'compileJava'`. It should be pretty obvious from the task name whether it's a compilation task or not.

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
```diff
@@ -233,13 +234,13 @@ public void execute(Manifest manifest) {
     }
 
     private static ModuleJvmArgsArgumentProvider newModuleJvmArgsArgumentProvider(
-            Project project, BaselineModuleJvmArgsExtension extension, Object classpath) {
+            Project project, BaselineModuleJvmArgsExtension extension, Object classpath, String taskName) {
```
pkoenig10: Might be better for callers to wrap values in `project.files(...)` to get some better type-safety here.

```suggestion
            Project project, BaselineModuleJvmArgsExtension extension, ConfigurableFileCollection classpath, String taskName) {
```

FinlayRJW: yep good idea updated

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
```diff
@@ -359,9 +373,10 @@ public final Iterable<String> asArguments() {
 
             if (isCompilation && !isBaselineJavaVersionEnabled) {
                 log.debug(
-                        "BaselineModuleJvmArgs not applying args to compilation task on {} due to lack of "
+                        "BaselineModuleJvmArgs not applying args to compilation task {} on project {} due to lack of "
                                 + "BaselineJavaVersion",
-                        getProjectPath());
+                        getTaskName().getOrNull(),
+                        getProjectPath().getOrNull());
```
CRogers: Can these ever be null? I'd just call `.get`

changelog/@unreleased/pr-3131.v2.yml
```diff
@@ -0,0 +1,8 @@
+type: fix
+fix:
+  description: |-
+    Fixes incorrect logging and restores correct handling of `--add-exports` and `--add-opens` in `BaselineModuleJvmArgs`
```
CRogers, line 4: It'd be really good to get a test for this.

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
```diff
@@ -173,15 +175,21 @@ public void execute(Task task) {
             });
 
             project.getTasks().withType(Test.class).configureEach(test -> {
-                ModuleJvmArgsArgumentProvider provider =
-                        newModuleJvmArgsArgumentProvider(project, extension, test.getClasspath());
```
CRogers, line 177: Wow, so calling `test.getClasspath()` eagerly evaluates stuff? I'm so used to being able to do this in Gradle: `task.getSomeFileCollection().from(otherTask.getClasspath())` and it just all work lazily...

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
```diff
@@ -173,15 +175,21 @@ public void execute(Task task) {
...
+                ModuleJvmArgsArgumentProvider provider = newModuleJvmArgsArgumentProvider(
+                        project,
+                        extension,
+                        project.files((Callable<FileCollection>) test::getClasspath),
```
felixdesouza, line 181: It might be worth using `Test::getStableClasspath`

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
```diff
@@ -173,15 +175,21 @@ public void execute(Task task) {
...
                 setTaskInputsFromExtension(test, extension);
             });
 
             project.getTasks().withType(JavaExec.class).configureEach(javaExec -> {
-                ModuleJvmArgsArgumentProvider provider =
-                        newModuleJvmArgsArgumentProvider(project, extension, javaExec.getClasspath());
+                ModuleJvmArgsArgumentProvider provider = newModuleJvmArgsArgumentProvider(
+                        project,
+                        extension,
+                        project.files((Callable<FileCollection>) javaExec::getClasspath),
```
felixdesouza: do the errors happen for both `Test` tasks and `JavaExec` tasks?

felixdesouza (commented):
Can we get some sort of test that verifies this fix?

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
CRogers: I was just looking at this [but it's `protected`](https://github.com/gradle/gradle/blob/2594bd957e6208aac6d7c7579a648d02f79ed3ea/platforms/jvm/testing-jvm/src/main/java/org/gradle/api/tasks/testing/Test.java#L1115). I'm also not sure what it's actually supposed to do.

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
CRogers: Hmm, I don't get why this was going wrong, calling `getClasspath()` on Test [doesn't actually evaluate anything](https://github.com/gradle/gradle/blob/2594bd957e6208aac6d7c7579a648d02f79ed3ea/platforms/jvm/testing-jvm/src/main/java/org/gradle/api/tasks/testing/Test.java#L1125)

felixdesouza: I don't think this is the case. I don't think anything is actually getting resolved. My guess is the fact that you can `setClasspath` and that might be happening via reflection somehow. You end up using a reference to a classpath that you don't ever see again.

This is probably why you have `stableClasspath` which captures whatever the latest field is when it's evaluated.

changelog/@unreleased/pr-3131.v2.yml
FinlayRJW: Yeah - I tried to make a test last week and failed, couldn't figure out how to repo locally, will have another go but can always do as a FLUP so we can unblock baseline upgrades

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
FinlayRJW: I think we are only seeing this in `Test` but figured it was better fix both incase it is also broken in `JavaExec`

FinlayRJW: Actually `JavaExec` doesn't have the same idea of a `stableClasspath` so I suspect we don't need to do any changes for it

FinlayRJW:
> Can we get some sort of test that verifies this fix?

Added a test

changelog/@unreleased/pr-3131.v2.yml
FinlayRJW: Test added now! not sure why I struggled so much to write one on Friday

CRogers (approved):

crogoz (approved):

PR #3135, opened 2025-06-23
Implement Error Prone `GradleExceptionUsage` to discourage throwing `GradleException`s in production code

kelvinou01:
## Before this PR

[`GradleException`s should only originate from within Gradle's internal code](https://github.com/palantir/gradle-incremental-configuration-cache/pull/1#discussion_r2155101552). This is [not immediately obvious from the docs](https://docs.gradle.org/current/javadoc/org/gradle/api/GradleException.html). 

## After this PR

==COMMIT_MSG==
Implement Error Prone `GradleExceptionUsage` to discourage throwing `GradleException`s in production code
==COMMIT_MSG==

### DIFF
--- a/gradle-baseline-java/src/main/java/com/palantir/baseline/extensions/BaselineErrorProneExtension.java
+++ b/gradle-baseline-java/src/main/java/com/palantir/baseline/extensions/BaselineErrorProneExtension.java
@@ -42,6 +42,7 @@ public abstract class BaselineErrorProneExtension {
             "ExecutorSubmitRunnableFutureIgnored",
             "ExtendsErrorOrThrowable",
             "FinalClass",
+            "GradleExceptionUsage",
             "IllegalSafeLoggingArgument",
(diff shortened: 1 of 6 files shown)


gradle-baseline-java/src/main/java/com/palantir/baseline/extensions/BaselineErrorProneExtension.java
```diff
@@ -99,6 +100,7 @@ public abstract class BaselineErrorProneExtension {
 
     /**
      * .
+     *
```
kelvinou01, line 103: Added by our formatter

CRogers:
We should put this in https://github.com/palantir/gradle-guide rather than baseline

kelvinou01:
As discussed in person, GradleExceptions aren't actually a bad thing for users to throw, so let's abandon this errorprone.

PR #3165, opened 2025-07-14
Bump error-prone to 2.40.0

CRogers:
## Before this PR
Internally, repos are being trying to take an upgrade of `com.github.ben-manes.caffeine:caffeine` to `3.2.2`. This release of caffeine [bumps the dependency of `error_prone_annotations` to 2.44.0](https://github.com/ben-manes/caffeine/compare/v3.2.1...v3.2.2#diff-697f70cdd88ba88fe77eebda60c7e143f6ad1286bca75017421e93ad84fb87dfR24).

Since we use gradle-consistent-versions and have a virtual platform for `com.google.errorprone:*`, this also bumps the version of `error_prone_check_api` used in the `annotationProcessor` configruation:

Screenshot 2025-07-14 at 13 27 05

The current latest `baseline-error-prone` that is running inside the build called the `ASTHelpers.findSuperMethod` which [got deleted in error-prone 2.39.0](https://github.com/google/error-prone/commit/175f4c87f827902d5b311681269f30c50d80f766) ([PR](https://github.com/google/error-prone/pull/5014#event-17407899713)). This causes this error:

```
java.lang.NoSuchMethodError: 'java.util.Optional com.google.errorprone.util.ASTHelpers.findSuperMethod(com.sun.tools.javac.code.Symbol$MethodSymbol, com.sun.tools.javac.code.Types)'
        at com.palantir.baseline.errorprone.ThrowSpecificity.matchMethod(ThrowSpecificity.java:68)
```

## After this PR

==COMMIT_MSG==
Upgrade error-prone to 2.40.0 and stop errors of the form when error-prone >2.39.0 is used:

```
java.lang.NoSuchMethodError: 'java.util.Optional com.google.errorprone.util.ASTHelpers.findSuperMethod(com.sun.tools.javac.code.Symbol$MethodSymbol, com.sun.tools.javac.code.Types)'
        at com.palantir.baseline.errorprone.ThrowSpecificity.matchMethod(ThrowSpecificity.java:68)
```
==COMMIT_MSG==

Upgrading allows people to upgrade dependencies that use error-prone-annotations >2.39.0.

## Possible downsides?
* `baseline-error-prone` uses a previously published version of `basline-error-prone` for it's own compilation. When we upgrade `error-prone` in `versions.props`, we upgrade the `error-prone` used by the previous `baseline-error-prone` in compilation. I've worked around this using `enforcedPlatform` (which might even be a nice trick to stop this kind of error happening in consumer repos - but needs more though).

---

We need to verify that the new checks added are reasonable and if we want them to autofix:

```
$ git diff --name-only --diff-filter A v2.38.0..v2.40.0
core/src/main/java/com/google/errorprone/bugpatterns/BooleanLiteral.java
core/src/main/java/com/google/errorprone/bugpatterns/PreferInstanceofOverGetKind.java
core/src/main/java/com/google/errorprone/bugpatterns/UnnecessaryQualifier.java
core/src/main/java/com/google/errorprone/bugpatterns/formatstring/LenientFormatStringUtils.java
core/src/test/java/com/google/errorprone/bugpatterns/BooleanLiteralTest.java
core/src/test/java/com/google/errorprone/bugpatterns/PreferInstanceofOverGetKindTest.java
core/src/test/java/com/google/errorprone/bugpatterns/UnnecessaryQualifierTest.java
docs/bugpattern/ExpensiveLenientFormatString.md
docs/bugpattern/UnnecessaryQualifier.md
```

* [x] [BooleanLiteral](https://github.com/google/error-prone/blob/v2.40.0/core/src/main/java/com/google/errorprone/bugpatterns/BooleanLiteral.java) ([Test](https://github.com/google/error-prone/blob/v2.40.0/core/src/test/java/com/google/errorprone/bugpatterns/BooleanLiteralTest.java))
  * This one seems fine and good, they even check for `Boolean.TRUE::equals` that we have a few instances of internally.
* [x] [PreferInstanceofOverGetKind](https://github.com/google/error-prone/blob/v2.40.0/core/src/main/java/com/google/errorprone/bugpatterns/PreferInstanceofOverGetKind.java) ([Test](https://github.com/google/error-prone/blob/v2.40.0/core/src/test/java/com/google/errorprone/bugpatterns/PreferInstanceofOverGetKindTest.java))
  * This is also a good change which improves code by enabling destructuring to happen. Going from equal to instanceof to destructing requires compilation twice though, we'll need to keep an eye on if this breaks upgrades for error-prone sources.
* [x] [UnnecessaryQualifier](https://github.com/google/error-prone/blob/v2.40.0/core/src/main/java/com/google/errorprone/bugpatterns/UnnecessaryQualifier.java) ([Test](https://github.com/google/error-prone/blob/v2.40.0/core/src/test/java/com/google/errorprone/bugpatterns/UnnecessaryQualifierTest.java))
  * Fixes something with Dagger injection. Willing to let it autopatch and see what happens. Lots of usages of `@Qualifer` internally.

---

We also need to check that suppressible-error-prone works with 2.40.0 - I tried it on [this PR](https://github.com/palantir/suppressible-error-prone/pull/18) and the test suites passed on [this build](https://app.circleci.com/jobs/github/palantir/suppressible-error-prone/966).


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ThrowSpecificity.java
```diff
@@ -65,7 +65,7 @@ public Description matchMethod(MethodTree tree, VisitorState state) {
             return Description.NO_MATCH;
         }
         Types types = state.getTypes();
-        if (ASTHelpers.findSuperMethod(ASTHelpers.getSymbol(tree), types).isPresent()) {
+        if (!ASTHelpers.findSuperMethods(ASTHelpers.getSymbol(tree), types).isEmpty()) {
```
esword, line 68: why this flip?

CRogers: `findSuperMethod` got deleted, we have to use `findSuperMethods` that returns a `Set` instead of a single value from the set. So checking the set is non-empty should be the same as the old logic.

esword (approved):

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ThrowSpecificity.java
esword: ah - didn't see the change of the "s"

PR #3173, opened 2025-07-16
Add PreferUncheckedIoException check

pkoenig10:
Prefer using `UncheckedIOException` or `SafeUncheckedIOException` when wrapping and rethrowing `IOException`. This helps consumers of these exceptions better better categorize errors.

```diff
  try {
      // do something
  } catch (IOException e) {
-     new RuntimeException("IO failed", e);
+     new UncheckedIOException("IO failed", e);
  }

  try {
      // do something
  } catch (IOException e) {
-     new SafeRuntimeException("IO failed", e);
+     new SafeUncheckedIoException("IO failed", e);
  }
```

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferUncheckedIoException.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferUncheckedIoException.java
@@ -37,7 +37,8 @@ import java.io.UncheckedIOException;
         link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
         linkType = BugPattern.LinkType.CUSTOM,
         severity = SeverityLevel.ERROR,
-        summary = "Prefer UncheckedIOException or SafeUncheckedIoException when wrapping IOException")
+        summary = "Prefer UncheckedIOException or SafeUncheckedIoException when wrapping IOException so consumers can"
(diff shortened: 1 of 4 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/PreferUncheckedIoException.java
```diff
@@ -0,0 +1,88 @@
...
+import com.sun.source.tree.NewClassTree;
+import java.io.IOException;
+import java.io.UncheckedIOException;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
+        severity = SeverityLevel.ERROR,
+        summary = "Prefer UncheckedIOException or SafeUncheckedIoException when wrapping IOException")
```
CRogers: Can you add a reason in the summary? It's useful for the devs who encounter them to understand why.

CRogers:
LGTM other than that

aldexis (dismissed):
Lgtm as well. One suggestion would be to add another test that error prone doesn't complain if we already use these classes, but defer to you (I don't think the test is super high value in itself, based on how error-prone works)

CRogers:
:+1:

PR #3196, opened 2025-07-29
Remove Preconditions RegexpSinglelineJava checkstyle rule

pkoenig10:
This one is a bit silly. It disallows

```java
import com.palantir.logsafe.Preconditions;

Preconditions.checkNotNull(foo);
```

but not

```java
import static com.palantir.logsafe.Preconditions.checkNotNull;

checkNotNull(foo);
```

The latter is far more common in our code (since our checkstyle rules allow this to be statically imported), so this rules is mostly useless.

Not to mention that the regex used here is quite complicated...


aldexis (approved):
Should we make this an error-prone instead?

PR #3197, opened 2025-07-29
bump spotless to version to `7.2.1`

FinlayRJW:
## Before this PR
see PJF PR https://github.com/palantir/palantir-java-format/pull/1384

We where seeing a bug where using spotless `6.22.0` and `scalafmt` would error:

```
A problem occurred configuring root project 'foundry-spark-root'.
> A problem occurred configuring project ':foundry-spark'.
   > You need to add a repository containing the '[org.scalameta:scalafmt-core_2.12:3.6.1]' artifact in 'foundry-spark/build.gradle'.
     E.g.: 'repositories { mavenCentral() }'
```

I believe this was due to spotless not being lazy enough / fully supporting configuration cache - this was fixed in version [`7.0.0`](https://github.com/diffplug/spotless/releases/tag/gradle%2F7.0.0)

## After this PR
Use the latest version of spotless to fix the bug. 

==COMMIT_MSG==
bump spotless to version to `7.2.1`
==COMMIT_MSG==

## Possible downsides?
The update order is important, if we release PJF before baseline then it wont merge and visa versa I think the best order is:
- Release PJF first
- update baseline to bump the version of spotless and set this version of PJF as the minimum 

This will mean that the update to PJF will be blocked unless you have the new version of baseline but the update of baseline wont be blocked as it will contain the correct minimum version of PJF


FinlayRJW:
tests pass locally with the new version of PJF so should be okay once that is released

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/LazyFormatterStep.java
```diff
@@ -56,4 +56,13 @@ public void readObject(ObjectInputStream input) throws IOException, ClassNotFoun
         FormatterStep serialized = (FormatterStep) input.readObject();
         delegate = () -> serialized;
     }
+
+    @Override
+    public void close() throws Exception {
```
CRogers, line 61: Is this method required by spotless now? If so, will upgrading PJF first before baseline bring in newer spotless and then error at runtime because this method is not implemented yet? Worth testing at least.

CRogers: Looks like `FormatterStep extends AutoCloseable` without a default method, so I think this will cause runtime issues and we'll need to upgrade baseline before PJF?

FinlayRJW: Okay so if you apply baseline first (without it having bump PJF):
```
      > 'com.diffplug.spotless.FormatterStep com.diffplug.spotless.FormatterStep.createLazy(java.lang.String, com.diffplug.spotless.ThrowingEx$Supplier, com.diffplug.spotless.ThrowingEx$Function)'

```

If you apply PJF first without it having baseline:
```
      > com/diffplug/spotless/SerializedFunction

```

CRogers (approved):

PR #3237, opened 2025-08-29
Remove DesignForExtension Checkstyle check

pkoenig10:
This check feel more annoying than it is useful. Abstract methods with non-final methods are very common (for example, Immutable classes with derived properties) and we generally don't add Javadoc to internal, non-library methods (adding Javadoc for simple accessors is somewhat tedious is any case).


aldexis:
I'm not necessarily in favor of removing this. I've personally found this check useful to remind me to make classes and/or methods final when they don't otherwise need to be. I wouldn't mind its description being somewhat different, but the reminder in itself feels somewhat useful to me. With that said, I don't feel extremely strongly about it and am fine with this change if the others want to remove it too.

PR #3244, opened 2025-09-03
[Error-prone] Add check to prevent usage of deprecated methods and fields, and disable `-Xlint:removal` by default

aldexis:
## Before this PR
[Internal discussion](https://pl.ntr/2wH).

When libraries want to change their public API, they typically rely on first deprecating the method/field/class before eventually removing it and breaking the ABI.

Some repositories rely on the `-Werror` and `-Xlint:deprecation` compiler flags to surface such cases so they can remove them. Similarly, just using `-Werror` would surface errors when using APIs that are marked `@Deprecated(forRemoval = true)` due to `-Xlint:removal` being on by default.

However, this has the downside of failing compilation and preventing library upgrades _as soon as they mark a method as deprecated_. This is problematic in the context of excavator, where we want to automate library upgrades, given this causes blocked excavators and actually _removes_ signal from within code that it is relying on a deprecated method.

Instead, we can leverage [suppressible-error-prone](https://github.com/palantir/suppressible-error-prone) by having a dedicated error-prone check that can flag when a deprecated part of the API is in use, and which can also be automatically suppressed by bots (leveraging the `for-rollout:` prefix).

This is beneficial because it
- Allows continued library upgrades, up until the actual break
- Provides clear signal to developers that they are relying on deprecated APIs (rather than having a blocked excavator PR among so many others)
- Prevents backsliding by making it clear to all developers that they are relying on a deprecated api, even if they haven't explicitly enabled `-Xlint:deprecation`

## After this PR
==COMMIT_MSG==
[Error-prone] Add check to prevent usage of deprecated methods and fields
==COMMIT_MSG==

## Possible downsides?
This might cause a lot of repos to get a ton of for-rollout suppressions for existing deprecated API usage. This is probably a good thing, unless the API is technically deprecated but unlikely to ever be removed. I'm unclear how common this pattern is, nor how we could go about checking this. Presumably, if this is a problem, we can just remove this check and clean up the auto-suppressions.

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ConjureEndpointDeprecatedForRemoval.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ConjureEndpointDeprecatedForRemoval.java
@@ -30,6 +30,8 @@ import com.sun.source.tree.Tree;
 
 @AutoService(BugChecker.class)
 @BugPattern(
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
         severity = BugPattern.SeverityLevel.ERROR,
         summary = "You should not use Conjure endpoints that are marked for removal as that may block"
                 + " upgrades in the near future. You can explicitly disable this check on a case-by-case basis using"
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DeprecatedApiUsage.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DeprecatedApiUsage.java
@@ -37,11 +37,14 @@ import com.sun.tools.javac.code.Symbol;
  * Instead, this check is meant to be auto-suppressed upon upgrades and prevent backsliding by unintentionally relying
  *   upon deprecated APIs.
  */
+@SuppressWarnings("BugPatternNaming")
 @AutoService(BugChecker.class)
 @BugPattern(
         severity = BugPattern.SeverityLevel.ERROR,
         summary = "Deprecated APIs should not be relied upon as they may be removed in a future release.",
-        altNames = {"deprecation", "removal"})
+        // Use deprecation as the main name for the check, for familiarity with the javac flag.
+        name = "deprecation",
+        altNames = {"DeprecatedApiUsage", "removal"})
 public final class DeprecatedApiUsage extends BugChecker
         implements BugChecker.MethodInvocationTreeMatcher,
                 BugChecker.MemberReferenceTreeMatcher,
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DeprecatedApiUsage.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DeprecatedApiUsage.java
@@ -20,13 +20,8 @@ import com.google.auto.service.AutoService;
 import com.google.errorprone.BugPattern;
 import com.google.errorprone.VisitorState;
 import com.google.errorprone.bugpatterns.BugChecker;
-import com.google.errorprone.matchers.Description;
 import com.google.errorprone.matchers.Matcher;
 import com.google.errorprone.matchers.Matchers;
-import com.google.errorprone.util.ASTHelpers;
-import com.sun.source.tree.MemberReferenceTree;
-import com.sun.source.tree.MemberSelectTree;
-import com.sun.source.tree.MethodInvocationTree;
 import com.sun.source.tree.Tree;
 import java.util.Optional;
 
@@ -48,10 +43,7 @@ import java.util.Optional;
         // Use deprecation as the main name for the check, for familiarity with the javac flag.
         name = "deprecation",
         altNames = "DeprecatedApiUsage")
-public final class DeprecatedApiUsage extends BugChecker
-        implements BugChecker.MethodInvocationTreeMatcher,
-                BugChecker.MemberReferenceTreeMatcher,
-                BugChecker.MemberSelectTreeMatcher {
+public final class DeprecatedApiUsage extends AbstractDeprecatedApiCheck {
 
     private static final String MESSAGE_DETAILS =
             " - this may be removed in a future release and prevent library upgrades. Note: This error comes from "
@@ -62,32 +54,15 @@ public final class DeprecatedApiUsage extends BugChecker
             Matchers.symbolMatcher((symbol, state) -> symbol.isDeprecated() && !symbol.isDeprecatedForRemoval());
 
     @Override
-    public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
-        return checkTree(tree, state);
+    protected boolean isDeprecationWarning(Tree tree, VisitorState state) {
+        return DEPRECATED_SYMBOL.matches(tree, state);
     }
 
     @Override
-    public Description matchMemberReference(MemberReferenceTree tree, VisitorState state) {
-        return checkTree(tree, state);
-    }
-
-    @Override
-    public Description matchMemberSelect(MemberSelectTree tree, VisitorState state) {
-        return checkTree(tree, state);
-    }
-
-    private Description checkTree(Tree tree, VisitorState state) {
-        if (!DEPRECATED_SYMBOL.matches(tree, state)) {
-            return Description.NO_MATCH;
-        }
-
-        Optional<String> qualifiedName = Optional.ofNullable(ASTHelpers.getSymbol(tree))
-                .map(s ->
-                        s.owner.getQualifiedName() + "#" + s.getQualifiedName().toString());
-        String description = qualifiedName
-                        .map(name -> String.format("%s is deprecated", name))
-                        .orElse("Deprecated API usage")
+    protected String getErrorDescription(Optional<String> qualifiedName) {
+        return qualifiedName
+                .map(name -> String.format("%s is deprecated", name))
+                .orElse("Deprecated API usage")
                 + MESSAGE_DETAILS;
-        return buildDescription(tree).setMessage(description).build();
     }
 }
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DeprecatedApiUsage.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DeprecatedApiUsage.java
@@ -38,6 +38,8 @@ import java.util.Optional;
 @SuppressWarnings("BugPatternNaming")
 @AutoService(BugChecker.class)
 @BugPattern(
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
         severity = BugPattern.SeverityLevel.ERROR,
         summary = "Deprecated APIs should not be relied upon as they may be removed in a future release.",
         // Use deprecation as the main name for the check, for familiarity with the javac flag.
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DeprecatedApiUsage.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DeprecatedApiUsage.java
@@ -42,7 +42,7 @@ import com.sun.tools.javac.code.Symbol;
         severity = BugPattern.SeverityLevel.ERROR,
         summary = "Deprecated APIs should not be relied upon as they may be removed in a future release.",
         altNames = {"deprecation", "removal"})
-public final class DeprecatedUsage extends BugChecker
+public final class DeprecatedApiUsage extends BugChecker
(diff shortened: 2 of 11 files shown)


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DeprecatedUsage.java
```diff
@@ -0,0 +1,85 @@
...
+ *   automatically suppressed by suppressible-error-prone and therefore block library upgrades as soon as a deprecation
+ *   is introduced that is being relied upon.
+ * Instead, this check is meant to be auto-suppressed upon upgrades and prevent backsliding by unintentionally relying
+ *   upon deprecated APIs.
+ */
+@AutoService(BugChecker.class)
+@BugPattern(
+        severity = BugPattern.SeverityLevel.ERROR,
+        summary = "Deprecated APIs should not be relied upon as they may be removed in a future release.",
+        altNames = {"deprecation", "removal"})
```
CRogers: This is nifty - means that `@SuppressWarnings("deprecation")` works.

CRogers: I wonder if we can make the canonical name for this check be `deprecation`??? That way it will suppress as `@SuppressWarnings("deprecation")` too rather than having both.

CRogers: I guess it would still end up as `for-rollout:deprecation` rather than `deprecation` (so won't stop the `-Xlint:deprecation` from firing if still there).

aldexis: I'm not sure what you mean by
>  it will suppress as @SuppressWarnings("deprecation") too rather than having both

Did you mean the auto-suppression? (in which case, it would use `"for-rollout:deprecation"` anyway, so I'm not sure if gives us much)

It would however be interesting to have `deprecation` as the default in that it is backwards-compatible with the compiler warning (if e.g. we want to go back to the compiler flags)

aldexis: How is https://github.com/palantir/gradle-baseline/pull/3244/commits/4e2dae481e292bad39f590244d10ac81067dfe19?

aldexis: > I guess it would still end up as for-rollout:deprecation rather than deprecation (so won't stop the -Xlint:deprecation from firing if still there).

That's true, but also devs would see `[deprecation]` in the error-prone failure, and thus would manually suppress with `"deprecation"` rather than `"DeprecatedUsage"`

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/DeprecatedApiUsageTest.java
```diff
@@ -0,0 +1,254 @@
...
+
+package com.palantir.baseline.errorprone;
+
+import com.google.errorprone.CompilationTestHelper;
+import org.junit.jupiter.api.Test;
+
+public class DeprecatedApiUsageTest {
+
+    @Test
+    public void testNotDeprecatedMethodCall() {
```
CRogers: Let's maintain our style of not prefixing with `test` and writing descriptive test names as snake_case_english_sentences

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DeprecatedApiUsage.java
```diff
@@ -0,0 +1,88 @@
...
+            return Description.NO_MATCH;
+        }
+
+        Symbol symbol = ASTHelpers.getSymbol(tree);
+        String qualifiedName = symbol == null
+                ? null
+                : symbol.owner.getQualifiedName() + "#"
+                        + symbol.getQualifiedName().toString();
+        String description = qualifiedName == null
+                ? "Deprecated API usage - this is dangerous as it may be removed in a future release."
```
CRogers: `dangerous` feels a bit strong

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DeprecatedApiUsage.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    private Description checkTree(Tree tree, VisitorState state) {
+        if (!DEPRECATED_SYMBOL.matches(tree, state)) {
+            return Description.NO_MATCH;
+        }
+
+        Symbol symbol = ASTHelpers.getSymbol(tree);
+        String qualifiedName = symbol == null
+                ? null
```
CRogers: Let's use `Optional`s instead of raw nulls here?

CRogers: `Optional.ofNullable(ASTHelpers.getSymbol(tree)).map(...).orElse("...")`

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DeprecatedApiUsage.java
aldexis: How do you feel about `this may be removed in a future release and prevent library upgrades`? Technically, it _is_ somewhat dangerous, since it might result in runtime P0s due to ABI breaks, but the warning is maybe a couple of steps too early to be saying it's `dangerous`

CRogers: That seems better

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/DeprecatedApiUsageTest.java
aldexis: We're not consistent everywhere in this repo, but I'll do it if that's our preferred stance

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/DeprecatedApiUsageTest.java
```diff
@@ -0,0 +1,313 @@
...
+public class DeprecatedApiUsageTest {
+
+    @Test
+    public void does_not_throw_on_non_deprecated_api_usage() {
+        helper().addSourceLines(
+                        "Test.java",
+                        // language=Java
+                        """
+                        import com.palantir.logsafe.*;
+                        import java.util.function.*;
```
CRogers: Can we avoid star imports? It makes it hard to see where the imports are from. Also, I don't think any `logsafe` or `java.util.function` or `java.util` types are used here?

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/DeprecatedApiUsageTest.java
```diff
@@ -0,0 +1,313 @@
...
+                        import java.util.*;
+
+                        class Helper {
+                          public static final String CONSTANT = "not deprecated";
```
CRogers, line 31: Why two indents out of curiosity? Surely this is harder than just doing a normal 4-space indent?

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/DeprecatedApiUsageTest.java
aldexis: Good point - I copy-pasted from another test, but this is not useful here. Fwiw though, I don't mind the star imports in some cases, because they can make the tests more readable as compared to multiple lines of import from the same package (they generally aren't particularly relevant to the test itself, so I'd rather reduce the number of extraneous lines)

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/DeprecatedApiUsageTest.java
aldexis: Error-prone [uses its own formatter](https://github.com/google/error-prone/blob/da03fea47dd31b1a63932d78f3d91fb383cce8d1/core/src/main/java/com/google/errorprone/bugpatterns/MisformattedTestData.java#L71-L79), and otherwise fails with `MisformattedTestData` 😢

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineErrorProne.java
```diff
@@ -63,6 +64,11 @@ private static void applyToJavaProject(Project project) {
                         new ConditionalPatchCheck(
                                 new IfModuleIsUsed("com.palantir.safe-logging", "logger"), "PreferSafeLogger"));
 
+        project.getTasks().withType(JavaCompile.class).configureEach(javaCompile -> {
+            // This is superseded by the DeprecatedForRemovalApiUsage error-prone check
+            javaCompile.getOptions().getCompilerArgs().add("-Xlint:-removal");
```
aldexis: I'm wondering whether this could also filter out `-Xlint:deprecation` here, or whether this would race with it being enabled

CRogers: Haven't tried it out, but I suspect it'll be whichever is last, and since we're in a plugin application we're almost always going to be first.

aldexis: Sounds good - I won't do this then, and will stick to an excavator to remove the flag instead

CRogers: I think this make sense anyway. We're essentially changing the default like this.

CRogers: You should use a `CommandLineArgumentProvider` here and not just addd to the `getCompilerArgs()` - the latter will force the evaluation of the existing command line args, which could break things.

CRogers: Ok, it doesn't force the value - that's `getAllCompilerArgs()` - but this method is marked `@ToBeReplacedByLazyProperty` so will change in the future, better to just always use the provider version.

aldexis: Ah my bad, I hadn't realized this was something different. This makes sense and I've updated the code

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/DeprecatedApiUsage.java
```diff
@@ -0,0 +1,68 @@
...
+ *   is introduced that is being relied upon.
+ * Instead, this check is meant to be auto-suppressed upon upgrades and prevent backsliding by unintentionally relying
+ *   upon deprecated APIs.
+ *
+ * Note that it explicitly does not handle APIs that are deprecated for removal, as those should be handled by
+ *   {@link DeprecatedForRemovalApiUsage}.
+ */
+@SuppressWarnings("BugPatternNaming")
+@AutoService(BugChecker.class)
+@BugPattern(
```
CRogers, line 40: Need to set the url here and in removal otherwise we get this:

> (see https://errorprone.info/bugpattern/deprecation)

aldexis: Good catch - I've added it and also added the same to `ConjureEndpointDeprecatedForRemoval` which was missing it

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ConjureEndpointDeprecatedForRemoval.java
```diff
@@ -30,6 +30,8 @@
 
 @AutoService(BugChecker.class)
 @BugPattern(
+        link = "https://github.com/palantir/gradle-baseline#baseline-error-prone-checks",
+        linkType = BugPattern.LinkType.CUSTOM,
```
aldexis, line 34: Drive-by fix

CRogers (approved):

baseline-error-prone/src/test/java/com/palantir/baseline/errorprone/DeprecatedApiUsageTest.java
Stephan202: Just a drive-by comment: if you'd like to clean up the repo by dropping those `test` prefixes, while not accidentally introducing a reserved keyword, see [this check](https://github.com/PicnicSupermarket/error-prone-support/blob/c23c3b7c1b0dba82f2f0f392589000f1fe09f25a/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/JUnitMethodDeclaration.java).

PR #3245, opened 2025-09-05
Add ImmutablesDefaultValue error-prone check

ghost:
This check enforces that default methods in immutable interfaces are
annotated with `@Value.Default` to ensure they are included as fields in
the generated immutable classes.

- Detects default methods in @Value.Immutable interfaces
- Handles `@Value.Style(defaultAsDefault = true)` configurations
- Supports meta-annotations with `@Value.Style`
- Provides automatic fixes to add missing `@Value.Default` annotations
- Can be suppressed when methods are auxiliary and not meant as fields


pkoenig10 (changes requested):
There are legitimate reason to want a default method that is not `@Value.Default`. It may be a cheap computation for which we do not want to create a field that eagerly computes the value.

PR #3253, opened 2025-09-12
[Error-prone] [Deprecation check] Ignore deprecations stemming from the current repository

aldexis:
## Before this PR
The check added in https://github.com/palantir/gradle-baseline/pull/3244 already had a built-in check to not flag deprecations from within the same class.

However, there are a number of cases and examples where the deprecation comes from different classes in the same project, or a different project in the same repo.

In both of these cases, flagging the deprecation doesn't truly matter, because a removal would have to be addressed immediately to fix compilation.

## After this PR

==COMMIT_MSG==
[Error-prone] [Deprecation check] Ignore deprecations stemming from the current repository
==COMMIT_MSG==

## Possible downsides?
Where I'm slightly debating is for deprecations from a different project in the same repository. While I've made this work for a majority of cases, this means developers won't realize they have a pending ABI break until they actually break it and fix it. This _could_ lead to consequences where they publish the ABI break and its fix in the same version of two distinct libraries, and where we'd end up with these two libraries with different versions at runtime, leading to a break.


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/AbstractDeprecatedApiCheck.java
```diff
@@ -73,11 +79,27 @@ private Description checkTree(Tree tree, VisitorState state) {
...
+        // This should however be good enough for well-behaved repositories.
+        Optional<URI> classFileUri = owningClass.map(c -> c.classfile).map(FileObject::toUri);
+        if (classFileUri.isPresent() && isRegularFileOnSystem(classFileUri.get())) {
+            if (classFileUri.get().getPath().contains("/classes")) {
+                // If the class file is a regular file on the local file system, and is within a /classes/ directory,
+                //   this means we're calling a deprecated API within the same repository, even though maybe not the
+                //   same project. We don't need to flag these usages, as breaks would be caught at compile time anyway.
                 return Description.NO_MATCH;
             }
         }
```
aldexis, line 109: See note in the PR description. I'm not entirely convinced this is desirable, because we'd have two libraries publishing here, which can be a vector for runtime ABI breaks. Specifically, if we hide the fact that the method is deprecated in the other project, we end up failing to fix it up until the actual break, which increases the likelihood that we'd have two different versions at runtime.

Concretely:
- Repo has `lib` and `transitive`
- `transitive` has a deprecated method, which `lib` uses
- `transitive` removes said method, which requires fixing it in `lib`, and both publish as `1.9.0`
- Downstream repo relies on `lib` and `transitive` `1.8.0`, but also pull in `transitive` in some other way, which bumps it to `1.9.0`, while `lib` stays on `1.8.0`

This is _not_ a problem if the `lib` and `transitive` use the same maven group, and the downstream repo is configured with e.g. `com.group:* = 1.8.0` as I believe this would force `lib` and `transitive` to be on the same version (e.g. `1.9.0` if `transitive` gets bumped transitively), but neither same group for same repo, nor platform dependencies are always true

aldexis: I also realized that this might happen if you depend _transitively_ on both `lib` and `transitive` through various dependencies. In that case, you wouldn't have a pin in your versions.props, and thus wouldn't have a platform constraint (unsure if that's the term) forcing both to be on the same version.

It's all more of an edge case, but I'd be inclined to remove this specific piece until we're proven that it is _still_ too noisy

CRogers: This can be a problem with any kind of cross project method usage, including non-public ones. We should (and are currently looking at) just ensuring the all the versions aligned for every library rather than worrying about it here tbh.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/AbstractDeprecatedApiCheck.java
```diff
@@ -88,9 +110,39 @@ private Description checkTree(Tree tree, VisitorState state) {
...
     }
 
-    private Optional<Name> getCurrentClass(VisitorState state) {
-        return Optional.ofNullable(ASTHelpers.findEnclosingNode(state.getPath(), ClassTree.class))
-                .map(ASTHelpers::getSymbol)
-                .map(Symbol::getQualifiedName);
+    private Optional<ClassSymbol> getOwningClass(Optional<Symbol> symbol) {
+        if (symbol.isEmpty()) {
+            return Optional.empty();
+        }
```
CRogers: I would just take a `Symbol` here and use `.flatMap` in the place that has an `Optional<Symbol>`. Just feels weird to take an `Optional` when empty is not a useful parameter to this method.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/AbstractDeprecatedApiCheck.java
```diff
@@ -88,9 +110,39 @@ private Description checkTree(Tree tree, VisitorState state) {
...
+            FileSystem fileSystem = path.getFileSystem();
+            if (!fileSystem.equals(FileSystems.getDefault())) {
+                return false;
+            }
+
+            // Check if it exists and is a regular file
+            return Files.exists(path) && Files.isRegularFile(path);
+        } catch (Exception e) {
+            return false;
+        }
```
CRogers, line 153: Is there a particular exception we think we'll run into here? If not, can we at least log an error here so we have a change of becoming aware of it?

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/AbstractDeprecatedApiCheck.java
```diff
@@ -73,11 +79,27 @@ private Description checkTree(Tree tree, VisitorState state) {
...
+            return Description.NO_MATCH;
+        }
+
+        // Note: the logic below will NOT work if the target dependency applies the "java" plugin rather than
+        //   the "java-library" plugin, because in that case the classfile that we get here will be within the
+        //   jar file itself, which makes it particularly tricky to distinguish from just regular jar dependencies.
+        // This should however be good enough for well-behaved repositories.
+        Optional<URI> classFileUri = owningClass.map(c -> c.classfile).map(FileObject::toUri);
+        if (classFileUri.isPresent() && isRegularFileOnSystem(classFileUri.get())) {
+            if (classFileUri.get().getPath().contains("/classes")) {
```
CRogers: Does this if need to be doubly nested? Just makes understanding the control flow a bit harder to understand I think.

gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineErrorProneIntegrationTest.groovy
```diff
@@ -127,6 +188,46 @@ class BaselineErrorProneIntegrationTest extends AbstractPluginTest {
...
+            }
+        '''.stripIndent())
+
+        file('lib/src/main/java/test/DeprecatedClass.java') << '''
+        package test;
+        public class DeprecatedClass {
+            @Deprecated(forRemoval = true)
+            static void deprecated() {}
+        }
+        '''.stripIndent()
```
CRogers: Everywhere in the tests, `.stripIndent(true)` so you can indent the text block to make it more readable.

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/AbstractDeprecatedApiCheck.java
aldexis: Alright, I'm convinced to ignore this across the entire repo then

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/AbstractDeprecatedApiCheck.java
aldexis: There could be some issues in `Paths.get(uri);` which I'm wary about
```
IllegalArgumentException – if preconditions on the uri parameter do not hold. The format of the URI is provider specific.
FileSystemNotFoundException – The file system, identified by the URI, does not exist and cannot be created automatically, or the provider identified by the URI's scheme component is not installed
SecurityException – if a security manager is installed and it denies an unspecified permission to access the file system
```
I'm also not finding any example of logging from within an error prone check in gradle-baseline, so I'm wondering whether we can/should do it 🤔

We _could_ just rethrow the exception, but I'm slightly concerned that this would fail upon rollout? (I guess this might be better than not knowing at all?)

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/AbstractDeprecatedApiCheck.java
aldexis: It does not - I've updated

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/AbstractDeprecatedApiCheck.java
CRogers: When inside the compiler you just use java.util.logging. It's basically the same as slf4j but you just a different `Logger`type/different levels.

gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineErrorProneIntegrationTest.groovy
```diff
@@ -120,13 +181,53 @@ class BaselineErrorProneIntegrationTest extends AbstractPluginTest {
...
+            }
+        '''.stripIndent(true)
+
+        def standardBuildFileForLibrary = standardBuildFile.replace("'java'", "'java-library'")
+        multiProject.addSubproject("lib", standardBuildFileForLibrary)
+        multiProject.addSubproject("app", standardBuildFile + '''
+            dependencies {
+                implementation project(':lib')
+            }
+        '''.stripIndent(true))
```
CRogers, line 206: The benefit of doing `.stripIndent(true)` is you can indent the text block so that it's a bit easier to read.

CRogers (commented):
LGTM except those comments

gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineErrorProneIntegrationTest.groovy
aldexis: Ah fair point. I'm updating

baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/AbstractDeprecatedApiCheck.java
aldexis: Thanks for the information! Updated!

CRogers (approved):

PR #3270, opened 2025-09-22
Ignore static variables in record array field check

shamith09:
## Before this PR
This check doesn't ignore static fields, so it emits a warning even in cases like these:
```
public record SomeRecord(String foo) {
    public static final String[] SOME_CONSTANT = new String[] { "foo", "bar" };
}
```
Static fields are not compared in `.equals`, so this code is safe and does not exhibit any weird behavior unless there is _also_ a non-static array field in the record. In this example, we should expect instances of `SomeRecord` to be equal if their `foo` fields are equal.

## After this PR
Static fields are ignored in this check. The presence of non-static array fields still triggers the warning.

==COMMIT_MSG==
==COMMIT_MSG==


kelvinou01:
### DO NOT MERGE (YET)

kelvinou01:

PR #3274, opened 2025-09-24
[Error-prone] [DeprecatedApiUsage] Flag on deprecated class usage

aldexis:
## Before this PR
If a class itself is deprecated and used, e.g. through a constructor, or through variable names, it is not currently flagged by the `DeprecatedApiUsage` check added in https://github.com/palantir/gradle-baseline/pull/3244

## After this PR
==COMMIT_MSG==
[Error-prone] [DeprecatedApiUsage] Flag on deprecated class usage
==COMMIT_MSG==

## Possible downsides?
This adds more cases where we're going to surface errors, so it could add more suppressions, and/or fail in the case of things like https://github.com/palantir/suppressible-error-prone/pull/160, so we should probably get https://github.com/palantir/gradle-baseline/pull/3253 and https://github.com/palantir/suppressible-error-prone/pull/160 merged first


baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/AbstractDeprecatedApiCheck.java
```diff
@@ -55,17 +57,22 @@ public final Description matchMemberReference(MemberReferenceTree tree, VisitorS
...
     @Override
     public final Description matchMemberSelect(MemberSelectTree tree, VisitorState state) {
+        return checkTree(tree, state);
+    }
+
+    private Description checkTree(Tree tree, VisitorState state) {
         if (isImportStatement(state)) {
             // We don't want to flag import statements, as those cannot be suppressed.
             return Description.NO_MATCH;
         }
```
aldexis, line 85: Moved the import statement down one level

CRogers (dismissed):

CRogers (approved):

PR #3276, opened 2025-09-25
Do not flag textBlock lineLength

crogoz:
## After this PR

In preparation for https://github.com/palantir/palantir-java-format/pull/1419. Makes sure that textBlock lineLength are not flagged as checkstyleErrors. Applied the same [regex as in the checkstyle repo.](https://github.com/mohitsatr/checkstyle/blob/7ddea2b02b219dcba8350ecfeb115b3c0d2c51aa/src/main/resources/google_checks.xml#L62-L66)

==COMMIT_MSG==
Do not flag textBlock lineLength
==COMMIT_MSG==

### DIFF
--- a/.baseline/checkstyle/checkstyle.xml
+++ b/.baseline/checkstyle/checkstyle.xml
@@ -45,6 +45,13 @@
         <property name="max" value="120"/>
         <property name="ignorePattern" value="^package.*|^import.*|a href|href|http://|https://|ftp://|\{@link"/>
     </module>
+
+    <!--  Suppression to prevent LineLength Check from flagging lines in Text-blocks -->
+    <module name="SuppressWithPlainTextCommentFilter">
+        <property name="checkFormat" value="LineLength"/>
+        <property name="offCommentFormat" value='^.*"""\s*$'/>
+        <property name="onCommentFormat" value='^\s*"""\s*(?:[,;]|.+)$'/>
+    </module>
     <module name="TreeWalker">
         <module name="SuppressionCommentFilter"/> <!-- baseline-gradle: README.md -->
         <module name="SuppressionCommentFilter">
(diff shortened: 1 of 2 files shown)


.baseline/checkstyle/checkstyle.xml
```diff
@@ -45,6 +45,13 @@
         <property name="max" value="120"/>
         <property name="ignorePattern" value="^package.*|^import.*|a href|href|http://|https://|ftp://|\{@link"/>
     </module>
+
+    <!--  Suppression to prevent LineLength Check from flagging lines in Text-blocks -->
```
esword: I'd describe why turning this off here - or link to why

felixdesouza:
baseline is just going to delete this when it comes by with the excavator, I think you need to add it to `custom-suppressions.xml`

and just to double check, this is for *this* repo right? not something that's going to be applied as part of baseline in other repos?

I imagine this won't be the case only if baseline-checkstyle reads from its own checkstyle, but I would have thought that a bit odd

felixdesouza:
if you *do* want this to be applied to all repos, then you probably need to put it in [`gradle-baseline-java-config`](https://github.com/palantir/gradle-baseline/blob/develop/gradle-baseline-java-config/resources/checkstyle/checkstyle.xml)

crogoz:
> if you do want this to be applied to all repos, then you probably need to put it in [gradle-baseline-java-config](https://github.com/palantir/gradle-baseline/blob/develop/gradle-baseline-java-config/resources/checkstyle/checkstyle.xml)

oh, thanks for noticing! yes, this was the target, addressed in https://github.com/palantir/gradle-baseline/pull/3276/commits/055356ba0fd6714756911d221e6e2e5fcf357086

felixdesouza:
👍

CRogers (approved):

PR #3279, opened 2025-09-29
[Error-prone] Add InlineMeInliner to default patch checks

aldexis:
## Before this PR
When libraries specify `@InlineMe` annotations like in https://github.com/palantir/deadlines-java/pull/188/files#diff-f8f90c9951b97fc3abe992cb927e3601222e7e3ce25efb471821039feab57432R179-R181, the current behavior would just suppress it ([example](https://github.com/palantir/dialogue/pull/2704)), when we could instead apply the proposed fix, which is more desirable.

## After this PR

==COMMIT_MSG==
[Error-prone] Add InlineMeInliner to default patch checks
==COMMIT_MSG==

## Possible downsides?
This means that external libraries could in theory change your code through the use of `InlineMe` annotations, but given they could also just change their own code, this doesn't change too much.


CRogers:
The only other issue could be external usages of `@InlineMe` not compiling when inlining, without us easily being a being able to fix. We have the same problem with Google's built in errorprones though, and can always revert this if it causes issues.

CRogers:
:+1:

PR #3286, opened 2025-10-02
[Error-prone] [Deprecation] Ignore deprecated classes coming from the same repo

aldexis:
## Before this PR
https://github.com/palantir/gradle-baseline/pull/3253 introduced the ability to ignore deprecations declared in the same repository.

https://github.com/palantir/gradle-baseline/pull/3274 also made us throw when using a deprecated class.

However, the first one missed the fact that the deprecated symbol might be a class symbol itself, and thus not have an owner (when not a nested class), which made us miss the information that it came from the same repo.

## After this PR

==COMMIT_MSG==
[Error-prone] [Deprecation] Ignore deprecated classes coming from the same repo
==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/AbstractDeprecatedApiCheck.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/AbstractDeprecatedApiCheck.java
@@ -127,7 +127,8 @@ public abstract class AbstractDeprecatedApiCheck extends BugChecker
 
     @Nullable
     private ClassSymbol getOwningClass(Symbol symbol) {
-        Symbol owner = symbol.owner;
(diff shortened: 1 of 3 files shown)


kelvinou01 (dismissed):
The fix looks good! 

Right now we're flagging deprecations aggressively "If we aren't be sure this is coming from the same repo, flag it". Perhaps this is good because it surfaces bugs like the one this PR is fixing. 

But maybe we want to be more defensive "flag this only if we're sure it comes from another repo"? To make sure we don't break people. Deferring to you on this as I don't have much context on our java infra "landscape".

aldexis:
This won't _break_ people. It will just surface more deprecations (which they should probably be taking care of anyway). It's at most annoying, but shouldn't break anybody, so I think that's the right trade-off

kelvinou01 (approved):

PR #3319, opened 2025-10-29
Use new testing framework for baseline-module-jvm-args plugin

CRogers:
## Before this PR
I'm going to add to this tests, so figured I might as well feed them through the AI test conversion manually.

## After this PR
New testing framework from http://github.com/palantir/gradle-plugin-testing/ is used. I also manually tidied up the tests a bit.

The AI did a reasonable job - although it did forget or didn't do several things I had to correct:
1. Used `apply plugin:` instead of plugins blocks
2. Stripped out all comments
3. Bad conversion of task outcome assertions
4. nit: Should be newline after a text block method for clarity


gradle-baseline-java/src/test/java/com/palantir/baseline/BaselineModuleJvmArgsIntegrationTest.java
```diff
@@ -0,0 +1,536 @@
...
+
+        rootProject
+                .buildGradle()
+                .edit(content ->
+                        content.replace("java.management/sun.management", "java.management/sun.management123"));
+
+        InvocationResult resultAfterChange = gradle.withArgs("jar").buildsSuccessfully();
+
+        assertThat(resultBeforeChange.task(":jar")).isPresent();
+        assertThat(resultAfterChange.task(":jar")).isPresent();
```
CRogers: I guess these should check they executed not just run.

gradle-baseline-java/src/test/java/com/palantir/baseline/BaselineModuleJvmArgsIntegrationTest.java
```diff
@@ -0,0 +1,550 @@
...
+                .filter(file -> file.getName().endsWith(".jar"))
+                .map(file -> {
+                    try {
+                        return new JarFile(file);
+                    } catch (IOException e) {
+                        throw new UncheckedIOException(e);
+                    }
+                })
+                .findFirst()
+                .orElseThrow();
```
crogoz, line 312: nit: we could create a helper method that finds the jar - since this is used in other tests as well

gradle-baseline-java/src/test/java/com/palantir/baseline/BaselineModuleJvmArgsIntegrationTest.java
```diff
@@ -0,0 +1,550 @@
...
+
+        InvocationResult result = gradle.withArgs("run").buildsSuccessfully();
+        assertThat(result.output()).doesNotContain("--add-exports");
+    }
+
+    @Test
+    void does_not_add_externally_defined_exports_to_the_jar_manifest(GradleInvoker gradle, RootProject rootProject)
+            throws IOException {
+
+        createJarWithExport(rootProject, "test.jar");
```
crogoz, line 375: nit: since we always have "test.jar" as the name of the empty jar we create, we could remove the name parameter and have this defined as a static variable that can get passed to the build.gradle as well.

crogoz (approved):

PR #3321, opened 2025-10-30
baseline-module-jvm-args: Extract code into methods & move tests into nested classes to group them

CRogers:
## Before this PR
I'm currently working on getting baseline-module-jvm-args to work with compiling using a higher version of a compiler and running with `--release`. I need to make quite a few changes to the code, but unfortunately there's so much nesting it's making it hard to write readable code. Additionally, there are so many tests that are not grouped together, which is confusing as junit does not run them in their source declared order.

## After this PR
I've extracted the code to methods. I haven't actually changed the code other than pressing cmd+alt+m to extract method in intellij.

I've grouped some of the tests together in using `@Nested`. The tests have not be changed in anyway.


FinlayRJW:
👍

PR #3324, opened 2025-10-31
baseline-module-jvm-args: Fix bug where compiler plugin `--add-exports` are added to `compilerArgs` instead of `forkArgs`; allow `--release` and `--add-exports` compiler args to be used together

CRogers:
## Before this PR
[Error Prone 2.43.0](https://github.com/google/error-prone/releases/tag/v2.43.0) has [made JDK 21 the minimum required JDK that Error Prone is run in](https://github.com/google/error-prone/issues/4867). The recommendation is to use `--release` so a higher version JDK can safely output java classes at a lower language level.

This is problematic for us:
* Currently, if we want to compile Java 17 code, we use a 17 JDK without the `--release` arg. We do not use a higher version (eg 21) JDK using `--release` as it is not compatible with `--add-exports` for system modules (ie modules shipped with the JDK). A decent amount of our code requires compiling with `--add-exports` (especially compiler plugins, more stuff internally).
* We have not yet moved to Java 21 for published libraries, because a couple of the (ever increasing) ecosystems where we use Java as part of the product we sell do not support 21 runtime yet. So we're stuck needing to compile our published libraries with Java 17
* Under this current setup, we cannot upgrade Error Prone. Unfortunately, open source libraries are updating their version of `error_prone_annotations` which is forcing our Error Prone to upgrade as we [force the Error Prone jar versions to be the same](https://github.com/palantir/suppressible-error-prone/pull/97). These OSS library upgrades are being blocked. Unfortunately, Error Prone regularly has breaking changes, even in the annotation jars, so skewing is not going to be safe for long. We also need to upgrade Error Prone to take bug fixes.

There is a bug in the existing implementation of `baseline-module-jvm-args`: it will currently put the `--add-exports` that are meant for the compiler process (ie used for running compiler plugins like Error Prone) in the `compilerArgs` that are meant for the code _being compiled_. This immediately causes every use of `--release` to break, as there is an `--add-exports` next to it.

Unfortunately, even with that bugfix, there is still a problem with using `--release`. We may need to have `--add-exports` compiler args at the same time as using `--release` (eg for jar containing custom error prone BugCheckers), which causes the javac error:
```
error: exporting a package from system module jdk.compiler is not allowed with --release
```

## After this PR

==COMMIT_MSG==
baseline-module-jvm-args: Fix bug where compiler plugin `--add-exports` are added to `compilerArgs` instead of `forkArgs`; allow `--release` and `--add-exports` compiler args to be used together
==COMMIT_MSG==

This is only a prerequisite for using the `--release` arg and a higher JDK, which will in a later PR.

### Fixed the `compilerArgs`/`forkOptions` bug

I've fixed the bug so that we put the `--add-exports` compiler plugin in the `forkOptions` rather than the `compilerOptions`.

### Compiler plugin

I've made a compiler plugin that works around the `--add-exports` + `--release` not used together built in check using reflection. It's a very simple plugin (change one single field).

Now we should be able to compile code with `--add-exports` + `--release` and hopefully never need to worry about this issue again (at least, unless our hack in the plugin breaks).

### **Always** fork the compiler

Previously, we had fork defaulted to false, so there was this horrible inconsistency and bug:

* If the requested java compiler was the _same_ major version as the JDK running the Gradle daemon: it would run the compilation tasks inside the main Gradle daemon, likely without the right `--add-exports` args, even if configured in `forkOptions`.
* If the requested java compiler was a _different_ major version as the JDK running the Gradle daemon: it would spin up a new worker daemon and actually apply the correct JVM args listed in `forkOptions`

Now we always fork the compiler so changing the compiler or daemon JDK do not have the behaviour change

## Possible downsides?
* Fixing the bug around `forkArgs` vs `compilerArgs` may break some repos where projects had not specified the exports they need, because there were `--add-exports` added spuriously into the compiler args. For example: https://github.com/palantir/suppressible-error-prone/pull/225.
  * We will just need to fix these as we go. Will mostly affect error prone check sources, mostly owned by the infra team.
* The compiler plugin is 100% crimes. Changing the compiler internals to get around (an even pretty silly) restriction may add maintenance if the compiler ever changes. It also adds to the WTF/confusing factor as our tooling behaves differently to "normal" java tooling.
  * Fortunately, the code we need to reflectively change has not changed since it was introduced 8 years ago. It should hopefully not change again.
  * I've given the compiler plugin a name that should instil fear and make it very obvious what it's doing if anyone looks at the compiler args

### Just used `--source` and `--target` and not use `--release`

A refresher:
* `--source`: Sets the source language level, eg can you use `sealed interface` or other source features?
* `--target`: Sets the level of the outputted bytecode in classfiles, eg can the compiler output some JVM instruction? Also sets the class files' major version.
* `--release N`: Sets `--source N` and `--target N` **and** the bootstrap classpath that is used, ie use major version N standard library. If you set `--release 17`, it will stop you using any standard library methods that are introduced after JDK 17.

So we could use a higher version JDK and just set `--source` and `--target` and not `--release`, and not face the issue with `--add-exports` used at the same time. However, this would allow devs to use a higher versioned standard library at compile time, which would then explode at runtime. This is not acceptable in my eyes. As an infra team, I'd rather go through a little pain to make our devex better and prevent production issues.

We could have had extra checks to the apis used (eg use animal sniffer), but this is more complexity and infra I'd like to avoid.

### Run Error Prone as a separate task

We could use a different task to run errorprones that can run with a different compiler version. There are benefits and downside to this that we have been thinking about. But this a big change that will delay fixing the blocked upgrades.

### DIFF
--- a/baseline-module-jvm-args-compiler-plugins/src/main/java/com/palantir/baseline/modulejvmargs/compilerplugins/AllowReleaseAndAddExportsToBeUsedTogetherByChangingCompilerInternalsUsingReflection.java
+++ b/baseline-module-jvm-args-compiler-plugins/src/main/java/com/palantir/baseline/modulejvmargs/compilerplugins/AllowReleaseAndAddExportsToBeUsedTogetherByChangingCompilerInternalsUsingReflection.java
@@ -25,9 +25,9 @@ import com.sun.tools.javac.util.Context;
 import java.lang.reflect.Field;
 
 /**
- * The JDK devs arbitrarily decided that you cannot use the `--release` arg along with `--add-exports`
+ * The JDK devs arbitrarily decided that you cannot use the {@code --release} arg along with {@code --add-exports}
  * if it's exporting a "system module" ie one that is shipped with the JDK. This is problematic, as if
- * we want to use `--release` to use a higher JDK to compile against a lower version of the Java
+ * we want to use {@code --release} to use a higher JDK to compile against a lower version of the Java
  * standard library, when someone uses BaselineModuleJvmArgs to add an exports or opens arg to
  * access system modules, they will immediately face this error and be unable to compile. A good example
  * of something that needs extra exports are compiler plugin eg any sort error-prone code, including
@@ -39,7 +39,7 @@ import java.lang.reflect.Field;
  * in Javac's Modules component to be true. We are very fortunate that the error does not actually get
  * created until after Plugin initialisation occurs
  * (<a href="https://github.com/openjdk/jdk/blame/a33aa65fbc70a91fe21e9016c393bb5a764cd75a/src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Modules.java#L1676">here</a>)
- * meaning we can change what we need to change during within a self-contained Plugin.
+ * meaning we can change what we need to change during within a self-contained {@code Plugin}.
  * <br><br>
  * The risk here is that the compiler changes such that we can no longer do this hack, particularly
  * if the timing of the error changes and happens to early before we can prevent it. However, the code
@@ -70,9 +70,16 @@ public final class AllowReleaseAndAddExportsToBeUsedTogetherByChangingCompilerIn
             // We can but pray that it does not change in the future.
             Field allowAccessIntoSystem = Modules.class.getDeclaredField("allowAccessIntoSystem");
             allowAccessIntoSystem.setAccessible(true);
+            // This requires --add-opens on jdk.compiler/com.sun.tools.javac.comp as it is
+            // reflectively changing code in that module
             allowAccessIntoSystem.setBoolean(modules, true);
         } catch (NoSuchFieldException | IllegalAccessException e) {
-            throw new RuntimeException(e);
+            throw new RuntimeException(
+                    "Failed to allow the `--add-exports` and `--release` options to be used together by reflectively"
+                            + " changing the compiler internals. This likely means the compiler"
+                            + " implementation has changed the code that makes this work need to be updated in "
+                            + "https://github.com/palantir/gradle-baseline",
+                    e);
         }
     }
 }
--- a/baseline-module-jvm-args-compiler-plugins/src/main/java/com/palantir/baseline/modulejvmargs/compilerplugins/AllowReleaseAndAddExportsToBeUsedTogetherByChangingCompilerInternalsUsingReflection.java
+++ b/baseline-module-jvm-args-compiler-plugins/src/main/java/com/palantir/baseline/modulejvmargs/compilerplugins/AllowReleaseAndAddExportsToBeUsedTogetherByChangingCompilerInternalsUsingReflection.java
@@ -25,7 +25,24 @@ import com.sun.tools.javac.util.Context;
 import java.lang.reflect.Field;
 
 /**
- * The name is intentionally long to instil the appropriate level of fear.
+ * The JDK devs arbitrarily decided that you cannot use the `--release` arg along with `--add-exports`
+ * if it's exporting a "system module" ie one that is shipped with the JDK. This is problematic, as if
+ * we want to use `--release` to use a higher JDK to compile against a lower version of the Java
+ * standard library, when someone uses BaselineModuleJvmArgs to add an exports or opens arg to
+ * access system modules, they will immediately face this error and be unable to compile. A good example
+ * of something that needs extra exports are compiler plugin eg any sort error-prone code, including
+ * custom BugCheckers. Or even this very compiler plugin (see build.gradle)!
+ * <br><br>
+ * This is a problem I, frankly, just don't want to have. So we "fix" it with this Javac plugin that
+ * will force the field
+ * <a href="https://github.com/openjdk/jdk/blame/a33aa65fbc70a91fe21e9016c393bb5a764cd75a/src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Modules.java#L150">allowAccessIntoSystem</a>
+ * in Javac's Modules component to be true. We are very fortunate that the error does not actually get
+ * created until after Plugin initialisation occurs
+ * (<a href="https://github.com/openjdk/jdk/blame/a33aa65fbc70a91fe21e9016c393bb5a764cd75a/src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Modules.java#L1676">here</a>)
+ * meaning we can change what we need to change during within a self-contained Plugin.
+ * <br><br>
+ * The name is intentionally long to instil the appropriate level of fear and make it obvious there is
+ * serious customisation happening to the compiler.
  */
 @AutoService(Plugin.class)
 public final class AllowReleaseAndAddExportsToBeUsedTogetherByChangingCompilerInternalsUsingReflection
--- a/baseline-module-jvm-args-compiler-plugins/src/main/java/com/palantir/baseline/modulejvmargs/compilerplugins/AllowReleaseAndAddExportsToBeUsedTogetherByChangingCompilerInternalsUsingReflection.java
+++ b/baseline-module-jvm-args-compiler-plugins/src/main/java/com/palantir/baseline/modulejvmargs/compilerplugins/AllowReleaseAndAddExportsToBeUsedTogetherByChangingCompilerInternalsUsingReflection.java
@@ -41,6 +41,11 @@ import java.lang.reflect.Field;
  * (<a href="https://github.com/openjdk/jdk/blame/a33aa65fbc70a91fe21e9016c393bb5a764cd75a/src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Modules.java#L1676">here</a>)
  * meaning we can change what we need to change during within a self-contained Plugin.
  * <br><br>
+ * The risk here is that the compiler changes such that we can no longer do this hack, particularly
+ * if the timing of the error changes and happens to early before we can prevent it. However, the code
+ * has not changed yet since it was introduced 8 years ago (at time of writing) so is hopefully
+ * unlikely to change.
+ * <br><br>
  * The name is intentionally long to instil the appropriate level of fear and make it obvious there is
  * serious customisation happening to the compiler.
  */
--- a/baseline-module-jvm-args-compiler-plugins/src/main/java/com/palantir/baseline/modulejvmargs/compilerplugins/AllowReleaseAndAddExportsToBeUsedTogetherByChangingCompilerInternalsUsingReflection.java
+++ b/baseline-module-jvm-args-compiler-plugins/src/main/java/com/palantir/baseline/modulejvmargs/compilerplugins/AllowReleaseAndAddExportsToBeUsedTogetherByChangingCompilerInternalsUsingReflection.java
@@ -1,0 +1,56 @@
+/*
+ * (c) Copyright 2025 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.baseline.modulejvmargs.compilerplugins;
+
+import com.google.auto.service.AutoService;
+import com.sun.source.util.JavacTask;
+import com.sun.source.util.Plugin;
+import com.sun.tools.javac.api.BasicJavacTask;
+import com.sun.tools.javac.comp.Modules;
+import com.sun.tools.javac.util.Context;
+import java.lang.reflect.Field;
+
+/**
+ * The name is intentionally long to instil the appropriate level of fear.
+ */
+@AutoService(Plugin.class)
+public final class AllowReleaseAndAddExportsToBeUsedTogetherByChangingCompilerInternalsUsingReflection
+        implements Plugin {
+
+    @Override
+    public String getName() {
+        return getClass().getSimpleName();
+    }
+
+    @Override
+    public void init(JavacTask task, String... _args) {
+        Context context = ((BasicJavacTask) task).getContext();
+        Modules modules = Modules.instance(context);
+
+        try {
+            // As of writing, this field has not been changed in 8 years:
+            // https://github.com/openjdk/jdk/blame/a33aa65fbc70a91fe21e9016c393bb5a764cd75a/
+            //      src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Modules.java#L150
+            // We can but pray that it does not change in the future.
+            Field allowAccessIntoSystem = Modules.class.getDeclaredField("allowAccessIntoSystem");
+            allowAccessIntoSystem.setAccessible(true);
+            allowAccessIntoSystem.setBoolean(modules, true);
+        } catch (NoSuchFieldException | IllegalAccessException e) {
+            throw new RuntimeException(e);
+        }
+    }
+}
--- a/gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
+++ b/gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
@@ -397,6 +397,39 @@ public abstract class BaselineModuleJvmArgs implements Plugin<Project> {
         @Inject
         protected abstract ProjectLayout getProjectLayout();
 
+        @Override
+        public final Iterable<String> asArguments() {
+            List<JarManifestModuleInfo> classpathInfo = collectClasspathInfo(getClasspath());
+            Stream<String> allExports = Stream.concat(
+                    getExports().get().stream(), classpathInfo.stream().flatMap(info -> info.exports().stream()));
+            Stream<String> allOpens = Stream.concat(
+                    getOpens().get().stream(), classpathInfo.stream().flatMap(info -> info.opens().stream()));
+
+            List<String> args = runtimeArgs(allExports, allOpens);
+
+            log.debug(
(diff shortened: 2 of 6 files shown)


baseline-module-jvm-args-compiler-plugins/src/main/java/com/palantir/baseline/modulejvmargs/compilerplugins/AllowReleaseAndAddExportsToBeUsedTogetherByChangingCompilerInternalsUsingReflection.java
```diff
@@ -0,0 +1,78 @@
...
+import com.google.auto.service.AutoService;
+import com.sun.source.util.JavacTask;
+import com.sun.source.util.Plugin;
+import com.sun.tools.javac.api.BasicJavacTask;
+import com.sun.tools.javac.comp.Modules;
+import com.sun.tools.javac.util.Context;
+import java.lang.reflect.Field;
+
+/**
+ * The JDK devs arbitrarily decided that you cannot use the `--release` arg along with `--add-exports`
```
felixdesouza: nit: 
```suggestion
 * The JDK devs arbitrarily decided that you cannot use the {@code --release} arg along with {@code --add-exports}
```

Unfortunately we're not on java 23 :'(

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
```diff
@@ -158,111 +156,148 @@ public void execute(Manifest manifest) {
...
-        javaCompile.getOptions().getCompilerArgumentProviders().add(provider);
+            SourceSet sourceSet, BaselineModuleJvmArgsExtension extension, JavaCompile javaCompile) {
+
+        // We will *always* fork the compiler - for both consistency and correctness:
+        // Consistency: when fork=false, Gradle will sometimes still make a forked Gradle worker daemon
+        //              for compilation! If the main Gradle daemon is running on the correct JDK major version,
+        //              Gradle will just use the current daemon. However, if the main Gradle daemon is
+        //              running with a different JDK Gradle will fork off a worker daemon to do the compilation.
+        //              There are material differences in behaviour when it is forked vs not forked, so
+        //              we always for consistency.
```
felixdesouza:
```suggestion
        //              we always fork for consistency.
```

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
```diff
@@ -158,111 +156,148 @@ public void execute(Manifest manifest) {
...
 
     private void configureJavadoc(Project project, SourceSet sourceSet, BaselineModuleJvmArgsExtension extension) {
-        TaskProvider<Task> javadocTaskProvider = null;
-        try {
-            javadocTaskProvider = project.getTasks().named(sourceSet.getJavadocTaskName());
-        } catch (UnknownTaskException e) {
-            // skip
+        if (!project.getTasks().getNames().contains(sourceSet.getJavadocTaskName())) {
+            // If there's no javadoc we can't configure it
+            return;
```
felixdesouza, line 255: do you not run into timing issues with this? and I suppose the old thing as well

baseline-module-jvm-args-compiler-plugins/src/main/java/com/palantir/baseline/modulejvmargs/compilerplugins/AllowReleaseAndAddExportsToBeUsedTogetherByChangingCompilerInternalsUsingReflection.java
```diff
@@ -0,0 +1,78 @@
...
+ * if it's exporting a "system module" ie one that is shipped with the JDK. This is problematic, as if
+ * we want to use `--release` to use a higher JDK to compile against a lower version of the Java
+ * standard library, when someone uses BaselineModuleJvmArgs to add an exports or opens arg to
+ * access system modules, they will immediately face this error and be unable to compile. A good example
+ * of something that needs extra exports are compiler plugin eg any sort error-prone code, including
+ * custom BugCheckers. Or even this very compiler plugin (see build.gradle)!
```
felixdesouza:
```suggestion
 * we want to use {@code --release} to use a higher JDK to compile against a lower version of the Java
 * standard library, when someone uses {@code BaselineModuleJvmArgs} to add an exports or opens arg to
 * access system modules, they will immediately face this error and be unable to compile. A good example
 * of something that needs extra exports are compiler plugin eg any sort error-prone code, including
 * custom BugCheckers. Or even this very compiler plugin (see {@code build.gradle})!
```

baseline-module-jvm-args-compiler-plugins/src/main/java/com/palantir/baseline/modulejvmargs/compilerplugins/AllowReleaseAndAddExportsToBeUsedTogetherByChangingCompilerInternalsUsingReflection.java
```diff
@@ -0,0 +1,78 @@
...
+ * <br><br>
+ * This is a problem I, frankly, just don't want to have. So we "fix" it with this Javac plugin that
+ * will force the field
+ * <a href="https://github.com/openjdk/jdk/blame/a33aa65fbc70a91fe21e9016c393bb5a764cd75a/src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Modules.java#L150">allowAccessIntoSystem</a>
+ * in Javac's Modules component to be true. We are very fortunate that the error does not actually get
+ * created until after Plugin initialisation occurs
+ * (<a href="https://github.com/openjdk/jdk/blame/a33aa65fbc70a91fe21e9016c393bb5a764cd75a/src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Modules.java#L1676">here</a>)
+ * meaning we can change what we need to change during within a self-contained Plugin.
```
felixdesouza:
```suggestion
 * meaning we can change what we need to change during within a self-contained {@link Plugin}.
```

baseline-module-jvm-args-compiler-plugins/src/main/java/com/palantir/baseline/modulejvmargs/compilerplugins/AllowReleaseAndAddExportsToBeUsedTogetherByChangingCompilerInternalsUsingReflection.java
```diff
@@ -0,0 +1,78 @@
...
+        Context context = ((BasicJavacTask) task).getContext();
+        Modules modules = Modules.instance(context);
+
+        try {
+            // As of writing, this field has not been changed in 8 years:
+            // https://github.com/openjdk/jdk/blame/a33aa65fbc70a91fe21e9016c393bb5a764cd75a/
+            //      src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Modules.java#L150
+            // We can but pray that it does not change in the future.
+            Field allowAccessIntoSystem = Modules.class.getDeclaredField("allowAccessIntoSystem");
+            allowAccessIntoSystem.setAccessible(true);
```
felixdesouza, line 72: perhaps worth writing this part requires `--opens` ?

baseline-module-jvm-args-compiler-plugins/src/main/java/com/palantir/baseline/modulejvmargs/compilerplugins/AllowReleaseAndAddExportsToBeUsedTogetherByChangingCompilerInternalsUsingReflection.java
```diff
@@ -0,0 +1,78 @@
...
+            allowAccessIntoSystem.setBoolean(modules, true);
+        } catch (NoSuchFieldException | IllegalAccessException e) {
+            throw new RuntimeException(e);
```
CRogers: I need to a good exception message here

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
```diff
@@ -111,19 +114,14 @@ private void applyToJavaProject(Project project) {
         });
 
         project.getTasks().withType(Test.class).configureEach(test -> {
-            ModuleJvmArgsArgumentProvider provider = newModuleJvmArgsArgumentProvider(
-                    project, extension, project.files((Callable<FileCollection>) test::getClasspath), test.getName());
-            test.getJvmArgumentProviders().add(provider);
+            test.getJvmArgumentProviders()
+                    .add(ModuleJvmArgsArgumentProvider.fromClasspathAndExtension(test, test::getClasspath));
             setTaskInputsFromExtension(test, extension);
```
felixdesouza, line 119: from IRL conversation, I don't think these are necessary anymore. I imagine not for task dependencies, but caching perhaps? since prior to  #3119 we weren't using the argument provider which with a managed type which is `@Nested`. Probably not required for this PR, but can be done independently of it, @FinlayRJW for SA.

felixdesouza: it would however require some sort of test

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
```diff
@@ -158,111 +156,148 @@ public void execute(Manifest manifest) {
...
+                        "baseline-module-jvm-args-compiler-plugins implementation version not found"));
+
+        project.getExtensions().getByType(SourceSetContainer.class).configureEach(sourceSet -> {
+            project.getDependencies()
+                    .add(
+                            sourceSet.getAnnotationProcessorConfigurationName(),
+                            "com.palantir.baseline:baseline-module-jvm-args-compiler-plugins:" + version);
+
+            project.getTasks().named(sourceSet.getCompileJavaTaskName(), JavaCompile.class, javaCompile -> {
+                javaCompile.getOptions().getCompilerArgumentProviders().add(new CommandLineArgumentProvider() {
```
felixdesouza, line 185: should this be a managed type as well? so that the version of compiler plugin flows in for caching purposes? same as the task inputs from above

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
```diff
@@ -158,111 +156,148 @@ public void execute(Manifest manifest) {
...
-                    @Override
-                    public void execute(Task task) {
-                        // The '--release' flag is set when BaselineJavaVersion is not used.
-                        if (!project.getPlugins().hasPlugin(BaselineJavaVersion.class)) {
-                            log.debug(
-                                    "BaselineModuleJvmArgs not applying args to compilation task {} on {} "
-                                            + "due to lack of BaselineJavaVersion",
-                                    task.getName(),
-                                    project.getPath());
-                            return;
```
felixdesouza, line 210: why is this not a thing anymore

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
```diff
@@ -158,111 +156,148 @@ public void execute(Manifest manifest) {
...
+                                "MinimalJavadocOptions implementation on '{}' was not CoreJavadocOptions, rather '{}'",
+                                javadoc.getPath(),
+                                options.getClass().getName());
+                        return;
                     }
-                });
 
-                setTaskInputsFromExtension(javadocTask, extension);
+                    Set<String> exportValuesRaw = ModuleJvmArgsArgumentProvider.fromJustExtensionForCompilation(javadoc)
+                            .configureWithClasspath(sourceSet::getAnnotationProcessorPath)
```
felixdesouza, line 275: talking IRL, don't necessarily need to do this? but requires a test as its a change in behaviour

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
```diff
@@ -158,111 +156,148 @@ public void execute(Manifest manifest) {
...
+
+                    log.debug(
+                            "BaselineModuleJvmArgs building {} with exports: {}",
+                            javadoc.getPath(),
+                            exportValuesSorted);
+
+                    if (!exportValuesSorted.isEmpty()) {
+                        coreOptions
+                                // options are automatically prefixed with '-' internally
+                                .addMultilineStringsOption("-add-exports")
```
felixdesouza, line 293: wat

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
```diff
@@ -365,70 +395,110 @@ public abstract static class ModuleJvmArgsArgumentProvider implements CommandLin
...
+
+        private Transformer<List<String>, List<JarManifestModuleInfo>> extractArgType(
+                Function<JarManifestModuleInfo, List<String>> extractor) {
+            return jarManifestModuleInfos -> jarManifestModuleInfos.stream()
+                    .flatMap(info -> extractor.apply(info).stream())
+                    .toList();
+        }
+
+        private static BaselineModuleJvmArgsExtension extension(Task task) {
+            return task.getProject().getExtensions().getByType(BaselineModuleJvmArgsExtension.class);
```
felixdesouza, line 503: surprised gradle-guide is not flagging you for this

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
```diff
@@ -365,70 +395,110 @@ public abstract static class ModuleJvmArgsArgumentProvider implements CommandLin
...
+                Task task, Callable<FileCollection> classpathCallable) {
+            return create(task).configureWithClasspath(classpathCallable);
+        }
+
+        public static ModuleJvmArgsArgumentProvider fromJustExtensionForCompilation(Task task) {
+            ModuleJvmArgsArgumentProvider argumentProvider = create(task);
+
+            argumentProvider.getExports().addAll(extension(task).exports());
+            // For compilation, `--add-opens` does nothing at all. Instead, each of the `opens` needs to
+            // become an export for compilation to work.
```
felixdesouza: this doesn't make sense? it sounds like an improperly configured extension i.e. you haven't properly specified your exports

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
```diff
@@ -365,70 +395,110 @@ public abstract static class ModuleJvmArgsArgumentProvider implements CommandLin
...
+            ModuleJvmArgsArgumentProvider provider =
+                    task.getProject().getObjects().newInstance(ModuleJvmArgsArgumentProvider.class);
+
+            provider.getDescription().set(task.getPath());
+
+            return provider;
+        }
+
+        /**
+         * The `getClasspath()` methods on many task types are not as lazy as you'd hope.
```
felixdesouza:
```suggestion
         * The {@code getClasspath()} methods on many task types are not as lazy as you'd hope.
```

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
```diff
@@ -365,70 +395,110 @@ public abstract static class ModuleJvmArgsArgumentProvider implements CommandLin
...
+         * Taking a Callable prevents the mistake of forcing the classpath too early.
```
felixdesouza:
```suggestion
         * Taking a {@link Callable} prevents the mistake of forcing the classpath too early.
```

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
felixdesouza: okay, I think I understand now. `--add-opens` when used at runtime, implies `--add-export` so you don't need to also specify `--add-export` at runtime. So if you used directly imported an encapsulated method (requires `--add-exports`) and also did deep reflection on it (requires `--add-opens`) you'd only need to specify `--add-opens`. 

This kinda sucks imo, but it is what it is.

felixdesouza: Can we update the comment however

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
CRogers: I don't think so - there are no inputs here, it's just a static list of strings we're adding to the args?

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
CRogers: We relied on baseline-java-versions to set the source compatibility, which would stop Gradle adding `--release` automatically (which would cause the error with `--add-exports`). Meaning that if you didn't have baseline-java-versions, using `--add-exports` would explode, so we just gave up if not used. We now have stopped the `--release`/`--add-exports` entirely, so it's safe to remove these blocks

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
CRogers: I don't think there's anything bad here? It's fine to do stuff with project at configuration time, and there is no lazy api for extensions.

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
CRogers: Yep, I don't like it either

felixdesouza:
👍

PR #3342, opened 2025-11-07
baseline-java-versions: `javaCompiler` property to compile Java code with a certain JDK version

CRogers:
## Before this PR
Currently when compiling Java code, if we want to target Java major version XX bytecode in class files, we'll use an XX Java compiler. We _do not_ use the `--release` flag (which would allow us to target a lower source/bytecode/jdk api from a higher versioned Java compiler) because it's is verboten to use `--release` with `--add-exports`, which happens in a number of our codebases.

[Error Prone 2.43.0](https://github.com/google/error-prone/releases/tag/v2.43.0) has [made JDK 21 the minimum required JDK that Error Prone is run in](https://github.com/google/error-prone/issues/4867). The recommendation is to use `--release` so a higher version JDK can safely output java classes at a lower language level. However, we can't do this due to the aforementioned issue with `--add-exports`. Similarly, we're stuck on Java 17 libraries for now, so can't just upgrade everything to Java 21. This makes us stuck, unable to upgrade Error Prone.

## After this PR
==COMMIT_MSG==
baseline-java-versions: Introduce an optional javaCompiler property to javaVersions which will determine which JDK is used to run the Java compiler. In this mode, --release is now used to target lower Java language versions/class file bytecode than the compiler.
==COMMIT_MSG==

We work around the `--release` and `--add-exports` being unable to be used together issue by not setting `--release` when there are exports/opens listed in `moduleJvmArgs`. 

I've expanded the tests quite a lot.

## Possible downsides?
The downside of not using `--release` is that if people use a JDK api that is newer than the compilation target, the compilation will succeed rather than fail. Mitigating factors are:
* IntelliJ still red-underlines the API usage, even if compilation succeeds
* Relatively few repos need `--add-exports` (<100 as of writing) and they are mainly owned by the Java infrastructure teams rather than regular developers.
So I'm hoping this risk *should* be manageable.

Additionally:
* The groovy and scala codepaths in this plugins were/are also entirely untested. I've added tests for Grooyv but left Scala be.
* Needs some more manual testing, just to be sure before rolling out.
* ~~We will need to excavate out the `javaCompiler` property as part of baseline upgrades (and have this ready *before this PR merges*)~~ I've changed it so it will maintain the old behaviour if `javaCompiler` is not set, so we can do stuff async


gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/javaversions/SubprojectBaselineJavaVersionsExtension.java
```diff
@@ -28,37 +28,42 @@ public SubprojectBaselineJavaVersionsExtension(Project project) {
...
     }
 
+    @Override
+    public final void setCompiler(int value) {
+        throwCannotSetFromSubproject();
+    }
+
     @Override
     public final void setLibraryTarget(int _value) {
-        throw throwCannotSetFromSubproject();
```
CRogers, line 33: Hilariously this is `throw`ing the result of a method that always throws.

README.md
```diff
@@ -444,15 +438,17 @@ The plugin is currently used on an opt-in basis. To use it, apply the plugin and
...
     libraryTarget = 11
     distributionTarget = 17
     runtime = 21
 }
 ```
 
 The configurable fields of the `javaVersions` extension are:
-* `libraryTarget`: (required) The Java version used for compilation of libraries that are published.
-* `distributionTarget`: (optional) The Java version used for compilation of code used within distributions, but not published externally. Defaults to the `libraryTarget` version.
+* `compiler`: (required) The version of the Java compiler used.
```
CRogers: I wonder if this should be `javaCompiler` to make clear this version is not being used for groovy/scala compilation

CRogers: I have made this required because if it is not, everything remains unset and Gradle defaults to using the daemon JDK version, which seems not ideal. But it does mean it always needs to be set, even if your project only has groovy/scala and no java.

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/javaversions/BaselineJavaVersion.java
```diff
@@ -153,45 +159,26 @@ public void execute(Task task) {
...
                 .set(getJavaLauncher(rootExtension, baselineConfiguredJavaToolchains, javaToolchainService, target)));
     }
 
-    private static void optOutOfReleaseFlagForGradle7(AbstractCompile compileTask, Provider<ChosenJavaVersion> target) {
-        if (GradleVersion.current().compareTo(GradleVersion.version("8.0")) >= 0) {
-            return;
-        }
-
-        // In Gradle <8, we need to set sourceCompatibility to opt out of '-release', allowing opens/exports to be used.
-        // https://github.com/gradle/gradle/issues/18824#issuecomment-1026909824
```
CRogers, line 184: _Pretty_ sure this is irrelevant after https://github.com/palantir/gradle-baseline/pull/3324. But really need some tests (currently the new tests do not even test against Gradle 7). I'm not even sure groovy/scala have `--add-exports` at compilation time.

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/javaversions/BaselineJavaVersion.java
```diff
@@ -149,49 +155,34 @@ public void execute(Task task) {
...
+            // run Groovy (in contrast to Java's use of `--release`).
             groovyCompileTask
                     .getJavaLauncher()
                     .set(getJavaLauncher(
                             rootExtension, baselineConfiguredJavaToolchains, javaToolchainService, target));
+
             groovyCompileTask
                     .getOptions()
                     .getCompilerArgumentProviders()
                     .add(new EnablePreviewArgumentProvider(target));
```
CRogers, line 159: Unsure if this actually does anything for Groovy compile. Same for Scala. Needs more testing. Might need it to be able to compile against preview APIs (maybe use Java preview source features when compiling with Java files with Groovy?)

CRogers: I guess not really relevant for this PR.

README.md
```diff
@@ -444,15 +438,17 @@ The plugin is currently used on an opt-in basis. To use it, apply the plugin and
 apply plugin: 'com.palantir.baseline-java-versions'
 
 javaVersions {
+    javaCompiler = 25
```
CRogers, line 285: We'll need to excavate this (`javaCompiler = 21`) out.

README.md
CRogers: Done

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/Baseline.java
```diff
@@ -55,7 +55,6 @@ public void apply(Project project) {
             proj.getPluginManager().apply(BaselineReproducibility.class);
             proj.getPluginManager().apply(BaselineClassUniquenessPlugin.class);
             proj.getPluginManager().apply(BaselineExactDependencies.class);
-            proj.getPluginManager().apply(BaselineReleaseCompatibility.class);
```
CRogers, line 60: This would set `--release` on Java <9, but we always set `--release` now so it's no longer needed.

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/javaversions/BaselineJavaVersion.java
```diff
@@ -112,16 +119,47 @@ public void execute(JavaToolchainSpec javaToolchainSpec) {
...
             setJavaCompiler(
-                    javaCompileTask, rootExtension, baselineConfiguredJavaToolchains, javaToolchainService, target);
+                    javaCompileTask, rootExtension, baselineConfiguredJavaToolchains, javaToolchainService, compiler);
             javaCompileTask.getOptions().getCompilerArgumentProviders().add(new EnablePreviewArgumentProvider(target));
-            optOutOfReleaseFlagForGradle7(javaCompileTask, target);
+
+            javaCompileTask
+                    .getOptions()
+                    .getRelease()
+                    .set(target.map(ChosenJavaVersion::asMajorVersion).map(value -> {
```
CRogers: Would use `Provider#filter` here instead, but requires Gradle 8.

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/javaversions/SubprojectBaselineJavaVersionsExtension.java
```diff
@@ -28,37 +28,42 @@ public SubprojectBaselineJavaVersionsExtension(Project project) {
...
     }
 
     @Override
     public final void setRuntime(String _value) {
-        throw throwCannotSetFromSubproject();
+        throwCannotSetFromSubproject();
     }
 
-    private RuntimeException throwCannotSetFromSubproject() {
+    private void throwCannotSetFromSubproject() {
```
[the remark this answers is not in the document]
CRogers: Dear lord no.

README.md
```diff
@@ -481,12 +477,14 @@ A sub-project can also explicitly override the default Java versions, but doing
 ```gradle
 // In a sub-project's build.gradle
 javaVersion {
+    javaCompiler = 17
```
CRogers, line 324: We will need to get gradle-jdks to add this to it's `jdkMajorVersionsToUse` as well as everything else. It currently [does this here](https://github.com/palantir/gradle-jdks/blob/19d1f68fc8b933df9536d00b5d1641ade134a830/gradle-jdks/src/main/java/com/palantir/gradle/jdks/ToolchainsPlugin.java#L83-L97). We should change it so that the `javaVersions` extension exposes some method that provides a `Provider<Set<JavaLanguageVersion>>` of all the java language versions used, rather than gradle-jdks knowing so much about baseline-java-version(s).

CRogers: This method/decoupling will be especially helpful when we come to rewrite baseline-java-versions in a way that supports isolated projects (will require changing the implementation dramatically).

README.md
```diff
@@ -444,15 +438,17 @@ The plugin is currently used on an opt-in basis. To use it, apply the plugin and
...
     libraryTarget = 11
     distributionTarget = 17
     runtime = 21
 }
 ```
 
 The configurable fields of the `javaVersions` extension are:
-* `libraryTarget`: (required) The Java version used for compilation of libraries that are published.
-* `distributionTarget`: (optional) The Java version used for compilation of code used within distributions, but not published externally. Defaults to the `libraryTarget` version.
+* `javaCompiler`: (required) The version of the Java compiler used.
```
crogoz: should we say that the javaCompiler version can be higher that the target versions because we are using the `--release` capability (cross-compiling capability) ? I am not sure if is too clear for people what all these versions mean/what it affects.

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/javaversions/BaselineJavaVersion.java
```diff
@@ -82,9 +84,14 @@ public void execute(JavaToolchainSpec javaToolchainSpec) {
 
             JavaToolchains baselineConfiguredJavaToolchains = new JavaToolchains(project, rootExtension);
 
-            // Compilation tasks (using target version)
+            // Compilation tasks (using java compiler version for the java compiler, but the targeting the target)
```
felixdesouza, line 87: this reads somewhat tautologically lol but also not sure what else you can say

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/javaversions/BaselineJavaVersion.java
```diff
@@ -149,49 +222,34 @@ public void execute(Task task) {
...
 
-            optOutOfReleaseFlagForGradle7(scalaCompileTask, target);
+            scalaCompileTask.getOptions().getCompilerArgumentProviders().add(new EnablePreviewArgumentProvider(target));
         });
 
         project.getTasks().withType(ScalaDoc.class).configureEach(scalaDoc -> scalaDoc.getJavaLauncher()
                 .set(getJavaLauncher(rootExtension, baselineConfiguredJavaToolchains, javaToolchainService, target)));
     }
 
-    private static void optOutOfReleaseFlagForGradle7(AbstractCompile compileTask, Provider<ChosenJavaVersion> target) {
```
felixdesouza, line 178: this is fine to delete because we're always setting source compatibility => on gradle 7 we always opt out of release so not running into the below issue?

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/javaversions/BaselineJavaVersionExtension.java
```diff
@@ -19,29 +19,41 @@
...
 import org.gradle.api.provider.Property;
+import org.gradle.jvm.toolchain.JavaLanguageVersion;
 
 /**
  * Extension named {@code javaVersion} used to set the
  * target and runtime java versions used for a single project.
  */
 public class BaselineJavaVersionExtension {
 
+    private final Property<JavaLanguageVersion> javaCompiler;
```
felixdesouza, line 33: surprised you didn't get dinged on this class re abstract types

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/javaversions/BaselineJavaVersionsExtension.java
```diff
@@ -40,6 +41,22 @@ public class BaselineJavaVersionsExtension implements BaselineJavaVersionsExtens
 
     @Inject
     public BaselineJavaVersionsExtension(Project project) {
+        this.javaCompiler = project.getObjects()
+                .property(JavaLanguageVersion.class)
+                .convention(project.provider(() -> {
```
felixdesouza: you'll need to amend the excavator that rolls this out to handle this? would using the same as distribution/runtime not be a more reasonable default?

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/javaversions/BaselineJavaVersionExtension.java
CRogers: It's already suppressed in the plugin. Eventually I'd like to get this check to work on the actual extension rather than where you create the extension. But requires annotating the extensions or something complex.

gradle-baseline-java/src/test/java/com/palantir/baseline/BaselineJavaVersionsIntegrationTest.java
```diff
@@ -265,6 +268,23 @@ void setting_library_target_to_preview_version_fails(GradleInvoker gradle, RootP
...
+                javaVersions {
+                    libraryTarget = 21
+                }
+                """);
+
+            subProject.mainSourceSet().java().writeClass(JAVA_11_COMPATIBLE_CODE);
+
+            InvocationResult result = gradle.withArgs("compileJava").buildsWithFailure();
+
+            result.assertThat().output().contains("`javaCompiler` property inside `javaVersions`");
```
felixdesouza: can you add a bit more to this error message 😅

gradle-baseline-java/src/test/java/com/palantir/baseline/BaselineModuleJvmArgsIntegrationTest.java
```diff
@@ -200,7 +215,7 @@ annotationProcessor project(':compilerPlugin')
                     // Previously, this plugin incorrectly put --add-exports on the compilerArgs (rather than
                     // the forkOptions that would apply to the compiler plugin in the compiler context).
                     // --add-exports is by default incompatible with `--release`.
-                    options.compilerArgumentProviders.add({ ['--release', '11'] } as CommandLineArgumentProvider)
+                    options.compilerArgumentProviders.add({ ['--release', '17'] } as CommandLineArgumentProvider)
```
felixdesouza, line 220: why the bump here

gradle-baseline-java/test-compiler-plugins/build.gradle
```diff
@@ -0,0 +1,30 @@
...
+moduleJvmArgs {
+    exports = [
+        'jdk.compiler/com.sun.source.util',
+        'jdk.compiler/com.sun.tools.javac.api',
+        'jdk.compiler/com.sun.tools.javac.util'
+    ]
+}
+
+// Solely for publishing to maven local
+apply plugin: 'maven-publish'
```
felixdesouza, line 22: do we want to exclude this from being published to maven central?

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/javaversions/BaselineJavaVersionsExtension.java
CRogers: Yes, I'm writing the excavator code now, which:

> We will need to excavate out the javaCompiler property as part of baseline upgrades (and have this ready before this PR merges)

I'd rather it be explicit than implicit. The implicit behaviour of `runtime`/`distributionTarget` is already confusing all the time internally. Especially so as it's generally not safe to just use the higher of `libraryTarget`/`distributionTarget` - even if you have `libraryTarget = 17` you can't use a 17 compiler with newer errorprone, you need to use 21.

gradle-baseline-java/test-compiler-plugins/build.gradle
CRogers: Because it's for tests only?

README.md
CRogers: Yeah, it's probably good to give this some more info, about the interaction with `--add-exports` too.

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/javaversions/BaselineJavaVersion.java
CRogers: Maybe `javaCompiler` and `target` are just named well? :p

gradle-baseline-java/src/test/java/com/palantir/baseline/BaselineJavaVersionIntegrationTest.java
```diff
@@ -101,32 +131,154 @@ task runMainClass(type: JavaExec) {
...
+                    runtime = 21
+                }
+                """);
+
+            rootProject.mainSourceSet().java().writeClass(JAVA_21_API_USAGE);
+
+            InvocationResult result = gradle.withArgs("compileJava").buildsWithFailure();
+
+            result.assertThat().output().contains("error: cannot find symbol");
+
```
felixdesouza, line 254: did you intend to leave out the release assertion?

gradle-baseline-java/src/test/java/com/palantir/baseline/BaselineModuleJvmArgsIntegrationTest.java
CRogers: 11 doesn't require module args (introduced in 16). It actually wasn't failing in certain cases.

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/javaversions/BaselineJavaVersion.java
CRogers: We are no longer opting out of release. Tbf, we should really run the tests again Gradle 7 too.

CRogers: If we still claim to support it.

gradle-baseline-java/src/test/java/com/palantir/baseline/BaselineJavaVersionIntegrationTest.java
```diff
@@ -225,6 +381,183 @@ void java_17_execution_succeeds_on_java_21(GradleInvoker gradle, RootProject roo
...
+            rootProject.buildGradle().append("""
+                javaVersion {
+                    javaCompiler = 21
+                    target = 17
+                }
+                """);
+
+            rootProject.mainSourceSet().java().writeClass(JAVA_21_SOURCE_FEATURE_CODE);
+
+            InvocationResult result = gradle.withArgs("javadoc").buildsWithFailure();
```
felixdesouza: can you assert on the task that actually failed? my guess is that this test doesn't actually matter because `javadoc` depends on `compileJava`

gradle-baseline-java/src/test/java/com/palantir/baseline/BaselineJavaVersionIntegrationTest.java
```diff
@@ -225,6 +381,183 @@ void java_17_execution_succeeds_on_java_21(GradleInvoker gradle, RootProject roo
...
+        @Test
+        void targeting_17_and_setting_java_compiler_to_21_outputs_17_bytecode(
+                GradleInvoker gradle, RootProject rootProject) {
+
+            // The groovy compiler needs to run with the JDK it's targeting, not with
+            // whatever we're setting for `javaCompiler` (a Java concept).
+
+            rootProject.buildGradle().append("""
+                javaVersion {
+                    javaCompiler = 21
```
felixdesouza, line 619: should we error instead?

gradle-baseline-java/src/test/java/com/palantir/baseline/BaselineJavaVersionIntegrationTest.java
```diff
@@ -225,6 +381,183 @@ void java_17_execution_succeeds_on_java_21(GradleInvoker gradle, RootProject roo
...
+                    .buildDir()
+                    .path()
+                    .resolve("classes/groovy/main/Main.class")
+                    .toFile();
+
+            assertBytecodeVersion(compiledClass, JAVA_17_BYTECODE, NOT_ENABLE_PREVIEW_BYTECODE);
+        }
+
+        @Test
+        void targeting_17_and_setting_java_compiler_to_21_does_let_you_use_21_jdk_apis(
```
felixdesouza:
```suggestion
        void targeting_17_and_setting_java_compiler_to_21_does_not_let_you_use_21_jdk_apis(
```

gradle-baseline-java/src/test/java/com/palantir/baseline/BaselineJavaVersionsIntegrationTest.java
CRogers: The whole error is currently:

> The `javaCompiler` property inside `javaVersions` in the root project must be set with a value

...do you think this needs to be expanded?

gradle-baseline-java/test-compiler-plugins/build.gradle
felixdesouza: yeah

gradle-baseline-java/src/test/java/com/palantir/baseline/BaselineJavaVersionsIntegrationTest.java
felixdesouza: That whole error makes the test make sense

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/javaversions/BaselineJavaVersionsExtension.java
felixdesouza: makes sense

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/javaversions/BaselineJavaVersion.java
felixdesouza: I mean, this is probably a 7.0.0 release anyway, could just sneak in a min gradle 8 in here as well idk

gradle-baseline-java/test-compiler-plugins/build.gradle
CRogers: It is not published to maven central - no `com.palantir.external-publish-jar`

gradle-baseline-java/src/test/java/com/palantir/baseline/BaselineJavaVersionIntegrationTest.java
CRogers: You're right, it's compilation that fails so it doesn't even get to javadoc

gradle-baseline-java/src/test/java/com/palantir/baseline/BaselineJavaVersionIntegrationTest.java
CRogers: A mistake, I've added it back

gradle-baseline-java/src/test/java/com/palantir/baseline/BaselineJavaVersionIntegrationTest.java
CRogers: Error out if there's no java compiler? I think this is fine to leave it in, it just won't be used. We don't error out when distributionTarget is set but not used.

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/javaversions/BaselineJavaVersion.java
```diff
@@ -268,61 +325,63 @@ private static Provider<JavaLauncher> getJavaLauncher(
...
+                            getProjectDisplayName().get(),
+                            javaCompiler,
                             target,
                             runtime);
 
             if (target.enablePreview() && !target.equals(runtime)) {
-                throw new GradleException(String.format(
-                        "Runtime Java version (%s) must be exactly the same as the compilation target (%s), because "
-                                + "--enable-preview is enabled. Otherwise Java will fail to start. See "
+                throw new RuntimeException(String.format(
```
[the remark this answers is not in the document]
pkoenig10: This seems misguided - `GradleException` is the base class of all exceptions thrown by Gradle. I don't think third-party plugins should be throwing `GradleException`.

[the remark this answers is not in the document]
CRogers: I agree and thought this was the point of view of the Gradle team as well - but couldn't actually find anything to back up that GradleExceptions should not be thrown by third party plugins. Still think they shouldn't throw `GradleException`. I reverted this change to just simplify the diff a bit, not because I think copilot is right here.

crogoz (dismissed):

crogoz (dismissed):

crogoz (dismissed):

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/javaversions/BaselineJavaVersion.java
```diff
@@ -120,21 +120,30 @@ public void execute(JavaToolchainSpec javaToolchainSpec) {
...
 
         project.getTasks().withType(JavaCompile.class).configureEach(javaCompileTask -> {
             setJavaCompiler(
-                    javaCompileTask, rootExtension, baselineConfiguredJavaToolchains, javaToolchainService, compiler);
+                    javaCompileTask,
+                    rootExtension,
+                    baselineConfiguredJavaToolchains,
+                    javaToolchainService,
+                    // If `javaCompiler` is explicitly set, we use that version for the compiler, else we fall
+                    // back to the old behaviour of
```
felixdesouza: incomplete comment?

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/javaversions/BaselineJavaVersion.java
```diff
@@ -120,21 +120,30 @@ public void execute(JavaToolchainSpec javaToolchainSpec) {
 
     private static void configureCompilationTasks(
             Project project,
-            Provider<JavaLanguageVersion> compiler,
+            Provider<JavaLanguageVersion> javaCompiler,
             Property<ChosenJavaVersion> target,
```
felixdesouza, line 124: talked offline: ensure that we make `compiler` and `target` the same if preview is enabled as a preview in `target` version is not guaranteed to exist in `compiler` version

felixdesouza (approved):

PR #3353, opened 2025-11-12
Add a instanceof check for sealed unions

kunalrkak:
## Before this PR

Add an errorprone check to steer developers away from using instanceof checks with sealed union types, instead pushing them towards switches and visitors, which are generally more efficient and also force exhaustive handling of all types.

## After this PR

==COMMIT_MSG==
Adds an errorprone check to flag instanceof check usage with sealed union types.
==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ConjureUnionExhaustiveSwitch.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ConjureUnionExhaustiveSwitch.java
@@ -30,8 +30,6 @@ import com.sun.source.tree.SwitchTree;
 import com.sun.source.tree.Tree;
 import com.sun.tools.javac.code.Type;
 import java.util.List;
-import javax.lang.model.element.ElementKind;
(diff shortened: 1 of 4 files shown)


pkoenig10:
This feels a little misguided. Given the existence of `default` cases, I don't think this check adds meaningful protection. There's not really a difference between these two implementations, yet this check would disallow the former and allow the latter.

```java
if (value instanceof Foo foo) {
    // do something
} else {
    // do something else
}
```

```java
switch (value) {
    case Foo foo -> {
        // do something
    },
    default -> {
        // do something else
    }
}
```

It's pretty common to have code that wants to handle exactly one case of a union and using a switch statement is less ergonomic for these cases. Another common pattern I see is "return if not instanceof":

```java
if (!(value instanceof Foo foo)) {
    // return early
}

// do something with foo
```

kelvinou01:
### DO NOT MERGE (YET)

kunalrkak:
We're not moving forward with this errorprone check, just the single check from #3344

PR #3363, opened 2025-11-18
Don't run annoying error-prones during local-development-iteration cycles

kelvinou01:
## Before this PR

Most error-prones give valuable pointers during the development process. Some error-prones just bog us down during development, but are valuable to prevent code quality regressions in a repo — let's call these annoying error-prones.

## After this PR

==COMMIT_MSG==
Don't run annoying error-prones during local-development-iteration cycles
==COMMIT_MSG==


pkoenig10:
This has been discussed multiple times before, see https://github.com/palantir/gradle-baseline/pull/2405. I don't think we want to do this.

For example, this means you won't get build cache hits for local compilation.

PR #3374, opened 2025-11-26
drop nullaway to 0.12.10 

FinlayRJW:
## Before this PR
In nullaway 0.12.12 there is a bug - https://github.com/uber/NullAway/issues/1351 - which can cause build failures

## After this PR
Drop to working version until fixed

==COMMIT_MSG==
==COMMIT_MSG==


felixdesouza:
👍

PR #3376, opened 2025-11-26
Remove javac plugin that allowed using `--release` and `--add-exports` together

CRogers:
## Before this PR
Javac bans you from using `--release` and `--add-exports` together as compiler args - it will error out with:

```
Compiler args: [--add-exports, jdk.compiler/com.sun.tools.javac.code=ALL-UNNAMED, --release, 17]
error: exporting a package from system module jdk.compiler is not allowed with --release
```

Unfortunately, if we want to use a higher compiler with a lower `--release`, you can't add any `--add-exports` - very annoying.

In [a previous PR](https://github.com/palantir/gradle-baseline/pull/3324), I was thought I was being Very Clever™ and wrote a compiler plugin that _I thought_ removed this restriction. More details why over there, but will would allow to use a higher compiler JDK and compiler lower code with `--release` and no downsides, even if `--add-exports` would be used.

However, it didn't actually work, and I had insufficient tests to prove it would work. While the plugin successfully got rid of the error message, actually trying to use a type from an exported package with `--release` set would not be found, eg:

```
Compiler args: [--add-exports, jdk.compiler/com.sun.tools.javac.code=ALL-UNNAMED, --release, 17]
...
src/main/java/com/Example.java:4: error: package com.sun.tools.javac.code does not exist
          com.sun.tools.javac.code.Symbol.class.toString();
```

The issue is that using `--release` changes how the Java compiler functions. Instead of looking at all the possible APIs it has access to (as it does when `--release` is _not_ set), it instead loads a "platform description" for the particular `--release <version>` from the `lib/ct.sym` file in the JDK home. These define the set of APIs that can be used at each release level. Notably, these "platform descriptions" _do not include APIs from any built in system modules_ like for example anything from `jdk.compiler/*`. So the compiler effectively thinks these built in modules do not exist, and fails to find the package/type.

Now, I think it is _possible_ to work around this restriction in the compiler plugin, but it is so much more involved I don't think it's a good a idea. You'd essentially have to reflectively replace the existing `JavaFileManager` which looks up these types in the platform description with a new implementation that will use the default, non-platform description `JavaFileManager` for system modules, and the existing platform description limited one for everything else.

## After this PR
==COMMIT_MSG==
Remove javac plugin that allowed using `--release` and `--add-exports` together as it doesn't actually work.
==COMMIT_MSG==

We just remove this plugin entirely.

In order to fix the original error-prone issue listed [here](https://github.com/palantir/gradle-baseline/pull/3324), I'm instead going to not use `--release` when `--add-exports` is used and instead fall back to `--source` and `--target`, which has downsides, but I think is better than the incredible amount of reflection I'd need to use to make the plugin work properly.

Another benefit is this plugin added a load of required jvm args to the forked compiler process, and these will no longer be added:

```
forkArgs: [--add-exports, jdk.compiler/com.sun.source.util=ALL-UNNAMED, --add-exports, jdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED, --add-exports, jdk.compiler/com.sun.tools.javac.comp=ALL-UNNAMED, --add-exports, jdk.compiler/com.sun.tools.javac.util=ALL-UNNAMED, --add-opens, jdk.compiler/com.sun.tools.javac.comp=ALL-UNNAMED]
```

 These could make projects that needed these same jvm args work without explicitly listing them, which is incorrect, and hide the bug until they're used elsewhere


felixdesouza (approved):
A shame, but makes sense

PR #3379, opened 2025-11-27
Inherit JDKs set up by gradle-jdks inside tests

CRogers:
## Before this PR
Currently, the baseline tests that deal with java compilation and JDKs require certain JDKs to be installed on the machine they're running on. In #3342, I want to start using 21 JDKs. However, the public circle image does not have 21 JDKs, meaning the tests fail.

## After this PR
We configure the tests to run with `gradle-jdks`, which will tell Gradle where the appropriate JDKs are.

The nifty part of the approach is that the tests actually inherit the metadata about the JDKs from the enclosing repo, rather than having to run code in each and every test, via the use of a symlink of the `gradle/jdks` directory.

## Possible downsides?
As discussed with @felixdesouza, this is actually probably not optimal. Inheriting the JDKs from the enclosing repo has a couple of problems:
1. We have to manually list the JDKs required by the tests in the enclosing repo, which is an extra bit of [connascence](https://en.wikipedia.org/wiki/Connascence) that shouldn't be necessary.
2. The versions of gradle-jdks used in the tests and the enclosing repo are tied together. Every time we've done something like this before, we've regretted it as it has stopped upgrades progressing.

The ideal would probably be something built in to the `gradle-plugin-testing` framework, where each test setup has gradle-jdks applied, then sets up the jdks, then runs the `gradle/gradle-jdks-setup.sh` setup script before the first (or perhaps every) Gradle invocation. This would be fully decoupled and the most "realistic" solution we could achieve without forgoing testkit and running `./gradlew` directly. It may have a performance cost though.

That said, even with these downsides, I think this is a strict improvement and we should merge it in, as it will unblock reviewing the next part of #3342.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -105,4 +105,7 @@ javaVersions {
 
 jdks {
     daemonTarget = 21
+
+    // Tests need these JDK versions installed
+    jdkMajorVersionsToUse().addAll([11, 17, 21].collect { JavaLanguageVersion.of(it)})
 }
--- a/gradle-baseline-java/src/test/java/com/palantir/baseline/BaselineJavaVersionIntegrationTest.java
+++ b/gradle-baseline-java/src/test/java/com/palantir/baseline/BaselineJavaVersionIntegrationTest.java
@@ -395,8 +395,8 @@ class BaselineJavaVersionIntegrationTest {
         void succeeds_when_all_runtimeClasspath_jars_are_compatible(GradleInvoker gradle, RootProject rootProject) {
             rootProject.buildGradle().append("""
                 javaVersion {
-                    target = 8
-                    runtime = 8
+                    target = 11
+                    runtime = 11
                 }
 
                 dependencies {
@@ -411,8 +411,8 @@ class BaselineJavaVersionIntegrationTest {
         void handles_gradleApi(GradleInvoker gradle, RootProject rootProject) {
             rootProject.buildGradle().append("""
                 javaVersion {
-                    target = 8
-                    runtime = 8
+                    target = 11
+                    runtime = 11
                 }
 
                 dependencies {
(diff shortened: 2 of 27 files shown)


gradle-baseline-java/src/test/java/com/palantir/baseline/BaselineJavaVersionIntegrationTest.java
```diff
@@ -395,8 +395,8 @@ implementation files(configurations.java17jar)
         void succeeds_when_all_runtimeClasspath_jars_are_compatible(GradleInvoker gradle, RootProject rootProject) {
             rootProject.buildGradle().append("""
                 javaVersion {
-                    target = 8
-                    runtime = 8
```
CRogers, line 399: `gradle-jdks` does not support Java 8. Java 8 is dead.

gradle-baseline-java/src/test/java/com/palantir/baseline/BaselineJavaVersionsIntegrationTest.java
```diff
@@ -234,61 +230,6 @@ void when_setupJdkToolchains_false_no_toolchains_are_configured_by_gradle_baseli
                             "-Porg.gradle.java.installations.auto-download=false")
                     .buildsWithFailure();
         }
-
-        @Test
-        void can_configure_a_jdk_path_to_be_used(GradleInvoker gradle, RootProject rootProject) {
```
CRogers, line 239: This is for a deprecated codepath virtually no repos use that I want to delete soon anyway, the test is too hard to get working properly imo, especially as it can't be run locally.

build.gradle
```diff
@@ -105,4 +105,7 @@ javaVersions {
 
 jdks {
     daemonTarget = 21
+
+    // Tests need these JDK versions installed
+    jdkMajorVersionsToUse().addAll([11, 17, 21].collect { JavaLanguageVersion.of(it)})
```
CRogers, line 110: Having to do this is a downside of the current approach, see the note in `Possible Downsides`.

crogoz (approved):

PR #3384, opened 2025-11-29
StrictUnusedVariable should respect -XoptIgnoreSuppressionAnnotations; @SuppressWarnings("UnusedVariable") no longer suppresses StrictUnusedVariable

kelvinou01:
## Before this PR

1. StrictUnusedVariable opts out of SuppressibleTreePathScanner's suppression mechanism, and checks for suppressions manually. This interferes with suppressible-error-prone's removeUnused mode, which relies on [modifying SuppressibleTreePathScanner](https://github.com/palantir/suppressible-error-prone/blob/develop/gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/transform/SuppressibleTreePathScannerClassVisitor.java). 
2. StrictUnusedVariable has a [quirky setup to deal with alt-names](https://github.com/palantir/gradle-baseline/pull/2599), so that`@SuppressWarnings("UnusedVariable")` suppresses it, but `errorProneOpotions.disable("UnusedVariable")` does not disable it. This also interferes with RemoveUnused mode, as it doesn't know `@SuppressWarnings("UnusedVariable")` is a suppression for StrictUnusedVariable

## After this PR

1. I think UnusedVariable (from which StrictUnusedVariable is mostly copied from) didn't use SuppressibleTreePathScanner because the class was new at the time (2022), but now we can switch to using SuppressibleTreePathScanner to [get our modifications](https://github.com/palantir/suppressible-error-prone/blob/develop/gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/transform/SuppressibleTreePathScannerClassVisitor.java)
2. It is reasonable that `@SuppressWarnings("UnusedVariable")` should only suppress UnusedVariable, `@SuppressWarnings("StrictUnusedVariable")` should only suppress StrictUnusedVariable, and `SuppressWarnings("unused")` is the catch-all suppression for all unused checks (including UnusedMethod). Let's implement this behavior so removeUnused mode can work with StrictUnusedVariable.

==COMMIT_MSG==
StrictUnusedVariable should respect `-XoptIgnoreSuppressionAnnotations`; `@SuppressWarnings("UnusedVariable")` no longer suppresses StrictUnusedVariable
==COMMIT_MSG==

### DIFF
--- a/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StrictUnusedVariable.java
+++ b/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/StrictUnusedVariable.java
@@ -67,10 +67,7 @@ import com.google.errorprone.bugpatterns.BugChecker;
 import com.google.errorprone.bugpatterns.UnusedVariable;
 import com.google.errorprone.fixes.SuggestedFix;
 import com.google.errorprone.fixes.SuggestedFixes;
-import com.google.errorprone.matchers.ChildMultiMatcher.MatchType;
 import com.google.errorprone.matchers.Description;
-import com.google.errorprone.matchers.Matcher;
-import com.google.errorprone.matchers.Matchers;
 import com.google.errorprone.suppliers.Suppliers;
 import com.google.errorprone.util.ASTHelpers;
 import com.sun.source.tree.AnnotationTree;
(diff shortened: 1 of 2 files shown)


kelvinou01:
When this rolls out, places with @SuppressWarnings("UnusedVariable") will additionally suppress for-rollout:StrictUnusedVariable. Then, when removeUnused mode is fixed, we can remove the unused suppressions on UnusedVariable.

kelvinou01:
### DO NOT MERGE (YET)

kelvinou01:

PR #3387, opened 2025-12-01
Move baseline-error-prone and baseline-null-away to palantir/baseline-error-prone

kelvinou01:
## Before this PR

FLUP to https://github.com/palantir/baseline-error-prone/pull/1

Error-prone failures are a frequent blocker of baseline upgrades. We want to decouple error-prone upgrades from baseline upgrades. 

## After this PR

- Move baseline-error-prone and baseline-null-away to palantir/baseline-error-prone
- Temporarily bring in baseline-error-prone so baseline-error-prone is still applied by baseline, and people can apply baseline-null-away without any additional deps

==COMMIT_MSG==
Move baseline-error-prone and baseline-null-away to palantir/baseline-error-prone
==COMMIT_MSG==


README.md
```diff
@@ -107,163 +106,6 @@ Generated Eclipse projects have default per-project code formatting rules as wel
 
 The Eclipse plugin is compatible with the following versions: Checkstyle 7.5+, JDK 1.7, 1.8
 
-
-## com.palantir.baseline-error-prone
-The `com.palantir.baseline-error-prone` plugin brings in the `net.ltgt.errorprone-javacplugin` plugin. The minimal setup is as follows:
```
CRogers, line 112: We should add a note (probably near the top) saying that baseline-error-prone now lives in it's own repo.

kelvinou01: gotcha

gradle-baseline-java/build.gradle
```diff
@@ -80,10 +79,8 @@ tasks.checkJUnitDependencies.enabled = false
 test {
     environment 'CIRCLE_ARTIFACTS', "${buildDir}/artifacts"
     environment 'CIRCLE_TEST_REPORTS', "${buildDir}/circle-reports"
-
-    // This sets the baselineErrorProneGradle property so that the tests know which version of baseline-error-prone
-    // to request from mavenLocal()
-    environment 'ORG_GRADLE_PROJECT_baselineErrorProneVersion', project.version
+    // TEMPHACK(okelvin): remove this to decouple baseline-error-prone
+    environment 'ORG_GRADLE_PROJECT_baselineErrorProneVersion', "0.2.0"
```
CRogers, line 83: Do we need this anymore?

CRogers: I would have thought it's only for tests that we moved

gradle-baseline-java/build.gradle
```diff
@@ -31,6 +29,11 @@ dependencies {
     implementation 'com.palantir.gradle.utils:environment-variables'
     implementation 'com.palantir.gradle.failure-reports:gradle-failure-reports-exceptions'
 
+    // TEMPHACK(okelvin): remove this to decouple baseline-error-prone
+    implementation 'com.palantir.baseline-error-prone:gradle-baseline-error-prone'
```
CRogers: It should probably be `api` to maintain backcompat (although we changed the package for the classes in baseline-error-prone)

gradle-baseline-java/build.gradle
kelvinou01: yes! so all the local tests which apply baseline (which apply baseline-error-prone) work

gradle-baseline-java/build.gradle
CRogers: So I guess it doesn't matter, the only affected repo is g-c internally.

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/Baseline.java
```diff
@@ -16,6 +16,7 @@
 
 package com.palantir.baseline.plugins;
 
+import com.palantir.baseline.BaselineErrorProne;
```
CRogers: We really should have put all the gradle plugin classes inside `baseline-error-prone` under a more descriptive package... something like `com.palantir.gradle.baselineerrorprone` - can we do that while we still can?

kelvinou01: yep lets do that in baseine-error-prone 0.3.0

CRogers:
:+1:

PR #3391, opened 2025-12-02
Mark `ExplicitDependency` as public

crogoz:
## Before this PR

When configuration cache was enabled for `checkUnusedDepedency` we would get an error of the format: 

```
Could not load the value of field `__explicitDependencies__` of task `:test-security:checkUnusedDependenciesMain` of type `com.palantir.baseline.tasks.CheckUnusedDependenciesTask`.
> Failed to create instance of com.palantir.baseline.tasks.CheckUnusedDependenciesTask$ExplicitDependency with args [com.google.code.findbugs, jsr305, null, jar]
Caused by: java.lang.NoSuchMethodException: com.palantir.baseline.tasks.CheckUnusedDependenciesTask$ExplicitDependency.<init>(java.lang.String,java.lang.String,java.lang.String,java.lang.String)
```

initial thread [here](https://pl.ntr/2yn)

## After this PR
Make `ExplicitDependency` public such that it can be instantiated by gradle. 

==COMMIT_MSG==
Mark `ExplicitDependency` as public
==COMMIT_MSG==

FLUPs: Add an errorProne for these cases.

### DIFF
--- a/gradle-baseline-java/src/main/java/com/palantir/baseline/tasks/CheckUnusedDependenciesTask.java
+++ b/gradle-baseline-java/src/main/java/com/palantir/baseline/tasks/CheckUnusedDependenciesTask.java
@@ -170,6 +170,7 @@ public abstract class CheckUnusedDependenciesTask extends DefaultTask {
         }));
     }
 
+    // public because it needs to be able to be accessed by Gradle when Configuration Cache is enabled.
     public record ExplicitDependency(
             String group,
             String name,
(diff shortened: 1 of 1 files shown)


gradle-baseline-java/src/main/java/com/palantir/baseline/tasks/CheckUnusedDependenciesTask.java
```diff
@@ -170,7 +170,7 @@ public final void withDeclaredDependenciesFrom(Provider<Configuration> configura
         }));
     }
 
-    protected record ExplicitDependency(
+    public record ExplicitDependency(
```
pkoenig10, line 174: Let's add a test to prevent regression. Or at least a comment explaining why this is public.

pkoenig10 (dismissed):

pkoenig10:
For context, this is the relevant Gradle issue: https://github.com/gradle/gradle/issues/35594

pkoenig10 (approved):

PR #3392, opened 2025-12-02
[Error-prone] [DeprecatedApiUsage] Suppress within deprecated methods/classes

aldexis:
## Before this PR
The Java compiler actually suppresses the deprecation warnings when within a deprecated method or class. This makes sense since deprecating using a deprecated api is a valid way to handle the deprecation if handling the deprecation requires doing an ABI break (e.g. by passing another argument).

## After this PR
==COMMIT_MSG==
[Error-prone] [DeprecatedApiUsage] Suppress within deprecated methods/classes
==COMMIT_MSG==

Note: This PR won't merge within this repo and instead needs to merge in https://github.com/palantir/baseline-error-prone once it's ready (cc @kelvinou01). I've opened it merely to have it somewhere while waiting for the new repo to be ready


CRogers:
DNM as we're extracting this code out to https://github.com/palantir/baseline-error-prone

kelvinou01:

CRogers:
Please recreate this PR on https://github.com/palantir/baseline-error-prone/

PR #3396, opened 2025-12-03
Migrate groovy test `BaselineExactDependenciesTest` to the new Java framework

crogoz:
## After this PR
Ran the migration locally following the instructions [here](https://github.palantir.build/excavator-checks/excavator-gradle/blob/develop/excavator/roomba/migrate-gradle-plugin-testing-junit.sh#L16-L19). Tested by comparing the old/new tests in Intellij. 

I want to use this test to check if the new changes in https://github.com/palantir/gradle-baseline/pull/3391 are compatible with CC.

==COMMIT_MSG==
Migrate groovy test `BaselineExactDependenciesTest` to the new Java framework
==COMMIT_MSG==


gradle-baseline-java/src/test/java/com/palantir/baseline/BaselineExactDependenciesTest.java
```diff
@@ -0,0 +1,446 @@
...
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+
+@GradlePluginTests
+@DisabledConfigurationCache
+public class BaselineExactDependenciesTest {
+
+    private void applyStandardPlugins(RootProject rootProject) {
+        rootProject.buildGradle().plugins().add("java");
+        rootProject.buildGradle().plugins().add("com.palantir.baseline-exact-dependencies");
```
crogoz: diff from the old framework: https://github.com/palantir/gradle-baseline/pull/3396/files#diff-033634a655e9d3f2af7704b1da55c54fe2f62fa7060ef6ed16b1e8c1531b618aL34-L35 (previously it was adding 2 plugins without applying them - not really sure what is the reason for this though). 

For the test https://github.com/palantir/gradle-baseline/pull/3396/files#diff-b026a686feb05404b8f5d356e0e84b44b388981aa92e7af591723af9f435e84fR308 we would end up with a root build.gradle like this - which is invalid - 
```
plugins {
  ...
  id 'consistent-versions' apply false
  id 'consistent-versions'
}
```

gradle-baseline-java/src/test/java/com/palantir/baseline/BaselineExactDependenciesTest.java
```diff
@@ -0,0 +1,446 @@
...
+
+    private String minimalJavaFile() {
+        return """
+            package pkg;
+            public class Foo { void foo() {} }
+            """;
+    }
+
+    @Test
+    public void both_tasks_vacuously_pass_with_no_dependencies(GradleInvoker gradle, RootProject rootProject) {
```
crogoz, line 51: one extra change from the old framework - the old framework was adding `--stacktrace` everywhere. This is done now by default in the new framework, so I removed the explicit `--stacktrace` from the runTasks calls.

gradle-baseline-java/src/test/java/com/palantir/baseline/BaselineExactDependenciesTest.java
```diff
@@ -0,0 +1,446 @@
...
+    }
+
+    /**
+     * Sets up a multi-module project with 2 subprojects. The root project has a transitive dependency
+     * on subProjectNoDeps and so checkImplicitDependencies should fail on it.
+     */
+    private void setupMultiProject(
+            RootProject rootProject, SubProject subProjectNoDeps, SubProject subProjectWithDeps) {
+        applyStandardPlugins(rootProject);
+        // Apply plugins to all projects individually
```
crogoz, line 383: before this was done in the `rootProject.allProjects {apply plugin ....}` but we don't have such an API atm. Do you think it is worth adding one ?

gradle-baseline-java/src/test/java/com/palantir/baseline/BaselineExactDependenciesTest.java
```diff
@@ -0,0 +1,446 @@
...
+    @Test
+    public void both_tasks_vacuously_pass_with_no_dependencies_when_entire_baseline_is_applied(
+            GradleInvoker gradle, RootProject rootProject) {
+        applyStandardPlugins(rootProject);
+        rootProject.buildGradle().plugins().add("com.palantir.baseline");
+        rootProject.buildGradle().append("""
+                repositories {
+                    mavenCentral()
+                    mavenLocal() // for baseline-error-prone
+                }
```
CRogers: weird indent

gradle-baseline-java/src/test/java/com/palantir/baseline/BaselineExactDependenciesTest.java
```diff
@@ -0,0 +1,446 @@
...
+        applyStandardPlugins(rootProject);
+        rootProject.mainSourceSet().java().writeClass(minimalJavaFile());
+
+        gradle.withArgs("checkUnusedDependencies", "checkImplicitDependencies").buildsSuccessfully();
+    }
+
+    @Test
+    public void both_tasks_vacuously_pass_with_no_dependencies_when_entire_baseline_is_applied(
+            GradleInvoker gradle, RootProject rootProject) {
+        applyStandardPlugins(rootProject);
```
CRogers: I wouldn't necessarily expect the AI migration to do any better that this (and definitely wouldn't want to block migration on it), but I kinda dislike this `standardBuildFile` approach. Imo it's a lot better to group similar tests into an `@Nested` which has `@BeforeEach` inside it which adds these.

gradle-baseline-java/src/test/java/com/palantir/baseline/BaselineExactDependenciesTest.java
CRogers: I don't think you can use `plugins` block inside `allprojects` - it has to be at the top. We can remove some duplication though:

```java
Stream.of(rootProject, subProjectNoDeps, subProjectWithDeps).forEach(...add plugins...)
```

I guess we could add a `rootProject.allProjects(project -> ...)` option to the testing framework... Maybe we wait until we see something similar again. I think outside of plugins people can just use `allpojects`

CRogers:
:+1:

CRogers:
Don't need changelog for repo-internal changes like test framework migration

gradle-baseline-java/src/test/java/com/palantir/baseline/BaselineExactDependenciesTest.java
crogoz: I was able to use just the @beforeEach (every test used the applyStandardPlugins) https://github.com/palantir/gradle-baseline/pull/3396/commits/7b4198d74b875377ba044f897e9c9c3229db0e28

Yes, I agree, it would be nicer to tell the LLM to do this change. I wonder though if it should be done in a separate commit ? This would deviate from the original code and it would make  reviewing the diff a bit harder to read.

CRogers: Yeah, I think it's probably best if we just let it keep the same structure as before, maybe we make an AI refactoring thing as a later to do this if we really desired

CRogers (approved):

PR #3399, opened 2025-12-03
Introduce an `allJavaVersionsUsed()` method to `javaVersions` extension

CRogers:
## Before this PR
`gradle-jdks` needs to know what JDKs are being used in a build to work out which JDKs it should actually set up. It currently does this by [deeply inspecting all the values inside the `javaVersions` and `javaVersion` extensions](https://github.com/palantir/gradle-jdks/blob/19d1f68fc8b933df9536d00b5d1641ade134a830/gradle-jdks/src/main/java/com/palantir/gradle/jdks/ToolchainsPlugin.java#L83-L97).

This is problematic, as [I'm about to add a `javaCompiler` option to `javaVersions`](https://github.com/palantir/gradle-baseline/pull/3342) it will need to know about. Currently, `gradle-jdks` is tightly coupled as it needs to know the internal structure of `baseline-java-versions`. When we come to add `javaCompiler`, if we set it to a JDK not already in used in a project (eg `javaCompiler = 25`) then `gradle-jdks` will not know to install it without it's own logic being updated.

If the structure of `baseline-java-versions` ever changes again, `gradle-jdks` will need updating, which is grim.

## After this PR
==COMMIT_MSG==
Introduce an `allJavaVersionsUsed()` method to `javaVersions` extension
==COMMIT_MSG==

`gradle-jdks` can now use this method instead and avoid the tight coupling. When we add `javaCompiler`, it will get the new JDK automatically with no code changes.

`gradle-jdks` PR that uses this new method: https://github.com/palantir/gradle-jdks/pull/622

### DIFF
--- a/gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/javaversions/BaselineJavaVersionsExtension.java
+++ b/gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/javaversions/BaselineJavaVersionsExtension.java
@@ -18,11 +18,15 @@ package com.palantir.baseline.plugins.javaversions;
 
 import com.palantir.gradle.utils.lazilyconfiguredmapping.LazilyConfiguredMapping;
 import java.util.Optional;
+import java.util.Set;
 import java.util.concurrent.atomic.AtomicReference;
 import javax.inject.Inject;
 import org.gradle.api.GradleException;
 import org.gradle.api.Project;
+import org.gradle.api.model.ObjectFactory;
 import org.gradle.api.provider.Property;
+import org.gradle.api.provider.Provider;
+import org.gradle.api.provider.SetProperty;
 import org.gradle.jvm.toolchain.JavaInstallationMetadata;
 import org.gradle.jvm.toolchain.JavaLanguageVersion;
 
@@ -30,7 +34,8 @@ import org.gradle.jvm.toolchain.JavaLanguageVersion;
  * Extension named {@code javaVersions} on the root project used to configure all java modules
  * with consistent java toolchains.
  */
-public class BaselineJavaVersionsExtension implements BaselineJavaVersionsExtensionSetters {
+public abstract class BaselineJavaVersionsExtension implements BaselineJavaVersionsExtensionSetters {
+    private final Project rootProject;
     private final Property<JavaLanguageVersion> libraryTarget;
     private final Property<ChosenJavaVersion> distributionTarget;
     private final Property<ChosenJavaVersion> runtime;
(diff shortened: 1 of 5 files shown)


gradle-baseline-java/src/test/java/com/palantir/baseline/BaselineJavaVersionsIntegrationTest.java
```diff
@@ -269,6 +267,101 @@ void setting_library_target_to_preview_version_fails(GradleInvoker gradle, RootP
...
+                javaVersion {
+                    target = 23
+                    runtime = 24
+                }
+                """);
+
+            InvocationResult result =
+                    gradle.withArgs("printAllJavaVersionsUsed").buildsSuccessfully();
+
+            result.assertThat().output().contains("allJavaVersionsUsed: [11, 17, 21, 23, 24]");
```
felixdesouza, line 322: nice

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/javaversions/BaselineJavaVersionsExtension.java
```diff
@@ -39,11 +44,15 @@ public class BaselineJavaVersionsExtension implements BaselineJavaVersionsExtens
     private final Property<Boolean> setupJdkToolchains;
 
     @Inject
-    public BaselineJavaVersionsExtension(Project project) {
-        this.libraryTarget = project.getObjects().property(JavaLanguageVersion.class);
-        this.distributionTarget = project.getObjects().property(ChosenJavaVersion.class);
-        this.runtime = project.getObjects().property(ChosenJavaVersion.class);
-        this.setupJdkToolchains = project.getObjects().property(Boolean.class);
+    protected abstract ObjectFactory getObjectFactory();
```
felixdesouza, line 47: not blocking but why not just convert all the properties since you've made it abstract

CRogers: Unfortunately we've coded ourselves into a corner with the existing code. There's a long standing gradle bug where you can't have:

```java
public abstract Property<String> getFoo();

public final void setFoo(int value) { ... }
```

...which is what we'd have to do here, in the most simple conversion. I tried it [here](https://github.com/palantir/gradle-baseline/commit/6956087dc640d72b08a4c80098de011f4096ae81) and you get the wonderful error message:

```
An exception occurred applying plugin request [id: 'com.palantir.baseline-java-versions']
> Failed to apply plugin 'com.palantir.baseline-java-versions'.
   > Could not create an instance of type com.palantir.baseline.plugins.javaversions.BaselineJavaVersionsExtension.
      > Could not generate a decorated class for type BaselineJavaVersionsExtension.
         > Cannot have abstract method BaselineJavaVersionsExtension.getDistributionTarget(): Property<ChosenJavaVersion>.
```

So it needs a little more thought than the straightforward change, and we should do this in another PR.

crogoz (approved):

PR #3416, opened 2025-12-08
baseline-java-versions: Ensure `checkJavaVersions` handles `javaCompiler` not being set

CRogers:
## Before this PR
In https://github.com/palantir/gradle-baseline/pull/3342 we introduced a `javaCompiler` parameter to let people control the Java compiler version use for the compiler process. At the last minute in the PR, we decided to make `javaCompiler` optional  so it would not be a breaking change. Despite testing with RCs up until that point, we did not test on a real repo's CI after the change to make it optional. The `checkJavaVersions` task cannot handle the optional property (we had no test for this) and fails with this error:

```
A problem was found with the configuration of task ':docker:checkJavaVersions' (type 'BaselineJavaVersion.CheckJavaVersionsTask').
  - In plugin 'com.palantir.baseline.plugins.BaselineCheckstyle' type 'com.palantir.baseline.plugins.javaversions.BaselineJavaVersion.CheckJavaVersionsTask' property 'javaCompilerVersion' doesn't have a configured value.
    
    Reason: This property isn't marked as optional and no value has been configured.
    
    Possible solutions:
      1. Assign a value to 'javaCompilerVersion'.
      2. Mark property 'javaCompilerVersion' as optional.
```

## After this PR

==COMMIT_MSG==
baseline-java-versions: Ensure `checkJavaVersions` handles `javaCompiler` not being set, avoiding this error:
> 'com.palantir.baseline.plugins.javaversions.BaselineJavaVersion.CheckJavaVersionsTask' property 'javaCompilerVersion' doesn't have a configured value.
==COMMIT_MSG==

Added a test to make sure this doesn't happen.

### DIFF
--- a/gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/javaversions/BaselineJavaVersion.java
+++ b/gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/javaversions/BaselineJavaVersion.java
@@ -19,6 +19,7 @@ package com.palantir.baseline.plugins.javaversions;
 import com.google.common.collect.Sets;
 import com.palantir.baseline.extensions.BaselineModuleJvmArgsExtension;
 import java.util.Collections;
+import java.util.Optional;
 import javax.inject.Inject;
 import org.gradle.api.Action;
 import org.gradle.api.DefaultTask;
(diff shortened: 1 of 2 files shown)


felixdesouza (approved):
fun

PR #3417, opened 2025-12-08
Don't apply `baseline-error-prone` if it already exists

kelvinou01:
## Before this PR

We are migrating to a separate baseline-error-prone repository. Firstly we entered an intermediary stage https://github.com/palantir/gradle-baseline/pull/3387 to not block contributions to gradle-baseline and baseline-error-prone. 

To reduce the risk of error-prone not being applied, lets enter the second stage of the rollout, where both `com.palantir.baseline` and `com.palantir.baseline-error-prone-root` applies `com.palantir.baseline-error-prone` to all projects if it isn't already applied: https://github.com/palantir/baseline-error-prone/pull/21

We don't want to apply the plugin twice because e.g. we cannot create the same extension twice. 

Then, when the dust settles, `com.palantir.baseline` can completely decouple from `baseline-error-prone`

## After this PR

==COMMIT_MSG==
Don't apply `baseline-error-prone` if it already exists
==COMMIT_MSG==

## Possible downsides?

No behavior changes expected.

### DIFF
--- a/gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/Baseline.java
+++ b/gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/Baseline.java
@@ -51,7 +51,9 @@ public final class Baseline implements Plugin<Project> {
             proj.getPluginManager().apply("com.palantir.baseline-eclipse");
             proj.getPluginManager().apply("com.palantir.baseline-idea");
             // TEMPHACK(okelvin): remove this to decouple baseline-error-prone
-            proj.getPluginManager().apply(BaselineErrorProne.class);
+            if (!proj.getPluginManager().hasPlugin("com.palantir.baseline-error-prone")) {
+                proj.getPluginManager().apply(BaselineErrorProne.class);
+            }
             proj.getPluginManager().apply(BaselineFormat.class);
             proj.getPluginManager().apply(BaselineEncoding.class);
             proj.getPluginManager().apply(BaselineReproducibility.class);


gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/Baseline.java
```diff
@@ -51,7 +51,9 @@ public void apply(Project project) {
             proj.getPluginManager().apply("com.palantir.baseline-eclipse");
             proj.getPluginManager().apply("com.palantir.baseline-idea");
             // TEMPHACK(okelvin): remove this to decouple baseline-error-prone
-            proj.getPluginManager().apply(BaselineErrorProne.class);
+            if (!proj.getPluginManager().hasPlugin("com.palantir.baseline-error-prone")) {
+                proj.getPluginManager().apply(BaselineErrorProne.class);
+            }
```
CRogers, line 56: I don't think we need to do this:
1. This only works if `baseline-error-prone` is applied before `baseline`, which isn't necessarily the case.
2. There's no harm in applying a plugin twice - if you try to apply a plugin that has already been applied, it's a no-op.

I think we can just leave the code as is and we can remove this line when we actually want to do the break

kelvinou01: > This only works if baseline-error-prone is applied before baseline, which isn't necessarily the case.
yep that's why we're doing the check in both plugins, and only rolling out baseline-error-prone after the version of baseline this PR represents

> There's no harm in applying a plugin twice - if you try to apply a plugin that has already been applied, it's a no-op.

as discussed in https://github.com/palantir/baseline-error-prone/pull/21 there is an error from creating duplicate baselineErrorProne extensions

kelvinou01: This problem seems to have magically resolved by itself locally 😄 closing this PR

PR #3429, opened 2025-12-17
Run checkstyle with the `javaCompiler` version

kelvinou01:
## Before this PR

Checkstyle would run with the `target` version, rather than the `javaCompiler` version. Meaning a project with target 11 and compiler 21 would fail with

```

* What went wrong:
Execution failed for task ':gradle-jdks-setup-common:checkstyleMain'.
> A failure occurred while executing org.gradle.api.plugins.quality.internal.CheckstyleAction
   > com/puppycrawl/tools/checkstyle/ant/CheckstyleAntTask has been compiled by a more recent version of the Java Runtime (class file version 61.0), this version of the Java Runtime only recognizes class file versions up to 55.0

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.
> Get more help at https://help.gradle.org.

BUILD FAILED in 5s
```

## After this PR

==COMMIT_MSG==
Run checkstyle with the `javaCompiler` version
==COMMIT_MSG==

### DIFF
--- a/gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/javaversions/BaselineJavaVersion.java
+++ b/gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/javaversions/BaselineJavaVersion.java
@@ -225,13 +225,17 @@ public final class BaselineJavaVersion implements Plugin<Project> {
         });
         // checkstyle.getJavaLauncher() was added in Gradle 7.5
         if (GradleVersion.current().compareTo(GradleVersion.version("7.5")) >= 0) {
+            // Checkstyle requires at least Java 17 to run, so we use the javaCompiler version, which is usually
+            // set high compared to the target version.
+            Provider<ChosenJavaVersion> checkstyleJavaVersion =
+                    javaCompiler.map(ChosenJavaVersion::of).orElse(target);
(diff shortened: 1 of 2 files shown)


gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/javaversions/BaselineJavaVersion.java
```diff
@@ -228,7 +228,10 @@ public void execute(Task task) {
             project.getTasks().withType(Checkstyle.class).configureEach(checkstyle -> checkstyle
                     .getJavaLauncher()
                     .set(getJavaLauncher(
-                            rootExtension, baselineConfiguredJavaToolchains, javaToolchainService, target)));
+                            rootExtension,
+                            baselineConfiguredJavaToolchains,
+                            javaToolchainService,
+                            javaCompiler.map(ChosenJavaVersion::of).orElse(target))));
```
CRogers: Needs tests for both `javaCompiler` being set and not. I'd add a comment summarising our discussion about this using `javaCompiler` as an imperfect proxy for what we should run checkstyle with, as it's likely to be higher than the `target` so we can run checkstyle that requires Java 17.

CRogers (approved):

PR #3453, opened 2026-01-16
Drop support for the Eclipse IDE

kelvinou01:
## Before this PR

The eclipse plugins are causing issues with Gradle 9. Since we don't officially support Eclipse as an IDE anymore, let's drop `com.palantir.baseline-eclipse` from baseline. 

## After this PR

==COMMIT_MSG==
Drop support for the Eclipse IDE
==COMMIT_MSG==


versions.props
```diff
@@ -41,7 +41,7 @@ org.unbroken-dome.gradle-plugins:gradle-testsets-plugin = 4.1.0
 # dependency-upgrader:OFF
 # Don't upgrade, we will remove this in a future release.
 # Users should depend on this plugin directly such that it will get updated by excavator.
-com.palantir.javaformat:gradle-palantir-java-format = 2.73.0
+com.palantir.javaformat:gradle-palantir-java-format = 2.85.0
```
kelvinou01: to bring in https://github.com/palantir/palantir-java-format/pull/1490

gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineConfigIntegrationTest.groovy
```diff
@@ -46,10 +46,9 @@ class BaselineConfigIntegrationTest extends AbstractPluginTest {
 
         then:
         with('--stacktrace', '--info', 'baselineUpdateConfig',
-                '-Pcom.palantir.baseline-format.eclipse',
                 '-Pcom.palantir.baseline-format.palantir-java-format').build()
         directory('.baseline').list().toList().toSet() == [
-                'checkstyle', 'copyright', 'eclipse', 'idea', 'spotless'
+                'checkstyle', 'copyright', 'idea'
```
kelvinou01: `spotless` is only used to store `spotless/eclipse.xml`

gradle-baseline-java/src/test/java/com/palantir/baseline/BaselineFormatIntegrationTest.java
```diff
@@ -177,32 +152,15 @@ void format_task_works_on_new_source_sets(GradleInvoker gradle, RootProject root
...
-
-        rootProject.file("src/foo/groovy/test/Test.java").overwrite(invalidJavaFile);
-
-        InvocationResult result = gradle.withArgs("format", "-Pcom.palantir.baseline-format.eclipse")
-                .buildsSuccessfully();
-
-        assertThat(result).task(":format").succeeded();
-        assertThat(result).task(":spotlessApply").succeeded();
-        rootProject.file("src/foo/groovy/test/Test.java").assertThat().hasContent(validJavaFile);
-    }
```
kelvinou01, line 204: no longer supported

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineFormat.java
```diff
@@ -172,18 +148,6 @@ private static void configureSpotlessJava(Project project, SpotlessExtension spo
             project.getPlugins().apply(PJF_PLUGIN); // provides the formatDiff task
         }
 
-        if (eclipseFormattingEnabled(project)) {
```
CRogers, line 175: Eclipse formatting is not actually the same as "eclipse support". It was an option to use the formatter from eclipse before we build palantir-java-format, even when people weren't actually using eclipse (eg CI). Might be worth leaving this in for now and removing later, we just need to check no one is using this (especially big old internal monorepo project)

README.md
```diff
@@ -6,12 +6,11 @@
...
 [![Gradle Plugin Portal](https://img.shields.io/maven-metadata/v/https/plugins.gradle.org/m2/com/palantir/baseline/com.palantir.baseline.gradle.plugin/maven-metadata.xml.svg?label=plugin&logo=gradle)](https://plugins.gradle.org/plugin/com.palantir.baseline)
 
-_Baseline is a family of Gradle plugins for configuring Java projects with sensible defaults for code-style, static analysis, dependency versioning, CircleCI and IntelliJ IDEA/Eclipse integration._
+_Baseline is a family of Gradle plugins for configuring Java projects with sensible defaults for code-style, static analysis, dependency versioning, CircleCI and IntelliJ IDEA integration._
 
 | Plugin                                             | Description            |
 |----------------------------------------------------|------------------------|
 | `com.palantir.baseline-idea`                       | Configures [Intellij IDEA](https://www.jetbrains.com/idea/) with code style and copyright headers
-| `com.palantir.baseline-eclipse`                    | Configures [Eclipse](https://www.eclipse.org/downloads/) with code style and copyright headers
```
CRogers, line 14: Lots of repos (around 60) apply `com.palantir.baseline-eclipse` manually. We should go and strip these out before merging this in so baseline upgrades don't error.

versions.props
CRogers: Do we need this version bump for interop reasons? palantir-java-format will be upgraded independently in other repos so we shouldn't need to bump it here for repos to get features; it just means that all gradle-baseline upgrade excavators will also be bumping PJF and may get blocked on PJF issues

.baseline/eclipse/dynamic/dotfile.checkstyle
CRogers, line 1: These are actually the files for `.baseline` for the gradle-baseline repo itself. I'd leave everything as it is in `.baseline` and when `gradle-baseline` gets `gradle-baseline` 7.0 it will automatically delete these.

CRogers: It's the ones in `gradle-baseline-java-config/resources/` we want to delete

versions.props
kelvinou01: > palantir-java-format will be upgraded independently in other repos so we shouldn't need to bump it here for repos to get features

yep good point! ill revert this bump

.baseline/eclipse/dynamic/dotfile.checkstyle
kelvinou01: yep, ive reverted this !

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineFormat.java
kelvinou01: I've reverted the changes related to the eclipse formatter

esword (approved):

PR #3595, opened 2026-04-24
Remove `LineLength` checkstyle

crogoz:
## Before this PR

Flagging lines like:  https://pl.ntr/2AN (123 characters exceeding 120 characters)

## After this PR
We are formatting using `palantir-java-format` which will format correctly (up to 120 characters when it can split the test (eg. by space)). The `LineLength` checks shouldn't be needed anymore.

==COMMIT_MSG==
Remove LineLength checkstyle
==COMMIT_MSG==


felixdesouza:
👍

PR #3606, opened 2026-05-12
Bump baseline-error-prone

kelvinou01:
## Before this PR

baseline-error-prone bump was failing with:

```
> Task :gradle-baseline-java:compileTestGroovy FAILED
An exception has occurred in the compiler (17.0.18). Please file a bug against the Java compiler via the Java bug reporting page (https://bugreport.java.com/) after checking the Bug Database (https://bugs.java.com/) for duplicates. Include your program, the following diagnostic, and the parameters passed to the Java compiler in your report. Thank you.
java.lang.UnsupportedClassVersionError: com/palantir/suppressibleerrorprone/timings/TimingsPlugin has been compiled by a more recent version of the Java Runtime (class file version 65.0), this version of the Java Runtime only recognizes class file versions up to 61.0
	at java.base/java.lang.ClassLoader.defineClass1(Native Method)
	at java.base/java.lang.ClassLoader.defineClass(ClassLoader.java:1017)
	at java.base/java.security.SecureClassLoader.defineClass(SecureClassLoader.java:150)
	at java.base/java.net.URLClassLoader.defineClass(URLClassLoader.java:524)
	at java.base/java.net.URLClassLoader$1.run(URLClassLoader.java:427)
	at java.base/java.net.URLClassLoader$1.run(URLClassLoader.java:421)
	at java.base/java.security.AccessController.doPrivileged(AccessController.java:712)
	at java.base/java.net.URLClassLoader.findClass(URLClassLoader.java:420)
```

## After this PR

Bump baseline-error-prone and fix failing tests


gradle-baseline-java/build.gradle
```diff
@@ -205,3 +205,8 @@ if (gradle.startParameter.isWriteDependencyLocks()) {
 gradleTestUtils {
     configurationCacheEnabled = true
 }
+
+javaVersion {
+    javaCompiler = 25
+    runtime = 21
```
kelvinou01, line 210: without `runtime=21` we fail with:

```
Execution failed for task ':gradle-baseline-java:checkRuntimeClasspathCompatible'.
> The runtime classpath has the following jars which contain classes that have too high a java language version. We're expecting the bytecode major version to be no more than 61 for java language version 17. Examples classes in each jar:
  
  /Users/okelvin/.gradle/caches/modules-2/files-2.1/com.palantir.baseline-error-prone/gradle-baseline-error-prone/0.17.0/975c81f451c34acf12d15d4e4f6ff4895ec5f12/gradle-baseline-error-prone-0.17.0.jar: com/palantir/gradle/baselineerrorprone/BaselineErrorProne$DisableXlintRemoval.class has bytecode major version 65
  /Users/okelvin/.gradle/caches/modules-2/files-2.1/com.palantir.suppressible-error-prone/gradle-suppressible-error-prone/2.31.0/34aaa8a9066b5f289205eed5200788e9232ae972/gradle-suppressible-error-prone-2.31.0.jar: com/palantir/gradle/suppressibleerrorprone/ConditionalPatchCheck.class has bytecode major version 65
  /Users/okelvin/.gradle/caches/modules-2/files-2.1/com.palantir.suppressible-error-prone/artifact-transforms/2.31.0/a936649dba0ccd76ecb20df91563f06937eb9b56/artifact-transforms-2.31.0.jar: com/palantir/gradle/suppressibleerrorprone/internal/SetupPreCompilationBytecodeManipulationPlugin.class has bytecode major version 65
```

kelvinou01: Seems like ^ is a false positive? I don't think `runtime` matters in this repo as it is a library

maybe `checkRuntimeClasspathCompatible` should only fire when `runtime` matters

gradle-baseline-java/src/test/groovy/com/palantir/baseline/BaselineCircleCiJavaIntegrationTests.groovy
```diff
@@ -31,7 +31,7 @@
 import org.junit.Test;
 import org.junit.rules.TemporaryFolder;
 
-public class BaselineCircleCiJavaIntegrationTests {
+class BaselineCircleCiJavaIntegrationTests {
```
kelvinou01, line 34: Renamed these files in the Groovy sourceset, as they were causing `GroovyCompile` tasks to fire `javac` 17

gradle-baseline-java/src/main/java/com/palantir/baseline/util/VersionsProps.java
```diff
@@ -68,7 +68,7 @@ class Builder extends VersionsProps_VersionForce_Builder {}
     }
 
     public static ParsedVersionsProps readVersionsProps(File propsFile) {
-        Preconditions.checkArgument(propsFile.exists(), "No " + propsFile.toPath() + " file found");
+        Preconditions.checkArgument(propsFile.exists(), "No %s file found", propsFile.toPath());
```
kelvinou01, line 71: [FormatStringShouldUsePlaceholders] Using a format string avoids string concatenation in the common case.

gradle-baseline-java/src/test/java/com/palantir/baseline/BaselineTestingIntegrationTest.java
```diff
@@ -31,7 +31,12 @@
 public class BaselineTestingIntegrationTest {
 
     private void standardBuildFile(RootProject rootProject) {
-        rootProject.buildGradle().plugins().add("java-library").add("com.palantir.baseline-testing");
+        rootProject
+                .buildGradle()
+                .plugins()
+                .add("java-library")
+                .add("com.palantir.baseline-testing")
+                .add("com.palantir.baseline-java-versions");
```
kelvinou01, line 39: We need this for `javaCompiler`, to set the javac version

gradle-baseline-java/build.gradle
crogoz: hmm, I think this would require consumers to be on `javaCompiler = 21`, which I guess everyone is using already. 

Any reason why we shouldn't set the runtime version in the root build.gradle ?

gradle-baseline-java/src/test/groovy/com/palantir/baseline/plugins/javaversions/ChosenJavaVersionTest.groovy
```diff
@@ -19,11 +19,6 @@
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
-import java.io.ByteArrayInputStream;
```
crogoz, line 22: nice!

PR #3637, opened 2026-06-12
Fix paths like `/generatedfoo/` being excluded from formatting

kelvinou01:
## Before this PR

gradle-baseline treated `/generatedfoo/` as a generated file, and excludes it from formatting

This was causing some files to be excluded from the formatter but still hit by suppressible-error-prone, leading to files being misformatted. 

## After this PR

Only treat paths with `/generated/` as generated files.

### DIFF
--- a/gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineFormat.java
+++ b/gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineFormat.java
@@ -47,7 +47,7 @@ class BaselineFormat extends AbstractBaselinePlugin {
     // TODO(dfox): remove this feature flag when we've refined the eclipse.xml sufficiently
     private static final String ECLIPSE_FORMATTING = "com.palantir.baseline-format.eclipse";
     private static final String PJF_PROPERTY = "com.palantir.baseline-format.palantir-java-format";
-    private static final String GENERATED_MARKER = File.separator + "generated";
+    private static final String GENERATED_MARKER = File.separator + "generated" + File.separator;
     private static final String PJF_PLUGIN = "com.palantir.java-format";
 
     @Override
--- a/gradle-baseline-java/src/test/java/com/palantir/baseline/BaselineFormatIntegrationTest.java
+++ b/gradle-baseline-java/src/test/java/com/palantir/baseline/BaselineFormatIntegrationTest.java
@@ -231,25 +231,13 @@ public class BaselineFormatIntegrationTest {
 
     @Test
     void format_does_not_skip_packages_that_merely_start_with_generated(GradleInvoker gradle, RootProject rootProject) {
-        // Regression test: the GENERATED_MARKER exclusion used to be a plain `/generated` substring match, which
-        // accidentally excluded any package whose name started with "generated" — e.g. `generatedxforms`,
-        // `generators`, `generation`. Such directories are not generated sources and must be formatted.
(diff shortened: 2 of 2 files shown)


gradle-baseline-java/src/test/java/com/palantir/baseline/BaselineFormatIntegrationTest.java
```diff
@@ -229,6 +229,37 @@ public class Test { Void test() {} }
...
+            package test.generatedxforms;
+            public class Test { Void test() {} }
+            """;
+        String formattedJavaFile = """
+            package test.generatedxforms;
+
+            public class Test {
+                Void test() {}
+            }
+            """;
```
kelvinou01: inline these

gradle-baseline-java/src/test/java/com/palantir/baseline/BaselineFormatIntegrationTest.java
```diff
@@ -229,6 +229,37 @@ public class Test { Void test() {} }
         assertThat(result).task(":spotlessJava").succeeded();
     }
 
+    @Test
+    void format_does_not_skip_packages_that_merely_start_with_generated(GradleInvoker gradle, RootProject rootProject) {
+        // Regression test: the GENERATED_MARKER exclusion used to be a plain `/generated` substring match, which
+        // accidentally excluded any package whose name started with "generated" — e.g. `generatedxforms`,
+        // `generators`, `generation`. Such directories are not generated sources and must be formatted.
```
kelvinou01: delete this

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineFormat.java
```diff
@@ -47,7 +47,7 @@ class BaselineFormat extends AbstractBaselinePlugin {
     // TODO(dfox): remove this feature flag when we've refined the eclipse.xml sufficiently
     private static final String ECLIPSE_FORMATTING = "com.palantir.baseline-format.eclipse";
     private static final String PJF_PROPERTY = "com.palantir.baseline-format.palantir-java-format";
-    private static final String GENERATED_MARKER = File.separator + "generated";
+    private static final String GENERATED_MARKER = File.separator + "generated" + File.separator;
```
crogoz, line 50: I don't think we need this anymore - we have this logic in: https://github.com/palantir/palantir-java-format/blob/develop/gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/SpotlessInterop.java#L53-L64

crogoz:
might be useful to expand the description - of what we are trying to solve (from the test it looks like we were excluding the modules that were called `generated*`)?

felixdesouza:
don't think you're going to have a good time with that? https://sourcegraph.palantir.build/search?q=context:%40fdesouza/infra-supported-orgs+file:/generated%5B%5E/%5D%2B+lang:Java+&patternType=keyword&sm=0

kelvinou01:
> don't think you're going to have a good time with that? https://sourcegraph.palantir.build/search?q=context:%40fdesouza/infra-supported-orgs+file:/generated%5B%5E/%5D%2B+lang:Java+&patternType=keyword&sm=0

oof.

kelvinou01:
Rather than include `/generatedfoo/` to be formatted, let's exclude `/generatedfoo/` from suppressible-error-prone

PR #3665, opened 2026-07-11
[`BaselineModuleJvmArgs`] Avoid capturing `SourceSet` to fix a configuration cache problem

gluxon:
## PR Stack

## Context

In a project on my team, I'm seeing a configuration cache issue that traces back the `doFirst` action added by `BaselineModuleJvmArgs`.

```
❯ ./gradlew assemble --configuration-cache

* What went wrong:

1 problem was found storing the configuration cache.
- Task `:<project>:javadoc` of type `org.gradle.api.tasks.javadoc.Javadoc`: cannot serialize object of type 'org.gradle.api.internal.tasks.DefaultSourceSet', a subtype of 'org.gradle.api.tasks.SourceSet', as these are not supported with the configuration cache.
  See https://docs.gradle.org/8.14.4/userguide/configuration_cache.html#config_cache:requirements:disallowed_types
```

## Changes

Saving a reference to the `sourceSet.getAnnotationProcessorPath()` gets past this error. There's a few other errors in this file that will need to be fixed separately.

### DIFF
--- a/gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
+++ b/gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
@@ -214,6 +214,10 @@ public abstract class BaselineModuleJvmArgs implements Plugin<Project> {
             return;
         }
 
+        // Grab a reference to the annotation processor path early to avoid capturing a reference to the full SourceSet
+        // (which is not Gradle configuration cache compatible) in the doFirst action below.
+        FileCollection annotationProcessorPath = sourceSet.getAnnotationProcessorPath();
+
         project.getTasks().named(sourceSet.getJavadocTaskName(), Javadoc.class, javadoc -> {
             // Javadoc task is horrible. There's no lazy properties/CommandLineArgumentProviders, so we have to
             // resort to this fresh hell of changing its configuration in a task action before it runs. Once
@@ -231,7 +235,7 @@ public abstract class BaselineModuleJvmArgs implements Plugin<Project> {
                     }
 
                     Set<String> exportValuesRaw = ModuleJvmArgsArgumentProvider.fromJustExtensionForCompilation(javadoc)
-                            .configureWithClasspath(sourceSet::getAnnotationProcessorPath)
+                            .configureWithClasspath(() -> annotationProcessorPath)
                             // Javadoc runs as *compilation* which means that everything that is normally an
                             // opens at runtime needs to be exports. Since we need to use the horrible
                             // addMultilineStringsOption, we just get all the arg fragments eg


gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
```diff
@@ -231,7 +235,7 @@ public void execute(Task _task) {
                     }
 
                     Set<String> exportValuesRaw = ModuleJvmArgsArgumentProvider.fromJustExtensionForCompilation(javadoc)
-                            .configureWithClasspath(sourceSet::getAnnotationProcessorPath)
+                            .configureWithClasspath(() -> annotationProcessorPath)
```
gluxon, line 238: I saw that the `configureWithClasspath` method intentionally takes a provider in order to encourage this value to be provided lazily.

I don't believe that's necessary here because it's atypical to completely overwrite underlying `FileCollection` object returned by `sourceSet.getAnnotationProcessorPath()`. Idiomatic Gradle usually involves mutating the `FileCollection` by calling `.from()` on it.

I can add a non-lazy overload of `configureWithClasspath` that takes a `FileCollection` directly if preferred. I figured given the Javadoc comment on `configureWithClasspath` that wouldn't be desirable.

CRogers: Hmmm, you can set the processor path, it's a method on `SourceSet`:
```java
void setAnnotationProcessorPath(FileCollection annotationProcessorPath);
```
It seems there is one internal Gradle plugin doing this. It's a bit more common to use `setRuntimeClasspath` and `setCompileClasspath`.

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
```diff
@@ -214,6 +214,10 @@ private void configureJavadoc(Project project, SourceSet sourceSet, BaselineModu
             return;
         }
 
+        // Grab a reference to the annotation processor path early to avoid capturing a reference to the full SourceSet
+        // (which is not Gradle configuration cache compatible) in the doFirst action below.
+        FileCollection annotationProcessorPath = sourceSet.getAnnotationProcessorPath();
```
CRogers, line 219: If we are OK with grabbing the `annotationProcessorPath` like this, we'd probably want to do this inside the `.named` just to make it a bit lazier and have more chance of being the right thing.

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
CRogers: I wonder if we could manually add a property to the Javadoc task and pass in what we need that way - something like this in the task configuration:

```java
javadoc.getInputs().property("add-exports-values", project.provider(() -> ...everything up to exportValuesSorted...)
```

...then inside `doFirst` we can just access `task.getInputs(). getProperties().get("add-exports-values")`. I think this should work, can't easily try it out in my current environment.

CRogers: Actually, I suspect this make #3665 an #3667 unnecessary as we wouldn't be calling the `task.getProject()` methods from inside a task action any more!

gluxon: > Hmmm, you can set the processor path, it's a method on `SourceSet`:
> 
> ```java
> void setAnnotationProcessorPath(FileCollection annotationProcessorPath);
> ```
> 
> It seems there is one internal Gradle plugin doing this. It's a bit more common to use `setRuntimeClasspath` and `setCompileClasspath`.

Ah that's my bad. Thank you for catching that. I should have checked that instead of assuming. 🤦🏻‍♂️

> I wonder if we could manually add a property to the Javadoc task and pass in what we need that way - something like this in the task configuration:

I like that idea!

PR #3666, opened 2026-07-12
[`BaselineModuleJvmArgs`] Pass extension directly rather than looking it up through the task

gluxon:
## PR Stack

## Problem

This fixes the last Gradle configuration cache problem I'm seeing in one of my projects.

```
* What went wrong:

1 problem was found storing the configuration cache.
- Task `:<project>:javadoc` of type `org.gradle.api.tasks.javadoc.Javadoc`: invocation of 'Task.project' at execution time is unsupported.
  See https://docs.gradle.org/8.14.4/userguide/configuration_cache.html#config_cache:requirements:use_project_during_execution
```

The problem is that at execution time, this method looks up the task's project in order to get a reference to the plugin's registered extension.

https://github.com/palantir/gradle-baseline/blob/959013868f0eb796709904bf92b45c5e5a37c897/gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java#L461-L463

## Changes

Instead of looking up the plugin extension in this roundabout way, let's pass around the extension reference directly.

### DIFF
--- a/gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
+++ b/gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
@@ -118,14 +118,14 @@ public abstract class BaselineModuleJvmArgs implements Plugin<Project> {
         project.getTasks().withType(Test.class).configureEach(test -> {
             test.getJvmArgumentProviders()
                     .add(ModuleJvmArgsArgumentProvider.fromClasspathAndExtension(
-                            getObjectFactory(), test, test::getClasspath));
+                            getObjectFactory(), test, extension, test::getClasspath));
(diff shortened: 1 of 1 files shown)


gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
```diff
@@ -245,7 +246,7 @@ public void execute(Task _task) {
                     }
 
                     Set<String> exportValuesRaw = ModuleJvmArgsArgumentProvider.fromJustExtensionForCompilation(
-                                    getObjectFactory(), javadoc)
+                                    getObjectFactory(), javadoc, extension)
```
gluxon, line 249: This is the line of code that's running at execution time and previously triggered a call for `task.getProject().getExtensions()`.

PR #3667, opened 2026-07-12
[`BaselineModuleJvmArgs`] Pass `ObjectFactory` directly instead of using `Project`

gluxon:
## PR Stack

## Problem

This fixes the next configuration cache error I'm seeing when running `javadoc` tasks.

```
* What went wrong:

2 problems were found storing the configuration cache, 1 of which seems unique.
- Task `:<project>:javadoc` of type `org.gradle.api.tasks.javadoc.Javadoc`: invocation of 'Task.project' at execution time is unsupported.
  See https://docs.gradle.org/8.14.4/userguide/configuration_cache.html#config_cache:requirements:use_project_during_execution
```

The problem is this `task.getProject()` call here happens during task execution time.

https://github.com/palantir/gradle-baseline/blob/959013868f0eb796709904bf92b45c5e5a37c897/gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java#L433

## Changes

I think the best fix here is to have the plugin `@Inject` an `ObjectFactory` and use that instead in the `doFirst` closure.

Unfortunately the call hierarchy of the `create` method is a bit intertwined. Changing its call signature requires changing a lot of upstream functions.

## Alternatives

I considered adding a new method overload of `create` that only `ModuleJvmArgsArgumentProvider.fromJustExtensionForCompilation` uses. That would be a more minimal change if preferred. I decided against it initially because it leaves around an extra overload for an unclear reason.

### DIFF
--- a/gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
+++ b/gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
@@ -46,6 +46,7 @@ import org.gradle.api.file.FileCollection;
 import org.gradle.api.java.archives.Manifest;
 import org.gradle.api.logging.Logger;
 import org.gradle.api.logging.Logging;
+import org.gradle.api.model.ObjectFactory;
 import org.gradle.api.provider.Property;
 import org.gradle.api.provider.Provider;
 import org.gradle.api.provider.ProviderFactory;
@@ -83,6 +84,9 @@ public abstract class BaselineModuleJvmArgs implements Plugin<Project> {
     @Inject
     protected abstract ConfigurationContainer getConfigurations();
 
+    @Inject
+    protected abstract ObjectFactory getObjectFactory();
+
     @Override
     @SuppressWarnings("checkstyle:MethodLength")
     public final void apply(Project project) {
@@ -113,13 +117,15 @@ public abstract class BaselineModuleJvmArgs implements Plugin<Project> {
 
         project.getTasks().withType(Test.class).configureEach(test -> {
             test.getJvmArgumentProviders()
-                    .add(ModuleJvmArgsArgumentProvider.fromClasspathAndExtension(test, test::getClasspath));
+                    .add(ModuleJvmArgsArgumentProvider.fromClasspathAndExtension(
+                            getObjectFactory(), test, test::getClasspath));
             setTaskInputsFromExtension(test, extension);
         });
 
         project.getTasks().withType(JavaExec.class).configureEach(javaExec -> {
             javaExec.getJvmArgumentProviders()
-                    .add(ModuleJvmArgsArgumentProvider.fromClasspathAndExtension(javaExec, javaExec::getClasspath));
+                    .add(ModuleJvmArgsArgumentProvider.fromClasspathAndExtension(
+                            getObjectFactory(), javaExec, javaExec::getClasspath));
             setTaskInputsFromExtension(javaExec, extension);
         });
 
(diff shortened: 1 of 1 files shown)


gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
```diff
@@ -432,9 +445,8 @@ public final ModuleJvmArgsArgumentProvider withExtraDescription(String descripti
             return this;
         }
 
-        private static ModuleJvmArgsArgumentProvider create(Task task) {
-            ModuleJvmArgsArgumentProvider provider =
-                    task.getProject().getObjects().newInstance(ModuleJvmArgsArgumentProvider.class);
+        private static ModuleJvmArgsArgumentProvider create(ObjectFactory objectFactory, Task task) {
+            ModuleJvmArgsArgumentProvider provider = objectFactory.newInstance(ModuleJvmArgsArgumentProvider.class);
```
gluxon, line 449: This specific line is the Gradle configuration cache fix. We no longer have a `task.getProject()` call here.

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
```diff
@@ -234,7 +244,8 @@ public void execute(Task _task) {
                         return;
                     }
 
-                    Set<String> exportValuesRaw = ModuleJvmArgsArgumentProvider.fromJustExtensionForCompilation(javadoc)
+                    Set<String> exportValuesRaw = ModuleJvmArgsArgumentProvider.fromJustExtensionForCompilation(
+                                    getObjectFactory(), javadoc)
```
gluxon, line 248: This is actually a bit odd, and I did want to point it out. Normally a `Task` uses `@Inject` to get an instance of `ObjectFactory`. Here the task is getting the injected `ObjectFactory` from the **plugin** above.

Unfortunately this is in a `doFirst` action on the Javadoc task, so I couldn't create a new task that directly `@Inject`'s the `ObjectFactory` instead of grabbing it from the plugin.

This didn't cause any issues in testing, but it's a bit non-standard and I wanted to call it out for review. For what it's worth, this is probably much better than what was happening before, which was the task looking up the project in order to grab the `ObjectFactory`.

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
```diff
@@ -83,6 +84,9 @@ public abstract class BaselineModuleJvmArgs implements Plugin<Project> {
     @Inject
     protected abstract ConfigurationContainer getConfigurations();
 
+    @Inject
+    protected abstract ObjectFactory getObjectFactory();
```
CRogers, line 88: I wonder if we can make a `ModuleJvmArgsArgumentProviderFactory` or something like that so we can avoid passing stuff around so much. I think this could also find the extension automatically... I will give it a go.

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
felixdesouza: do you want to just make this non-static?

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
CRogers: I got something going at https://github.com/palantir/gradle-baseline/compare/develop...callumr/cc-mja-with-factory

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
CRogers: I think this is fine - the plugin is configuring the task, and in the configuration using an object factory it has grabbed from the plugin class.

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java
felixdesouza: `ExtensionContainer` is a little tricky, the injection depends on the context iirc. If other things can have `ExtensionContainer`s you can get into weird things where have the extension container for the managed type itself, or it's on the task or it's on the project etc.

CRogers:
I think if [this works](https://github.com/palantir/gradle-baseline/pull/3665/changes#r3571572394) we'd fix the issue at the javadoc and perhaps not even need these changes

PR #3726, opened 2026-09-21
Emit info log for location of implicit dependency

pkoenig10:
It can sometimes be difficult to understand where an implicit dependency is coming from. For example, when the type is only used in an intermediate variable in a fluent expression.

```java
boolean anyExist = fooIds.stream()
     // getFoo looks like:
     //     Optional<Foo> getFoo(FooId id);
     //  
     // This class has a dependency on `Foo`, even though it's never explicitly imported
    .map(fooReader::getFoo)
    .mapMulti(Optional::ifPresent)
    .findAny();
```

The Maven dependency analyzer reports information about the usage location., but Baseline simply discards it.

This PR refactors the implementation to preserve this information and emit it in an `INFO` log which can be viewed by running the task with `--info`.


gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineExactDependencies.java
```diff
@@ -135,10 +137,11 @@ static String checkUnusedDependenciesNameForSourceSet(SourceSet sourceSet) {
     }
 
     /** Given a {@code com/palantir/product/Foo.class} file, what other classes does it import/reference. */
-    public static Stream<String> referencedClasses(File classFile) {
```
schlosna, line 138: Is there anything calling this public method? Do we need to deprecate and provide an adapted version that transforms `dependencyUsages` into referenced classes? I don't see any internal usages in my search, but not sure if there's anything external consuming this API

gradle-baseline-java/src/main/java/com/palantir/baseline/tasks/CheckImplicitDependenciesTask.java
```diff
@@ -83,19 +84,37 @@ public final void checkImplicitDependencies() {
...
-                .filter(artifacts -> artifacts.stream().noneMatch(this::shouldIgnore))
-                .filter(artifacts -> artifacts.stream().noneMatch(declaredArtifacts::contains))
-                // Select a single deterministic artifact for the suggestion
-                .map(artifacts -> artifacts.stream().min(ARTIFACT_COMPARATOR))
-                .<ResolvedArtifact>mapMulti(Optional::ifPresent)
+        List<ResolvedArtifact> usedButUndeclared = dependencyUsages()
+                .map(usage -> {
+                    List<ResolvedArtifact> artifacts = BaselineExactDependencies.INDEXES
+                            .classToArtifacts(usage.getDependencyClass())
+                            .collect(Collectors.toList());
```
schlosna: nit
```suggestion
                            .toList();
```

schlosna (commented):
Does this significantly change the runtime of the dependencies tasks as it looks like CI is timing out?

pkoenig10:
> Does this significantly change the runtime of the dependencies tasks as it looks like CI is timing out?

Mmm pushing the `noneMatch` checks up in every iteration of the loop is the cause.

I was being a bit lazy here - let me do a slightly larger refactor to avoid this penalty.

gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineExactDependencies.java
pkoenig10: I don't see anything, and I don't think we care. This is a de-facto internal method. The only reason it's public is because we made the unfortunate decision to put the tasks in a different package.

pkoenig10:
> Does this significantly change the runtime of the dependencies tasks as it looks like CI is timing out?

I don't see how it would be - AFAICT we're doing the exact same work we were before.

pkoenig10:
Oh the `Set` deduplication when constructing `Set<List<ResolvedArtifact>> necessaryArtifacts` is probably doing a lot of heavy lifting here.

I was being a bit lazy - left me refactor this a bit further.

